Issue #2, opened 2019-01-15
Boostrap project with Create React App


willdurand:
We want to use [Create React App](https://github.com/facebook/create-react-app) for this project. The project created with CRA should be configured to use:

- ~~SCSS: https://facebook.github.io/create-react-app/docs/adding-a-sass-stylesheet~~ (moved to https://github.com/mozilla/addons-code-manager/issues/22)
- ~~TypeScript: https://facebook.github.io/create-react-app/docs/adding-typescript~~ (moved to https://github.com/mozilla/addons-code-manager/issues/23)
- ~~React Router: https://facebook.github.io/create-react-app/docs/adding-a-router~~ (moved to https://github.com/mozilla/addons-code-manager/issues/24)

PR #20, opened 2019-01-15
Start with the default Create React App

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/2

I wanted to start with the (mostly) default Create React App although I did stop it from opening the site in Chrome 😛 

I figure it will be easier for us to collaborate by patching in customization on top of this.


willdurand:
> I wanted to start with the (mostly) default Create React App although I did stop it from opening the site in Chrome 😛

Good call, thanks. I think we can configure to open FF but... it is annoying.

willdurand:
I went ahead and merged this patch so that we can submit new patches based on this one.

PR #28, opened 2019-01-16
Example of CSS modules

### LINKED ISSUE #1: Do we want to use CSS modules?
See: https://facebook.github.io/create-react-app/docs/adding-a-css-modules-stylesheet

The main benefit is:

> CSS Modules let you use the same CSS class name in different files without worrying about naming clashes.

See also: https://css-tricks.com/css-modules-part-1-need/

> With CSS Modules, it’s a guarantee that all the styles for a single component:
>  - Live in one place
>  - Only apply to that component and nothing else

### LINKED ISSUE #19: Add enzyme for testing
Add [enzyme](https://airbnb.io/enzyme/) and helpers like in https://github.com/mozilla/addons-frontend/

willdurand:
~~⚠️ depends on #27~~
Fixes #1
Fixes #19 

---

Here it how it works (second commit):

- rename `.scss` file to `.module.scss`
- import it as a "JS-like module"
- use it

No need to add a scope to all CSS classes (hence shorter class names). It is guaranteed that there won't be any name collision. In production build, names can be shortened to reduce the amount of HTML sent to the client.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,9 +16,14 @@
     "typescript": "3.2.4"
   },
   "devDependencies": {
+    "@types/enzyme": "^3.1.15",
+    "@types/enzyme-adapter-react-16": "^1.0.3",
+    "codecov": "^3.1.0",
+    "enzyme": "^3.8.0",
+    "enzyme-adapter-react-16": "^1.7.1",
+    "jest-enzyme": "7.0.1",
     "prettier": "1.16.0",
     "pretty-quick": "^1.8.1",
-    "codecov": "^3.1.0",
     "type-coverage": "^1.6.2"
   },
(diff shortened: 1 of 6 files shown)


bobsilverberg:
This is pretty cool. I like it! 👍

willdurand:
> I like it but how do we reference classes in an enzyme wrapper query?

Good call, I pushed another commit to add enzyme support and wrote a better test case for the `App` component :)

kumar303:
> src/setupTests.ts

I was thinking that we should use `.tsx` for everything. That way we won't have to worry about renaming the file later if we need to put a React component in it (and we won't have to think about it).

PR #29, opened 2019-01-16
Add prettier

### LINKED ISSUE #9: Add Prettier and the `prettier-dev` command
We want to use Prettier to auto-format our code with the addons-frontend config:

- https://github.com/mozilla/addons-frontend/blob/e5a284072da5dc30d0613d3edf38eb5e70d67699/.prettierrc
- https://github.com/mozilla/addons-frontend/blob/e5a284072da5dc30d0613d3edf38eb5e70d67699/.prettierignore

We also want to reuse addons-frontend's `prettier-dev` command, which can be found here: https://github.com/mozilla/addons-frontend/blob/e5a284072da5dc30d0613d3edf38eb5e70d67699/package.json#L35

willdurand:
Fixes #9

---

This PR adds `prettier` as well as the yarn scripts we use in addons-frontend.


willdurand:
Sorry, I just rebased to add a new job to travis-ci. Everything looks good now :)

.prettierignore
```diff
@@ -0,0 +1,13 @@
...
+Dockerfile
+# exclude these directories
+/bin/
+/dist/
+/docker/
+/node_modules/
+# white-list files we want to process
+!*.js
+!*.md
+!*.scss
```
bobsilverberg: I know this is separate from the TS PR, but should this also white-list .ts and .tsx files?

willdurand: yep, we should. I'll add that before merging if the TS PR gets merged

willdurand:
Below is the new output of Travis CI with this patch:

screen shot 2019-01-17 at 11 42 08

 /cc @kumar303

Issue #6, opened 2019-01-15
Add Redux to the project


willdurand:
⚠️ react-redux v6 and the current impl. in addons-frontend are not compatible because the `store` prop does not get injected as before. Since we want to reuse `shallowUntilTarget()` and the testing pattern described in https://hacks.mozilla.org/2018/04/testing-strategies-for-react-and-redux/, we should make sure both can play nice together.

PR #36, opened 2019-01-16
Configure React Router for Create React App

### LINKED ISSUE #24: Configure React Router for Create React App
See https://facebook.github.io/create-react-app/docs/adding-a-router

bobsilverberg:
Fixes #24 

I tested this using an example from https://reacttraining.com/react-router/web/guides/quick-start and it seems to work. I didn't create any actual routes as we don't have any pages to serve yet.

I think we'll probably want to upgrade the router to use [connected-react-router](https://github.com/supasate/connected-react-router), but that should be done after we add redux, which I believe is coming in #6.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,11 +1,12 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import {HashRouter} from 'react-router-dom';
+import { BrowserRouter } from 'react-router-dom';
 import './index.css';
 import App from './App';
 
-ReactDOM.render((
-  <HashRouter>
+ReactDOM.render(
+  <BrowserRouter>
     <App />
-  </HashRouter>
-), document.getElementById('root') as HTMLElement);
+  </BrowserRouter>,
+  document.getElementById('root') as HTMLElement,
+);
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,6 +1,11 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
+import {HashRouter} from 'react-router-dom';
 import './index.css';
 import App from './App';
 
-ReactDOM.render(<App />, document.getElementById('root') as HTMLElement);
+ReactDOM.render((
+  <HashRouter>
+    <App />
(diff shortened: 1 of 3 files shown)


kumar303:
> I think we'll probably want to upgrade the router to use connected-react-router, 

Can you file an issue for that?

bobsilverberg:
> Can you file an issue for that?

#48

src/index.tsx
```diff
@@ -1,6 +1,11 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
+import {HashRouter} from 'react-router-dom';
 import './index.css';
 import App from './App';
 
-ReactDOM.render(<App />, document.getElementById('root') as HTMLElement);
+ReactDOM.render((
+  <HashRouter>
```
willdurand: Why should we use `HashRouter`? (compared to `BrowserRouter`)

bobsilverberg: I'm honestly not sure. I've read the docs, as well as a number of blog posts, and none of them seem to accurately explain the difference. In one place I saw someone mention that if you're running an SPA which does not require communication with a server for its functionality, then `HashRouter` makes it simpler. From everything I've read, other than that one mention, it doesn't seem like either would make much of a difference.

Also, I believe when we convert to `connected-react-router` it becomes moot, as there are no longer two different routers to choose between.

willdurand: So the `HashRouter` uses URLs like `/#foo/1`, which is convenient because you don't need to setup anything. The "server" serves `/` only and there is no need to hack anything to redirect all URLs to `/`. But URLs are fake from an HTTP point of view because everything after `#` is hidden from anything but the browser.

The `BrowserRouter` uses the HTML5 API, which is better because it uses proper URLs (real URLs), but it requires the server (that hosts the SPA) to redirect all URLs to `/`. This is pretty common use case nowadays, so it not a problem usually. It just works. I believe we should be using this one.

bobsilverberg: Ok, I'll change it to `BrowserRouter`.

PR #42, opened 2019-01-17
Enhance renovate config

willdurand:
This PR updates the renovate config to support our pinning strategy (similar to addons-frontend).


willdurand:
I merged this because it does not really need a review and I don't want more renovate PRs :D

PR #53, opened 2019-01-21
Add react-diff-view library

### LINKED ISSUE #10: Add react-diff-view library
[This library](https://github.com/otakustay/react-diff-view) will be used to write one of the main screens of this app.

⚠️ v2 of this library is about to be released, and it would be a good idea to use it directly. More information at: https://github.com/otakustay/react-diff-view/issues/24#issuecomment-453703703

bobsilverberg:
Fixes #10 

This doesn't do much, but it does add a component that will allow us to embed a `Diff` component into a page/other component. I tested it by temporarily adding my new `DiffView` component into `App.tsx` and it rendered successfully (although it was admittedly very ugly to look at).

I imagine this component will be iterated on as we actually start using it, but I thought it worthwhile to get feedback on what I've done with TS and with the tests.

Note that the change to the README file was made by `prettier` on my system, so I've included those changes in this commit.


src/components/DiffView.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+  expect(root).toHaveProp('viewType', viewType);
+});
+
+it('passes parsed diff information to DiffView', () => {
+  const parsedDiff = parseDiff(basicDiff)[0];
+  const root = shallow(<DiffView diff={basicDiff} />);
+
+  expect(root).toHaveProp('diffType', parsedDiff.type);
+  expect(root).toHaveProp('hunks', parsedDiff.hunks);
+});
```
willdurand: I think we should wrap all the `it` blocks + the two `const` statements into a `describe` block, likely `describe(__filename, () => {});`.

src/components/DiffView.tsx
```diff
@@ -0,0 +1,29 @@
+/// <reference path="DiffView.d.ts"/>
```
willdurand: How about creating a `DiffView` folder?

src/components/DiffView.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'react-diff-view';
```
willdurand: Shouldn't we put more types into this file?

src/components/DiffView.test.tsx
bobsilverberg: Yes, that makes sense.

src/components/DiffView.tsx
bobsilverberg: Oh, I thought I did that, but I didn't. I guess I only meant to do that. Thanks!

bobsilverberg: I also renamed the `.tsx` file from `DiffView.tsx` to `index.tsx` to match what we did for addons-frontend. I'm not sure if we want to follow that model, or if we'd rather have the component tsx files named after the components (i.e., `DiffView.tsx`) as I did originally.

src/components/DiffView.d.ts
bobsilverberg: TBH, I wasn't sure what should be put in this file, vs. in the file for the `DiffView` component. My understanding is that this file is meant to contain types that describe the third-party `react-diff-view` package. I guess we should add types for anything from that package that we are currently accessing?

src/components/DiffView.tsx
bobsilverberg: > How do you use Jest typeahead to find its tests? I assume that currently you can type DiffView and find the test, right?

I didn't check that, as there are so few tests I have just been either running them all, or defaulting to what needs to be run based in my changes. If I type `Diff` it does find and run my test, but I cannot see the typeahead output due to the issue you filed about that.

bobsilverberg:
After spending most of the day trying to figure out exactly how to create a usable `d.ts` file for `react-diff-view` and also get TS to recognize it (the documentation was unhelpful as were the voluminous blog posts and SO articles I read), I finally have something that works.

I've updated this patch with a working `d.ts` file. It's not complete, and it may very well include some things we won't actually need, as I don't think were totally sure exactly how we're going to interact with the `react-diff-view` library, but at least it's an example that can be used moving forwards.

I'm going to add some comments to the code as there is some questionable/strange stuff in there.

src/components/DiffView/index.d.ts
```diff
@@ -0,0 +1,38 @@
+declare module 'react-diff-view' {
+  export interface HunkInfo {
```
bobsilverberg: Note that we probably have no need to document the type for `HunkInfo` as it's never directly used by our code. We receive an array of `HunkInfo` objects from `parseDiff` and then pass them to the `Diff` component via a prop, but what they contain is irrelevant. We could possibly just use `type Hunks = Array<object>` instead of `type Hunks = Array<HunkInfo>` at line 12, below.

That's also the reason I documented `changes` as `Array<object>` here, rather than providing a concrete type for what a `change` is, because it's so deep that our code never really cares about it.

src/components/DiffView/index.d.ts
```diff
@@ -0,0 +1,38 @@
...
+    newStart: number;
+    oldLines: number;
+    newLines: number;
+    changes: Array<object>;
+    isPlain: boolean;
+  }
+
+  export type Hunks = Array<HunkInfo>;
+
+  export interface DiffInfo {
```
bobsilverberg: We don't make use of most of these properties, so maybe we only need to/should document the ones we actually care about? Those would be `hunks` and `type` in this case.

src/components/DiffView/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+export interface Props {
+  diff: string;
+  viewType: ViewType;
+}
+
+const diffViewDefaultProps = {
+  viewType: 'unified' as ViewType,
+};
+
+interface DiffViewSFC extends React.SFC<Props> {
```
bobsilverberg: I had to do this monkey business with the `diffViewDefaultProps` const and this extra interface to get TS to recognize the fact that we're providing a default value for `viewType` via `defaultProps`. Otherwise it doesn't recognize it which causes type problems. I spent quite a bit of time trying to figure this one out, so maybe there's a better solution. I believe this is just needed for an SFC and if we were creating a class it might "just work", and if so maybe that's a reason to use a class even though we don't currently need one.

willdurand: What is SFC?

bobsilverberg: Stateless Functional Component

bobsilverberg: It's from a code snippet I found which explained how to make this work.

src/components/DiffView.tsx
kumar303: ok, if you rename it to `index.tsx` I'd just want to make sure the same thing happens. It probably does.

bobsilverberg: Sorry, rename *what* to `index.tsx`? If you mean the main component file (not the test file), then that's already done.

kumar303: Oh, I see. Any reason why you didn't rename `DiffView/DiffView.test.tsx` to `DiffView/index.test.tsx`? My concern was that type ahead would still let you type `DiffView` to select that test (it should).

src/components/DiffView/DiffView.test.tsx
```diff
@@ -0,0 +1,46 @@
+/// <reference path="index.d.ts"/>
+
+import { shallow } from 'enzyme';
+import React from 'react';
+import { parseDiff } from 'react-diff-view';
+
+import DiffView from '.';
+
+describe(__filename, () => {
+  const basicDiff = `diff --git a/src/components/DiffView.test.tsx b/src/components/DiffView.test.tsx
```
willdurand: How about using fixture files for the diffs? This one is ~10 lines long, but we might have larger diffs in the future. We could maybe version diff files instead?

src/components/DiffView/index.d.ts
```diff
@@ -0,0 +1,38 @@
+declare module 'react-diff-view' {
+  export interface HunkInfo {
+    content: string;
+    oldStart: number;
+    newStart: number;
+    oldLines: number;
+    newLines: number;
+    changes: Array<object>;
```
willdurand: Can we use something else than `object` (a more precise type?)

src/components/DiffView/index.tsx
```diff
@@ -0,0 +1,33 @@
+/// <reference path="index.d.ts"/>
+
+import * as React from 'react';
+import { Diff, ViewType, parseDiff } from 'react-diff-view';
+export interface Props {
```
willdurand: Why do we export the interface here?

src/components/DiffView/index.tsx
willdurand: ok. Can we write something a bit simpler?

```diff
diff --git a/src/components/DiffView/index.tsx b/src/components/DiffView/index.tsx
index 258be44..b9682dd 100644
--- a/src/components/DiffView/index.tsx
+++ b/src/components/DiffView/index.tsx
@@ -1,22 +1,15 @@
 /// <reference path="index.d.ts"/>
 
 import * as React from 'react';
-import { Diff, ViewType, parseDiff } from 'react-diff-view';
-export interface Props {
-  diff: string;
-  viewType: ViewType;
-}
-
-const diffViewDefaultProps = {
-  viewType: 'unified' as ViewType,
-};
+import { Diff, DiffProps, ViewType, parseDiff } from 'react-diff-view';
 
-interface DiffViewSFC extends React.SFC<Props> {
-  defaultProps: typeof diffViewDefaultProps;
+interface Props extends DiffProps {
+  diff: string;
 }
 
-const DiffView: DiffViewSFC = (props) => {
+const DiffView = (props: Props) => {
   const { diff, viewType } = props;
+
   const parsedDiff = parseDiff(diff)[0];
 
   return (
@@ -28,6 +21,4 @@ const DiffView: DiffViewSFC = (props) => {
   );
 };
 
-DiffView.defaultProps = diffViewDefaultProps;
-
 export default DiffView;
```

src/components/DiffView/index.d.ts
```diff
@@ -0,0 +1,38 @@
+declare module 'react-diff-view' {
```
willdurand: It looks like we don't need all the `export`:

```diff
diff --git a/src/components/DiffView/index.d.ts b/src/components/DiffView/index.d.ts
index a5eb974..ac851f1 100644
--- a/src/components/DiffView/index.d.ts
+++ b/src/components/DiffView/index.d.ts
@@ -1,5 +1,5 @@
 declare module 'react-diff-view' {
-  export interface HunkInfo {
+  interface HunkInfo {
     content: string;
     oldStart: number;
     newStart: number;
@@ -9,9 +9,9 @@ declare module 'react-diff-view' {
     isPlain: boolean;
   }
 
-  export type Hunks = Array<HunkInfo>;
+  type Hunks = Array<HunkInfo>;
 
-  export interface DiffInfo {
+  interface DiffInfo {
     oldPath: string;
     newPath: string;
     hunks: Hunks;
@@ -24,15 +24,15 @@ declare module 'react-diff-view' {
     type: string;
   }
 
-  export function parseDiff(text: string, options?: object): Array<DiffInfo>;
+  function parseDiff(text: string, options?: object): Array<DiffInfo>;
 
-  export type ViewType = 'split' | 'unified';
+  type ViewType = 'split' | 'unified';
 
-  export interface DiffProps {
+  interface DiffProps {
     diffType: string;
     hunks: Hunks;
     viewType: ViewType;
   }
 
-  export class Diff extends React.Component<DiffProps, any> {}
+  class Diff extends React.Component<DiffProps, any> {}
 }
```

src/components/DiffView/DiffView.test.tsx
bobsilverberg: Yes, that's a good idea. The reason I just inlined it here is that I actually expect that the need for this will go away. I imagine that we'll have helpers that give us response data from the APIs, much as we have for addons-frontend, and at least one of those would contain diff information (I believe), so when we have that we won't need to store a sample diff elsewhere.

So because this is temporary, I felt like it made sense to just inline it.

src/components/DiffView/index.d.ts
bobsilverberg: I commented on this in the comments I added with my latest push. I don't believe we ever access this `changes` prop in our app. It's just something that is embedded in a `Hunk`, so I wasn't sure that it made sense to fully document it. I suppose, on the other hand, we don't know that we'll *never* make use of it, so maybe I should fully document it? Is that what you suggest?

src/components/DiffView/index.tsx
bobsilverberg: I exported everything by default, thinking we might need it in the tests. I will revisit what actually needs to be exported.

src/components/DiffView.tsx
bobsilverberg: I was a bit worried about having so many files called `index.*`. In addons-frontend we named our tests using the component names, but I guess that was because they all lived in the same folder (as opposed to beside each component). I've changed the test to be called `index.test.tsx` and Jest still seems to be able to find it with `p`, so I guess that's a better approach.

src/components/DiffView/index.tsx
bobsilverberg: The reason that works (for the `defaultProps` part) is that you're using a `class` instead of a functional component. I commented on this code earlier saying that we didn't need this monkey business if we used a class instead of a function, even though we don't need a class in this case. I gather you're saying that you'd prefer we use a class, even though one isn't needed, to be able to write simpler code?

bobsilverberg:
I've addressed most of the change requests, although I've left the inline diff in the test as per my response to your comment. I can change that if you feel strongly about it, but as I think it's going to be temporary, just inlining it seems like the thing that will require the least amount of effort if we do need to remove it later.

src/components/DiffView/index.tsx
willdurand: > to be able to write simpler code?

Yep, that sounds better to me.

src/components/DiffView/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+  viewType: DiffProps['viewType'];
+}
+
+class DiffView extends React.Component<Props> {
+  static defaultProps = {
+    viewType: 'unified',
+  };
+
+  render() {
+    const { diff, viewType } = this.props as Props;
```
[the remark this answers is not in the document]
bobsilverberg: There seems to be a problem with the `type-coverage` VSCode plugin, in that it's reporting these warnings as if TS doesn't understand that the type of `this.props` is `Props`. These problems are not reported by the CLI `type-coverage` tool, so I'll remove the type hint for now, as it still does not eliminate all of the warnings issued by the plugin.

I cannot find any way to configure the plugin, so at this point I'm not sure how to tell it not to issue these warnings. I opened https://github.com/plantain-00/type-coverage/issues/12 about the plugin.

bobsilverberg:
I have updated the patch to address the latest batch of comments.

src/components/DiffView/index.d.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  type ViewType = 'split' | 'unified';
+
+  type DiffProps = {
+    diffType: string;
+    hunks: Hunks;
+    viewType: ViewType;
+  };
+
+  class Diff extends React.Component<DiffProps, any> {}
```
kumar303: I think this could be `React.Component<DiffProps, {}>` to indicate empty state, right? It doesn't matter because we won't reach into its state anyway but our type coverage might be slightly happier.

src/components/DiffView/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+class DiffView extends React.Component<Props> {
+  static defaultProps = {
+    viewType: 'unified',
+  };
+
+  render = () => {
+    const { diff, viewType } = this.props;
+    const files = parseDiff(diff);
+
+    const renderFile = ({ hunks, type }: DiffInfo, index: number) => (
```
kumar303: TS can figure out `DiffInfo` if you pass it inline:

```diff
diff --git a/src/components/DiffView/index.tsx b/src/components/DiffView/index.tsx
index 88d020a..ba72c6c 100644
--- a/src/components/DiffView/index.tsx
+++ b/src/components/DiffView/index.tsx
@@ -17,11 +17,9 @@ class DiffView extends React.Component<Props> {
     const { diff, viewType } = this.props;
     const files = parseDiff(diff);
 
-    const renderFile = ({ hunks, type }: DiffInfo, index: number) => (
+    return <div>{files.map(({ hunks, type }, index: number) => (
       <Diff diffType={type} hunks={hunks} key={index} viewType={viewType} />
-    );
-
-    return <div>{files.map(renderFile)}</div>;
+    ))}</div>;
   };
 }
 
```

src/components/DiffView/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+      <Diff diffType={type} hunks={hunks} key={index} viewType={viewType} />
+    );
+
+    return <div>{files.map(renderFile)}</div>;
```
kumar303: It would be nice to have a test for multiple `<Diff>` hunks.

src/components/DiffView/index.tsx
bobsilverberg: Yep, that works.

src/components/DiffView/index.tsx
bobsilverberg: The problem I am having is that I cannot get `parseDiff` to generate output with multiple files. That's actually why I wrote the code without support for multiple files in the first place. Even using the sample diff at https://github.com/otakustay/react-diff-view/blob/v2.0.0-alpha.1/src/__test__/cases.js#L16, when I pass that into `parseDiff` it only extracts info for the first file.

This might be a bug with `parseDiff` that we're bound to come across when we try to use this library with actual data from our API, but I don't think it's worth trying to solve at this point. As I think I've already mentioned, this entire component will probably be replaced when we start trying to use this library to do actual things with our actual data. I only implemented it as a test to prove that we can at least use the `Diff` component to render something.

So while I agree that it makes sense to have code here that deals with possible multiple files, I don't think we should spend a lot of effort creating a test for it when the supporting code (`parseDiff`) may itself have a bug. Does this make sense?

kumar303: Huh, weird. If you get a minute, it might be helpful to file a bug about it.

bobsilverberg: Good idea, done at https://github.com/otakustay/react-diff-view/issues/33

PR #56, opened 2019-01-23
Introduce authentication code

### LINKED ISSUE #3: Config: should we use `.env` (dotenv, built-in with CRA)?
Instead of the [heavy configuration used in addons-frontend](https://github.com/lorenwest/node-config), we could maybe switch to something simpler, more intuitive and more cross-language: **dotenv**.

See: https://facebook.github.io/create-react-app/docs/adding-custom-environment-variables#what-other-env-files-can-be-used

### LINKED ISSUE #8: Setup Jest as testing framework
We want to use [Jest](https://jestjs.io/) for writing/running tests.

See: https://jestjs.io/docs/en/tutorial-react#setup-with-create-react-app

### LINKED ISSUE #14: Write common API layer
The common API layer should be the base of all API calls, i.e. the so-called `callApi` in addons-frontend. We can probably reuse its interface/design/code (fixing a few annoying things along the way).

This layer will talk to the Django HTTP API and we'll have to setup authentication for that too (with base62 tokens IIRC).

willdurand:
Fixes #3 
Fixes #14 
Likely fixes #8 too

---

~~This PR is more a POC than anything else so far but it shows how we could/should do authentication and the issues implied by the current addons-server's authentication layer (see: https://github.com/mozilla/addons-server/issues/10491).~~ Given the feedback, I went ahead and finished this work (adding tests, etc.)

This PR contains a proxy, very similar to the proxy used in addons-frontend. ~~The API host is `addons-dev.allizom.org` in (local) development and the FxA config is `local`.~~

I wrote a basic `callApi()` function, very simple but functional.

I use https://github.com/jefflau/jest-fetch-mock for mocking the HTTP layer (we could switch to something else, I just needed a basic mock library).

**Update:**

There are new commands to run the app in dev/prod mode with different envs (dev, stage, prod): https://github.com/mozilla/addons-code-manager/pull/56#issuecomment-457273914

There is now a redux integration and we can log in/log out from -dev or stage locally, in dev or prod modes.

There are a lot of missing test cases, this is normal given that it is a POC. If we go that way, then the last iteration will be to write more test cases.

The server code is written in plain old JS because it runs with `node`. We could use `ts-node`, maybe, but that was easier/faster for me to write this server without TS. We can totally convert this file to a TS file now though, assuming `ts-node` is good enough. We should keep in mind that this code will run on our servers, so better write something both simple and fast.

We can choose another library to mock `fetch`.


willdurand:
I just pushed a new commit with a server/proxy, integrated with CRA (no eject). It passes the auth token from the cookie to the pure client-side app by injecting it in the HTML (similar to addons-frontend). It works for both the "development" and "production" builds.

- development mode can be run with `yarn dev`
- production mode can be run with `yarn start`

You have to browse `http://localhost:3000/` in both cases (even though CRA says it listens to 3100). There is a lot of things to make better, but it is a POC :)

/cc @kumar303 @bobsilverberg

willdurand:
I just added a few more commands (and some changes):

- `yarn dev` runs the project in development mode, using the -dev API
- `yarn stage` runs the project in development mode, using the stage API
- `yarn start-dev` runs the project in production mode, using the -dev API
- `yarn start-stage` runs the project in production mode, using the stage API

Authentication works for each of these 4 commands.

`yarn start` (production mode, production API) does not work locally but should work in an hosted environment (hopefully).

willdurand:
> I am confused about one thing - does this approach mean that we do require a server for the app now? I was under the impression that we were trying to avoid the need for a server, if possible.

That looks like the only viable option we have unfortunately.. Given that we know that this mechanism works and is rather secure on addons-frontend. 

> Also, I tried running this locally and had issues. When I first started the app via yarn dev, and browsed to localhost:3000, I saw myself as logged in, I guess because I was already logged in to FxA for the frontend locally.

Yep, same cookie. 

> So that made it seem like it was working, but then when I logged out (by closing this app's server and starting up addons-frontend locally and then logging out), I was unable to log in again via the app. Maybe that's expected and that's not part of what's working yet?

This should work 😔. You should try to delete all your cookies in the devtools to be sure.

Also, can you give me the http calls made in the network panel? (html only, with persistent logs)

Do you see any new cookie created?

> Should I be accessing the app via localhost:3000 or localhost:3100, both are listed in the console after starting the app with yarn dev, and both seem to display the app. I cannot seem to get either working with login though.

localhost:3000

I moved CRA to port 3100, which is why you see it in the console and it shows the app. But that will not work because the API proxy won't be configured. We can probably output a better message in the console.

src/api/index.ts
```diff
@@ -0,0 +1,35 @@
...
+  const headers = {} as Headers;
+  const rootElement = document.getElementById('root');
+  const authToken = rootElement ? rootElement.dataset.authToken : null;
+
+  if (authToken && authToken !== '__AUTH_TOKEN__') {
+    headers['Authorization'] = `Bearer ${authToken}`;
+  }
+
+  const response = await fetch(
+    `${process.env.REACT_APP_API_HOST}/api/v4${endpoint}`,
```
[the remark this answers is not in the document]
willdurand: Probably yes, either in the config or as an argument at least.

src/api/index.ts
```diff
@@ -0,0 +1,35 @@
+type CallApiParams = {
+  endpoint: string;
+  method?: 'GET' | 'POST' | 'PUT' | 'PATCH' | 'DELETE';
+};
+
+type Headers = {
+  [name: string]: string;
+};
+
+export const callApi = async ({ endpoint, method = 'GET' }: CallApiParams) => {
```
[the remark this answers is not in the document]
willdurand: Good question. I think we'll likely need a type in the future.

src/api/index.test.ts
```diff
@@ -0,0 +1,50 @@
...
+        },
+      );
+    });
+
+    it('adds a trailing slash to the endpoint if there is none', async () => {
+      fetchMock.mockResponse(JSON.stringify({}));
+
+      await callApi({ endpoint: '/foo' });
+
+      expect(fetchMock.mock.calls[0][0]).toEqual(
```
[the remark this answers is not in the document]
willdurand: I want to assert the first argument only (the endpoint value)

src/App.test.tsx
```diff
@@ -5,6 +5,8 @@ import App from './App';
 import styles from './App.module.scss';
 
 it('renders without crashing', () => {
+  fetchMock.mockResponse(JSON.stringify({}));
```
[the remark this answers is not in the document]
willdurand: Yes, because I added an api call directly in a lifecycle method, so it is invoked in this test case. That's a demo, but better have a green test suite :p

willdurand:
Note: once we have redux, we could create an initial store containing the token value either in the server-side code or in the client-side code.

README.md
```diff
@@ -53,12 +53,36 @@ This runs [Prettier][] to automatically format the entire codebase.
 
 This runs [Prettier][] on only your changed files. This is intended for development.
 
-### `yarn start`
```
kumar303: I don't see `yarn start` documented. Can you add that?

willdurand:
@bobsilverberg ok so I think I fixed the issue you encountered while trying to log in after a log out. The cache headers were not correct (coming from CRA) but it is fixed now.

willdurand:
I added a commit with the code to log out a user.

README.md
willdurand: I don't really know how to document it yet because it is more an internal command now, i.e. it compiles and starts the app in production mode (but not for local usage). Maybe we should actually remove it for now?

src/api/index.test.ts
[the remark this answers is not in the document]
bobsilverberg: I understand that, but it still seems like using the `toHaveBeenCalledWith` API here would be better than reaching into the `calls` array, imo.

src/App.test.tsx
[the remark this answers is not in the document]
bobsilverberg: If I comment out this line the test still passes.

src/App.tsx
```diff
@@ -1,24 +1,67 @@
 import * as React from 'react';
-import logo from './logo.svg';
+
 import styles from './App.module.scss';
+import * as api from './api';
+import Login from './components/Login';
+
+interface Props {}
```
bobsilverberg: I believe we decided to use `type` for everything, except when `interface` is specifically needed. I'm going to update my open PR to make that change and it should probably be reflected in this PR as well.

src/App.tsx
```diff
@@ -1,24 +1,67 @@
...
+
+interface State {
+  profile: null | {
```
bobsilverberg: Nit: I know this is just for demo purposes, so maybe it would be overkill, but if this were production code I believe we'd want a `Profile` type defined.

src/api/index.tsx
```diff
@@ -0,0 +1,48 @@
+type CallApiParams = {
+  endpoint: string;
+  method?: 'GET' | 'POST' | 'PUT' | 'PATCH' | 'DELETE';
```
bobsilverberg: Do you think it's worth using an `enum` for this? You could probably make used of the `enum` at line 14, below.

src/api/index.test.ts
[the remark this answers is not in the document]
willdurand: I don't like this assertion either to be honest, but I don't really want to duplicate all the parameter values..

src/App.test.tsx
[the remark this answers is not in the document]
willdurand: Yep, but it should also output a warning about a promise not resolved (or something like this)

src/App.tsx
willdurand: agreed

src/App.tsx
willdurand: agreed

src/api/index.tsx
willdurand: agreed

willdurand:
Thanks @bobsilverberg, I am going to change everything you asked/suggested me to change. I will probably wait until #59 gets merged and rebase on top of that.

README.md
kumar303: If it's for production then we definitely need to document it.

willdurand: Yep, but it is not really ready yet. There is no `Dockerfile`, etc. I'd rather file an issue to create those commands when we'll have hosted envs.

willdurand:
mmh, looks like something is broken after the rebase.

willdurand:
> mmh, looks like something is broken after the rebase.
> 
> **EDIT:** I did not update the endpoint in the `App` component (it should be `accounts/profile`, not `accounts/authenticate`)

Fixed!

src/api/index.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+        headers: {},
+      });
+    });
+
+    it('adds a trailing slash to the endpoint if there is none', async () => {
+      fetchMock.mockResponse(JSON.stringify({}));
+
+      await callApi({ apiState, endpoint: '/foo' });
+
+      expect(fetchMock.mock.calls[0][0]).toEqual(`/api/v4/foo/`);
```
kumar303: Aww, this won't be as nice as `sinon.assert.calledWith()`. I like the sinon matchers. I guess the typescript and jest integration is nice with your approach, though.

src/api/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+  try {
+    const response = await fetch(`/api/${version}${endpoint}`, {
+      method,
+      headers,
+      credentials: credentials ? 'include' : undefined,
+    });
+
+    return await response.json();
+  } catch (error) {
+    return {
```
kumar303: It looks like error handling isn't fully hooked up (I don't see tests for it) so I suggest throwing an error for now just so we don't miss it.

src/api/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const headers = {} as Headers;
+  if (apiState.authToken) {
+    headers['Authorization'] = `Bearer ${apiState.authToken}`;
+  }
+
+  try {
+    const response = await fetch(`/api/${version}${endpoint}`, {
+      method,
+      headers,
+      credentials: credentials ? 'include' : undefined,
```
kumar303: I'm confused about credentials (I always have been on `addons-frontend`). I think we should not let it be configurable and always set it to some value -- either `include` or `undefined` -- until we run into problems. By not passing credentials, the API won't be able to recognize its own cookies. Does it need to see its own cookies? I'm not sure. If I toggle `credentials` on and off, the current log out call you added here still works. I guess that's because no cookies are involved.

src/components/Login/index.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
```
kumar303: This component is more like `LoginButton`, no?

server/server.js
```diff
@@ -0,0 +1,119 @@
...
+  );
+}
+
+app.listen(app.get('port'), () => {
+  console.log(
+    '\nApp is running at http://localhost:%d/ in %s mode',
+    app.get('port'),
+    app.get('env'),
+  );
+  console.log('Press CTRL-C to stop\n');
```
kumar303: It would be nice to get the [console clear](https://github.com/facebook/create-react-app/blob/91f54afb4086b30e1a9244390658dd6f711b855e/packages/react-dev-utils/clearConsole.js) feature back from CRA. Without it, TypeScript errors will hang around in the console after fixing them, which is confusing.

src/index.tsx
```diff
@@ -6,14 +6,26 @@ import { Provider } from 'react-redux';
...
 
 const store = configureStore();
 
+const rootElement = document.getElementById('root') as HTMLElement;
+const authToken = rootElement ? rootElement.dataset.authToken : null;
+
+if (
+  authToken &&
+  authToken.length &&
+  authToken !== process.env.REACT_APP_AUTH_TOKEN_PLACEHOLDER
```
kumar303: Does this case mean that the placeholder was not replaced? If so, this should throw an error because it means someone is running the app incorrectly.

src/index.tsx
```diff
@@ -6,14 +6,26 @@ import { Provider } from 'react-redux';
...
+) {
+  store.dispatch(apiActions.setAuthToken({ authToken }));
```
kumar303: I'd rather see this happen in `App` within `componentDidMount()`. That will make it easier to test.

src/reducers/api.tsx
```diff
@@ -0,0 +1,34 @@
+import { Reducer } from 'redux';
+import { ActionType, createAction, getType } from 'typesafe-actions';
+
+type SetAuthTokenPayload = {
+  authToken: string;
```
kumar303: `string | null`, right? I'm thinking of when logging out will need to reset `authToken`.

kumar303:
After closing down the server (`yarn dev`) with ^C, I noticed a process like this pegging my CPU at a constant 69%:

```
kumar            79759  69.0  0.7  2922268 116496 s001  R    11:05AM  18:47.95 /Users/kumar/.nvm/versions/node/v10.15.0/bin/node --max-old-space-size=2048 /Users/kumar/dev/addons-code-manager/node_modules/fork-ts-checker-webpack-plugin-alt/lib/service.js
```

I'm not totally sure what's going on but I have nothing running so I'm suspicious that maybe the proxy didn't shut down the CRA process cleanly. Could you look into that?

`dtruss` only shows `SYSCALL(args) 		 = return` and nothing else so I guess it's stuck in some non-IO loop.

kumar303:
Yep, more evidence that the proxy didn't shut down CRA correctly. When I killed the process I saw this in the shell:

```
[0] PORT=3100 BROWSER=none react-scripts start exited with code SIGINT
```

server/server.js
```diff
@@ -0,0 +1,119 @@
...
+  // Serve all the static files but `index.html`.
+  app.use(
+    express.static(path.join(__dirname, '..', 'build'), { index: false }),
+  );
+
+  // This handles all the incoming requests.
+  app.get('/*', (req, res) => {
+    const html = injectAuthenticationToken(req, indexHtml);
+
+    res.set('cache-control', 'private, no-store');
```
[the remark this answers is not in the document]
willdurand: I guess you specifically commented on the `no-store` directive, which is indeed used on addons-frontend to avoid issues with the serialized redux state. In this project, we need this directive to always read a valid auth token.

I thought about changing the etag but that seemed more work. I believe it's doable though, so if there are perf issues, we could do that.

src/api/index.test.tsx
willdurand: Yep

server/server.js
```diff
@@ -0,0 +1,119 @@
+const fs = require('fs');
```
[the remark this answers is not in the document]
willdurand: Yep

src/api/index.tsx
willdurand: In general, we should not send the cookies, especially because the API does not read them.

>  If I toggle `credentials` on and off, the current log out call you added here still works. I guess that's because no cookies are involved.

Ah? Really? The API returns a `set-cookie` header (to remove the cookies) and, without `credentials: include`, this header does not seem to be taken into account by the browser: no cookie gets removed without `credentials: include`.

src/index.tsx
willdurand: Yep, that works for me but it means: no API call can be made in `componentDidMount()` anymore.

server/server.js
willdurand: Not sure this is feasible with `concurrently` (or any other similar tool).

You should see a success message after an error though:

```
[0] Compiled successfully!
```

willdurand:
> After closing down the server (`yarn dev`) with ^C, I noticed a process like this pegging my CPU at a constant 69%:

Weird, `concurrently` should kill both processes and I added the `--kill-others` option to make sure both processes would be killed if one of them dies or exits.

src/api/index.tsx
willdurand: fixed

src/App.test.tsx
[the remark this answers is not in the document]
willdurand: removed

src/api/index.ts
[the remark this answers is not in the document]
willdurand: fixed

src/api/index.ts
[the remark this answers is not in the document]
willdurand: added a return type

src/api/index.test.ts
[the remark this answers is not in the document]
willdurand: fixed with a better jest syntax

bobsilverberg:
I just tried running the latest version and there seems to be a problem with it sending repeated requests:

jan-28-2019 10-52-51

bobsilverberg:
I also noticed that after clicking the "Log out" button, the browser does not update and in fact continues to issue those requests, and it is only after refreshing the browser that I appear to be logged out and the requests stop being issued.

willdurand:
> I just tried running the latest version and there seems to be a problem with it sending repeated requests:

ah oops. I think that's because of the change in `componentDidUpdate()` :(

src/api/index.spec.tsx
```diff
@@ -0,0 +1,144 @@
...
+      expect(fetch).toHaveBeenCalledWith(`/api/v4/foo/`, {
+        headers: {},
+        method: 'GET',
+      });
+    });
+
+    it('adds a trailing slash to the endpoint if there is none', async () => {
+      await callApiWithDefaultApiState({ endpoint: '/foo' });
+
+      expect(fetch).toHaveBeenCalledWith(`/api/v4/foo/`, expect.any(Object));
```
bobsilverberg: Maybe a helper for `toHaveBeenCalledWithURL`, which then checks the URL and does the `expect.any(Object)` thing would be nice. That would replace this more complex code in a number of the following tests.

Another option would be to create an object with the expected default values, e.g., `{ headers: {}, method: 'GET' }` in this case, and then pass that object in as the second argument to `toHaveBeenCalledWith`, and use the spread operator if you need to override one of those properties for a specific test.

I'm not sure how keen I am about having this `expect.any(Object)` in the tests.

willdurand:
@bobsilverberg sorry for that, I introduced an infinite loop causing the two issues you mentioned. This is fixed now.

src/api/index.spec.tsx
```diff
@@ -0,0 +1,144 @@
...
+    });
+
+    it('accepts an HTTP method', async () => {
+      const method = HttpMethod.POST;
+
+      await callApiWithDefaultApiState({ endpoint: '/foo/', method });
+
+      expect(fetch).toHaveBeenCalledWith(
```
bobsilverberg: A helper could be used here as well, `toHaveBeenCalledWithProperties`, which ignores the URL (by passing `expect.any(String)` as the first argument), and accepts an object with the expected properties. This could be combined with my suggestion about a default properties object which could be used to compose the argument to this new helper.

src/components/App/index.test.tsx
```diff
@@ -1,16 +1,56 @@
...
+    const dispatch = jest.spyOn(store, 'dispatch');
+
+    render({ authToken, store });
+
+    expect(dispatch).toHaveBeenCalledWith(
+      apiActions.setAuthToken({ authToken }),
+    );
+  });
+
+  it('does not dispatch setAuthToken on mount when authToken is null', () => {
```
bobsilverberg: I see you fixed the infinite loop, but could we have a test that would have caught that?

src/components/App/index.tsx
```diff
@@ -46,8 +123,9 @@ class App extends React.Component<
 
 const mapStateToProps = (state: ApplicationState): PropsFromState => {
   return {
+    apiState: state.api,
     toggledOn: state.example.toggledOn,
   };
 };
 
-export default connect(mapStateToProps)(App);
+export default connect(mapStateToProps)(AppBase);
```
bobsilverberg: Why are we calling it `AppBase` instead of just `App`?

src/configureStore.spec.tsx
```diff
@@ -0,0 +1,9 @@
+import configureStore from './configureStore';
+
+describe(__filename, () => {
+  it('registers all the reducers', () => {
+    const store = configureStore();
+
+    expect(Object.keys(store.getState())).toEqual(['api', 'example']);
```
bobsilverberg: Nit: It seems like this might be a hassle to maintain, but I guess it's worth having.

src/reducers/api.spec.tsx
```diff
@@ -0,0 +1,21 @@
+import reducer, { actions, initialState } from './api';
+
+describe(__filename, () => {
+  describe('reducer', () => {
+    it('initializes the state', () => {
+      const state = reducer(undefined, { type: 'SOME_ACTION' } as any);
```
bobsilverberg: Why not just `reducer(undefined, {})`, or does TS get mad about that?

README.md
```diff
@@ -39,7 +39,13 @@ This builds the app for production to the `build` folder.
...
-See the section about [deployment](https://facebook.github.io/create-react-app/docs/deployment) for more information.
+Note: for this project, we use a node server (see `scripts/server.js`).
+
+### `yarn dev`
+
+This runs the app in the development mode, with the [-dev AMO API](https://addons-server.readthedocs.io/).
+
+Open [http://localhost:3000](http://localhost:3000) to view it in the browser. The page will reload if you make edits. You will also see any lint errors in the console.
 
 ### `yarn eject`
```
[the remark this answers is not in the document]
kumar303: Yes, it is a command that can be run by a developer so it should be documented. Also, it can be useful to temporarily eject (then revert) if you want to figure out how something is configured. 

The only commands that I think are not worth documenting are those run by machines (such as `yarn test-ci`). There is a note in the readme about undocumented commands.

src/components/App/index.test.tsx
kumar303: Personally I think regression tests like that cause more harm than good because they are often slow and flaky, depending on how they are implemented it. They are hard to get right.

src/components/App/index.tsx
kumar303: Once we port over our linting rules from frontend you'll find out 😄  There's one in there that doesn't let you name a non-default export identical to the module (iirc). This is a good rule because it's easy to accidentally mix up named and default imports.

.env
```diff
@@ -0,0 +1,7 @@
+# This is the placeholder used in the `index.html` file to inject the
+# authentication token on runtime.
+REACT_APP_AUTH_TOKEN_PLACEHOLDER=__AUTH_TOKEN__
+
+REACT_APP_API_HOST=https://addons.mozilla.org
+REACT_APP_FXA_CONFIG=amo
+REACT_APP_AUTHENTICATION_COOKIE=frontend_auth_token
```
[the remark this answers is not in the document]
willdurand: This is the cookie name used by addons-frontend and we want to reuse that.

src/reducers/api.spec.tsx
willdurand: I believe so and that would make sense given that a redux action should have at least a `type`.

scripts/server.js
```diff
@@ -0,0 +1,143 @@
...
+  // Replace the placeholder in the `index.html` file with the token.
+  return html.replace(
+    `"${process.env.REACT_APP_AUTH_TOKEN_PLACEHOLDER}"`,
+    JSON.stringify(authToken || '').replace(/</g, '\\u003c'),
+  );
+};
+
+// We use a proxy to forward API requests to REACT_APP_API_HOST (i.e. the
+// AMO/addons-server API). This is useful for local development.
+if (process.env.REACT_APP_REQUIRE_PROXY === 'true') {
```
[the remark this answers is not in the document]
willdurand: We can't really rely on `NODE_ENV` because `NODE_ENV` is not our configuration environment, it is eith `development` or `production`, and it means more compiled or not compiled code than anything else :/

Adding a warning makes sense though.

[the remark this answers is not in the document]
willdurand: yep

src/api/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+    });
+
+    return await response.json();
+  } catch (error) {
+    return {
```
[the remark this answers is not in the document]
willdurand: But then, everyone has to try/catch and handle the error? While here, this is done once for all? I know errors are mostly network issues, so callers will already have to deal with 4xx vs 2xx responses, unless we improve this function to return an error object when the response is not OK.

src/api/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  try {
+    const response = await fetch(`/api/${version}${endpoint}`, {
+      method,
+      headers,
+      credentials: credentials ? 'include' : undefined,
+    });
+
+    return await response.json();
```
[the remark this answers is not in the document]
willdurand: Your patch shows a different implementation than what you wrote and it is more aligned with my previous comment:

> [...] unless we improve this function to return an error object when the response is not OK.

```diff
+    throw new Error(`Unexpected status for ${method} ${endpoint}: ${response.status}`);
```

Why not returning this error instead?

This would allow callers to write:

```js
const data = await callApi({ endpoint: '/foo' });

if (!data.error) {
  // happy path
} else {
  // error handling, like showing the error to the user
}
```

It is much more concise and throwing in async functions always lead to leaky promises (and this famous "node unhandled Promise Rrjections" warnings).

We could log a warning when `!response.ok` so that we get the best of both worlds: a nice API and developer feedback if something went wrong.

src/api/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const headers = {} as Headers;
+  if (apiState.authToken) {
+    headers['Authorization'] = `Bearer ${apiState.authToken}`;
+  }
+
+  try {
+    const response = await fetch(`/api/${version}${endpoint}`, {
+      method,
+      headers,
+      credentials: credentials ? 'include' : undefined,
```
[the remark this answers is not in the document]
willdurand: >     * Default `credentials` to `true` in `callApi()` so that all endpoints get session cookies by default. Most of them will need session cookies anyway.

I can understand why `omit` breaks but I don't think we need to send the cookies all the time for XHR requests.

> Setting credentials: undefined here still lets me log out in Firefox Release. I see the cookies when I log in and see them get deleted when logging out. I think that maybe some polyfill magic is involved here, doing surprising things when credentials is undefined.

I'd like to better understand why this happens, because I really tried to not add `credentials` when implementing the logout function.

scripts/server.js
```diff
@@ -0,0 +1,143 @@
...
+const injectAuthenticationToken = (req, html) => {
+  // Try to retrieve the authentication cookie, which contains the token.
+  const authToken = req.universalCookies.get(
+    process.env.REACT_APP_AUTHENTICATION_COOKIE,
+  );
+
+  // Replace the placeholder in the `index.html` file with the token.
+  return html.replace(
+    `"${process.env.REACT_APP_AUTH_TOKEN_PLACEHOLDER}"`,
+    JSON.stringify(authToken || '').replace(/</g, '\\u003c'),
```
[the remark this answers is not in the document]
willdurand: I wasn't happy with the coverage either, so I went ahead and added some easy coverage.

src/api/index.tsx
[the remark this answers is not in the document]
willdurand: Okay so I am not sure to understand why everything works without `credentials` actually. I remove everything and login/logout still works, huh. On this basis, I removed all the code around `credentials` since we don't need it.

src/api/index.tsx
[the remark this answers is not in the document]
willdurand: > It is much more concise and throwing in async functions always lead to leaky promises (and this famous "node unhandled Promise Rrjections" warnings).
> 
> We could log a warning when `!response.ok` so that we get the best of both worlds: a nice API and developer feedback if something went wrong.

I've implemented this suggestion to show what I was thinking

willdurand:
Added more test cases, fixed most of the things you asked. The only remaining point is the `callApi` behavior, which I think is okay now (cf. comments above).

src/server/index.spec.js
```diff
@@ -0,0 +1,70 @@
...
+  describe('createServer', () => {
+    const rootPath = path.join(__dirname, 'fixtures');
+
+    describe('NODE_ENV=production', () => {
+      const env = {
+        NODE_ENV: 'production',
+      };
+
+      it('creates an Express server', async () => {
+        const server = request(createServer({ env, rootPath }));
```
kumar303: Each time supertest creates a server it gets bound to an ephemeral port. That's about all the docs say. I'm pretty sure that awaiting the promise will shut down the server but I'm not totally sure. We need to shut the server down after each test. 

On the frontend we used `.end()` to do this but in this newer version of supertest, it prints a warning saying that you shouldn't mix that with promises. If you have any ideas, let me know, but I guess we can just ride with it.

src/server/index.spec.js
```diff
@@ -0,0 +1,70 @@
...
+        universalCookies: {
+          get: () => token,
+        },
+      };
+      const env = { REACT_APP_AUTH_TOKEN_PLACEHOLDER: '__PLACEHOLDER__' };
+      const html = `<div data-token="${env.REACT_APP_AUTH_TOKEN_PLACEHOLDER}">`;
+
+      const htmlWithToken = injectAuthenticationToken(req, html, env);
+
+      expect(htmlWithToken).toEqual(`<div data-token="${token}">`);
```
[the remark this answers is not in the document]
willdurand: ah oops, yep. I borrowed this extra `replace` call from the redux's doc to avoid injections: https://redux.js.org/recipes/server-rendering

src/server/index.spec.js
willdurand: > I'm pretty sure that awaiting the promise will shut down the server but I'm not totally sure. We need to shut the server down after each test.

Reading the docs, it is done, no? I can't find leaky processes.

kumar303: Yes, I think it's working

PR #57, opened 2019-01-23
Minor fixes for prettier

willdurand:
This PR:

- reformats the README file
- renames the `LICENSE` file to avoid a prettier error


README.md
```diff
@@ -84,4 +82,4 @@ You can learn more in the [Create React App documentation](https://facebook.gith
 To learn React, check out the [React documentation](https://reactjs.org/).
 
 [prettier]: https://prettier.io/
-[TypeScript]: https://www.typescriptlang.org/
+[typescript]: https://www.typescriptlang.org/
```
bobsilverberg: You've changed it to `typescript` here, but not changed the references in the doc.

willdurand: Looks like it works: https://github.com/mozilla/addons-code-manager/tree/minor-fixes-prettier

bobsilverberg: It does work, but I thought we might want to be consistent with the casing anyway. 🤷‍♂️

willdurand: I think prettier enforces that for url keys because that's the spec. But TypeScript should be written with capital T and S.

kumar303:
If this was how prettier wanted it formatted, how come the previous builds didn't fail? I don't see a failure in the original README patch nor any of the patches to update prettier.

bobsilverberg:
Is it possible that our tests don't target `.md` files?

willdurand:
> If this was how prettier wanted it formatted, how come the previous builds didn't fail? I don't see a failure in the original README patch nor any of the patches to update prettier.

Correct, I fixed that in #60, sorry 😞

Issue #55, opened 2019-01-22
Configure Redux devtools


kumar303:
After https://github.com/mozilla/addons-code-manager/issues/6, we need to hook up the app for use with Redux devtools.

Issue #17, opened 2019-01-15
Set up code linting


kumar303:
Set up our common linting rules similar to [addons-frontend](https://github.com/mozilla/addons-frontend/blob/d69965f80b615a2ba57adb777e56f33f73d97afd/.eslintrc).

We will be using TypeScript instead of Flow but TS linting can be done in a separate issue.

willdurand:
Note: `tslint` is recommended over `eslint` when using TypeScript.

kumar303:
As discussed in https://github.com/mozilla/addons-code-manager/pull/59/files#r250526915 let's be sure to add a lint rule to enforce arrow functions everywhere.

PR #59, opened 2019-01-23
Set up Redux

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/6, fixes https://github.com/mozilla/addons-code-manager/issues/55

Here are some notes:

* There is some code in here that I used for testing (the toggle button) but I didn't write actual tests for it. I guess that will mess up our coverage. I'm not sure yet if it's worth adding tests just for that.
* The example code will be good for reference (mostly to figure out typing) but I thought we'd delete it once we replace it with something real.
* Try to play around with making mistakes, like incorrect `mapStateToProps` return values or reducer errors. TypeScript figures everything out without needing too many hints! This is pretty cool. **If you can figure out a way to break type safety, let me know.**
* I mostly used https://github.com/resir014/react-redux-typescript-example as a guide

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "jest-enzyme": "7.0.1",
     "prettier": "1.16.1",
     "pretty-quick": "^1.8.1",
-    "redux-devtools-extension": "2.13.7",
+    "redux-devtools-extension": "^2.13.7",
     "type-coverage": "^1.6.2"
   },
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "react-router-dom": "4.3.1",
     "react-scripts": "2.1.3",
     "redux": "4.0.1",
+    "redux-devtools-extension": "2.13.7",
     "redux-logger": "3.0.6",
     "typesafe-actions": "3.0.0",
     "typescript": "3.2.4"
@@ -30,7 +31,6 @@
     "jest-enzyme": "7.0.1",
     "prettier": "1.16.1",
     "pretty-quick": "^1.8.1",
-    "redux-devtools-extension": "^2.13.7",
     "type-coverage": "^1.6.2"
   },
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "node-sass": "4.11.0",
     "react": "16.7.0",
     "react-dom": "16.7.0",
-    "react-redux": "6.0.0",
+    "react-redux": "5.1.1",
     "react-router-dom": "4.3.1",
     "react-scripts": "2.1.3",
     "redux": "4.0.1",
--- a/package.json
+++ b/package.json
@@ -8,11 +8,17 @@
     "@types/react": "16.7.20",
     "@types/react-dom": "16.0.11",
     "@types/react-router-dom": "4.3.1",
+    "@types/react-redux": "7.0.0",
+    "@types/redux-logger": "3.0.6",
+    "@types/webpack-env": "1.13.6",
     "node-sass": "4.11.0",
     "react": "16.7.0",
     "react-dom": "16.7.0",
+    "react-redux": "6.0.0",
     "react-router-dom": "4.3.1",
     "react-scripts": "2.1.3",
+    "redux": "4.0.1",
+    "redux-logger": "3.0.6",
     "typescript": "3.2.4"
   },
   "devDependencies": {
@@ -24,6 +30,7 @@
     "jest-enzyme": "7.0.1",
     "prettier": "1.16.1",
     "pretty-quick": "^1.8.1",
+    "redux-devtools-extension": "2.13.7",
     "type-coverage": "^1.6.2"
   },
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -10,7 +10,6 @@
     "@types/react-router-dom": "4.3.1",
     "@types/react-redux": "7.0.0",
     "@types/redux-logger": "3.0.6",
-    "@types/webpack-env": "1.13.6",
     "node-sass": "4.11.0",
     "react": "16.7.0",
     "react-dom": "16.7.0",
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
     "@types/node": "10.12.18",
     "@types/react": "16.7.20",
     "@types/react-dom": "16.0.11",
-    "@types/react-router-dom": "4.3.1",
     "@types/react-redux": "7.0.0",
+    "@types/react-router-dom": "4.3.1",
     "@types/redux-logger": "3.0.6",
     "node-sass": "4.11.0",
     "react": "16.7.0",
@@ -18,6 +18,7 @@
     "react-scripts": "2.1.3",
     "redux": "4.0.1",
     "redux-logger": "3.0.6",
+    "typesafe-actions": "3.0.0",
     "typescript": "3.2.4"
   },
   "devDependencies": {
--- a/src/App.test.tsx
+++ b/src/App.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { mount, shallow } from 'enzyme';
+import { shallow } from 'enzyme';
 
 import App from './App';
 import styles from './App.module.scss';
--- a/src/App.test.tsx
+++ b/src/App.test.tsx
@@ -1,11 +1,15 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { mount, shallow } from 'enzyme';
 
 import App from './App';
 import styles from './App.module.scss';
+import configureStore from './configureStore';
 
 it('renders without crashing', () => {
-  const root = shallow(<App />);
+  const store = configureStore();
+  // TODO: Use shallowUntilTarget()
+  // https://github.com/mozilla/addons-code-manager/issues/15
+  const root = shallow(<App />, { context: { store } }).shallow();
 
   expect(root).toHaveClassName(styles.container);
   expect(root.find(styles.header)).toHaveLength(1);
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,11 +6,15 @@ import logo from './logo.svg';
 import styles from './App.module.scss';
 import { ExampleState, toggleOn, toggleOff } from './reducers/example';
 
-interface Props {
+interface PublicProps {}
+
+interface PropsFromState {
   toggleState: ExampleState['toggle'];
 }
 
-class App extends React.Component<Props & ConnectedReduxProps, {}> {
+class App extends React.Component<
+  PublicProps & PropsFromState & ConnectedReduxProps
+> {
   handleToggleClick = (event: React.SyntheticEvent<HTMLButtonElement>) => {
     const { dispatch, toggleState } = this.props;
     event.preventDefault();
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,36 @@
 import * as React from 'react';
+import { connect } from 'react-redux';
+
+import { ApplicationState, ConnectedReduxProps } from './configureStore';
 import logo from './logo.svg';
 import styles from './App.module.scss';
+import { ExampleState, toggleOn, toggleOff } from './reducers/example';
 
-class App extends React.Component {
+interface Props {
+  toggleState: ExampleState['toggle'];
+}
+
+class App extends React.Component<Props & ConnectedReduxProps, {}> {
+  handleToggleClick = (event: React.SyntheticEvent<any>) => {
+    const { dispatch, toggleState } = this.props;
+    event.preventDefault();
+
+    if (toggleState === 'on') {
+      dispatch(toggleOff());
+    } else {
+      dispatch(toggleOn());
+    }
+  };
+
   render() {
+    const { toggleState } = this.props;
+
     return (
       <div className={styles.container}>
         <header className={styles.header}>
           <img src={logo} className={styles.logo} alt="logo" />
           <p>
-            Edit <code>src/App.js</code> and save to reload.
+            Edit <code>src/App.tsx</code> and save to reload.
           </p>
           <a
             className={styles.link}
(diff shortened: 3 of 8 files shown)


src/reducers/example.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export function toggleOn(): ToggleOnAction {
+  return {
+    type: TOGGLE_ON,
+    payload: {},
+  };
+}
+
+interface ToggleOffAction {
```
willdurand: I am curious, why is this an `interface` and not a `type`?

src/App.tsx
```diff
@@ -1,15 +1,36 @@
...
+import { ApplicationState, ConnectedReduxProps } from './configureStore';
 import logo from './logo.svg';
 import styles from './App.module.scss';
+import { ExampleState, toggleOn, toggleOff } from './reducers/example';
+
+interface Props {
+  toggleState: ExampleState['toggle'];
+}
+
+class App extends React.Component<Props & ConnectedReduxProps, {}> {
```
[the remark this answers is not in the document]
bobsilverberg: I kind of like the use of the `&` here, making it more explicit that there are two types contributing to the props of the class. I guess it's really a matter of taste - it doesn't seem like one approach is strictly better than the other.

src/App.tsx
```diff
@@ -19,10 +40,25 @@ class App extends React.Component {
...
+            </button>
+          </p>
         </header>
       </div>
     );
   }
 }
 
-export default App;
+function mapStateToProps(state: ApplicationState) {
```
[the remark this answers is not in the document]
bobsilverberg: I would be happy with doing this as well.

src/App.tsx
```diff
@@ -1,15 +1,36 @@
 import * as React from 'react';
+import { connect } from 'react-redux';
+
+import { ApplicationState, ConnectedReduxProps } from './configureStore';
 import logo from './logo.svg';
 import styles from './App.module.scss';
+import { ExampleState, toggleOn, toggleOff } from './reducers/example';
+
+interface Props {
```
bobsilverberg: There's a pattern presented in https://medium.com/@resir014/redux-4-typescript-2-9-a-type-safe-approach-7f073917b803 which uses a `PropsFromState` interface to define any props coming from `mapStateToProps`, which can then be intersectioned with any props specific to the component. In this example it would be overkill, but I thought it to be a pretty nice pattern, making it very obvious exactly from whence props were originating. When I first saw it I thought it seemed nicer than our `InternalProps` that we use in addons-frontend, which could include props from multiple sources.

What do you think about adopting a convention like that?

src/reducers/example.tsx
```diff
@@ -0,0 +1,54 @@
+import { Reducer } from 'redux';
+
+const TOGGLE_ON = 'TOGGLE_ON';
```
bobsilverberg: What do you think about using an `enum` for these? `enum`s are one of the things I really liked about TS when I was reading more about it.

src/reducers/example.tsx
```diff
@@ -0,0 +1,54 @@
...
+const TOGGLE_OFF = 'TOGGLE_OFF';
+
+export interface ExampleState {
+  toggle: 'on' | 'off';
+}
+
+interface ToggleOnAction {
```
bobsilverberg: Have you looked at [typesafe-actions](https://github.com/piotrwitek/typesafe-actions)? I've seen it mentioned in a few of the examples I've looked at for Redux+TS and it looks pretty good. I think it could cut down on verbosity of defining actions.

src/reducers/example.tsx
bobsilverberg: I was trying to figure out when one would use a `type` vs an `interface` and cam across quite a bit of inaccurate information. [This post](https://medium.com/@martin_hotell/interface-vs-type-alias-in-typescript-2-7-2a8f1777af4c) seems to provide an up to date description of the issue. It seems to me that for the most part they are actually interchangeable, so it might be best to standardize on always using one in particular unless the specific application requires the other. What do you think?

bobsilverberg:
Ugh, I did that think where I respond to existing comments in the diff view, rather than the conversation view, so my comments appear disconnected from the comment I was replying to. I think if you view them in the diff view they are understandable, but they come off as kind of confusing when viewed in the conversation view. 😢

src/reducers/example.tsx
willdurand: yep I also, think we should adopt ENUM for the redux actions.

package.json
```diff
@@ -8,11 +8,17 @@
     "@types/react": "16.7.20",
     "@types/react-dom": "16.0.11",
     "@types/react-router-dom": "4.3.1",
+    "@types/react-redux": "7.0.0",
+    "@types/redux-logger": "3.0.6",
+    "@types/webpack-env": "1.13.6",
```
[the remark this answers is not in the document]
kumar303: Oh, actually, I can remove it now. In an earlier patch I was referencing `module.hot` (it was needed for that) for reloading reducers but I discovered the reloading was already supported in CRA (I think? I dunno, it works).

src/App.test.tsx
```diff
@@ -3,9 +3,13 @@ import { shallow } from 'enzyme';
 
 import App from './App';
 import styles from './App.module.scss';
+import configureStore from './configureStore';
```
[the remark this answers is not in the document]
kumar303: If CRA let's us add them, sure. I'm not too worried about it since the tests will be co-located and I don't think the app will get as big as addons-frontend.

src/reducers/example.tsx
kumar303: Thanks for the link, Bob. I propose that we use `type` everywhere because of this:

> [declaration merging] doesn’t work with type aliases, because type is an unique type entity ( for both global or module scope )

[This](https://medium.com/@martin_hotell/interface-vs-type-alias-in-typescript-2-7-2a8f1777af4c) is saying that you can merge interfaces but *you cannot merge* types. We probably won't need to merge interfaces so we don't need that feature. The reason I think `type` is superior here is that someone could copy/paste the same interface name by mistake and not get a compilation error:

```js
interface LogInAction {
  type: 'LOG_IN';
  payload: {};
}

// oops, this is a typo maybe caused by copy/paste.
interface LogInAction {
  type: 'LOG_OUT_ACTION';
  payload: {};
}
```

src/reducers/example.tsx
kumar303: ok, I switched to `enum` for action types but it makes them longer to type. What's the advantage?

src/App.tsx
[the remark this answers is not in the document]
bobsilverberg: For consistency, imo.

src/reducers/example.tsx
bobsilverberg: 👍 for using `type` everywhere. Death to `interface`!

src/reducers/example.tsx
kumar303: I had not looked at this. I like it! Let's use it. It cut the code size down by nearly 50% and I think it still reads well. The API is well documented, too.

```
2 files changed, 12 insertions(+), 21 deletions(-)
```

kumar303:
> Ugh, I did that thing where I respond to existing comments in the diff view

Yep, this is a pretty bad github bug

src/App.tsx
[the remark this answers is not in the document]
kumar303: that's the only reason?

src/App.tsx
kumar303: Yeah, I think we should adopt the `PropsFromState` idea. I switched to that. It's really just for code readability, not for type hinting (which is what we needed it for in Flow). TypeScript can already figure out mapped props automatically.

src/App.tsx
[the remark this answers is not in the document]
willdurand: If TS does that automatically, then there is maybe no need for this extra type?

As for the extend on the interface, it is different than & and it looks like the main advantage of interfaces is to be able to extend them.

src/App.tsx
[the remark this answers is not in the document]
willdurand: that's more than enough for me, plus `this` is correctly bound all the time, no need to think about that either.

src/reducers/example.tsx
willdurand: I also saw this problem about merging interfaces and was about to suggest the use of types everywhere. Yet, there are probably valid use cases for interfaces and we should maybe learn more about why interfaces are encouraged in TS. Also, it is very unfortunate to have two options for that..

src/App.tsx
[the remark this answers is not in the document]
kumar303: Alright, I'll change them to arrow functions. Consistency doesn't seem like the strongest rationale to me but if you both feel strongly, let's do it 🤷‍♀️  I will just need a lint rule otherwise I'll forget 😀 https://github.com/mozilla/addons-code-manager/issues/17#issuecomment-457371415

src/App.tsx
willdurand: I thought adopting TS was also about typing less useless code? I don't really want to write code that "TypeScript can already figure out".

kumar303:
Thanks, it's ready for another look. 

@bobsilverberg the new approach with `typesafe-actions` doesn't even need `enum` so that works for me.

src/App.tsx
kumar303: Well, I think it's nice as documentation to glance at `PropsFromState` and know *immediately* that those are props I should not set. TypeScript would give you an error if you try to set it, anyway, but I think it's still nice for readability.

src/App.tsx
```diff
@@ -19,10 +39,25 @@ class App extends React.Component {
...
 }
 
-export default App;
+const mapStateToProps = (state: ApplicationState) => {
+  return {
+    toggledOn: state.example.toggledOn,
+  };
+};
+
+export default connect(mapStateToProps)(App);
```
bobsilverberg: Hmm, don't we need to do something like:

```
const App: React.ComponentType<PublicProps> = connect(mapStateToProps)(AppBase);
export default App;
```

In order to ensure type safety when using this component? Even if TS is smart enough to understand which props come from state (and I'm not sure how it can do that - I guess by magic), we still have the `ConnectedReduxProps` in the props listed for the default component.

src/App.tsx
```diff
@@ -19,10 +39,25 @@ class App extends React.Component {
...
         </header>
       </div>
     );
   }
 }
 
-export default App;
+const mapStateToProps = (state: ApplicationState) => {
+  return {
+    toggledOn: state.example.toggledOn,
```
bobsilverberg: I did a test here where I added an additional property to this object, and the TS compiler didn't complain. Shouldn't we be trying to enforce type safety in this case as well, or is that just something we choose not to worry about?

src/App.tsx
kumar303: We do not! TypeScript figures everything out. Here are some examples:

Passing an unknown prop named `foo` to a component:

```
Type error: Type '{ foo: string; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<Pick<PropsFromState & ConnectedReduxProps<AnyAction>, never>, any, any>> & Readonly<{ children?: ReactNode; }> & Readonly<...>'.
  Property 'foo' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<Pick<PropsFromState & ConnectedReduxProps<AnyAction>, never>, any, any>> & Readonly<{ children?: ReactNode; }> & Readonly<...>'.  TS2322

    13 |   <Provider store={store}>
    14 |     <BrowserRouter>
  > 15 |       <App foo="bar" />
       |        ^
    16 |     </BrowserRouter>
    17 |   </Provider>,
    18 |   document.getElementById('root') as HTMLElement,
```

Passing an  *already mapped* prop to a component:

```
Type error: Type '{ toggledOn: boolean; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<Pick<PropsFromState & ConnectedReduxProps<AnyAction>, never>, any, any>> & Readonly<{ children?: ReactNode; }> & Readonly<...>'.
  Property 'toggledOn' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<Pick<PropsFromState & ConnectedReduxProps<AnyAction>, never>, any, any>> & Readonly<{ children?: ReactNode; }> & Readonly<...>'.  TS2322

    13 |   <Provider store={store}>
    14 |     <BrowserRouter>
  > 15 |       <App toggledOn={true} />
       |        ^
    16 |     </BrowserRouter>
    17 |   </Provider>,
    18 |   document.getElementById('root') as HTMLElement,
```

Passing a special Redux prop (like `dispatch`):

```
Type error: Type '{ dispatch: () => void; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<Pick<PropsFromState & ConnectedReduxProps<AnyAction>, never>, any, any>> & Readonly<{ children?: ReactNode; }> & Readonly<...>'.
  Property 'dispatch' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<Pick<PropsFromState & ConnectedReduxProps<AnyAction>, never>, any, any>> & Readonly<{ children?: ReactNode; }> & Readonly<...>'.  TS2322

    13 |   <Provider store={store}>
    14 |     <BrowserRouter>
  > 15 |       <App dispatch={() => {}} />
       |        ^
    16 |     </BrowserRouter>
    17 |   </Provider>,
    18 |   document.getElementById('root') as HTMLElement,
```

I don't know how it figured all of this out but my gut instinct is that the type definitions imported from `react-redux` have been implemented to do all the right things. This gives me a warm fuzzy feeling that TypeScript "just works" a lot better than Flow.

src/App.tsx
kumar303: Good catch. I guess when I tried it, I only changed the value of `toggledOn` (which produced an error). I added a type hint for the return type of `PropsFromState` which addresses it.

src/App.tsx
willdurand: Ok, I don't necessarily agree because, if the language is smart and actually better than we expected, I don't think we should be too verbose, mostly because we were doing that to solve a Flow limitation before (in this case, because I believe we would not have introduced `InternalProps` if Flow was behaving like TS here).

Anyway, let's see how it goes first.

src/App.tsx
willdurand: >  This gives me a warm fuzzy feeling that TypeScript "just works" a lot better than Flow.

💯

bobsilverberg:
This is a good point, @willdurand. If it's true that we never would have
introduced `InternalProps` if it weren't for Flow needing them for proper
type safety, then we probably don't need to call these out here as a
separate type. You've convinced me!

PR #60, opened 2019-01-24
Fix prettier-ci

### LINKED ISSUE #61: `prettier-ci` does not fail the build
`prettier-ci` should fail the build when there are formatting issues. This is not the case right now, see: https://github.com/mozilla/addons-code-manager/pull/57#issuecomment-456968830

willdurand:
Fixes #61

---

This makes the build fail if there is a prettier error.


kumar303:
Thanks for digging into this

PR #64, opened 2019-01-25
Source code clean-up

willdurand:
This PR removes a few useless files and fix minor things like title,
etc. It also moves the App logic to a directory.

This should ease my work on #56.


public/index.html
```diff
@@ -8,11 +8,6 @@
       content="width=device-width, initial-scale=1, shrink-to-fit=no"
     />
     <meta name="theme-color" content="#000000" />
-    <!--
-      manifest.json provides metadata used when your web app is added to the
-      homescreen on Android. See https://developers.google.com/web/fundamentals/web-app-manifest/
-    -->
-    <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
     <!--
       Notice the use of %PUBLIC_URL% in the tags above.
```
bobsilverberg: Nit: It is now a single `tag`, rather than multiple `tags`.

src/components/App/index.test.tsx
```diff
@@ -1,9 +1,9 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 
-import App from './App';
-import styles from './App.module.scss';
-import configureStore from './configureStore';
+import App from '.';
+import styles from './styles.module.scss';
+import configureStore from '../../configureStore';
```
bobsilverberg: Maybe it's time to start using absolute paths, if that's an option? All the `../../` in this patch are a bit annoying.

PR #74, opened 2019-01-29
Configure ESLint

### LINKED ISSUE #17: Set up code linting

### LINKED ISSUE #75: Add a lint rule to disallow .ts files, enforce .tsx
We want to always create `.tsx` files, not `.ts` files. This reduces the mental burden of trying to figure out which file to create. This lint rule is important because our lint command in https://github.com/mozilla/addons-code-manager/pull/74 will *ignore* `.ts` files.

willdurand:
Fixes #17 
Fixes #75

---

There are 3 warnings due to JSX not being supported by ESLint anymore: https://eslint.org/blog/2015/03/eslint-0.17.0-released#changes-to-jsxreact-handling.

I think it could maybe be fixed automagically by upgrading some of our dependencies, but it requires some work on https://github.com/mozilla/eslint-config-amo (which is on my todo-list).


src/configureStore.tsx
```diff
@@ -1,5 +1,13 @@
...
+import {
+  Action,
+  AnyAction,
+  Dispatch,
+  Reducer,
+  Store,
+  applyMiddleware,
+  combineReducers,
+  createStore,
+} from 'redux';
```
bobsilverberg: 😆

src/configureStore.tsx
```diff
@@ -1,5 +1,13 @@
-import { Dispatch, Action, AnyAction } from 'redux';
-import { Store, applyMiddleware, combineReducers, createStore } from 'redux';
+import {
+  Action,
+  AnyAction,
+  Dispatch,
+  Reducer,
```
bobsilverberg: I can't decide whether I think it's nicer, or more confusing, that we now import types the same as code. 🤷‍♂️

kumar303:
If it's not too crazy, how about squeezing this in? https://github.com/prettier/eslint-config-prettier Otherwise it could be done in a separate PR. Disabling the rules made moot by prettier would be nice and could possibly speed up our linting.

package.json
```diff
@@ -25,17 +25,23 @@
...
+    "eslint-plugin-amo": "^1.9.1",
+    "jest-enzyme": "^7.0.1",
     "prettier": "1.16.1",
     "pretty-quick": "^1.8.1",
     "type-coverage": "^1.6.2"
   },
   "scripts": {
     "build": "react-scripts build",
     "eject": "react-scripts eject",
+    "eslint": "eslint --ext .js --ext tsx src/",
```
kumar303: Does this still lint `.ts` files? It doesn't look like it.

package.json
```diff
@@ -45,9 +51,6 @@
     "type-coverage": "type-coverage",
     "typecheck": "tsc"
   },
-  "eslintConfig": {
-    "extends": "react-app"
```
kumar303: Do we still want this for anything?

src/configureStore.tsx
kumar303: > I can't decide whether I think it's nicer, or more confusing, that we now import types the same as code. 🤷‍♂️

I think it's nice. It illustrates how TypeScript is a first class language, not something bolted-on to the JS.

src/configureStore.tsx
```diff
@@ -13,7 +21,7 @@ export type ApplicationState = {
   example: ExampleState;
 };
 
-const createRootReducer = () => {
+const createRootReducer = (): Reducer<ApplicationState> => {
```
kumar303: TS can already infer this from `return combineReducers<ApplicationState>`. Did you need this for something? If I remove it, I don't get any lint errors.

.eslintrc
```diff
@@ -0,0 +1,37 @@
+{
+  "extends": [
+    "amo",
+    "plugin:amo/recommended",
+    "plugin:jest/recommended",
+    "plugin:@typescript-eslint/recommended"
```
kumar303: I see that they clobbered some of our AMO preferences. I discovered that `no-used-vars` is a warning but it should be an error. Maybe there are more.

```diff
diff --git a/.eslintrc b/.eslintrc
index cbae414..f70c925 100644
--- a/.eslintrc
+++ b/.eslintrc
@@ -25,7 +25,8 @@
     ],
     // We decided to prefer types over interfaces.
     "@typescript-eslint/prefer-interface": "off",
-    "@typescript-eslint/explicit-member-accessibility": "off"
+    "@typescript-eslint/explicit-member-accessibility": "off",
+    "@typescript-eslint/no-unused-vars": "error"
   },
   "settings": {
     "import/resolver": {
```

src/index.tsx
```diff
@@ -1,3 +1,4 @@
+/* global: document */
```
kumar303: I see you added this to the rc file so this can be deleted

src/reducers/example.tsx
```diff
@@ -5,7 +5,7 @@ type TogglePayload = {};
 
 export const actions = {
   toggle: createAction('TOGGLE', (resolve) => {
-    return () => resolve({} as TogglePayload);
```
kumar303: I see that this fails a rule but we need to provide type hints here (it's the way `typesafe-actions` wants us to do it). Maybe there's another way?

Here is a hypothetical problem that we would need the typehint to catch for us (the resolved action payload is invalid):

```diff
diff --git a/src/reducers/example.tsx b/src/reducers/example.tsx
index d2c6c48..2667eb2 100644
--- a/src/reducers/example.tsx
+++ b/src/reducers/example.tsx
@@ -1,19 +1,21 @@
 import { Reducer } from 'redux';
 import { ActionType, createAction, getType } from 'typesafe-actions';
 
-type TogglePayload = {};
+type TogglePayload = { thing: 'string' };
 
 export const actions = {
   toggle: createAction('TOGGLE', (resolve) => {
-    return () => resolve({});
+    return () => resolve({ thing: 'something', aTypo: 'should-fail' });
   }),
 };
 
 export type ExampleState = {
+  thing: string,
   toggledOn: boolean;
 };
 
 const initialState: ExampleState = {
+  thing: '',
   toggledOn: false,
 };
 
@@ -23,7 +25,7 @@ const reducer: Reducer<ExampleState, ActionType<typeof actions>> = (
 ): ExampleState => {
   switch (action.type) {
     case getType(actions.toggle):
-      return { ...state, toggledOn: !state.toggledOn };
+      return { ...state, toggledOn: !state.toggledOn, thing: action.payload.thing };
     default:
       return state;
   }
```


Using a type hint will catch the error:

```diff
diff --git a/src/reducers/example.tsx b/src/reducers/example.tsx
index d2c6c48..7cc8cca 100644
--- a/src/reducers/example.tsx
+++ b/src/reducers/example.tsx
@@ -1,19 +1,21 @@
 import { Reducer } from 'redux';
 import { ActionType, createAction, getType } from 'typesafe-actions';
 
-type TogglePayload = {};
+type TogglePayload = { thing: 'string' };
 
 export const actions = {
   toggle: createAction('TOGGLE', (resolve) => {
-    return () => resolve({});
+    return () => resolve({ thing: 'something', aTypo: 'should-fail' } as TogglePayload);
   }),
 };
 
 export type ExampleState = {
+  thing: string,
   toggledOn: boolean;
 };
 
 const initialState: ExampleState = {
+  thing: '',
   toggledOn: false,
 };
 
@@ -23,7 +25,7 @@ const reducer: Reducer<ExampleState, ActionType<typeof actions>> = (
 ): ExampleState => {
   switch (action.type) {
     case getType(actions.toggle):
-      return { ...state, toggledOn: !state.toggledOn };
+      return { ...state, toggledOn: !state.toggledOn, thing: action.payload.thing };
     default:
       return state;
   }
```

```
/Users/kumar/dev/addons-code-manager/src/reducers/example.tsx
Type error: Conversion of type '{ thing: "something"; aTypo: string; }' to type 'TogglePayload' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Types of property 'thing' are incompatible.
    Type '"something"' is not comparable to type '"string"'.  TS2352

     6 | export const actions = {
     7 |   toggle: createAction('TOGGLE', (resolve) => {
  >  8 |     return () => resolve({ thing: 'something', aTypo: 'should-fail' } as TogglePayload);
       |                          ^
     9 |   }),
    10 | };
    11 |
```

src/configureStore.tsx
willdurand: > I think it's nice. It illustrates how TypeScript is a first class language, not something bolted-on to the JS.

💯

package.json
willdurand: I thought we banned `.ts` files?

src/configureStore.tsx
willdurand: No error but a warning:

```
  26:27  warning  Missing return type on function      @typescript-eslint/explicit-function-return-type
```

.eslintrc
willdurand: Agreed!

My problem currently is that this rule reports false-positive when vars are used in JSX.

src/reducers/example.tsx
willdurand: right, but this is useless here, no? There is nothing in the payload? I don't even think we need to add the `resolve` function.

willdurand: That was useful to implement the `api` action though, so thanks!

willdurand:
I pushed a new commit with more types.. The ESLint TS config wanted moaar!

package.json
kumar303: Ah, yes we did! Amen. I filed https://github.com/mozilla/addons-code-manager/issues/75 so we don't forget to disallow `.ts` files.

src/reducers/example.tsx
kumar303: Yes, it's useless in the real code but I was just pointing out how we will need to use this pattern (or some alternative) in the future. It was a little different in your auth patch since payload was the input parameter and could be typed that way.

willdurand:
rebased again

willdurand:
> I found the issue related to the JSX and false-positives, a fix has been merged < 24h ago: [typescript-eslint/typescript-eslint#161](https://github.com/typescript-eslint/typescript-eslint/pull/161) and I confirm it works.

I added another commit to use the canary release of the plugin.

kumar303:
> If it's not too crazy, how about squeezing this in? https://github.com/prettier/eslint-config-prettier 

I deferred it to https://github.com/mozilla/addons-code-manager/issues/84

src/api/index.tsx
```diff
@@ -24,17 +27,19 @@ export const callApi = async ({
   endpoint,
   method = HttpMethod.GET,
   version = 'v4',
-}: CallApiParams): Promise<object | { error: Error }> => {
```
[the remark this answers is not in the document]
willdurand: yep, but I think I needed to reuse the type somewhere

src/api/index.tsx
```diff
@@ -24,17 +27,19 @@ export const callApi = async ({
...
+}: CallApiParams): Promise<CallApiResponse> => {
   if (!endpoint.startsWith('/')) {
+    // eslint-disable-next-line no-param-reassign
```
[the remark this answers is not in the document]
willdurand: Better, thanks!

src/api/index.spec.tsx
```diff
@@ -1,12 +1,15 @@
...
   actions as apiActions,
+  initialState as defaultApiState,
 } from '../reducers/api';
 import configureStore from '../configureStore';
 
+import { HttpMethod, callApi, logOutFromServer } from '.';
+
 describe(__filename, () => {
-  const getApiState = ({ authToken = '12345' } = {}) => {
+  const getApiState = ({ authToken = '12345' } = {}): ApiState => {
```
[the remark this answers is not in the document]
willdurand: I tend to agree that adding all those types seems overkill. Let's try without this rule :)

src/index.tsx
```diff
@@ -1,3 +1,4 @@
+/* global document */
```
kumar303: Wasn't this in the rc file before? I think it would be safe to allow all browser globals since that's where the app will run. We could use [`env: browser`](https://eslint.org/docs/user-guide/configuring#specifying-environments), too.

src/reducers/api.spec.tsx
```diff
@@ -3,6 +3,7 @@ import reducer, { actions, initialState } from './api';
 describe(__filename, () => {
   describe('reducer', () => {
     it('initializes the state', () => {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
kumar303: I like this rule 👍

src/reducers/example.tsx
kumar303: I think we need an example of how to accomplish this pattern. I don't know if this is the best way but I'd say maybe keep this as is with a comment disabling the rule. If you do want to delete it, you can delete the type definition for `TogglePayload`, too (why didn't unused vars catch that? Hmm).

package.json
```diff
@@ -50,6 +54,8 @@
     "build": "react-scripts build",
     "dev": "cat .env.common-local .env.dev > .env.local && yarn start-dev",
     "eject": "react-scripts eject",
+    "eslint": "eslint --ext .js --ext tsx src/",
```
kumar303: I thought more about this. I think we do need to include `.ts` files for declarations.

bobsilverberg: I just noticed this. Why is it `.js` (with a dot) and `tsx` (without a dot)?

src/index.tsx
willdurand: forgot about the `env` config... thanks!

src/reducers/example.tsx
willdurand: removed everything

package.json
willdurand: fixed, sorry

willdurand:
I fixed all the comments, thanks!

package.json
```diff
@@ -53,6 +57,8 @@
     "build": "react-scripts build",
     "dev": "cat .env.common-local .env.dev > .env.local && yarn start-dev",
     "eject": "react-scripts eject",
+    "eslint": "eslint --ext js --ext tsx src/",
```
kumar303: I think we also need `--ext ts` for declaration files, right?

src/api/index.tsx
```diff
@@ -24,42 +26,48 @@ export const callApi = async ({
...
     return {
       error,
     };
   }
 };
 
-export const logOutFromServer = async (apiState: ApiState) => {
+export const logOutFromServer = async (
+  apiState: ApiState,
+): Promise<CallApiResponse> => {
```
kumar303: You can remove the explicit return type now

package.json
kumar303: We do indeed. However, there is a meta problem: `amo/only-tsx-files` doesn't seem to fail for the *existence* of a `.ts` file. D'oh.

```
$ eslint --ext js --ext tsx --ext ts src/

/Users/kumar/dev/addons-code-manager/src/@types/react-diff-view/index.d.ts
   1:1   error  Only use `.tsx` file extensions in a TypeScript project                        amo/only-tsx-files
  16:14  error  Array type using 'Array<ChangeInfo>' is forbidden. Use 'ChangeInfo[]' instead  @typescript-eslint/array-type
  25:16  error  Array type using 'Array<HunkInfo>' is forbidden. Use 'HunkInfo[]' instead      @typescript-eslint/array-type
  40:55  error  Array type using 'Array<DiffInfo>' is forbidden. Use 'DiffInfo[]' instead      @typescript-eslint/array-type
  50:9   error  'Diff' is defined but never used                                               @typescript-eslint/no-unused-vars
  50:22  error  'React' is not defined                                                         no-undef

/Users/kumar/dev/addons-code-manager/src/react-app-env.d.ts
  1:1  error  Expected exception block, space or tab after '//' in comment  spaced-comment
  2:1  error  Only use `.tsx` file extensions in a TypeScript project       amo/only-tsx-files

/Users/kumar/dev/addons-code-manager/src/setupTests.ts
   1:1   error  Only use `.tsx` file extensions in a TypeScript project                                        amo/only-tsx-files
   1:1   error  'enzyme' should be listed in the project's dependencies, not devDependencies                   import/no-extraneous-dependencies
   2:1   error  'enzyme-adapter-react-16' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
   3:1   error  'jest-fetch-mock' should be listed in the project's dependencies, not devDependencies          import/no-extraneous-dependencies
   4:1   error  'jest-enzyme' should be listed in the project's dependencies, not devDependencies              import/no-extraneous-dependencies
  10:1   error  Expected 1 empty line after require statement not followed by another require                  import/newline-after-import
  10:22  error  'jest-fetch-mock' should be listed in the project's dependencies, not devDependencies          import/no-extraneous-dependencies
  14:3   error  'fetchMock' is not defined                                                                     no-undef

✖ 16 problems (16 errors, 0 warnings)
```

kumar303: > amo/only-tsx-files doesn't seem to fail for the existence of a .ts file. 

Actually, wait, I think just adding `--ext ts` will address this

willdurand: arf, fixed.

src/@types/react-diff-view/index.d.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint amo/only-tsx-files: 0, @typescript-eslint/no-unused-vars: 0 */
```
kumar303: Maybe we should rename it to `react-diff-view/index.d.tsx` if that's allowed?

willdurand: I tried but it did not work.. Let's try again, otherwise I'll fix that in the rule (later).

willdurand: It does not work, `tsc` does not recognize the module anymore.

willdurand: see: https://github.com/mozilla/eslint-plugin-amo/issues/17

kumar303: Ah, ok, I thought `.d.ts` might be special

PR #81, opened 2019-01-30
test.js => spec.js

### LINKED ISSUE #67: Rename test files from `.test.js` to `.spec.js`

willdurand:
Fixes #67


willdurand:
Auto-merging as it is only 2 renamed files.

PR #82, opened 2019-01-30
Use Node 10

### LINKED ISSUE #68: Make Node 10 the default
Everyone else is doing it:
* https://github.com/mozilla/addons-frontend/pull/7417
* https://github.com/mozilla/addons-frontend/issues/7418

Since we will be running a server as of https://github.com/mozilla/addons-code-manager/pull/56 we should be on Node 10.

willdurand:
Fixes #68

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,10 @@ language: node_js
 _aliases:
   - &node-prod
     # This is the NodeJS version we run in production.
-    node_js: '8'
+    node_js: '10'
   - &node-next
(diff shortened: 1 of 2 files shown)


willdurand:
> Heh, I noticed that I was already developing with node 10 in here. Should we add a `.nvmrc` for that reason?

I would rather not, I don't use Node 10. We could add an `engine` section in `package.json` if you want to throw an error if Node 10 is not used though.

PR #83, opened 2019-01-30
Add tests for deps

### LINKED ISSUE #78: Enforce version etiquette for package dependencies
We need to enforce this etiquette:

* Production dependencies in `package.json` are always exact versions (i.e. no caret). We already set this as the default in [`.yarnrc`](https://github.com/mozilla/addons-code-manager/blob/b075147c1c02b68f2c24ee426f3c3368ee94391f/.yarnrc).
* Development dependencies in `package.json` should allow for automatic patch upgrades (version numbers should have a caret). We will need to manage a list of exceptions, like for `prettier` which must be pinned to an exact version.

We can do this in a unit test, similar to what we did for https://github.com/mozilla/addons-frontend/issues/3391

willdurand:
Fixes #78


src/packageJson.spec.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+type KeyValue = { [key: string]: string };
+
+describe(__filename, () => {
+  const skipDevDeps = ['prettier'];
+
+  Object.keys(packageJson.devDependencies).forEach((key) => {
+    it(`should have devDependencies[${key}] version prefixed with "^"`, () => {
+      if (!skipDevDeps.includes(key)) {
+        expect((packageJson.devDependencies as KeyValue)[key]).toEqual(
```
kumar303: `expect().toMatch()` would be a bit shorter

src/packageJson.spec.tsx
```diff
@@ -0,0 +1,25 @@
...
+          expect.stringMatching(/^(\^|git)/),
+        );
+      }
+    });
+  });
+
+  Object.keys(packageJson.dependencies).forEach((key) => {
+    it(`should have dependencies[${key}] version prefixed with a number`, () => {
+      expect((packageJson.dependencies as KeyValue)[key]).toEqual(
```
kumar303: here, too

Issue #86, opened 2019-01-30
`yarn dev` presents confusing output when fixing TypeScript errors


kumar303:
After https://github.com/mozilla/addons-code-manager/pull/56 landed, `yarn dev` will run both the server and `create-react-app` (CRA) process within the same stdout stream. This means CRA will no longer clear the stdout buffer when files change and that makes it confusing to fix a TypeScript error -- you still see the old error until it scrolls off the screen.

I'd like to bring back the default CRA experience where the stdout buffer gets cleared on file changes.

willdurand:
You are free to run the two commands in two separate consoles, we can document that.

I prefer the "one command to run everything" approach.

willdurand:
Sorry I did not want to close this issue.

PR #88, opened 2019-01-30
Use stmux for yarn dev to multiplex process output

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/86

The `yarn dev` command now multiplexes all developer-y things which preserves the pleasant experience of how `create-react-app` clears stdout as files change.

<details>
<summary>yarn dev</summary>

screenshot 2019-01-31 10 55 30

</details>

<hr>

For making use of wider screen area, there's a `yarn dev-2col` command, too.

<details>
<summary>yarn dev-2col</summary>

screenshot 2019-01-31 11 17 54

</details>


kumar303:
@willdurand @bobsilverberg Before I add documentation, what do you think? Is it too prescriptive? If so, we could just expose `yarn stmux` and document the way to build the commands yourself. I do like the idea of having a single way for any team member to run all the things.

package.json
```diff
@@ -38,31 +38,35 @@
...
     "prettier": "1.16.2",
     "pretty-quick": "^1.8.1",
+    "stmux": "^1.6.1",
     "supertest": "^3.4.2",
     "type-coverage": "^1.8.0"
   },
   "scripts": {
     "build": "react-scripts build",
-    "dev": "cat .env.common-local .env.dev > .env.local && yarn start-dev",
+    "build-dev-config": "cat .env.common-local .env.dev > .env.local",
```
willdurand: I suggest `build-local-dev-config`, it is lengthy but explicit.

package.json
```diff
@@ -38,31 +38,35 @@
...
     "prettier": "prettier --write '**'",
     "prettier-ci": "prettier --list-different '**' || (echo '\n\nThis failure means you did not run `yarn prettier-dev` before committing\n\n' && exit 1)",
     "prettier-dev": "pretty-quick --branch master",
+    "react-scripts-start": "PORT=3100 BROWSER=none react-scripts start",
     "serve": "node scripts/server.js",
+    "serve-dev": "NODE_ENV=development yarn serve",
     "stage": "cat .env.common-local .env.stage > .env.local && yarn start-dev",
     "start": "yarn build && NODE_ENV=production yarn serve",
-    "start-dev": "concurrently --kill-others 'PORT=3100 BROWSER=none react-scripts start' 'NODE_ENV=development yarn serve'",
     "start-local-dev": "cat .env.common-local .env.dev > .env.local && yarn start",
```
willdurand: You could reuse `build-local-dev-config` here now.

package.json
```diff
@@ -38,31 +38,35 @@
...
+    "stmux": "^1.6.1",
     "supertest": "^3.4.2",
     "type-coverage": "^1.8.0"
   },
   "scripts": {
     "build": "react-scripts build",
-    "dev": "cat .env.common-local .env.dev > .env.local && yarn start-dev",
+    "build-dev-config": "cat .env.common-local .env.dev > .env.local",
+    "dev": "yarn build-dev-config && yarn stmux [ 'yarn serve-dev' : 'yarn react-scripts-start' ]",
+    "dev-all": "yarn build-dev-config && yarn stmux [ 'yarn test' .. 'tsc -w' .. [ 'yarn react-scripts-start' : 'yarn serve-dev' ] ]",
```
willdurand: Can we stack all the panels instead?

```diff
diff --git a/package.json b/package.json
index 1692e77..de6b7db 100644
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "build": "react-scripts build",
     "build-dev-config": "cat .env.common-local .env.dev > .env.local",
     "dev": "yarn build-dev-config && yarn stmux [ 'yarn serve-dev' : 'yarn react-scripts-start' ]",
-    "dev-all": "yarn build-dev-config && yarn stmux [ 'yarn test' .. 'tsc -w' .. [ 'yarn react-scripts-start' : 'yarn serve-dev' ] ]",
+    "dev-all": "yarn build-dev-config && yarn stmux [ 'yarn test' : 'tsc -w' : 'yarn react-scripts-start' : 'yarn serve-dev' ]",
     "eject": "react-scripts eject",
     "prettier": "prettier --write '**'",
     "prettier-ci": "prettier --list-different '**' || (echo '\n\nThis failure means you did not run `yarn prettier-dev` before committing\n\n' && exit 1)",

```

screen shot 2019-01-31 at 10 49 41

scripts/server.js
```diff
@@ -39,7 +39,9 @@ app.listen(app.get('port'), () => {
       }).`,
       app.get('port'),
     );
-    console.log('\n  Press CTRL-C to stop');
+    console.log('');
+    console.log('  Press CTRL-a-k to stop');
+    console.log('  Press CTRL-a-? for help');
```
willdurand: That must be... me, but I cannot exit `stmux` easily 😞 

Note that it should still be `CTRL-C` to stop the server, because that's only one process in one stmux panel. I'd suggest adding a new sentence for exiting `stmux` (and keep CTRL-C).

willdurand: Ah ok, got it. So, `stmux` uses `CMD` and not `CTRL` on MacOS. `CMD-a-k` exits properly now, no need for `killall` anymore 😅

bobsilverberg: I cannot get any of these commands to work for me to exit everything. I'm on OS X. What is the exact keystroke combination I should be using?

bobsilverberg:
This is fantastic! Thanks @kumar303!

scripts/server.js
willdurand: While maintaining `CMD`, I hit `a` then `k`

kumar303: Huh. For me in iTerm on macOS, control works as expected. I press and hold control, press a, let go of control, and press k.

willdurand: Heh. I should have written "I can easily imagine issues with various **Mac**OS" in my review comment 😅

kumar303:
@willdurand @bobsilverberg thanks for taking an early look. I added docs and settled on two main commands. If in the future we don't end up using them or we all use custom commands 😅 then we could consider removing them.

* `yarn dev` now stacks everything in one column.
* `yarn dev-2col` puts `yarn test` in its own column since scrollback will be important there.

scripts/server.js
```diff
@@ -13,6 +13,10 @@ const app = createServer({
   rootPath: path.join(__dirname, '..'),
 });
 
+const printDivider = () => {
+  console.log('\n' + '#'.repeat(60) + '\n');
```
willdurand: `70` was used so that it was longer than the text inside.

kumar303: I just reduced the width of all text (in a separate commit). It won't be perfect on every screen but keeping it less wide makes it slightly more versatile.

kumar303:
I think snyk is up to date but it probably don't like running on a branch or something

PR #90, opened 2019-01-30
Add a test for a diff with multiple files

### LINKED ISSUE #89: Add a test for a diff with multiple files
This is a follow-up to #10. For some reason when developing that I wasn't able to get `parseDiff` to work with multiple files. Now, suddenly, I am unable to reproduce that problem. I don't know what I did wrong the first time, but I can now parse a diff with multiple files, so we should add a test for that to `/DiffView/index.spec.tsx`.

bobsilverberg:
Fixes #89


bobsilverberg:
Amazing, a PR that didn't require a single rebase to deal with merge conflicts! I'm merging it quick!

PR #95, opened 2019-01-31
Implement server code in TS

### LINKED ISSUE #71: Implement server code in TypeScript
We are about to ship some server logic in https://github.com/mozilla/addons-code-manager/pull/56/ but it's not type safe. This logic will be very important to get right so let's port it to TypeScript. This will probably require a new webpack entry point and other annoying things.

willdurand:
Fixes #71

---

I use `ts-node` in production, which is good enough. The only overhead seems to be on the initial startup time, then it is alright. It is simpler to use `ts-node` over anything else so that we can decouple our server code into different TS files and still be able to run the server from a unique entrypoint and without any bundler. That's the main motivation (and also the reason I mentioned `ncc` in the issue)

I use `ts-node-dev` in developement, which watches our server code and restarts it automatically on change. It does not restart the server if frontend code is changed, which is great. It also means we have full reload in dev, no matter which code is changed :tada:

The `.node-dev.json` file has been added to configure the following features in dev mode:

- clear the console on reload–that will be a great addition combined to #88
- do not use the OS/term notification system: this is enabled by default but I don't really like having popups every time the server code is reloaded

We need to compile the server code for `commonjs`, hence the compilation options in the npm scripts.

I used the `as XXX` trick to mute TSC in the test files because I did not want to write full `env` objects, but maybe I should?

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,9 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@types/express": "4.16.1",
+    "@types/http-proxy": "1.16.2",
+    "@types/http-proxy-middleware": "0.19.2",
     "@types/jest": "23.3.13",
     "@types/loglevel": "1.5.3",
     "@types/node": "10.12.21",
@@ -28,8 +31,10 @@
     "redux": "4.0.1",
     "redux-devtools-extension": "2.13.7",
     "redux-logger": "3.0.6",
+    "ts-node": "8.0.2",
     "typesafe-actions": "3.0.0",
     "typescript": "3.2.4",
+    "universal-cookie": "3.0.7",
     "universal-cookie-express": "3.0.7"
   },
   "devDependencies": {
@@ -53,6 +58,7 @@
     "pretty-quick": "^1.8.1",
     "stmux": "^1.6.1",
     "supertest": "^3.4.2",
+    "ts-node-dev": "^1.0.0-pre.32",
     "type-coverage": "^1.8.0"
   },
   "scripts": {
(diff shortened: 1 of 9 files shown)


src/server/index.tsx
```diff
@@ -1,14 +1,30 @@
...
+} from 'universal-cookie-express';
+
+export type ServerEnvVars = {
+  NODE_ENV: string;
+  PORT: number;
+  REACT_APP_API_HOST: string;
+  REACT_APP_AUTHENTICATION_COOKIE: string;
+  REACT_APP_AUTH_TOKEN_PLACEHOLDER: string;
+  REACT_APP_CRA_PORT: number;
+  REACT_APP_USE_INSECURE_PROXY: boolean;
```
kumar303: Are you sure that this is a `boolean` and not a `string`? I can't find any docs to suggest that this is the case.

scripts/server.tsx
```diff
@@ -1,15 +1,16 @@
...
 
 // Load environment configuration.
 process.env.NODE_ENV = process.env.NODE_ENV || 'development';
 require('react-scripts/config/env');
 
 const app = createServer({
-  env: process.env,
+  // This is risky but also expected because `process.env` is only defined at
+  // runtime.
+  env: (process.env as any) as ServerEnvVars,
```
kumar303: Hmm, I don't know what the right way to do this is. Well, I can't think of a better way than what you have.

package.json
```diff
@@ -56,18 +60,19 @@
     "prettier": "prettier --write '**'",
     "prettier-ci": "prettier --list-different '**' || (echo '\n\nThis failure means you did not run `yarn prettier-dev` before committing\n\n' && exit 1)",
     "prettier-dev": "pretty-quick --branch master",
-    "serve": "node scripts/server.js",
+    "serve": "NODE_ENV=production ts-node -O '{\"module\": \"commonjs\", \"typeRoots\": [\"./src/@types\"]}' scripts/server.tsx",
```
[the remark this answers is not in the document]
willdurand: `ts-node-dev` uses `ts-node`. I was thinking about declaring it as a dependency (eslint will complain soon) but I was not sure to know how to keep `ts-node` and `ts-node-dev` in sync.

src/server/index.tsx
```diff
@@ -1,16 +1,32 @@
 /* eslint @typescript-eslint/no-var-requires: 0, no-console: 0,
    global-require: 0, amo/only-tsx-files: 0 */
```
kumar303: You can remove `amo/only-tsx-files` and probably `global-require`, too.

PR #96, opened 2019-01-31
Add loglevel lib

### LINKED ISSUE #85: Adding a logging abstraction
We closed https://github.com/mozilla/addons-code-manager/issues/13 because we're not doing SSR but we still need to add something that can be called like `log.info()` so that lint rules such as `amo/only-log-strings` catch the right mistakes. Also, I think we'll want to disable debug logging in production.

willdurand:
Fixes #85

### DIFF
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -1,3 +1,5 @@
+import log from 'loglevel';
+
 import { ApiState } from '../reducers/api';
 
 export enum HttpMethod {
@@ -56,8 +58,7 @@ export const callApi = async ({
 
     return await response.json();
   } catch (error) {
-    // eslint-disable-next-line no-console
-    console.debug(error);
+    log.debug(error);
 
     return {
       error,
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -58,7 +58,7 @@ export const callApi = async ({
 
     return await response.json();
   } catch (error) {
-    log.debug(error);
+    log.debug('Error caught in callApi():', error);
 
     return {
       error,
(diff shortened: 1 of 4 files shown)


src/index.tsx
```diff
@@ -2,12 +2,17 @@ import React from 'react';
...
 import { Provider } from 'react-redux';
+import log from 'loglevel';
 
 import './styles.scss';
 import App from './components/App';
 import configureStore from './configureStore';
 import { actions as apiActions } from './reducers/api';
 
+if (process.env.NODE_ENV === 'production') {
+  log.setLevel(log.levels.INFO);
```
kumar303: This should probably be `log.setLevel(..., false)` to disable persistence. Otherwise, the docs say that the lib will try to store the level in cookies or localStorage. I don't think we need that.

src/index.tsx
```diff
@@ -2,12 +2,17 @@ import React from 'react';
...
+}
```
kumar303: Can you add an `else` branch that configures debug logging? The docs don't say what the default level will be so I think it's better to make it explicit.

src/api/index.tsx
```diff
@@ -51,7 +53,7 @@ export const callApi = async ({
 
     return await response.json();
   } catch (error) {
-    console.error(error);
+    log.error(error);
```
kumar303: This can probably be `log.debug(error)` (maybe with a string prefix explaining that `callApi` caught the error).

src/index.tsx
willdurand: oh, the level gets persisted! Is that a problem though?

willdurand:
Updated!

PR #101, opened 2019-01-31
Create a users reducer

### LINKED ISSUE #100: Create a users reducer
I'm working on #97 and it's getting pretty big, so I thought I'd split it up by creating the users reducer in a separate patch. That's this.

bobsilverberg:
Fixes #100


src/reducers/users.spec.tsx
```diff
@@ -0,0 +1,116 @@
+import reducer, {
+  actions,
+  getCurrentUser,
+  getUserById,
+  getUserByName,
+  initialState,
+} from './users';
+import { fakeUser } from '../helpers';
```
kumar303: I know that we are co-locating tests, but what about putting any test specific module in `src/__tests__`? That seems to be a standard location. Just keep in mind that `src/setupTests.tsx` is a magical path so it can't be moved, IIRC.

src/reducers/users.tsx
```diff
@@ -0,0 +1,109 @@
...
+};
+
+export const actions = {
+  loadCurrentUser: createAction('LOAD_CURRENT_USER', (resolve) => {
+    return (payload: { user: User }) => resolve(payload);
+  }),
+  loadUser: createAction('LOAD_USER', (resolve) => {
+    return (payload: { user: User }) => resolve(payload);
+  }),
+  logOut: createAction('LOG_OUT', (resolve) => {
```
kumar303: I believe this can be simplified to `logOut: createAction('LOG_OUT')`

src/reducers/users.tsx
```diff
@@ -0,0 +1,109 @@
...
+    return () => resolve({});
+  }),
+};
+
+export type UsersState = {
+  currentUserId: UserId | null;
+  byId: { [id: number]: User };
+  byName: { [name: string]: UserId };
```
kumar303: I agree with @willdurand that we should only focus on log in / out for this patch. We probably won't ever need to load multiple users. I think you can remove `byName` and all related actions. We can always add them back in later if we need them.

src/reducers/users.tsx
```diff
@@ -0,0 +1,109 @@
...
+  name: string;
+  email?: string;
+  permissions?: Array<string>;
+};
+
+export const actions = {
+  loadCurrentUser: createAction('LOAD_CURRENT_USER', (resolve) => {
+    return (payload: { user: User }) => resolve(payload);
+  }),
+  loadUser: createAction('LOAD_USER', (resolve) => {
```
kumar303: In regards to my last comment, I can't think of why we would need this action so I suggest deleting it for now.

src/reducers/users.tsx
```diff
@@ -0,0 +1,109 @@
...
+  byName: {},
+};
+
+type LoadUserIntoStateParams = {
+  isCurrentUser?: boolean;
+  state: UsersState;
+  user: User;
+};
+
+export const loadUserIntoState = ({
```
kumar303: If you agree with my suggestion about removing the `LOAD_USER` action then you won't need this helper function and the reducer can become a lot simpler.

src/reducers/users.tsx
```diff
@@ -0,0 +1,109 @@
+import invariant from 'invariant';
+import { Reducer } from 'redux';
+import { ActionType, createAction, getType } from 'typesafe-actions';
+
+type UserId = number;
+
+export type User = {
```
kumar303: This is only a subset of the [user response](https://addons-server.readthedocs.io/en/latest/topics/api/accounts.html#account) which is fine but I think you should follow our typical pattern for this where `User` is the internal representation and `ExternalUser` is a representation of what the API will respond with. The action functions will except `ExternalUser` and the reducer would store it as `createInternalUser(payload.user)`.

bobsilverberg:
Thanks for the reviews @willdurand and @kumar303. I had originally thought about only storing the current user, but then I thought about the fact that every time a user performs a review, they will be reviewing code from *another* user. And they might want to look at other reviews for the same user, so I thought that being able to accommodate the existence of multiple users in the app would be a good thing.

I'm fine to delete all the extra code I wrote, but before doing so I just wanted to double check that you're sure we don't want to accommodate that at this point.

willdurand:
> I'm fine to delete all the extra code I wrote, but before doing so I just wanted to double check that you're sure we don't want to accommodate that at this point.

I am sure we don't need this right now :)

src/reducers/users.spec.tsx
bobsilverberg: When I moved the helper to `__tests__`, Jest now tries to run the file and I get the error: `Your test suite must contain at least one test`. I tried adding a `testPathIgnorePatterns` to `package.json`, but `create-react-app` complains that that is not allowed without ejecting.

What do you suggest?

src/reducers/users.tsx
bobsilverberg: I was trying to keep it simple by not even referencing any of the data we don't need from the API, but sure, I can go the `ExternalUser` approach.

bobsilverberg:
I have addressed the issues. The only outstanding issue is that `Jest` is running `src/__tests__/helpers.tsx` as a test, and is therefore failing. I'm haven't been able to figure out a way to get it to ignore that file.

willdurand:
> I have addressed the issues. The only outstanding issue is that `Jest` is running `src/__tests__/helpers.tsx` as a test, and is therefore failing. I'm haven't been able to figure out a way to get it to ignore that file.

Can we move this file to `src/test-helpers.tsx` maybe? We can't change Jest's config so helper functions cannot be in a test file, unless there are other tests in it. If we really want to go this way, we could maybe call `jest.skip()` in the helpers file, but that's not super... nice.

src/reducers/users.spec.tsx
kumar303: ok, then I suggest keeping the file where it is but renaming it to `testHelpers.js`

kumar303: ... or even `test-helpers.tsx` ! I just saw that.

src/reducers/users.spec.tsx
```diff
@@ -0,0 +1,62 @@
...
+  initialState,
+} from './users';
+import { fakeUser } from '../test-helpers';
+
+describe(__filename, () => {
+  describe('reducer', () => {
+    it('initializes the state', () => {
+      const state = reducer(undefined, { type: 'SOME_ACTION' } as any);
+
+      expect(state).toEqual(initialState);
```
kumar303: fwiw, I don't think these types of tests are very helpful. We have a strongly recognized pattern for creating reducers and they always have a default case that returns the state. I'm. pretty sure we'd catch a mistake there in code review.

PR #105, opened 2019-02-01
Configure Sentry

### LINKED ISSUE #26: Set up Sentry for errors
We want to hook up https://sentry.prod.mozaws.net/operations/ for error reporting. We'll need some help from ops to set up a destination (see #146) and then we'll need to configure the app to report errors to Sentry with RavenJS (this issue).

willdurand:
Fixes #26

---

This is an example of what we need. Sentry project at: https: //sentry.prod.mozaws.net/operations/addons-code-manager-test/

Once we have the hosted environments, I suppose we would just have to create the Sentry projects and add the DSNs to each `.env` file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "loglevel": "1.6.1",
     "node-sass": "4.11.0",
     "prop-types": "15.6.2",
+    "raven-js": "3.27.0",
     "react": "16.8.0",
     "react-bootstrap": "1.0.0-beta.5",
     "react-diff-view": "2.0.0-alpha.1",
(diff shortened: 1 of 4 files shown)


src/index.tsx
```diff
@@ -16,6 +17,10 @@ if (process.env.NODE_ENV === 'production') {
   log.setLevel(log.levels.DEBUG, false);
 }
 
+if (process.env.REACT_APP_SENTRY_DSN) {
+  Raven.config(process.env.REACT_APP_SENTRY_DSN).install();
```
kumar303: This needs to configure `{logger: 'client-js'}` so we can tell it apart from server logs

https://github.com/mozilla/addons-frontend/blob/533c052acc959664baa817c4b69aa76d83b5dc34/src/core/client/base.js#L38

package.json
```diff
@@ -63,7 +64,7 @@
   },
   "scripts": {
     "build": "react-scripts build",
-    "build-local-dev-config": "cat .env.common-local .env.dev > .env.local",
+    "build-local-dev-config": "cat .env.dev .env.common-local > .env.local",
```
kumar303: Why did these get swapped? Do we need to swap the others?

src/index.tsx
willdurand: Which server logs? Because, we don't log anything on the server.

package.json
willdurand: Mmmh, probably. I think the `.env.common-local` file should be appended so that it overrides other variables. For example, I want the `test` sentry project locally, rather the the sentry project for the hosted dev.

src/index.tsx
kumar303: Yes but we may log things on the server in the future now that we have to run the token injection.

willdurand: Ah, ok. My take would be to leave things as is for now and to configure a `{ logger: 'server' }` when we'll need one.

kumar303: I still think having `logger` defined everywhere (client and server) makes the filtering easier but if you want to add that later, it's fine.

Issue #12, opened 2019-01-15
Add a lib for handling side effects (redux-thunk, redux-loop, ???)


willdurand:
We want to use [redux-thunk](https://github.com/reduxjs/redux-thunk) to compose actions. redux-thunk is a redux middleware that allows to dispatch functions instead of plain old actions. Such functions get the `dispatch` function as first argument and then, one can do whatever should be done.

kumar303:
> redux-loop/redux-loop#131 😞

I'd have to try it out but if the Redux developer tool currently does *not* trigger redux-loop side effects, that might be a good thing! Triggering side effects in the developer tool was the exact problem I mentioned with sagas: you can't step through your application state one action at a time.

kumar303:
Let's say you were using an app with sagas. The developer tool recorded these actions:

* `SHOW_USER_PAGE`
* `FETCH_USER` -> this triggers the saga `fetchUserAndFriends()` which dispatches more actions:
  * `LOAD_USER`
  * `FETCH_FRIENDS`
  * `LOAD_USER_FRIENDS`

Let's also say that `FETCH_USER` renders a loading screen and you wanted to use time travel to jump back to that action and use hot reloading to develop the UX. As soon as you time travel to `FETCH_USER`, it re-runs the saga, makes the API request again, and dispatches more actions. 

The Redux developer tool *does* account for this problem and will sort of ignore the side effect actions. However, it still fills up the timeline with side effect actions which makes it very hard to reason about state changes. 

You may be able to fulfill the use case I described above of working on the loading screen but you would have a very hard time understanding the user flow that caused a bug in, say, the final `LOAD_USER_FRIENDS` action.

I'd like to avoid this problem in a side effect library. I don't know if it's fully avoidable, though.

willdurand:
**Disclaimer:** I haven't used the redux-loop lib.

The thing I already dislike about react-loop is that now `reducers` are responsible for two things:

1. ensuring data consistency and integrity
2. initiating API calls (and dealing with errors)

That will make reducers super complex. We found errors in our addons-frontend reducers while keeping them as "pure" as possible, so while I can see why redux-loop exists, I have mixed feelings.

We wrote sagas with more than one API call in addons-frontend, I wonder how we would do that with redux-loop, in an elegant way I mean.

kumar303:
Yeah, I have similar reservations. I'd like to learn more about redux-loop. Trying to write a prototype with multiple API calls would be a good test of it. 

[...]

PR #136, opened 2019-02-04
Access the user profile from the redux store

### LINKED ISSUE #97: Move the profile data into the Redux store
Currently the `profile` data is stored in the component state for `App`, but really this should be in the Redux store. This issue is to move it there, which will include creating a reducer for it.

bobsilverberg:
Fixes #97 

I'm having trouble getting the test for the logOut button working, to confirm that it dispatches `userActions.logOut`. I believe my code should be working, but jest reports that the mock was not called. I wonder if the problem is that I am using the `dispatch` method on the store object to do a real dispatch in `storeWithUser` and then after that trying to replace it with a mock via `jest.spyOn`? I need the original dispatch though, or the button won't appear.

Any advice anyone has would be appreciated.


src/components/Navbar/index.spec.tsx
```diff
@@ -1,69 +1,104 @@
...
-      const root = render({ onLogOut: onLogOutMock, profile: createProfile() });
+    it('calls logOutFromServer when clicked', () => {
+      const logOutFromServerMock = jest.fn();
+      const root = render({
+        _logOutFromServer: logOutFromServerMock,
+        store: storeWithUser(),
+      });
+
+      root.find(`.${styles.logOut}`).simulate('click');
+      expect(logOutFromServerMock).toHaveBeenCalled();
```
kumar303: I haven't tried running the code but `this.logOut()` on the class is async so that's probably the problem. We rarely had to deal with this on `addons-frontend` because sagas were such a testable abstraction. 

Until we have a testable abstraction (which hopefully will happen in https://github.com/mozilla/addons-code-manager/issues/12), you could split it into two tests. One that verifies `this.logOut()` gets called and another that calls `root.instance().logOut()` directly so that it can wait for the promise before making assertions.

src/components/App/index.tsx
```diff
@@ -51,47 +40,32 @@ class AppBase extends React.Component<Props, State> {
...
+      const response = (await api.callApi({
         apiState,
         endpoint: '/accounts/profile/',
-      })) as State['profile'];
+      })) as ExternalUser;
 
-      if (profile && profile.name) {
-        // eslint-disable-next-line react/no-did-update-set-state
-        this.setState({ profile });
+      if (response && response.name) {
```
willdurand: I suppose we should look for `response.error` and deal with that too.

src/components/Navbar/index.spec.tsx
willdurand: Can't we `await` the call above? I think we can do things like this because Jest definitely supports async code.

bobsilverberg: That works @willdurand, thanks!

src/components/App/index.tsx
bobsilverberg: Well, I think this is temporary code, so I'm not sure we need to make it perfect at this point. Also, What we're doing here is setting the user to "logged in" if the response contains a `name`. Wouldn't `name` be missing on an error, and therefore our code would work and not log the user in? We'd obviously want some messaging around a failed log in like that, but I'm not sure we should add that now, when I imagine this changing in the future.

willdurand: That's correct, and I also think we'll iterate over this code again (i.e. no need to apply changes here)

I was thinking that maybe it would make more sense to test `response.error` than `response.name` but 🤷‍♂️

bobsilverberg: Ah, I didn't realize that's what you were suggesting. Is it okay to just leave it as is for now?

willdurand: yep. (and no, I was not suggesting anything initially, I was thinking out loud)

src/components/Navbar/index.spec.tsx
```diff
@@ -1,69 +1,104 @@
...
+    });
+
+    it('dispatches userActions.logOut when clicked', async () => {
+      const store = storeWithUser();
+      const dispatch = jest.spyOn(store, 'dispatch');
+
+      const root = render({ store });
+
+      await root.find(`.${styles.logOut}`).simulate('click');
```
kumar303: Enzyme does not provide any support for async event handlers so this probably isn't doing what you think it is

src/components/Navbar/index.spec.tsx
```diff
@@ -1,69 +1,104 @@
...
+    it('calls logOutFromServer when clicked', () => {
+      const logOutFromServerMock = jest.fn();
+      const root = render({
+        _logOutFromServer: logOutFromServerMock,
+        store: storeWithUser(),
+      });
 
       root.find(`.${styles.logOut}`).simulate('click');
-      expect(onLogOutMock).toHaveBeenCalled();
+      expect(logOutFromServerMock).toHaveBeenCalled();
```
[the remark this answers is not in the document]
bobsilverberg: Yes, it's unfortunate that the trick from that SO article doesn't work. FTR we do have `@types/enzyme` in our project.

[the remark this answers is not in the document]
kumar303: Ah, yes, we do have `@types/enzyme`

src/components/Navbar/index.tsx
```diff
@@ -1,19 +1,44 @@
...
+  logOut = async () => {
+    const { _logOutFromServer, apiState, dispatch } = this.props;
+
+    await _logOutFromServer(apiState);
+    dispatch(userActions.logOut());
+  };
+
   render() {
-    const { onLogOut, profile } = this.props;
+    const { profile } = this.props;
```
kumar303: It would be nice if this component had a loading state while it fetches the user profile so I filed https://github.com/mozilla/addons-code-manager/issues/144

willdurand: I submitted #143 for that yesterday (proposal)

PR #137, opened 2019-02-04
Set up the test suite for shallowUntilTarget rendering

### LINKED ISSUE #15: Set up the test suite for shallowUntilTarget rendering
Once we have https://github.com/mozilla/addons-code-manager/issues/8 , set the test suite up for shallow rendering. Since we'll be using Redux, we will need a strategy like [shallowUntilTarget](https://github.com/mozilla/addons-frontend/search?q=shallowUntilTarget&unscoped_q=shallowUntilTarget).

Here is a write-up of what worked well in the addons-frontend test suite: https://hacks.mozilla.org/2018/04/testing-strategies-for-react-and-redux/

bobsilverberg:
Fixes #15 

I had some trouble with types, so there will probably be some comments and suggestions around those. In addition to that I changed the signature so it accepts an object instead of individual arguments. I think it's cleaner, and that's what we tend to do most of the time, but it is a bit more of a hassle to call.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -13,7 +13,7 @@
     // https://github.com/typescript-eslint/typescript-eslint/blob/5b0b3d9edbcb3ab588a34c431037d9deece30824/packages/eslint-plugin/docs/rules/indent.md#options
     "@typescript-eslint/indent": ["error", 2],
     "import/no-extraneous-dependencies": ["error", {
-        "devDependencies": ["**/*.spec.*", "src/setupTests.tsx"]
+        "devDependencies": ["**/*.spec.*", "src/setupTests.tsx", "src/test-helpers.tsx"]
       }
     ],
     // We use `.tsx` instead of `.jsx` for files with JSX inside.
--- a/package.json
+++ b/package.json
@@ -3,7 +3,6 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@types/common-tags": "1.8.0",
     "@types/express": "4.16.1",
     "@types/http-proxy": "1.16.2",
     "@types/http-proxy-middleware": "0.19.2",
@@ -17,7 +16,6 @@
     "@types/react-router-dom": "4.3.1",
     "@types/redux-logger": "3.0.6",
     "bootstrap": "4.2.1",
-    "common-tags": "1.8.0",
     "express": "4.16.4",
     "history": "4.7.2",
     "http-proxy-middleware": "0.19.1",
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@types/common-tags": "1.8.0",
     "@types/express": "4.16.1",
     "@types/http-proxy": "1.16.2",
     "@types/http-proxy-middleware": "0.19.2",
@@ -16,6 +17,7 @@
     "@types/react-router-dom": "4.3.1",
     "@types/redux-logger": "3.0.6",
     "bootstrap": "4.2.1",
+    "common-tags": "1.8.0",
     "express": "4.16.4",
     "history": "4.7.2",
     "http-proxy-middleware": "0.19.1",
--- a/src/components/App/index.spec.tsx
+++ b/src/components/App/index.spec.tsx
@@ -1,4 +1,3 @@
-import { shallow } from 'enzyme';
 import React from 'react';
 import { Container } from 'react-bootstrap';
 import { Store } from 'redux';
@@ -8,9 +7,13 @@ import styles from './styles.module.scss';
 import configureStore from '../../configureStore';
 import { actions as apiActions } from '../../reducers/api';
 import { actions as userActions } from '../../reducers/users';
-import { createContextWithFakeRouter, fakeUser } from '../../test-helpers';
+import {
+  createContextWithFakeRouter,
+  fakeUser,
+  shallowUntilTarget,
+} from '../../test-helpers';
 
-import App from '.';
+import App, { AppBase } from '.';
 
 describe(__filename, () => {
   type RenderParams = {
@@ -31,15 +34,11 @@ describe(__filename, () => {
       },
     };
 
-    // TODO: Use shallowUntilTarget()
-    // https://github.com/mozilla/addons-code-manager/issues/15
-    const root = shallow(<App authToken={authToken} />, context)
-      // withRouter HOC
-      .shallow(context)
-      // connect HOC
-      .shallow(context)
-      // base component
-      .shallow();
+    const root = shallowUntilTarget({
+      componentInstance: <App authToken={authToken} />,
+      targetComponent: AppBase,
+      options: { shallowOptions: { context } },
+    });
 
     return root;
   };
--- a/src/components/App/index.spec.tsx
+++ b/src/components/App/index.spec.tsx
@@ -27,11 +27,8 @@ describe(__filename, () => {
   }: RenderParams = {}) => {
     const contextWithRouter = createContextWithFakeRouter();
     const context = {
-      ...contextWithRouter,
-      context: {
-        ...contextWithRouter.context,
-        store,
-      },
+      ...contextWithRouter.context,
+      store,
     };
 
     const root = shallowUntilTarget({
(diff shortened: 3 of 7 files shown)


.eslintrc
```diff
@@ -13,7 +13,7 @@
     // https://github.com/typescript-eslint/typescript-eslint/blob/5b0b3d9edbcb3ab588a34c431037d9deece30824/packages/eslint-plugin/docs/rules/indent.md#options
     "@typescript-eslint/indent": ["error", 2],
     "import/no-extraneous-dependencies": ["error", {
-        "devDependencies": ["**/*.spec.*", "src/setupTests.tsx"]
+        "devDependencies": ["**/*.spec.*", "src/setupTests.tsx", "src/test-helpers.tsx"]
```
willdurand: Why is this needed?

src/test-helpers.spec.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+import { shallowUntilTarget } from './test-helpers';
+
+describe(__filename, () => {
+  describe('shallowUntilTarget', () => {
+    function ExampleBase() {
+      return <div>Example component</div>;
+    }
+
+    function wrapper() {
```
willdurand: We could probably switch to arrow functions in this file.

src/test-helpers.spec.tsx
```diff
@@ -0,0 +1,138 @@
+/* eslint-disable react/no-multi-comp */
```
willdurand: Why is this rule disabled?

.eslintrc
bobsilverberg: Without it I get "2:1   error    'enzyme' should be listed in the project's dependencies, not devDependencies" on `/src/test-helpers.tsx`. I assume this is  because `ESLint` thinks this file is a prod file and not a test-only file, and this fixes that.

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@types/common-tags": "1.8.0",
```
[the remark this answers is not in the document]
bobsilverberg: Sure, it's just used for an error message but I can probably do without it.

bobsilverberg:
Thanks for the review @willdurand. I've addressed the comments and pushed a new commit.

src/components/App/index.spec.tsx
```diff
@@ -19,11 +19,11 @@ describe(__filename, () => {
...
   }: RenderParams = {}) => {
-    // TODO: Use shallowUntilTarget()
-    // https://github.com/mozilla/addons-code-manager/issues/15
-    const root = shallow(<App authToken={authToken} />, {
-      context: { store },
-    }).shallow();
+    const root = shallowUntilTarget({
+      componentInstance: <App authToken={authToken} />,
+      targetComponent: AppBase,
+      options: { shallowOptions: { context: { store } } },
```
willdurand: This is interesting (not related to your patch): we set the `store` to the `context` whereas in addons-frontend, it was injected as a prop, which caused issues with the react-redux v6.

src/test-helpers.tsx
```diff
@@ -25,3 +27,54 @@ export const fakeUser: ExternalUser = Object.freeze({
...
+ *
+ * The `componentInstance` parameter is a React component instance.
+ * Example: <MyComponent {...props} />
+ *
+ * The `targetComponent` parameter is the React class (or function) that
+ * you want to retrieve from the component tree.
+ */
+type ShallowUntilTargetParams = {
+  componentInstance: React.ReactElement<any>;
+  targetComponent: any;
```
willdurand: Can we do better than `any` here?

src/components/App/index.spec.tsx
bobsilverberg: If this works, does that mean we aren't blocked on upgrading to react-redux v6?

willdurand: We're still blocked by Enzyme 😞

src/test-helpers.tsx
bobsilverberg: I tried a bunch of different things and couldn't seem to figure out what would work. I've now found that `Function` works. I'm not sure if that is 100% accurate, but it's definitely better than `any`.

willdurand: mmh, looking at some docs, I'd suggest `JSXElementConstructor<any>` (`any` refers to the props).

willdurand:
This needs to be rebased now :)

bobsilverberg:
> This needs to be rebased now :)

Hooray! 🎉

src/components/App/index.spec.tsx
```diff
@@ -23,22 +25,15 @@ describe(__filename, () => {
   }: RenderParams = {}) => {
     const contextWithRouter = createContextWithFakeRouter();
     const context = {
-      ...contextWithRouter,
```
bobsilverberg: @willdurand please take a look at this. I didn't understand why you were creating an object called `context` that only contains a single property called `context`, so I removed that. It could be that there might be something other than the `context` property in the object returned by `createContextWithFakeRouter()` at some point, but for now there is not, so the extra wrapping didn't really make sense to me.

willdurand: `createContextWithFakeRouter()` is supposed to return:

```
{
  context: {},
  childContextTypes: {}
}
```

bobsilverberg: Thanks @willdurand, I've put it back the way it was.

src/components/App/index.spec.tsx
```diff
@@ -30,15 +32,11 @@ describe(__filename, () => {
...
-    const root = shallow(<App authToken={authToken} />, context)
-      // withRouter HOC
-      .shallow(context)
-      // connect HOC
-      .shallow(context)
-      // base component
-      .shallow();
+    const root = shallowUntilTarget({
+      componentInstance: <App authToken={authToken} />,
+      targetComponent: AppBase,
```
willdurand: I would prefer to keep the old signature, we don't need object properties for the instance and the target component.

src/test-helpers.spec.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+describe(__filename, () => {
+  describe('shallowUntilTarget', () => {
+    const ExampleBase = () => {
+      return <div>Example component</div>;
+    };
+
+    const wrapper = () => {
+      return function Wrapper(WrappedComponent: any) {
+        return function InnerWrapper(props: object) {
```
willdurand: arrow function?

src/test-helpers.tsx
```diff
@@ -80,3 +81,54 @@ export const createContextWithFakeRouter = ({
...
+    maxTries?: number;
+    shallowOptions?: object;
+    _shallow?: typeof shallow;
+  };
+};
+
+export const shallowUntilTarget = ({
+  componentInstance,
+  targetComponent,
+  options: { maxTries = 10, shallowOptions = {}, _shallow = shallow } = {},
```
willdurand: We don't want to assign an empty object here because that does not make much sense. It will fail if we don't specify some arguments, so I'd like to see the old signature (only the 3rd parameter is an object).

src/test-helpers.spec.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+        render() {
+          return <div>example of using lifecycle methods</div>;
+        }
+      }
+
+      const Example = compose(wrapper())(LifecyleExample);
+
+      const root = shallowUntilTarget(<Example />, LifecyleExample, {
+        shallowOptions: { lifecycleExperimental: true },
```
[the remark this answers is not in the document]
bobsilverberg: You're right, that test is no longer valid. I changed it to use the `disableLifecycleMethods` option and I believe it is now valid once again.

src/test-helpers.spec.tsx
```diff
@@ -0,0 +1,118 @@
...
+        _shallow: shallowStub,
+      });
+
+      expect(shallowStub).toHaveBeenCalledWith(instance, shallowOptions);
+    });
+
+    it('lets you pass options to the final shallow()', () => {
+      const componentDidUpdate = jest.fn();
+
+      // eslint-disable-next-line react/no-multi-comp
```
kumar303: We should disable `react/no-multi-comp` in all test files. It's perfectly reasonable to put multiple components in test files.

willdurand: I just filed: https://github.com/mozilla/addons-code-manager/issues/154

PR #141, opened 2019-02-05
Add ESLint reporter for jest

### LINKED ISSUE #140: Run lint checks along with unit tests
I really liked how in addons-frontend, when you ran `yarn test` you also saw any ESLint errors. Let's set that up for code-manager as well.

bobsilverberg:
Fixes #140 

This is just an initial crack at getting this to work. I simply brought a couple of files over from addons-frontend, and added a new yarn command called `test-lint`, which runs the Jest tests and also the ESLint checks.

It's a bit hacky, but the only thing I could get to work, which I picked up from https://github.com/facebook/create-react-app/issues/2474#issuecomment-321837707.

The Jest reporter and utils files are still `js` files, as opposed to `tsx` files, and I guess we'll want to rewrite them as `tsx` files, as long as Jest allows that.

Just submitting this for initial thoughts and feedback.

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "stmux": "stmux -nM -t addons-code-manager -e 'error,!0 errors'",
     "storybook": "start-storybook -p 9001 -c stories/setup",
     "storybook-smoke-test": "yarn storybook --smoke-test",
-    "test": "react-scripts test",
+    "test": "react-scripts test --reporters=default --reporters=`pwd`/src/jest-reporters/eslint-check.js",
     "test-ci": "yarn test --coverage && codecov",
     "type-coverage": "type-coverage",
     "typecheck": "tsc"
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "start-local-dev": "yarn build-local-dev-config && yarn start",
     "start-local-prod": "cat .env.common-local > .env.local && yarn start",
     "start-local-stage": "cat .env.common-local .env.stage > .env.local && yarn start",
-    "stmux": "stmux -nM -t addons-code-manager -e 'error,!0 errors'",
+    "stmux": "stmux -nM -t addons-code-manager -e 'error,!0 errors,!no errors'",
     "storybook": "start-storybook -p 9001 -c stories/setup",
     "storybook-smoke-test": "yarn storybook --smoke-test",
     "test": "react-scripts test --reporters=default --reporters=`pwd`/src/jest-reporters/eslint-check.js",
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "stmux": "stmux -nM -t addons-code-manager -e 'error,!0 errors,!no errors'",
     "storybook": "start-storybook -p 9001 -c stories/setup",
     "storybook-smoke-test": "yarn storybook --smoke-test",
-    "test": "react-scripts test --reporters=default --reporters=`pwd`/src/jest-reporters/eslint-check.js",
+    "test": "react-scripts test --reporters=default --reporters='<rootDir>/src/jest-reporters/eslint-check.js'",
     "test-ci": "yarn test --coverage && codecov",
     "type-coverage": "type-coverage",
     "typecheck": "tsc"
--- a/src/jest-reporters/eslint-check.js
+++ b/src/jest-reporters/eslint-check.js
@@ -1,0 +1,61 @@
+/* eslint-disable no-console */
+/* eslint-disable amo/only-tsx-files */
+/* eslint-disable @typescript-eslint/no-var-requires */
+// eslint-disable-next-line import/no-extraneous-dependencies
+const { CLIEngine } = require('eslint');
+
+const { getChangedFiles } = require('./utils');
+
+const NO_ESLINT_ENV_VAR = 'NO_ESLINT';
+
+class EslintCheckReporter {
+  constructor() {
+    this.eslint = new CLIEngine();
+    this.eslintOutput = null;
+  }
+
+  isDisabled() {
+    return process.env[NO_ESLINT_ENV_VAR] === '1';
+  }
+
+  async onRunStart() {
+    if (this.isDisabled()) {
+      return;
+    }
+
+    const files = await getChangedFiles();
+
+    if (!files) {
+      throw new Error(`Failed to retrieve files in the eslint check reporter.`);
+    }
+
+    const report = this.eslint.executeOnFiles(files);
+
+    if (report.errorCount === 0 && report.warningCount === 0) {
+      // All good.
+      this.eslintOutput = null;
+    } else {
+      this.eslintOutput = CLIEngine.getFormatter()(report.results);
+    }
+  }
+
+  getLastError() {
+    if (this.isDisabled()) {
+      return undefined;
+    }
+
+    console.log('');
+    if (this.eslintOutput) {
+      console.log(this.eslintOutput);
+      console.log(
+        `Set ${NO_ESLINT_ENV_VAR}=1 in the environment to disable eslint checks`,
+      );
+      return new Error('eslint errors');
+    }
+    console.log('Eslint: no errors 💄 ✨');
+
+    return undefined;
+  }
+}
+
+module.exports = EslintCheckReporter;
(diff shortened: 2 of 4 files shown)


kumar303:
Do you think we still need this after getting `yarn dev` and `yarn dev-2col` with a separate window showing lint output? For me, this solved the problem of not seeing lint errors while running tests so I think we need a jest reporter. 

Also, I think this jest reporter is problematic for a couple reasons. First, the timing of when the linting runs is out of sync (at least it was on addons-frontend). Sometimes you'll see lint errors for something you just fixed because it runs too early (I'm not sure why). Secondly, when you are preparing a big patch because you don't want to commit just yet, it starts to slow down the test suite since many files get linted each time you touch a file. This is especially apparent if you're using a jest filter to run a single test. This kind of use case is important because the jest filter allows you to get very quick test feedback.

willdurand:
There is no (built-in) watcher for ESLint though, so another window won't help much.

kumar303:
Doesn't the `yarn react-scripts-start` command show linting? The [docs](https://facebook.github.io/create-react-app/docs/setting-up-your-editor#displaying-lint-output-in-the-editor) suggest it does:

> You should see the linter output right in your terminal as well as the browser console. 

Admittedly, I can't get it to show any lint errors for me. Hmm.

bobsilverberg:
It definitely doesn't show up anywhere currently, which is why I added this issue. I agree with @kumar303 about the current issues with the jest reporter, but I still feel like it's better than nothing. I'm surprised that nobody has created a watcher for ESLint. I agree that having that in its own window in the multi-window terminal view would be ideal.

kumar303:
> It definitely doesn't show up anywhere currently

We broke something https://github.com/mozilla/addons-code-manager/issues/152

kumar303:
Here's how to fix the TS failures:

```diff
diff --git a/tsconfig.json b/tsconfig.json
index ad6a34b..f09b2df 100644
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -25,6 +25,7 @@
     "stories"
   ],
   "exclude": [
-    "stories/setup/webpack.config.js"
+    "stories/setup/webpack.config.js",
+    "src/__tests__/jest-reporters/"
   ]
 }
```

src/__tests__/jest-reporters/utils.js
```diff
@@ -0,0 +1,45 @@
...
+      return line;
+    })
+    .map((line) => line.split(' '))
+    // .flatMap()
+    .reduce((chunks, chunk) => chunks.concat(chunk), [])
+    // We assume no filename can be smaller than 3 to filter the short format
+    // statuses: https://git-scm.com/docs/git-status#_short_format.
+    .filter((chunk) => chunk.length > 3)
+    // Allow directories OR JS/JSX files
+    .filter((filename) => /(\/|\.jsx?)$/.test(filename));
```
kumar303: Needs moar TypeScript

```diff
diff --git a/src/__tests__/jest-reporters/utils.js b/src/__tests__/jest-reporters/utils.js
index 587385b..2e56f7d 100644
--- a/src/__tests__/jest-reporters/utils.js
+++ b/src/__tests__/jest-reporters/utils.js
@@ -22,7 +22,7 @@ const filterFileNamesFromGitStatusOutput = (output) => {
     // statuses: https://git-scm.com/docs/git-status#_short_format.
     .filter((chunk) => chunk.length > 3)
     // Allow directories OR JS/JSX files
-    .filter((filename) => /(\/|\.jsx?)$/.test(filename));
+    .filter((filename) => /(\/|\.[tj]sx?)$/.test(filename));
 
   return files;
 };
```

src/jest-reporters/eslint-check.js
```diff
@@ -0,0 +1,61 @@
+/* eslint-disable no-console */
+/* eslint-disable amo/only-tsx-files */
+/* eslint-disable @typescript-eslint/no-var-requires */
+// eslint-disable-next-line import/no-extraneous-dependencies
```
bobsilverberg: Note that I really should have been able to address this by adding an entry into `devDependencies` at https://github.com/mozilla/addons-code-manager/blob/master/.eslintrc#L16, but that doesn't work for some reason.

bobsilverberg:
This is ready for another look. I believe I have eliminated all of the errors. I didn't make any changes to the reporter or utils file, other than to make sure that `.ts` and `.tsx` files are also linted.

src/jest-reporters/eslint-check.js
kumar303: > I really should have been able to address this by adding an entry into `devDependencies`...

Huh, I couldn't get it to work either.

src/jest-reporters/eslint-check.js
```diff
@@ -0,0 +1,61 @@
...
+
+    console.log('');
+    if (this.eslintOutput) {
+      console.log(this.eslintOutput);
+      console.log(
+        `Set ${NO_ESLINT_ENV_VAR}=1 in the environment to disable eslint checks`,
+      );
+      return new Error('eslint errors');
+    }
+    console.log('Eslint: no errors 💄 ✨');
```
kumar303: stmux will think this is an error unless you say it's OK:

```diff
diff --git a/package.json b/package.json
index 6990a39..106966e 100644
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "start-local-dev": "yarn build-local-dev-config && yarn start",
     "start-local-prod": "cat .env.common-local > .env.local && yarn start",
     "start-local-stage": "cat .env.common-local .env.stage > .env.local && yarn start",
-    "stmux": "stmux -nM -t addons-code-manager -e 'error,!0 errors'",
+    "stmux": "stmux -nM -t addons-code-manager -e 'error,!0 errors,!no errors'",
     "storybook": "start-storybook -p 9001 -c stories/setup",
     "storybook-smoke-test": "yarn storybook --smoke-test",
     "test": "react-scripts test --reporters=default --reporters=`pwd`/src/jest-reporters/eslint-check.js",
```

src/__tests__/jest-reporters/utils.js
willdurand: I believe we should not even have the `j`, only `tsx?`

package.json
```diff
@@ -90,7 +90,7 @@
     "stmux": "stmux -nM -t addons-code-manager -e 'error,!0 errors'",
     "storybook": "start-storybook -p 9001 -c stories/setup",
     "storybook-smoke-test": "yarn storybook --smoke-test",
-    "test": "react-scripts test",
+    "test": "react-scripts test --reporters=default --reporters=`pwd`/src/jest-reporters/eslint-check.js",
```
willdurand: Can we not add `\`pwd\``?

src/__tests__/jest-reporters/utils.js
bobsilverberg: I don't see the harm in linting `js` files if they exist. For example, this patch adds two `js` files and we probably want them linted, don't we?

package.json
bobsilverberg: Sorry, I don't understand the comment.

src/jest-reporters/eslint-check.js
bobsilverberg: I don't know whether to 👍 this or to 👎 this. 😉

package.json
willdurand: Can you use a relative path instead of using the `pwd` program inlined with back ticks.

src/__tests__/jest-reporters/utils.js
willdurand: ah, right.

package.json
bobsilverberg: I tried and couldn't get it to work. I found that `pwd` trick in a SO article and it seems to be the only thing that *does* work. Do you want to try to see if you can find a different solution?

willdurand: Did you try `<rootDir>` by chance?

```suggestion
    "test": "react-scripts test --reporters=default --reporters='<rootDir>/src/jest-reporters/eslint-check.js'",
```

bobsilverberg: Thanks @willdurand, that does work!

willdurand:
why did this PR get merged without any tests? Can someone file an issue to port the tests from https://github.com/mozilla/addons-frontend/tree/master/tests/unit/jest-reporters? Thanks

bobsilverberg:
> why did this PR get merged without any tests? Can someone file an issue to port the tests from [mozilla/addons-frontend:tests/unit/jest-reporters@`master`](https://github.com/mozilla/addons-frontend/tree/master/tests/unit/jest-reporters)? Thanks

Oops! I forgot about tests. I guess I merged it because it was approved. My bad. I'll file that issue.

PR #143, opened 2019-02-05
Add loading state

### LINKED ISSUE #144: Render a loading state while fetching the user profile

* Click the Log In button and sign in through Firefox Accounts
* See that you are now logged in
* Refresh the page

### What happened?

When the page first loads, the button says *Log In* then changes very quickly to *Log Out* once the user profile loads from the API.

The button should begin rendering in a loading state while the user profile is being fetched. It should not begin by saying *Log In*.

willdurand:
Fixes #144

---

As discussed in #134, I added a loading state until the user profile gets loaded. It is only displayed when there is an authentication token, otherwise we see the "please log in" message.

For some reasons, the catch-all route was not working as intended (404). I fixed that but the behavior is different: the "please log in" message has precedence over the 404 (which is ok I believe).

Note: I decided to add this code in `App` so that it is done once for all.

### Loading

screen shot 2019-02-05 at 19 17 55

### 404

screen shot 2019-02-05 at 19 15 02

### Please log in

screen shot 2019-02-05 at 19 14 52

### Browse page

screen shot 2019-02-05 at 19 14 58

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@types/classnames": "2.2.7",
     "@types/express": "4.16.1",
     "@types/http-proxy": "1.16.2",
     "@types/http-proxy-middleware": "0.19.2",
@@ -16,6 +17,7 @@
     "@types/react-router-dom": "4.3.1",
     "@types/redux-logger": "3.0.6",
     "bootstrap": "4.2.1",
+    "classnames": "2.2.6",
     "express": "4.16.4",
     "history": "4.7.2",
     "http-proxy-middleware": "0.19.1",
--- a/src/components/App/index.spec.tsx
+++ b/src/components/App/index.spec.tsx
@@ -12,6 +12,7 @@ import {
   fakeUser,
   shallowUntilTarget,
 } from '../../test-helpers';
+import Navbar from '../Navbar';
 
 import App, { AppBase } from '.';
 
@@ -42,10 +43,11 @@ describe(__filename, () => {
   };
 
   it('renders without an authentication token', () => {
-    const root = render({ authToken: '' });
+    const root = render({ authToken: null });
 
     expect(root.find(Container)).toHaveClassName(styles.container);
     expect(root.find(`.${styles.content}`)).toHaveLength(1);
+    expect(root.find(Navbar)).toHaveLength(1);
   });
 
   it('renders with an empty authentication token', () => {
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -16,6 +17,7 @@
     "@types/react-router-dom": "4.3.1",
     "@types/redux-logger": "3.0.6",
     "bootstrap": "4.2.1",
+    "classnames": "2.2.6",
```
bobsilverberg: I was wondering how long it would be until someone added `classnames`. ;-)

src/components/App/index.tsx
```diff
@@ -38,6 +39,7 @@ export type Profile = {
 };
 
 type State = {
+  loading: boolean;
```
bobsilverberg: PR #136 removed the state entirely from this component. I think maybe we should put this `loading` flag in the redux store, rather than adding back in a component state. What do you think?

package.json
willdurand: 😄

src/components/App/index.tsx
willdurand: We could probably now that #136 has landed.

willdurand:
@bobsilverberg thanks, I think I made it simpler.

willdurand:
I just added one more test case.

src/components/App/index.spec.tsx
```diff
@@ -41,13 +41,26 @@ describe(__filename, () => {
...
-  it('renders without crashing', () => {
-    const root = render();
+  it('renders without an authentication token', () => {
+    const root = render({ authToken: '' });
 
     expect(root.find(Container)).toHaveClassName(styles.container);
     expect(root.find(`.${styles.content}`)).toHaveLength(1);
   });
 
+  it('renders with an empty authentication token', () => {
```
bobsilverberg: This test appears to be the same as the one above it.

src/components/App/index.tsx
```diff
@@ -65,43 +67,60 @@ export class AppBase extends React.Component<Props> {
...
     );
   }
 }
 
-const mapStateToProps = (state: ApplicationState): PropsFromState => {
+const mapStateToProps = (
+  state: ApplicationState,
+  ownProps: PublicProps,
+): PropsFromState => {
+  const profile = getCurrentUser(state.users);
```
bobsilverberg: Nit: You could just inline these in the returned object below.

src/components/App/index.tsx
```diff
@@ -65,43 +67,60 @@ export class AppBase extends React.Component<Props> {
...
+    ) : (
+      <p className={styles.loginMessage}>
+        {gettext('Please log in to continue.')}
+      </p>
+    );
 
     return (
       <React.Fragment>
-        <Navbar />
+        {!loading && <Navbar />}
```
bobsilverberg: This logic doesn't seem to be covered by a test.

src/components/App/index.spec.tsx
willdurand: oops

PR #148, opened 2019-02-06
Add more tests for the server

### LINKED ISSUE #133: Add tests for the server in development mode
The server code executed in development mode is not unit tested yet (because it is tricky). Let's find a way to add some coverage to it, because it would be useful!

willdurand:
Fixes #133 

---

If that works for everyone, we could add some more test cases for the API proxy.

### DIFF
--- a/src/server/index.spec.tsx
+++ b/src/server/index.spec.tsx
@@ -1,6 +1,8 @@
 /* eslint @typescript-eslint/no-object-literal-type-assertion: 0 */
+import http from 'http';
 import path from 'path';
 
+import express from 'express';
 import request from 'supertest';
 import { RequestWithCookies } from 'universal-cookie-express';
 import Cookies from 'universal-cookie';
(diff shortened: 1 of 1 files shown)


willdurand:
Up!

src/server/index.spec.tsx
```diff
@@ -106,5 +108,102 @@ describe(__filename, () => {
...
+      let proxyServer: http.Server;
+
+      const devEnv = {
+        NODE_ENV: 'development',
+        REACT_APP_CRA_PORT: 60123,
+      };
+
+      beforeEach(() => {
+        proxy = express();
+        proxyServer = proxy.listen(devEnv.REACT_APP_CRA_PORT);
```
kumar303: This needs some kind of error handling, otherwise proxy errors show up as unhandled promises (the tests still fail with expected errors, though).

screenshot 2019-02-13 14 58 59

This is the best I can think of:

```diff
diff --git a/src/server/index.spec.tsx b/src/server/index.spec.tsx
index 1c94e11..8a96302 100644
--- a/src/server/index.spec.tsx
+++ b/src/server/index.spec.tsx
@@ -121,6 +121,9 @@ describe(__filename, () => {
       beforeEach(() => {
         proxy = express();
         proxyServer = proxy.listen(devEnv.REACT_APP_CRA_PORT);
+        proxyServer.on('error', (error) => {
+          console.error('proxy error', error);
+        });
       });
 
       afterEach(() => {
```

After that, I don't see unhandled promise rejections.

src/server/index.spec.tsx
```diff
@@ -106,5 +108,102 @@ describe(__filename, () => {
...
+      });
+
+      afterEach(() => {
+        proxyServer.close();
+      });
+
+      it('creates an Express server that uses a proxy', async () => {
+        const proxyContent = 'Hello, I am a proxy';
```
kumar303: This is more like `proxiedContent`, right? Can you make it more realistic? Perhaps: `Hello, I am the create-react-app server` ... or something.

src/server/index.spec.tsx
kumar303: Maybe this is obvious because of the env var but I was confused at how `createServer` was able to connect to this. I would suggest adding a comment saying something like: This is meant to simulate the create-react-app dev server

src/server/index.spec.tsx
```diff
@@ -106,5 +108,102 @@ describe(__filename, () => {
...
+      let proxy: express.Application;
+      let proxyServer: http.Server;
+
+      const devEnv = {
+        NODE_ENV: 'development',
+        REACT_APP_CRA_PORT: 60123,
+      };
+
+      beforeEach(() => {
+        proxy = express();
```
kumar303: Would it make sense to call it `fakeCRAserver` or something specific like that?

src/server/index.spec.tsx
willdurand: :+1:

src/server/index.spec.tsx
willdurand: correct, thanks

src/server/index.spec.tsx
```diff
@@ -106,5 +108,115 @@ describe(__filename, () => {
...
+          // eslint-disable-next-line no-console
+          console.error('proxy error', error);
+        });
+      });
+
+      afterEach(() => {
+        proxyServer.close();
+      });
+
+      it('creates an Express server that uses a proxy', async () => {
```
kumar303: Can this be more specific? Maybe: it can proxy the create-react-app server

src/server/index.spec.tsx
```diff
@@ -106,5 +108,115 @@ describe(__filename, () => {
...
+        const proxiedContent = 'Hello, I am the create-react-app server';
+        // Configure the proxy
```
kumar303: I'm not sure this is accurate. Maybe: Configure the proxied server ?

src/server/index.spec.tsx
```diff
@@ -106,5 +108,115 @@ describe(__filename, () => {
...
+
+        const response = await server.get('/');
+
+        expect(_injectAuthenticationToken).toHaveBeenCalled();
+        expect(response.text).toContain(expectedHTML);
+      });
+
+      it('does not configure the `cache-control` header of non-HTML proxy responses', async () => {
+        const proxiedContent = { foo: 'bar' };
+        // Configure the proxy to return JSON content
```
kumar303: Configure the proxied server ... ?

src/server/index.spec.tsx
```diff
@@ -106,5 +108,115 @@ describe(__filename, () => {
...
+        expect(response.text).toContain(JSON.stringify(proxiedContent));
+        expect(response.header).not.toHaveProperty(
+          'cache-control',
+          'private, no-store',
+        );
+      });
+
+      it('does not call injectAuthenticationToken() for non-HTML proxy responses', async () => {
+        const proxiedContent = { foo: 'bar' };
+        // Configure the proxy to return JSON content
```
kumar303: Same here

PR #149, opened 2019-02-06
Add test file for the LoginButton component

### LINKED ISSUE #147: Add coverage for the `LoginButton` component
The `LoginButton` component is still used and should have a test file.

willdurand:
Fixes #147

### DIFF
--- a/src/components/LoginButton/index.tsx
+++ b/src/components/LoginButton/index.tsx
@@ -4,9 +4,17 @@ import { Button } from 'react-bootstrap';
 import { gettext } from '../../utils';
 import styles from './styles.module.scss';
 
-export class LoginButtonBase extends React.Component {
+type PublicProps = {
+  fxaConfig: string;
+};
+
+export class LoginButtonBase extends React.Component<PublicProps> {
+  static defaultProps = {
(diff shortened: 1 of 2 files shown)


src/components/LoginButton/index.tsx
```diff
@@ -4,9 +4,17 @@ import { Button } from 'react-bootstrap';
...
+export class LoginButtonBase extends React.Component<PublicProps> {
+  static defaultProps = {
+    fxaConfig: process.env.REACT_APP_FXA_CONFIG,
+  };
+
   getFxaURL() {
-    const fxaConfig = process.env.REACT_APP_FXA_CONFIG;
+    const { fxaConfig } = this.props;
 
     return `/api/v4/accounts/login/start/?config=${fxaConfig}&to=/`;
```
bobsilverberg: Nit: I suppose you could extract this out into a function, e.g., `createFxaURL` which then gets exported so that the test doesn't have to copy this template string and the test could just ensure that the `href` property of the `Button` is the result of calling `createFxaURL`.

But that's probably overkill for something as simple as this.

willdurand: I'll leave it for now but I tend to agree with you (it'd be nice but it's also overkill :p)

Issue #39, opened 2019-01-17
Implement file browsing


EnTeQuAk:
This may need to be splitted into more issues.

File-browsing consists of two parts:

* A file-tree on the (currently) left side that shows all files
* A file-content part on the (currently) right side that shows the content of the currently selected file

API: https://addons-server.readthedocs.io/en/latest/topics/api/reviewers.html#browse

This requires a `version_id` that should be applied through the URL the user clicks on the reviewer tools page.

<img src="https://user-images.githubusercontent.com/139033/51308927-8cddfc00-1a43-11e9-9c5e-88516b07ee2a.png" width="240">

![screenshot from 2019-01-17 10-24-34](https://user-images.githubusercontent.com/139033/51308318-21475f00-1a42-11e9-9d60-01d005a6ffe6.png)

kumar303:
Thanks @EnTeQuAk 🎉 ; I broke it into separate issues:
* https://github.com/mozilla/addons-code-manager/issues/107
* https://github.com/mozilla/addons-code-manager/issues/108
* https://github.com/mozilla/addons-code-manager/issues/109

PR #153, opened 2019-02-06
Start playing with Treebeard

willdurand:
Supports #39 

---

This PR adds:

- `react-treebeard`, the most "okay" react tree library: https://github.com/storybooks/react-treebeard
- a `FileTree` component that renders any tree coming from the browse API

Known limitations:

- the API types are partial
- API errors are not handled at all
- files are not sorted (we should have directories first, then files IMHO)
- I store the response in the component's state until we get a reducer for that (but that's not a lot of code anyway)
- the interface (public props) of the `FileTree` will likely change, in part because the reducer will likely allow this component to be connected (or at least to receive more precise props from the `Browse` component)
- it uses the default style + a fix, but we could make it prettier, later

Use these links to try this patch:

- http://localhost:3000/en-US/firefox/files/browse/1541786/
- http://localhost:3000/en-US/firefox/files/browse/1541520/

## Screenshots

screen shot 2019-02-06 at 18 18 56
screen shot 2019-02-06 at 18 18 53

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "react-redux": "5.1.1",
     "react-router-dom": "4.3.1",
     "react-scripts": "2.1.3",
+    "react-treebeard": "3.1.0",
     "redux": "4.0.1",
     "redux-devtools-extension": "2.13.8",
     "redux-logger": "3.0.6",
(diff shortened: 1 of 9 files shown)


willdurand:
So.. Treebeard is flawed. It uses mutation to change a node state in the tree, it is broken, not maintained and difficult to customize. Switching to something else with more control...

willdurand:
Please see #160 no

PR #159, opened 2019-02-07
Add support for query params in callApi

### LINKED ISSUE #157: Add query string support to the `callApi` function
We'll need to pass query parameters to the `callApi` function soon.

willdurand:
Fixes #157

---

This PR adds basic support for query parameters.

### DIFF
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -15,6 +15,7 @@ type CallApiParams = {
   method?: HttpMethod;
   version?: string;
   apiState: ApiState;
+  query?: { [key: string]: string };
 };
 
 type CallApiResponse = object | { error: Error };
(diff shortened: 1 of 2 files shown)


src/api/index.tsx
```diff
@@ -15,6 +15,7 @@ type CallApiParams = {
   method?: HttpMethod;
   version?: string;
   apiState: ApiState;
+  query?: { [key: string]: string };
```
bobsilverberg: Nit: Do you think maybe `params` would be a better name for this?

willdurand: `params` is a bit too generic to me. `query` could be generic but in this context, it is more meaningful. `queryParameters` would be too long, and `qs` or `queryString` would be a bad name for an object value.

willdurand:
> Thanks for doing this @willdurand. I noticed that we need this for the `browse` API for selecting individual files, and I was going to defer that, but now with this I guess I'll include that in my first iteration.

Yeah, I started to play with that so that clicking a file in the tree list (cf. #153) shows its content on the right panel:

screen shot 2019-02-07 at 11 54 31

That's why I needed this patch :D

kumar303:
Sorry that I didn't have time to review this. I filed a request for changes but it's not high priority: https://github.com/mozilla/addons-code-manager/issues/164

PR #162, opened 2019-02-07
Sort files in FileTree

### LINKED ISSUE #161: Sort files in the file tree
Supports #107 

---

The tree list (list of files on the left panel) should be sorted. We should have directories first, then files and everything sorted alphabetically.

willdurand:
~~⚠️ depends on #160~~
Fixes #161

---

Before:

screen shot 2019-02-07 at 15 12 29

After:

screen shot 2019-02-07 at 16 36 02

### DIFF
--- a/src/components/FileTree/index.spec.tsx
+++ b/src/components/FileTree/index.spec.tsx
@@ -1,6 +1,7 @@
-import { Version, VersionEntryType } from '../../reducers/versions';
+import { Version, createInternalVersionEntry } from '../../reducers/versions';
+import { fakeVersionEntry } from '../../test-helpers';
 
-import { buildFileTree } from '.';
+import { buildFileTree, DirectoryNode } from '.';
 
 describe(__filename, () => {
   describe('buildFileTree', () => {
@@ -8,9 +9,9 @@ describe(__filename, () => {
       const versionId = '1234';
       const entries: Version['entries'] = [];
 
-      const data = buildFileTree(versionId, entries);
+      const tree = buildFileTree(versionId, entries);
 
-      expect(data).toEqual({
+      expect(tree).toEqual({
         id: `root-${versionId}`,
         name: versionId,
         children: [],
@@ -18,22 +19,19 @@ describe(__filename, () => {
     });
 
     it('converts a non-directory entry to a file node', () => {
-      const versionId = 'some-name';
+      const versionId = '1234';
 
       const filename = 'some-filename';
       const entries = [
-        {
-          depth: 0,
-          type: VersionEntryType.text,
+        createInternalVersionEntry({
+          ...fakeVersionEntry,
           filename,
-          path: filename,
-          modified: '2019-01-01',
-        },
+        }),
       ];
 
-      const data = buildFileTree(versionId, entries);
+      const tree = buildFileTree(versionId, entries);
 
-      expect(data.children).toEqual([
+      expect(tree.children).toEqual([
         {
           id: entries[0].path,
           name: filename,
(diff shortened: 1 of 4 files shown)


willdurand:
See also: https://github.com/mozilla/addons-code-manager/pull/163

willdurand:
This one is ready too. I refactored some test cases as discussed in https://github.com/mozilla/addons-code-manager/pull/160#discussion_r255537903

src/reducers/versions.tsx
```diff
@@ -148,17 +148,24 @@ export const createInternalVersionFile = (
   };
 };
 
+export const createInternalVersionEntry = (
+  entry: ExternalVersionEntry,
+): VersionEntry => {
+  return {
+    type: getVersionEntryType(entry),
```
bobsilverberg: Nit: can we alphabetize these?

src/reducers/versions.tsx
```diff
@@ -148,17 +148,24 @@ export const createInternalVersionFile = (
...
+    filename: entry.filename,
+    modified: entry.modified,
+    path: entry.path,
+  };
+};
+
 export const createInternalVersion = (version: ExternalVersion): Version => {
   return {
     entries: Object.keys(version.file.entries).map((nodeName) => {
       const entry = version.file.entries[nodeName];
```
bobsilverberg: Nit: we don't strictly need this `const`.

src/components/FileTree/index.tsx
```diff
@@ -10,14 +10,38 @@ type FileNode = {
...
 type TreeNode = FileNode | DirectoryNode;
 
+const recursiveSortInPlace = (node: DirectoryNode): void => {
+  node.children.sort((a, b) => {
+    if ('children' in a && 'children' in b) {
+      return a.name.localeCompare(b.name);
+    }
+
+    if ('children' in a === false && 'children' in b === false) {
+      return a.name.localeCompare(b.name);
```
bobsilverberg: This is the same thing as the result of the other `if` test above. Do you want to `||` them, or do you think it's better this way because it's more readable?

src/components/FileTree/index.spec.tsx
```diff
@@ -1,39 +1,37 @@
...
+      const versionId = '1234';
 
       const filename = 'some-filename';
       const entries = [
-        {
-          depth: 0,
-          type: VersionEntryType.text,
+        createInternalVersionEntry({
+          ...fakeVersionEntry,
           filename,
```
bobsilverberg: Maybe include `directory: false` just to be explicit?

src/components/FileTree/index.spec.tsx
```diff
@@ -42,56 +40,51 @@ describe(__filename, () => {
...
+          directory: true,
           filename: directory,
-          path: directory,
-          modified: '2019-01-01',
-        },
-        {
+        }),
+        createInternalVersionEntry({
+          ...fakeVersionEntry,
           depth: 1,
```
bobsilverberg: As above, maybe include `directory: false` to be explicit?

src/components/FileTree/index.spec.tsx
```diff
@@ -112,27 +105,24 @@ describe(__filename, () => {
...
           filename: directoryName,
           path: `${directoryName}/${directoryName}`,
-          modified: '2019-01-01',
-        },
-        {
+        }),
+        createInternalVersionEntry({
+          ...fakeVersionEntry,
           depth: 2,
-          type: VersionEntryType.text,
```
bobsilverberg: As above, maybe include `directory: false` to be explicit?

src/components/FileTree/index.spec.tsx
```diff
@@ -156,5 +146,188 @@ describe(__filename, () => {
...
+      const versionId = '1234';
+      const entries = [
+        createInternalVersionEntry({
+          ...fakeVersionEntry,
+          directory: true,
+          filename: 'B',
+        }),
+        createInternalVersionEntry({
+          ...fakeVersionEntry,
+          filename: 'A',
```
bobsilverberg: As above, maybe include `directory: false` to be explicit?

src/reducers/versions.tsx
willdurand: yep

src/reducers/versions.tsx
willdurand: ah, correct

src/components/FileTree/index.tsx
willdurand: I was thinking it was more readable that way

src/components/FileTree/index.spec.tsx
willdurand: good call

src/components/FileTree/index.spec.tsx
willdurand: good call

src/components/FileTree/index.spec.tsx
willdurand: :+1:

PR #166, opened 2019-02-07
Implement API and reducer code for the browse API

### LINKED ISSUE #142: Implement API and reducer code for the browse API
Implement code to support interacting with the API documented at https://addons-server.readthedocs.io/en/latest/topics/api/reviewers.html#browse.

bobsilverberg:
Fixes #142

### DIFF
--- a/src/api/index.spec.tsx
+++ b/src/api/index.spec.tsx
@@ -5,7 +5,7 @@ import {
 } from '../reducers/api';
 import configureStore from '../configureStore';
 
-import { HttpMethod, callApi, logOutFromServer } from '.';
+import { HttpMethod, callApi, getVersionFile, logOutFromServer } from '.';
 
 describe(__filename, () => {
   const getApiState = ({ authToken = '12345' } = {}) => {
@@ -141,6 +141,36 @@ describe(__filename, () => {
       expect(fetch).toHaveBeenCalledWith(
         '/api/v4/url/?foo=1&bar=abc',
         expect.any(Object),
+      );
+    });
+  });
+
+  describe('getVersionFile', () => {
+    it('calls the API to retrieve default version file information', async () => {
+      const versionId = 123;
+
+      await getVersionFile({ apiState: defaultApiState, versionId });
+
+      expect(fetch).toHaveBeenCalledWith(
+        `/api/v4/reviewers/browse/${versionId}/`,
+        {
+          headers: {},
+          method: HttpMethod.GET,
+        },
+      );
+    });
+    it('calls the API to retrieve information for a specific version file', async () => {
+      const filename = 'test.js';
+      const versionId = 123;
+
+      await getVersionFile({ apiState: defaultApiState, filename, versionId });
+
+      expect(fetch).toHaveBeenCalledWith(
+        `/api/v4/reviewers/browse/${versionId}/?file=${filename}`,
+        {
+          headers: {},
+          method: HttpMethod.GET,
+        },
       );
     });
   });
--- a/src/api/index.spec.tsx
+++ b/src/api/index.spec.tsx
@@ -160,13 +160,13 @@ describe(__filename, () => {
       );
     });
     it('calls the API to retrieve information for a specific version file', async () => {
-      const filename = 'test.js';
+      const path = 'test.js';
       const versionId = 123;
 
-      await getVersionFile({ apiState: defaultApiState, filename, versionId });
+      await getVersionFile({ apiState: defaultApiState, path, versionId });
 
       expect(fetch).toHaveBeenCalledWith(
-        `/api/v4/reviewers/browse/${versionId}/?file=${filename}`,
+        `/api/v4/reviewers/browse/${versionId}/?file=${path}`,
         {
           headers: {},
           method: HttpMethod.GET,
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -76,7 +76,7 @@ export const callApi = async ({
   }
 };
 
-type GetVersionParams = {
+type GetVersionFileParams = {
   apiState: ApiState;
   filename?: string;
   versionId: number;
@@ -86,12 +86,11 @@ export const getVersionFile = async ({
   apiState,
   filename = '',
   versionId,
-}: GetVersionParams) => {
+}: GetVersionFileParams) => {
   return callApi({
     apiState,
     endpoint: `reviewers/browse/${versionId}`,
     query: filename ? { file: filename } : undefined,
-    method: HttpMethod.GET,
   });
 };
 
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -76,6 +76,25 @@ export const callApi = async ({
   }
 };
 
+type GetVersionParams = {
+  apiState: ApiState;
+  filename?: string;
+  versionId: number;
+};
+
+export const getVersionFile = async ({
+  apiState,
+  filename = '',
+  versionId,
+}: GetVersionParams) => {
+  return callApi({
+    apiState,
+    endpoint: `reviewers/browse/${versionId}`,
+    query: filename ? { file: filename } : undefined,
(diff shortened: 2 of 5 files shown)


src/api/index.spec.tsx
```diff
@@ -145,6 +145,36 @@ describe(__filename, () => {
     });
   });
 
+  describe('getVersionFile', () => {
```
willdurand: I thought we decided not to use specialized API functions given that it would provide little value.

src/api/index.tsx
```diff
@@ -76,6 +76,25 @@ export const callApi = async ({
...
+export const getVersionFile = async ({
+  apiState,
+  filename = '',
+  versionId,
+}: GetVersionParams) => {
+  return callApi({
+    apiState,
+    endpoint: `reviewers/browse/${versionId}`,
+    query: filename ? { file: filename } : undefined,
+    method: HttpMethod.GET,
```
willdurand: This is the default IIRC

src/api/index.spec.tsx
bobsilverberg: I thought we just decided not to create additional files for them, and to put them all into `api/index.tsx` for now. The value I see is that we get some type safety around the call because of the params, and also we have centralization of API calls. I assume the other method would be to inline the call to `callApi` inside other files (whatever our equivalent to sagas will be) and that could result in them sprinkled around the code.

src/api/index.tsx
bobsilverberg: Fair point. I'll remove it.

src/reducers/versions.tsx
```diff
@@ -0,0 +1,218 @@
...
+  created: string;
+  id: number;
+  size: number;
+};
+
+export enum VersionEntryType {
+  image,
+  binary,
+  text,
+  directory,
```
willdurand: Great! Can we uppercase them? Those values look like constants to me.

src/reducers/versions.tsx
```diff
@@ -0,0 +1,218 @@
...
+  }),
+};
+
+export type VersionsState = {
+  versionInfo: {
+    [versionId: number]: Version;
+  };
+  versionFiles: {
+    [versionId: number]: {
+      [filename: string]: VersionFile;
```
willdurand: I don't think this will work because the filename is not unique. It should probably be an `id`, but I don't think it exists. Yet, `path` exists and it is unique, so we probably want to use that.

src/reducers/versions.tsx
```diff
@@ -0,0 +1,218 @@
...
+      };
+    }),
+    id: version.id,
+    reviewed: version.reviewed,
+    version: version.version,
+  };
+};
+
+export const getVersionFile = (
+  filename: string,
```
willdurand: I believe it should not be `filename` but `path` here or `file` (which I like better).

src/reducers/versions.tsx
EnTeQuAk: You could use the sha256 hash for added uniqueness and easier serializability. But the path should work too.

src/reducers/versions.tsx
bobsilverberg: They are constants, but I thought the code using them looked better as `VersionEntryType.image` and `VersionEntryType.binary` as opposed to `VersionEntryType.IMAGE` and `VersionEntryType.BINARY`. Can we adopt a convention that when we use `enum` we can use camelCased names?

src/reducers/versions.tsx
bobsilverberg: Thanks @EnTeQuAk. We want to use something that will be available in `VersionEntry`, and I don't think we have a need to store the `sha256` for our internal VersionEntries, so I think `path` makes the most sense.

src/reducers/versions.tsx
bobsilverberg: I went with `path` because that's what it's called when stored internally for `VersionEntry`.

src/reducers/versions.tsx
willdurand: yeah, I like path too

bobsilverberg:
Thanks for the review @willdurand. I've updated the patch to address your issues.

src/api/index.spec.tsx
willdurand: fair enough

src/reducers/versions.tsx
willdurand: sure

src/api/index.spec.tsx
```diff
@@ -145,6 +145,36 @@ describe(__filename, () => {
...
+      const versionId = 123;
+
+      await getVersionFile({ apiState: defaultApiState, filename, versionId });
+
+      expect(fetch).toHaveBeenCalledWith(
+        `/api/v4/reviewers/browse/${versionId}/?file=${filename}`,
+        {
+          headers: {},
+          method: HttpMethod.GET,
+        },
```
willdurand: You could omit this since it is covered by another test case, but it is also fine if you want to highlight the fact that passing a file does not change the `headers`.

src/api/index.tsx
```diff
@@ -76,6 +76,24 @@ export const callApi = async ({
   }
 };
 
+type GetVersionFileParams = {
+  apiState: ApiState;
+  filename?: string;
```
willdurand: so this should be `file` or `path`, no? I'd go for `file` but I already hear you saying that `path` is what's used in the rest of this patch.

src/api/index.tsx
```diff
@@ -76,6 +76,24 @@ export const callApi = async ({
...
+  versionId: number;
+};
+
+export const getVersionFile = async ({
+  apiState,
+  filename = '',
```
willdurand: why is there a default empty string value?

src/reducers/versions.tsx
```diff
@@ -0,0 +1,218 @@
+import { Reducer } from 'redux';
+import { ActionType, createAction, getType } from 'typesafe-actions';
+
+type VersionId = number;
+
+type LocalizedString = {
+  [lang: string]: string;
+};
```
willdurand: I have the feeling that this could be reused somewhere else in the future, but no need to change anything right now

src/reducers/versions.tsx
```diff
@@ -0,0 +1,218 @@
...
+          ...state.versionFiles,
+          [version.id]: {
+            ...state.versionFiles[version.id],
+            [path]: createInternalVersionFile(version.file),
+          },
+        },
+      };
+    }
+    default:
+      return state;
```
willdurand: this is not covered by a test case 🤓

src/api/index.tsx
bobsilverberg: The docs for the API just say `file`, so I assumed it was `filename`, but I guess it must be `path` because `filename` might not be unique. I guess I'll use `path` to keep it consistent with the rest of the code.

src/api/index.tsx
bobsilverberg: Not sure. I've removed it.

src/reducers/versions.tsx
bobsilverberg: I was thinking the same thing, about both. :-)

src/reducers/versions.tsx
bobsilverberg: Yes, I noticed this. I believe this is because we decided to remove the tests for reducers that simply check that `initialState` is returned when an invalid action is passed. Maybe we want to add something like that back in so that we can have 100% coverage? @kumar303 what do you think?

willdurand: ah right, I remember a comment mentioning that such test cases had little value. I do agree. We could also add a comment to ignore the coverage.

bobsilverberg:
Thanks for the reviews @willdurand. I've addressed the remaining comments and I will merge this once it passes CI so you can use it for #160.

PR #168, opened 2019-02-08
Add more tests for the FileTree and some style

### LINKED ISSUE #158: Switch to "light" Bootstrap theme
I like the dark theme but our main component (react-diff-view) uses a light theme and it is usually easier to style a light theme than a dark one, so I think we should switch to a light theme. WDYT?

### LINKED ISSUE #169: Style tree
Supports #107

---

The tree list (list of files on the left panel) should be styled. Let's used some Bootstrap components and FontAwesome icons.

### LINKED ISSUE #183: Extract FileTreeNode component
In #160, we introduce a file tree. The code to render one single node is embedded into the `FileTree` component, which makes it difficult to test. This issue is about fixing that by extracting the code to render a node into its own component and to add tests for that.

willdurand:
~~⚠️ depends on #160~~
Fixes #158  
Fixes #169
Fixes #183

---


willdurand:
This one is ready too :)

willdurand:
> First impression: It's pretty, but I think maybe it's not quite compact enough. A lot less nodes fit on a single screen now.

That's Bootstrap style. I can maybe reduce the height of each node? In any cases, we should rely on Bootstrap so that's the first step :)

src/components/FileTreeNode/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+      };
+    }
+
+    return (
+      <React.Fragment>
+        <ListGroup.Item {...listGroupItemProps}>
+          <span style={{ paddingLeft: `${level * 20}px` }}>
+            {isFolder ? (
+              <React.Fragment>
+                <FontAwesomeIcon icon={isExpanded ? 'folder-open' : 'folder'} />
```
bobsilverberg: I'm not sure I love `react-fontawesome`. What does it buy us over using `Font Awesome` with just basic classes and css? This seems like a lot of code to be adding just to be able to use a React component instead of assigning a className. If we'd need a `span` or something else in its place then I guess it makes sense, but would it be possible to do something just with css using `::before`?

src/components/FileTreeNode/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+      isFolder,
+      isExpanded,
+      getToggleProps,
+      hasChildNodes,
+      renderChildNodes,
+    } = this.props;
+
+    let listGroupItemProps = {
+      className: makeClassName(styles.node, {
+        [styles.directoryNode]: isFolder,
```
bobsilverberg: Why do we need this? It looks like it's just for testing, but it also looks like the tests could confirm the same thing based on the icon, and don't really need to use this class just to prove that a node is a directory node.

willdurand:
@bobsilverberg I pushed a commit to show how we can adjust the style for the Bootstrap list. Here is the result:

Before:

screen shot 2019-02-11 at 17 49 16

After:

screen shot 2019-02-11 at 17 48 25

src/components/FileTreeNode/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+      }),
+    };
+
+    if (isFolder) {
+      listGroupItemProps = {
+        ...listGroupItemProps,
+        ...getToggleProps(),
```
bobsilverberg: This doesn't seem to be covered by a test.

src/components/FileTreeNode/index.tsx
willdurand: FA allows us to use lots of icons easily. We don't have UX so I don't want to pull random icons and end up with an inconsistent UI. Bootstrap and FA work well together.

We will likely need some other icons for some metadata, maybe for the errors vs warnings, links, etc. FA has all of that, the license is permissive and, again, all icons are consistent.

Also, those icons work on all devices/OS/browsers 😛

src/components/FileTreeNode/index.tsx
willdurand: Ah good catch, it's missing an assertion.

src/components/FileTreeNode/index.tsx
willdurand: If we get rid of the icon or change it, tests should be updated. It's less probable with a CSS class, no? Also, I guess we'll end up styling those directories a bit in the near future :p

src/components/FileTreeNode/index.tsx
bobsilverberg: My issue wasn’t with the use of font awesome, but with the react package and using a component to render the icon. I thought maybe we could use font awesome in conjunction with css to add icons to existing elements rather than needing to include an extra component.

willdurand: ah, ooops. It's true that the setup is weird but it's the recommended react way.. We could maybe create our own component instead but I am not sure if that'd be better.
I haven't checked but it is possible that the lib inlines the SVGs.

Issue #174, opened 2019-02-08
Do we want to add comments for ESLint warnings?


bobsilverberg:
When running `ESLint` on the current code base, we get the following warnings:

```
/addons-code-manager/src/test-helpers.spec.tsx
  14:33  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any

/addons-code-manager/src/test-helpers.tsx
  170:41  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
  171:48  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any

✖ 3 problems (0 errors, 3 warnings)
```

which all have to do with `ShallowUntilTarget`. I think our use of `any` in those cases makes sense, but it's a bit annoying to see these warnings every time we run `ESLint`.

How do folks feel about adding code comments to the files to ignore those specific warnings in those cases?

willdurand:
I think we should make this rule throw an error instead of a warning. We can still disable the rule if needed. The patch for this issue will determine whether each current warning needs to be fixed or ignored.

kumar303:
> I think we should make this rule throw an error instead of a warning. We can still disable the rule if needed.

I agree. This will make the rule much more helpful.

I think warnings are dangerous because they are easy to ignore and they clutter up the terminal which makes solving errors more difficult.

willdurand:
This has been fixed in #238.

PR #176, opened 2019-02-08
Use redux-thunk for side effects

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/12

After some consideration (see below), redux-thunk is a pretty good option. This adds redux-thunk for handling side effects.

<hr>

*Original PR description:*

As part of https://github.com/mozilla/addons-code-manager/issues/12, here's an example of using `redux-thunk` so we can assess it.

Here are some questions worth asking:
- How testable is this?
- Does it allow our application behavior to be traced through one place?
- [maybe more questions]

Known issues:
- ~~The `(dispatch as ThunkDispatch)` type hint seems like it should be unnecessary since we have types for our middleware (see https://github.com/reduxjs/redux-thunk/issues/103) but I couldn't get the error to go away without this type hint, even when doing the same module override.~~
- I don't like how the *execution* of `requestLogOut()` is not captured in action replay. When we used sagas, it was possible to replay the stack of actions and re-execute each saga. This reduces traceability. As an example, let's say we were given a stack of actions by someone who encountered an error thrown in a thunk. The stack of actions would not let us reproduce that error.

### DIFF
--- a/src/components/Navbar/index.tsx
+++ b/src/components/Navbar/index.tsx
@@ -2,12 +2,15 @@ import * as React from 'react';
 import { Button, Navbar } from 'react-bootstrap';
 import { AnyAction } from 'redux';
 import { connect } from 'react-redux';
-import { ThunkDispatch } from 'redux-thunk';
 
 import { gettext } from '../../utils';
 import LoginButton from '../LoginButton';
 import { logOutFromServer } from '../../api';
-import { ApplicationState, ConnectedReduxProps } from '../../configureStore';
+import {
+  ApplicationState,
+  ConnectedReduxProps,
+  ThunkDispatch,
+} from '../../configureStore';
 import { ApiState } from '../../reducers/api';
 import {
   User,
@@ -35,10 +38,7 @@ export class NavbarBase extends React.Component<Props> {
 
   logOut = () => {
     const { _requestLogOut, apiState, dispatch } = this.props;
-    // dispatch(_requestLogOut());
-    (dispatch as ThunkDispatch<ApplicationState, undefined, AnyAction>)(
-      _requestLogOut(),
-    );
+    (dispatch as ThunkDispatch)(_requestLogOut());
   };
 
   render() {
--- a/src/components/Navbar/index.tsx
+++ b/src/components/Navbar/index.tsx
@@ -1,20 +1,16 @@
 import * as React from 'react';
 import { Button, Navbar } from 'react-bootstrap';
-import { AnyAction } from 'redux';
 import { connect } from 'react-redux';
 
 import { gettext } from '../../utils';
 import LoginButton from '../LoginButton';
-import { logOutFromServer } from '../../api';
 import {
   ApplicationState,
   ConnectedReduxProps,
   ThunkDispatch,
 } from '../../configureStore';
-import { ApiState } from '../../reducers/api';
 import {
   User,
-  actions as userActions,
   getCurrentUser,
   requestLogOut,
 } from '../../reducers/users';
@@ -25,7 +21,6 @@ type PublicProps = {
 };
 
 type PropsFromState = {
-  apiState: ApiState;
   profile: User | null;
 };
 
@@ -37,7 +32,7 @@ export class NavbarBase extends React.Component<Props> {
   };
 
   logOut = () => {
-    const { _requestLogOut, apiState, dispatch } = this.props;
+    const { _requestLogOut, dispatch } = this.props;
     (dispatch as ThunkDispatch)(_requestLogOut());
   };
 
@@ -66,7 +61,6 @@ export class NavbarBase extends React.Component<Props> {
 
 const mapStateToProps = (state: ApplicationState): PropsFromState => {
   return {
-    apiState: state.api,
     profile: getCurrentUser(state.users),
   };
 };
--- a/src/components/Navbar/index.tsx
+++ b/src/components/Navbar/index.tsx
@@ -4,11 +4,7 @@ import { connect } from 'react-redux';
 
 import { gettext } from '../../utils';
 import LoginButton from '../LoginButton';
-import {
-  ApplicationState,
-  ConnectedReduxProps,
-  ThunkDispatch,
-} from '../../configureStore';
+import { ApplicationState, ConnectedReduxProps } from '../../configureStore';
 import { User, getCurrentUser, requestLogOut } from '../../reducers/users';
 import styles from './styles.module.scss';
 
--- a/src/components/Navbar/index.tsx
+++ b/src/components/Navbar/index.tsx
@@ -35,7 +35,8 @@ export class NavbarBase extends React.Component<Props> {
 
   logOut = () => {
     const { _requestLogOut, apiState, dispatch } = this.props;
-    (dispatch as ThunkDispatch<ApplicationState, null, AnyAction>)(
+    // dispatch(_requestLogOut());
+    (dispatch as ThunkDispatch<ApplicationState, undefined, AnyAction>)(
       _requestLogOut(),
     );
   };
--- a/src/components/Navbar/index.tsx
+++ b/src/components/Navbar/index.tsx
@@ -29,7 +29,7 @@ export class NavbarBase extends React.Component<Props> {
 
   logOut = () => {
     const { _requestLogOut, dispatch } = this.props;
-    (dispatch as ThunkDispatch)(_requestLogOut());
+    dispatch(_requestLogOut());
   };
 
   render() {
--- a/src/components/Navbar/index.tsx
+++ b/src/components/Navbar/index.tsx
@@ -1,6 +1,8 @@
 import * as React from 'react';
 import { Button, Navbar } from 'react-bootstrap';
+import { AnyAction } from 'redux';
 import { connect } from 'react-redux';
+import { ThunkDispatch } from 'redux-thunk';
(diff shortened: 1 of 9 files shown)


kumar303:
@willdurand @bobsilverberg let me know what you think of thunks! I added questions and known issues to the PR description.

willdurand:
>     * The `(dispatch as ThunkDispatch)` type hint seems like it should be unnecessary since we have types for our middleware (see [reduxjs/redux-thunk#103](https://github.com/reduxjs/redux-thunk/issues/103)) but I couldn't get the error to go away without this type hint, even when doing the same module override.

It is a dumb question but: isn't it that `ConnectedReduxProps` declares `dispatch` as `Dispatch` and not `ThunkDispatch`?

willdurand:
> I don't like how the _execution_ of `requestLogOut()` is not captured in action replay. When we used sagas, it was possible to replay the stack of actions and re-execute each saga. This reduces traceability. As an example, let's say we were given a stack of actions by someone who encountered an error thrown in a thunk. The stack of actions would not let us reproduce that error.

This could be done by dispatching an action at the beginning of the thunk as shown here: https://github.com/mozilla/addons-code-manager/issues/12#issuecomment-459351292.

kumar303:
> This could be done by dispatching an action at the beginning of the thunk

How would that solve it? When replying actions, you would just be replaying the first action, *not* re-executing the entire thunk function.

kumar303:
> isn't it that `ConnectedReduxProps` declares `dispatch` as `Dispatch` and not `ThunkDispatch`?

Yes, that is correct and that is the problem. A thunk is a function, not an action, so if you edit the code in this patch to do `dispatch(_requestLogOut())` then you get an error saying that `_requestLogOut()` doesn't appear to be an action (indeed it's not).

willdurand:
> How would that solve it? When replying actions, you would just be replaying the first action, _not_ re-executing the entire thunk function.

I am not sure to follow. You want the execution of the thunk to be captured. What do you expect precisely? I read it as "I want to know that the thunk has been executed". Which other information is missing otherwise?

willdurand:
>  A thunk is a function, not an action

The middleware changes the behavior so that you can dispatch both, so why do reflecting that in the type? `dispatch: Action | Thunk`?

src/reducers/users.spec.tsx
```diff
@@ -53,4 +55,36 @@ describe(__filename, () => {
...
+      const thunk = requestLogOut({ _logOutFromServer: jest.fn(), ...params });
+
+      const store = configureStore();
+      const dispatch = jest.fn();
+
+      return {
+        dispatch,
+        thunk: () => thunk(dispatch, () => store.getState(), undefined),
+        store,
+      };
```
willdurand: I like this helper function!

src/test-helpers.spec.tsx
```diff
@@ -11,6 +11,7 @@ describe(__filename, () => {
     };
 
     const wrapper = () => {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
willdurand: I guess this resolves #174 :D

kumar303:
> You want the execution of the thunk to be captured. What do you expect precisely?

I expect to preserve this key principle of Redux as much as possible:

> All of the behavior of your application can be traced through one place, and that behavior can be easily broken apart and composed back together.

When not relying on side effects, this principle can be achieved by simply dispatching actions:

* `dispatch LOG_IN` -> all component code to render a logged in state is executed
* `dispatch LOAD_USER` -> all component code to render user data is executed

We can replay these two actions and execute the exact code that led us to these UI states. We cannot replay the click handler functions which *dispatched* each action but as long as those handler functions do nothing but dispatch actions then the risk of breakage there is minimal.

Let's say we needed to rely on side effects and we implemented them with sagas. Dispatching actions would look more like this:

* `dispatch REQUEST_LOG_OUT` ->
  * *code in saga*: make an API request to log out
  * *code in saga*: `dispatch LOG_OUT` or `dispatch ERROR`

We can now replay the action `REQUEST_LOG_OUT` and execute the exact code that led us to all UI states, including the code that interacted with the API. This is appealing to me because it would theoretically let us reproduce a bug inside a saga given just a stack of actions. It also helps one to understand how the app works because you can always trace behavior to an action. The risk of breakage within side effect handlers (such as a saga) is *not* minimal. There are lots of things that could break so it's important to have traceability here.

Here's how it would look with a thunk:

* *user clicks log out button*
  * *code in thunk*: make an API request to log out
  * *code in thunk*: `dispatch LOG_OUT` or `dispatch ERROR`

We can replay the `LOG_OUT` or `ERROR` actions but we cannot replay the API request when given a stack of actions.

I'm going to try `redux-loop` next to see if it lets us achieve traceability.

willdurand:
Thanks for the detailed reply, now I get it.

> We can replay the `LOG_OUT` or `ERROR` actions but we cannot replay the API request when given a stack of actions.

As per https://github.com/mozilla/addons-code-manager/issues/12#issuecomment-454448197, this is exactly what you wanted to avoid though:

>> As soon as you time travel to `FETCH_USER`, it re-runs the saga, makes the API request again, and dispatches more actions.

That being said, I understand that replaying a thunk is difficult *but* given a state and a stack of actions, you will still be able to reproduce what has happened, without firing the API call. I mean, in terms of data flow, you would get the same end state.

Unless there is an issue with the `fetch` implementation that is not caught in a `try/catch`, I think replaying the HTTP call is actually not needed. Redux actions describe the result of a side-effect, so replaying these actions should be enough to retrieve the same state, even if those side-effects did not fire again.

The redux devtools allow to load a state and to replay a stack of actions, so if we can record the stack of actions, we could also retain the initial state and either go forward or backward (if we can capture the end state too for example)

kumar303:
> As per https://github.com/mozilla/addons-code-manager/issues/12#issuecomment-454448197, [triggering side effects in the developer tool] is exactly what you wanted to avoid though

Yes, that's true. It's because I mostly want to use the developer tool to develop screens and sometimes I like to hop around between states. We had the habit of rendering important screens using the same action that triggers a saga and that made it hard to time travel. You could do it but the stack of actions would keep growing and growing which made it hard to move forward after moving backward.

You raised a good point in that conversation about how you *do* want to see side effects in the developer tool. I get that, too. Ideally, I'd like a checkbox to temporarily disable side effects.

> ...replaying a thunk is difficult _but_ given a state and a stack of actions, you will still be able to reproduce what has happened, without firing the API call. I mean, in terms of data flow, you would get the same end state.

Yes, but all of the code executed in between is very crucial to the program flow, especially in trying to track down unexpected errors. The same would go for rendering a component: when replaying an action I would very much want to see an error thrown by `render()`, for example. Otherwise, what's the point of using Redux?

I would like to store a stack of actions as Sentry breadcrumbs one day 💡 In dev only, [of course](https://techcrunch.com/2019/02/07/apple-glassbox-apps/)! 😎

src/components/Navbar/index.tsx
```diff
@@ -4,37 +4,32 @@ import { connect } from 'react-redux';
...
   };
 
-  logOut = async () => {
-    const { _logOutFromServer, apiState, dispatch } = this.props;
-
-    await _logOutFromServer(apiState);
-    dispatch(userActions.logOut());
+  logOut = () => {
+    const { _requestLogOut, dispatch } = this.props;
+    (dispatch as ThunkDispatch)(_requestLogOut());
```
bobsilverberg: I saw your comment about this and yes, it would be better if we could find a way to not have to do this type hinting. It seems like the component dispatching the action shouldn't even need to know if it's a "plain" action or a thunk action.

src/reducers/users.spec.tsx
```diff
@@ -53,4 +55,36 @@ describe(__filename, () => {
...
+    };
+
+    it('calls logOutFromServer', async () => {
```
bobsilverberg: I was a bit concerned about how we'd test the thunks but these seems to be a pretty good approach.

kumar303:
> It is a dumb question but: isn't it that `ConnectedReduxProps` declares `dispatch` as `Dispatch` and not `ThunkDispatch`?

Not dumb at all! The declaration of `ConnectedReduxProps` just needed an update -- that was the problem and I somehow misread your comment when I first replied to it. Thanks. 😓

kumar303:
@bobsilverberg @willdurand I think we should go with thunks. I still dislike how thunks are not traceable given actions alone but since they don't require generator functions, they are probably our best option, IMO. Could you take another look at the patch? I have abandoned my attempt at getting redux-loop to work (https://github.com/mozilla/addons-code-manager/pull/229).

src/test-helpers.tsx
```diff
@@ -197,3 +204,79 @@ export const shallowUntilTarget = (
...
+ * };
+ *
+ * You can replace this thunk for testing as:
+ *
+ * const fakeThunk = createFakeThunk();
+ * render({ _doLogout: fakeThunk.createThunk });
+ *
+ * You can make an assertion that it was called like:
+ *
+ * expect(dispatch).toHaveBeenCalledWith(fakeThunk.thunk);
```
willdurand: ❤️

kumar303:
Thanks for the review. @bobsilverberg I'm going to merge this to keep it going but let me know if it needs any follow-ups from your point of view. We can always revisit this side effect strategy as we begin using it more.

PR #189, opened 2019-02-12
Add add-on info to the `versions` reducer

### LINKED ISSUE #180: Add add-on info to the `versions` reducer
See this addons-server issue: https://github.com/mozilla/addons-server/issues/10624

The `versions` reducer should probably store the `addon` data.

bobsilverberg:
Fixes #180 

I'm wondering, though, what we want to do about localized strings. Making use of this new `VersionAddon` object will be the first time we have to deal with them (for the `name` property). The API returns localized strings to us, i.e., an object with a translation for different locales, but maybe that's not what we want?

I believe in addons-frontend we set up the API call so it would include a language and we'd only get the string back for the language we requested. Although we don't think we're going to support multiple languages in the app, it seems like this is probably the route we should go here. Otherwise our code is going to have to deal with these localized string objects.

What do people think?

### DIFF
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -82,7 +82,7 @@ describe(__filename, () => {
 
   describe('createInternalVersionAddon', () => {
     it('creates a VersionAddon', () => {
-      const addon = { ...fakeVersionAddon, id: 1234 };
+      const addon = fakeVersionAddon;
 
       expect(createInternalVersionAddon(addon)).toEqual({
         iconUrl: addon.icon_url,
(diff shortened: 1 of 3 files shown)


kumar303:
I guess this will need an update after https://github.com/mozilla/addons-code-manager/pull/202

bobsilverberg:
> I guess this will need an update after #202

👍 I'll rebase it on top of that one and update it.

bobsilverberg:
Ok, this has been rebased and should be ready for a review.

src/reducers/versions.spec.tsx
```diff
@@ -78,6 +80,19 @@ describe(__filename, () => {
     });
   });
 
+  describe('createInternalVersionAddon', () => {
+    it('creates a VersionAddon', () => {
+      const addon = { ...fakeVersionAddon, id: 1234 };
```
willdurand: Do you need to change `id`?

bobsilverberg: I guess not. I'll remove this.

Issue #164, opened 2019-02-07
Make query param handling more robust


kumar303:
As a follow-up to https://github.com/mozilla/addons-code-manager/issues/157, I'd like to see some improvements that will help us scale the codebase for future work:

1. Support query string values that have spaces or special characters. This should be as easy as:

```diff
diff --git a/src/api/index.tsx b/src/api/index.tsx
index ff6c939..9b58a53 100644
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -1,3 +1,4 @@
+import url from 'url';
 import log from 'loglevel';
 
 import { ApiState } from '../reducers/api';
@@ -45,11 +46,7 @@ export const callApi = async ({
   }
 
   if (Object.keys(query).length) {
-    const queryString = Object.keys(query)
-      .map((k) => `${k}=${query[k]}`)
-      .join('&');
-
-    adjustedEndpoint = `${adjustedEndpoint}?${queryString}`;
+    adjustedEndpoint = `${adjustedEndpoint}${url.format(query)}`;
   }
 
   try {
```

2. Make sure `undefined` or `null` object values do not pass empty parameters to the API, [like this](https://github.com/mozilla/addons-frontend/blob/aabbd8097c58ae1bc5bb48bc942a63bf6bfd2e12/src/core/api/index.js#L30). That has triggered Django errors in the past.

3. When testing query string parameters, we cannot rely on object key order, [as this does](https://github.com/mozilla/addons-code-manager/blob/9b93ab5627578426eafed58dde14735dd0a15ed3/src/api/index.spec.tsx#L142). It will *probably* work but it's non-deterministic and it's easy to avoid, anyway. Let's add [a helper similar to this](https://github.com/mozilla/addons-frontend/blob/aabbd8097c58ae1bc5bb48bc942a63bf6bfd2e12/tests/unit/helpers.js#L909-L931) to make query string assertions deterministic.

AlexandraMoga:
@willdurand I understand that this issue  and issue #666 were connected. I've verified that the error identified in #666 is no longer reproducible. 

I've also verified that navigation by file path is still working in the browse and compare pages.

If you think there is more to be verified here, please let me know. Otherwise, I will mark this issue as verified fixed.

willdurand:
> If you think there is more to be verified here, please let me know. Otherwise, I will mark this issue as verified fixed.

Looks good, thanks!

PR #202, opened 2019-02-13
Send an explicit language with API requests

### LINKED ISSUE #192: Send an explicit language with API requests
We're going to have to deal with localized strings for some properties, which look like this: `{ 'en-US': 'some text' }`. Probably the best thing to do is to always pass `en-US` as a language to the API, and then, when processing strings, always use the string that corresponds to `en-US`. For now we can make that a constant as we don't have plans to support any other languages.

This issue is about making sure we always pass a default language to the API.

bobsilverberg:
Fixes #192 

Note that after making this change, the data I am getting back from the API no longer contains localized strings. I believe this is because the `v4` API is currently configured to return simple strings, and not objects, in order to be backwards compatible with addons-frontend, and that there is a different endpoint we can use to get the localized strings (something like `dev4`, iirc). 

@eviljeff What is the endpoint we need to use to get localized strings back from the API, even if we pass in a `lang`, and is it currently configured to work with `addons-code-manager`?

### DIFF
--- a/.env
+++ b/.env
@@ -1,7 +1,8 @@
+REACT_APP_API_HOST=https://addons.mozilla.org
 # This is the placeholder used in the `index.html` file to inject the
 # authentication token on runtime.
 REACT_APP_AUTH_TOKEN_PLACEHOLDER=__AUTH_TOKEN__
-
-REACT_APP_API_HOST=https://addons.mozilla.org
-REACT_APP_FXA_CONFIG=amo
 REACT_APP_AUTHENTICATION_COOKIE=frontend_auth_token
+# This is the language that will be sent with all API requests.
+REACT_APP_DEFAULT_API_LANG=en-US
+REACT_APP_FXA_CONFIG=amo
--- a/src/api/index.spec.tsx
+++ b/src/api/index.spec.tsx
@@ -1,14 +1,15 @@
 /* global fetchMock */
+import configureStore from '../configureStore';
 import {
   actions as apiActions,
   initialState as defaultApiState,
 } from '../reducers/api';
-import configureStore from '../configureStore';
 
 import { HttpMethod, callApi, getVersion, logOutFromServer } from '.';
 
 describe(__filename, () => {
-  const defaultQueryString = `?lang=${process.env.REACT_APP_DEFAULT_API_LANG}`;
+  const defaultLang = process.env.REACT_APP_DEFAULT_API_LANG;
+  const defaultVersion = 'v4dev';
 
   const getApiState = ({ authToken = '12345' } = {}) => {
     const store = configureStore();
@@ -24,20 +25,34 @@ describe(__filename, () => {
       return callApi({ apiState: defaultApiState, endpoint: '/', ...params });
     };
 
-    it('calls the API using the default language', async () => {
-      await callApiWithDefaultApiState({ endpoint: '/foo/' });
+    it('calls the API with the expected defaults', async () => {
+      await callApiWithDefaultApiState();
 
-      expect(fetch).toHaveBeenCalledWith(`/api/v4/foo/${defaultQueryString}`, {
-        headers: {},
-        method: 'GET',
-      });
+      expect(fetch).toHaveBeenCalledWith(
+        `/api/${defaultVersion}/?lang=${defaultLang}`,
+        {
+          headers: {},
+          method: 'GET',
+        },
+      );
+    });
+
+    it('calls the API using the default language', async () => {
+      await callApiWithDefaultApiState();
+
+      expect(fetch).toHaveBeenCalledWith(
+        expect.urlWithTheseParams({
+          lang: process.env.REACT_APP_DEFAULT_API_LANG,
+        }),
+        expect.any(Object),
+      );
     });
 
     it('adds a trailing slash to the endpoint if there is none', async () => {
       await callApiWithDefaultApiState({ endpoint: '/foo' });
 
       expect(fetch).toHaveBeenCalledWith(
-        `/api/v4/foo/${defaultQueryString}`,
+        expect.stringMatching('/foo/'),
         expect.any(Object),
       );
     });
@@ -46,7 +61,7 @@ describe(__filename, () => {
       await callApiWithDefaultApiState({ endpoint: 'foo/' });
 
       expect(fetch).toHaveBeenCalledWith(
-        `/api/v4/foo/${defaultQueryString}`,
+        expect.stringMatching('/foo/'),
         expect.any(Object),
       );
     });
@@ -70,7 +85,7 @@ describe(__filename, () => {
       await callApiWithDefaultApiState({ endpoint: '/', version });
 
       expect(fetch).toHaveBeenCalledWith(
-        `/api/${version}/${defaultQueryString}`,
+        expect.stringMatching(`/api/${version}/`),
         expect.any(Object),
       );
     });
@@ -81,7 +96,7 @@ describe(__filename, () => {
       await callApiWithDefaultApiState({ endpoint: '/', lang });
 
       expect(fetch).toHaveBeenCalledWith(
-        `/api/v4/?lang=${lang}`,
+        expect.urlWithTheseParams({ lang }),
         expect.any(Object),
       );
     });
@@ -148,7 +163,7 @@ describe(__filename, () => {
 
       expect(response).toHaveProperty(
         'error',
-        new Error(`Unexpected status for GET /${defaultQueryString}: 400`),
+        new Error(`Unexpected status for GET /?lang=${defaultLang}: 400`),
       );
     });
 
@@ -158,7 +173,7 @@ describe(__filename, () => {
       await callApiWithDefaultApiState({ endpoint: '/url', query });
 
       expect(fetch).toHaveBeenCalledWith(
-        `/api/v4/url/${defaultQueryString}&foo=1&bar=abc`,
+        expect.urlWithTheseParams(query),
         expect.any(Object),
       );
     });
@@ -172,11 +187,10 @@ describe(__filename, () => {
       await getVersion({ apiState: defaultApiState, addonId, versionId });
 
       expect(fetch).toHaveBeenCalledWith(
-        `/api/v4/reviewers/addon/${addonId}/versions/${versionId}/${defaultQueryString}`,
-        {
-          headers: {},
-          method: HttpMethod.GET,
-        },
+        expect.stringMatching(
+          `/api/${defaultVersion}/reviewers/addon/${addonId}/versions/${versionId}/`,
+        ),
+        expect.any(Object),
       );
     });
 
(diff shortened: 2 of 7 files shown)


src/api/index.tsx
```diff
@@ -39,6 +41,9 @@ export const callApi = async ({
     adjustedEndpoint = `${adjustedEndpoint}/`;
   }
 
+  // Add the lang parameter to the querystring
+  adjustedEndpoint = `${adjustedEndpoint}?lang=${lang}`;
```
kumar303: Why not just set `query.lang = lang` ?

src/api/index.spec.tsx
```diff
@@ -139,7 +158,7 @@ describe(__filename, () => {
       await callApiWithDefaultApiState({ endpoint: '/url', query });
 
       expect(fetch).toHaveBeenCalledWith(
-        '/api/v4/url/?foo=1&bar=abc',
+        `/api/v4/url/${defaultQueryString}&foo=1&bar=abc`,
```
kumar303: I'd rather not keep investing in this testing style for two reasons:

1. As you can see here, you had to fix every test that compared query strings even though they had nothing to do with the `lang` feature.

2. Objects are hash tables. There is [no guarantee](https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order) that their keys will be in order.

The solution we had in addons-frontend solved both of these issues and I filed https://github.com/mozilla/addons-code-manager/issues/164 to add it to the code-manager. Since this patch doubles down on what we had previously, I'd like to see the assertions switch to a [helper like this](https://github.com/mozilla/addons-frontend/blob/aabbd8097c58ae1bc5bb48bc942a63bf6bfd2e12/tests/unit/helpers.js#L909-L931) at the very least. That helper lets you only check the necessary parameters and would address both points above.

kumar303:
> Note that after making this change, the data I am getting back from the API no longer contains localized strings.

What do you mean by localized strings? Do you get the simple strings or the objects? (Both would be localized strings.)

src/api/index.tsx
bobsilverberg: Because that would be too simple! 🍅

bobsilverberg:
> What do you mean by localized strings? Do you get the simple strings or the objects? (Both would be localized strings.)

I mean simple strings. Yes, they are strings that have been localized. My point was that I'm getting simple strings and not objects, but I believe we expect to get objects.

kumar303:
Working with simple strings should be fine, right? Why would we need objects? As long as the simple string falls back to the default locale when it's not in `en-US` then it should be fine (I think it does).

bobsilverberg:
> Working with simple strings should be fine, right? Why would we need objects? As long as the simple string falls back to the default locale when it's not in `en-US` then it should be fine (I think it does).

It would be fine with me, and certainly makes things simpler for the app, but my understanding was that the newest version of the API will never return simple strings and will always return objects. Maybe I'm wrong about that?

willdurand:
> It would be fine with me, and certainly makes things simpler for the app, but my understanding was that the newest version of the API will never return simple strings and will always return objects. Maybe I'm wrong about that?

`v4dev` is the API version to get the localized objects.

eviljeff:
yeah, `/api/v4/` (and /v3/) returns (and sets) simple strings anytime you send the lang param.  `/api/v4dev/` has the change I added to always return (and set) an object of lang:string values.  

IMO, if you're starting a new codebase it would be more future-proof to handle the v4dev style response.  (I'm biased because I spent the time writing the code but I think (a) it's more consistent and (b) you actually know what locale you got back rather than, for example, asking for `fr` and getting back `ar` and just assuming it's `fr`, and (c) setting values is less edge-casey, though perhaps not a problem here).  

We can make `/api/v4dev/` available in stage (and prod) if that helps - or promote it to `/api/v5/` even, though last time that stranded addons-frontend on the previous version because it couldn't handle the new style response.

kumar303:
Good point, @eviljeff. We should aim to support the future-style which will be to always deal with localized objects. 

@bobsilverberg my suggestion is this: if there's an easy way to use localized objects in the current state of things (deployed API versions, etc), go for it. If there's not an easy way, just go with simple strings and file a follow-up bug to convert everything to localized objects as soon as we can.

bobsilverberg:
@kumar303 It looks like we can just use `v4dev` for now, and we get objects instead of strings, so that's what I'm going to do.

bobsilverberg:
Ok, I've switched to the `v4dev` API and I've added a Jest matcher for `urlWithTheseParams`. One problem I'm having is that the Typescript compiler is not recognizing my definition for `urlWithTheseParams` for the tests for the API code. It is recognizing the definition for the tests for `jest-matchers/index.tsx` though, which I accomplished by adding the `jest.d.ts` file.

I've tried moving that file around, and read a number of issues, posts and SO articles explaining how to get this to work, including https://github.com/americanexpress/jest-image-snapshot/issues/33, https://stackoverflow.com/questions/43667085/extending-third-party-module-that-is-globally-exposed and others, and none of those things seem to work.

If anyone can figure out how to do this I would really appreciate it.

src/api/index.spec.tsx
bobsilverberg: I introduced a jest matcher to deal with this.

.env
```diff
@@ -1,7 +1,8 @@
...
 # This is the placeholder used in the `index.html` file to inject the
 # authentication token on runtime.
 REACT_APP_AUTH_TOKEN_PLACEHOLDER=__AUTH_TOKEN__
-
-REACT_APP_API_HOST=https://addons.mozilla.org
-REACT_APP_FXA_CONFIG=amo
 REACT_APP_AUTHENTICATION_COOKIE=frontend_auth_token
+# This is the language that will be sent with all API requests.
+REACT_APP_DEFAULT_API_LANG=en-US
+REACT_APP_FXA_CONFIG=amo
```
willdurand: Any reason to change the order of the variables?

src/api/index.tsx
```diff
@@ -28,8 +29,9 @@ export const callApi = async ({
   apiState,
   endpoint,
   method = HttpMethod.GET,
-  version = 'v4',
+  version = 'v4dev',
```
willdurand: Maybe we could also move this to a `REACT_APP_DEFAULT_API_VERSION` variable, WDYT? Given that it is done for the `lang` parameter, that would avoid an hardcoded value in the test file.

src/api/index.tsx
```diff
@@ -44,13 +46,14 @@ export const callApi = async ({
     headers.Authorization = `Bearer ${apiState.authToken}`;
   }
 
-  if (Object.keys(query).length) {
-    const queryString = Object.keys(query)
-      .map((k) => `${k}=${query[k]}`)
-      .join('&');
+  // Add the lang parameter to the querystring
+  const queryWithLang = Object.assign({ lang }, query);
```
willdurand: How about:

```js
const queryWithLang = {
   lang,
   ...query,
};
```

src/jest-matchers/index.spec.tsx
```diff
@@ -0,0 +1,70 @@
+const baseUrl = 'http://example.com';
+const param1 = 'param1';
+const param2 = 'param2';
+const testParams = { param1, param2 };
+
+describe('.urlWithTheseParams', () => {
```
willdurand: We should probably have a top-level `describe` block with `__filename`. I am surprised that ESLint did not catch that 🤔 (cf. https://github.com/mozilla/eslint-plugin-amo#describe-with-filename)

src/jest-matchers/index.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+  test('passes when URL includes all params', () => {
```
willdurand: could you use `it`?

src/jest-matchers/index.spec.tsx
willdurand: Note for myself: that's maybe because this file does not have a `module` scope for some reasons, maybe because there is no import/export so its scope is something else...

.env
bobsilverberg: I put them in alphabetical order. Without a specific reason for an order, I feel like that's a good practice.

src/api/index.tsx
bobsilverberg: This is nice, but I reversed the order, so that `lang` will override anything in `query`.

src/jest-matchers/index.spec.tsx
bobsilverberg: This file is actually causing a problem for the same reason you describe in your second comment. I'm getting the following `tsc` error:

```
src/jest-matchers/index.spec.tsx:1:1 - error TS1208: Cannot compile namespaces when the '--isolatedModules' flag is provided.

1 describe(__filename, () => {
  ~~~~~~~~
```

It seems like this happens because `tsc` thinks this is a global module. I can fix it by changing `isolatedModules` to `false` in `compilerOptions` in `tsconfig.json`, but I'm not sure if we want to do that. It can also be fixed by including an `import` or `export` in the test file, but we don't actually need either of those.

What do you suggest?

bobsilverberg: Note, I found these solutions via https://stackoverflow.com/questions/33977054/cannot-compile-namespaces-when-the-isolatedmodules-flag-is-provided

bobsilverberg:
Thanks for the review @willdurand. This is ready for another look. It still includes the `tsc` failure I discussed at https://github.com/mozilla/addons-code-manager/pull/202#discussion_r258671667 though.

src/jest-matchers/index.spec.tsx
bobsilverberg: I dealt with this by just adding `export {};` at the bottom of the file.

src/api/index.spec.tsx
```diff
@@ -1,13 +1,16 @@
...
 import {
   actions as apiActions,
   initialState as defaultApiState,
 } from '../reducers/api';
-import configureStore from '../configureStore';
 
 import { HttpMethod, callApi, getVersion, logOutFromServer } from '.';
 
 describe(__filename, () => {
+  const defaultLang = process.env.REACT_APP_DEFAULT_API_LANG as string;
```
willdurand: why does that need `as string`?

src/jest-matchers/index.spec.tsx
```diff
@@ -0,0 +1,76 @@
...
+      expect(() =>
+        expect(
+          `${baseUrl}?param2=${param2}&param1=${param1}`,
+        ).not.urlWithTheseParams(testParams),
+      ).toThrow('expected');
+    });
+  });
+});
+
+export {};
```
willdurand: We should add a comment to explain why this is needed, i.e. because there is no other import/export, TS fails to parse it correctly.

src/api/index.spec.tsx
bobsilverberg: I think I needed it at one point, but I don't anymore. It was probably a red herring. I've removed it.

PR #213, opened 2019-02-15
Add tests for the insecure proxy

### LINKED ISSUE #212: Add tests for the REACT_APP_USE_INSECURE_PROXY server logic
We should add more tests for the server logic. This is similar to #133 but this issue focuses on the `REACT_APP_USE_INSECURE_PROXY` feature.

willdurand:
Fixes #212

### DIFF
--- a/src/server/index.spec.tsx
+++ b/src/server/index.spec.tsx
@@ -3,7 +3,7 @@ import http from 'http';
 import path from 'path';
 
 import express from 'express';
-import request from 'supertest';
+import request, { SuperTest, Test } from 'supertest';
 import { RequestWithCookies } from 'universal-cookie-express';
 import Cookies from 'universal-cookie';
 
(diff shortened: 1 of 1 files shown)


src/server/index.spec.tsx
```diff
@@ -107,6 +107,81 @@ describe(__filename, () => {
...
+              rootPath,
+            }),
+          );
+
+          const response = await server.get('/api/foo');
+
+          expect(response.text).toEqual(apiResponseBody);
+        });
+
+        it('removes the secure attribute of cookies sent by the API', async () => {
```
kumar303: It would be nice to also get a test for the else branch of this statement:

https://github.com/mozilla/addons-code-manager/blob/8836cba68a3f561b1bc46d53c5453a5d5805e6b3/src/server/index.tsx#L79

Currently the else branch does not seem to be executed by the test suite.

willdurand: ah right, thanks!

PR #214, opened 2019-02-15
Add story for FileNodeTree

### LINKED ISSUE #191: Add story for the FileTreeNode component
We should add a storybook story for the `FileTreeNode` component.

willdurand:
Fixes #191

### DIFF
--- a/stories/FileTreeNode.stories.tsx
+++ b/stories/FileTreeNode.stories.tsx
@@ -1,0 +1,52 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+import FileTreeNode from '../src/components/FileTreeNode';
+
+const getProps = ({
+  hasChildNodes = false,
+  isExpanded = false,
+  isFolder = false,
+  level = 0,
+  nodeName = 'node name',
+  renderChildNodes = () => ({}),
+} = {}) => {
+  return {
+    getToggleProps: () => ({
+      onClick: () => {},
+      onKeyDown: () => {},
+      role: 'button',
+      tabIndex: 0,
+    }),
+    hasChildNodes,
+    isExpanded,
+    isFolder,
+    level,
+    node: {
+      id: 'node-id',
+      name: nodeName,
+    },
+    renderChildNodes,
+  };
+};
+
+storiesOf('FileTreeNode', module)
+  .add('file node', () => <FileTreeNode {...getProps()} />)
+  .add('directory node', () => (
+    <FileTreeNode {...getProps({ isFolder: true })} />
+  ))
+  .add('directory, expanded, without children', () => (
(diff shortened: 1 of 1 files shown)


stories/setup/config.ts
```diff
@@ -1,10 +1,15 @@
...
+import { library } from '@fortawesome/fontawesome-svg-core';
+import { fas } from '@fortawesome/free-solid-svg-icons';
 
 // Include application styles.
 import '../../src/styles.scss';
 // Apply some custom styles to storybook.
 import './styles.scss';
 
+// Import all the "free solid FontAwesome" icons
+library.add(fas);
```
kumar303: This has the potential to get out of sync with the real app so I'd rather see it moved to a shared function. I think all app configuration can be moved to a shared function. Storybook doesn't need all of this but it should be harmless to run the code:

```diff
diff --git a/src/index.tsx b/src/index.tsx
index caa7053..76faff4 100644
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -11,22 +11,8 @@ import './styles.scss';
 import App from './components/App';
 import configureStore from './configureStore';
 
-if (process.env.NODE_ENV === 'production') {
-  // The second parameter prevents the log level to be persisted in a cookie or
-  // localStorage.
-  log.setLevel(log.levels.INFO, false);
-} else {
-  log.setLevel(log.levels.DEBUG, false);
-}
-
-if (process.env.REACT_APP_SENTRY_DSN) {
-  Raven.config(process.env.REACT_APP_SENTRY_DSN, {
-    logger: 'client',
-  }).install();
-}
-
-// Import all the "free solid FontAwesome" icons
-library.add(fas);
+configureApplication();
 
 const store = configureStore();
 
```

willdurand: ah thanks, I wanted to file an issue for that too or, at least, leave a comment about the ever-growing code in `src/index.tsx`, which isn't super testable...

willdurand: I will probably try to do that first, likely in a separate patch.

willdurand:
> Also, this could be done in a follow-up: I prefer to see stories all on one page when possible so one could use hot reloading to develop changes more easily. We were using [Checkfront/react-storybook-addon-chapters](https://github.com/Checkfront/react-storybook-addon-chapters) for this on the frontend but maybe there is a simpler way.

That makes sense. I did not see it in the current setup, so I thought we would keep things simple and use what Storybook recommended.

stories/setup/config.ts
willdurand: Patch here: https://github.com/mozilla/addons-code-manager/pull/228

willdurand:
> **Edit:** I filed #226

I'll take care of it in a separate patch, unless someone disagrees.

willdurand:
I am working on #226 right now, good timing. Thanks!

Issue #182, opened 2019-02-11
Handle API errors in App, Browse, and Compare components


willdurand:
We did not really handle API errors so far, mostly because we don't have an error layer for that. Let's discuss this layer here.

- Do we want to introduce a layer similar to the error handler code in addons-frontend?
- Do we want to do something else? If so, what?

bobsilverberg:
I do think we want something, but I think we should keep it as simple as possible for now. I don't have an immediate suggestion, other than that.

EnTeQuAk:
As part of this discussion, I'm thinking about adding on-demand extraction for files that may have failed extraction during upload (for some reason, or these are legacy / unlisted add-ons or Mozilla signed add-ons). And in addition to that, I'll double-review possible errors that I'll try to forward as detailed as possible to the API so that reviewers get properly notified.

We should handle errors, luckily we can include much internal information to give reviewers a better sense of what failed given the importance of review errors should be clear and informative and ideally actionable.

willdurand:
This has been partially fixed in #181.

PR #218, opened 2019-02-15
Add better return types for the API functions

### LINKED ISSUE #181: Add better return types to API functions
Instead of using the `callApi` default return type, we could use more specific return types. I think we can achieve that by making `callApi` a bit more generic:

```js
export const callApi<ReturnType> = async (...): Promise<CallApiResponse<ReturnType>> => {
  //...
}
```

willdurand:
Fixes #181 

---

This PR allows us to write code like this now: https://github.com/mozilla/addons-code-manager/pull/218/files#diff-eaeb6d1753d7786e20e90d2089ffe11e. In other words, we now have precise return types for our API calls.

The PR also adds new thunks and a bunch of test cases.

### DIFF
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -1,6 +1,8 @@
 import log from 'loglevel';
 
 import { ApiState } from '../reducers/api';
+import { ExternalUser } from '../reducers/users';
+import { ExternalVersion } from '../reducers/versions';
 
 export enum HttpMethod {
   DELETE = 'DELETE',
@@ -19,20 +21,37 @@ type CallApiParams = {
   lang?: string;
 };
 
-type CallApiResponse = object | { error: Error };
+type ErrorResponseType = { error: Error };
+
+type CallApiResponse<SuccessResponseType> =
+  | SuccessResponseType
+  | ErrorResponseType;
+
+// This function below is a user-defined type guard that we use to check
+// whether a callApi response object is successful or unsuccessful. We use
+// `any` on purpose because we don't know the exact type for
+// `SuccessResponseType`.
+export const isErrorResponse = (
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  arg: CallApiResponse<any>,
+): arg is ErrorResponseType => {
+  return arg.error !== undefined;
+};
 
 type Headers = {
   [name: string]: string;
 };
 
-export const callApi = async ({
+// For the `extends {}` part, please see:
+// https://github.com/Microsoft/TypeScript/issues/4922
+export const callApi = async <SuccessResponseType extends {}>({
   apiState,
   endpoint,
   method = HttpMethod.GET,
   version = process.env.REACT_APP_DEFAULT_API_VERSION,
   query = {},
   lang = process.env.REACT_APP_DEFAULT_API_LANG as string,
-}: CallApiParams): Promise<CallApiResponse> => {
+}: CallApiParams): Promise<CallApiResponse<SuccessResponseType>> => {
   let adjustedEndpoint = endpoint;
   if (!adjustedEndpoint.startsWith('/')) {
     adjustedEndpoint = `/${adjustedEndpoint}`;
@@ -96,7 +115,7 @@ export const getVersion = async ({
   addonId,
   versionId,
 }: GetVersionParams) => {
-  return callApi({
+  return callApi<ExternalVersion>({
     apiState,
     endpoint: `reviewers/addon/${addonId}/versions/${versionId}`,
     query: path ? { file: path } : undefined,
@@ -104,9 +123,16 @@ export const getVersion = async ({
 };
 
 export const logOutFromServer = async (apiState: ApiState) => {
-  return callApi({
+  return callApi<{}>({
     apiState,
     endpoint: 'accounts/session',
     method: HttpMethod.DELETE,
+  });
+};
+
+export const getCurrentUserProfile = async (apiState: ApiState) => {
+  return callApi<ExternalUser>({
+    apiState,
+    endpoint: '/accounts/profile/',
   });
 };
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -50,7 +50,7 @@ export const callApi = async <SuccessResponseType extends {}>({
   method = HttpMethod.GET,
   version = process.env.REACT_APP_DEFAULT_API_VERSION,
   query = {},
-  lang = process.env.REACT_APP_DEFAULT_API_LANG as string,
+  lang = process.env.REACT_APP_DEFAULT_API_LANG,
 }: CallApiParams): Promise<CallApiResponse<SuccessResponseType>> => {
   let adjustedEndpoint = endpoint;
   if (!adjustedEndpoint.startsWith('/')) {
@@ -69,8 +69,8 @@ export const callApi = async <SuccessResponseType extends {}>({
     [key: string]: string;
   };
 
-  // Add the lang parameter to the querystring
-  const queryWithLang: QueryWithLang = { ...query, lang };
+  // Add the lang parameter to the query string.
+  const queryWithLang: QueryWithLang = lang ? { ...query, lang } : query;
 
   const queryString = Object.keys(queryWithLang)
     .map((k) => `${k}=${queryWithLang[k]}`)
--- a/src/components/App/index.spec.tsx
+++ b/src/components/App/index.spec.tsx
@@ -6,9 +6,13 @@ import { Route } from 'react-router-dom';
 import styles from './styles.module.scss';
 import configureStore from '../../configureStore';
 import { actions as apiActions } from '../../reducers/api';
-import { actions as userActions } from '../../reducers/users';
+import {
+  actions as userActions,
+  fetchCurrentUserProfile,
+} from '../../reducers/users';
 import {
   createContextWithFakeRouter,
+  createFakeThunk,
   fakeUser,
   shallowUntilTarget,
 } from '../../test-helpers';
@@ -18,13 +22,15 @@ import App, { AppBase } from '.';
 
 describe(__filename, () => {
   type RenderParams = {
+    _fetchCurrentUserProfile?: typeof fetchCurrentUserProfile;
     store?: Store;
     authToken?: string | null;
   };
 
   const render = ({
-    store = configureStore(),
+    _fetchCurrentUserProfile,
     authToken = 'some-token',
+    store = configureStore(),
   }: RenderParams = {}) => {
     const contextWithRouter = createContextWithFakeRouter();
     const context = {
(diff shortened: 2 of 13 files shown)


src/api/index.tsx
```diff
@@ -87,17 +106,24 @@ export const getVersionFile = async ({
...
+  return callApi<ExternalVersion>({
     apiState,
     endpoint: `reviewers/browse/${versionId}`,
     query: path ? { file: path } : undefined,
   });
 };
 
 export const logOutFromServer = async (apiState: ApiState) => {
-  return callApi({
+  return callApi<{}>({
```
kumar303: Hmm, I think it's actually an empty response. Maybe that doesn't matter.

src/components/App/index.tsx
```diff
@@ -55,12 +54,9 @@ export class AppBase extends React.Component<Props> {
...
 
     if (!profile && prevProps.apiState.authToken !== apiState.authToken) {
-      const response = (await api.callApi({
-        apiState,
-        endpoint: '/accounts/profile/',
-      })) as ExternalUser;
+      const response = await getCurrentUserProfile(apiState);
 
-      if (response && response.name) {
+      if (!isErrorResponse(response)) {
```
kumar303: This seems like the wrong place to defend against errors. It's awkward and adds unnecessary complexity. I know your code is no worse than the code it replaces but I would rather see something like this:

```diff
diff --git a/src/api/index.spec.tsx b/src/api/index.spec.tsx
index 3597be2..69c95d2 100644
--- a/src/api/index.spec.tsx
+++ b/src/api/index.spec.tsx
@@ -210,6 +210,8 @@ describe(__filename, () => {
 
   describe('getCurrentUserProfile', () => {
     it('calls the API to retrieve the current logged-in user profile', async () => {
+      // Something like this needs to get applied everywhere.
+      fetchMock.mockResponse(JSON.stringify({}));
       await getCurrentUserProfile(defaultApiState);
 
       expect(fetch).toHaveBeenCalledWith(`/api/v4/accounts/profile/`, {
diff --git a/src/api/index.tsx b/src/api/index.tsx
index a75f91d..02a8e21 100644
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -71,6 +71,7 @@ export const callApi = async <SuccessResponseType extends {}>({
     adjustedEndpoint = `${adjustedEndpoint}?${queryString}`;
   }
 
+  let responseData;
   try {
     const response = await fetch(`/api/${version}${adjustedEndpoint}`, {
       method,
@@ -85,14 +86,19 @@ export const callApi = async <SuccessResponseType extends {}>({
       );
     }
 
-    return await response.json();
+    responseData = await response.json();
   } catch (error) {
-    log.debug('Error caught in callApi():', error);
-
-    return {
+    responseData = {
       error,
     };
   }
+
+  if (isErrorResponse(responseData)) {
+    log.debug('Error returned from callApi():', responseData.error);
+    throw new Error(`callApi() error: ${responseData.error}`);
+  }
+
+  return responseData;
 };
 
 type GetVersionFileParams = {
```

There might be cases where the code executing `callApi()` will expect an error object but I don't think we have that code yet. When we need it, we can add a parameter like `expectError: true` or something.

src/api/index.tsx
```diff
@@ -18,19 +20,36 @@ type CallApiParams = {
...
+  | SuccessResponseType
+  | ErrorResponseType;
+
+// This function below is a user-defined type guard that we use to check
+// whether a callApi response object is successful or unsuccessful. We use
+// `any` on purpose because we don't know the exact type for
+// `SuccessResponseType`.
+export const isErrorResponse = (
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  arg: CallApiResponse<any>,
```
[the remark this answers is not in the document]
willdurand: No, because otherwise TS would have expectations on the shape of `arg`.

src/api/index.tsx
willdurand: Indeed, the API returns a 204 but I didn't know how to express that. Should we use `null`?

src/components/App/index.tsx
willdurand: I still believe we should not `throw` in `callApi`, but this depends on how we plan to handle errors. I guess #182 just got a higher priority now :)

willdurand:
> This seems useful for detecting errors but I'd like to not keep investing in all of this code that ignores errors. I suggested an approach that would throw all errors but if you think we should tackle that in a follow-up instead just let me know.

Yes, it is time to resolve #182 now.

willdurand:
> Yes, it is time to resolve #182 now.

@kumar303 one idea built on top of this PR is this: https://github.com/mozilla/addons-code-manager/commit/8fbbb9dc85d1dd5b3977a1704359b6540bec6bc6

screen shot 2019-02-18 at 14 22 50

We dispatch all (known) errors and store them into a centralized `notifications` reducer, as "error" notifications. We could use the same reducer for other notifications, e.g., `success` messages.

kumar303:
> @kumar303 one idea built on top of this PR is this: [8fbbb9d](https://github.com/mozilla/addons-code-manager/commit/8fbbb9dc85d1dd5b3977a1704359b6540bec6bc6)
> ...
> We dispatch all (known) errors and store them into a centralized `notifications` reducer, as "error" notifications.

I like this approach. 

I think ultimately we should go with that approach but there may be some complexities in how we deal with rendering / dismissing notices.

How about taking a baby step towards that approach like this?

```js
    if (isErrorResponse(response)) {
      log.error(`TODO: handle this error response: ${response.error}`);
    } else {
      dispatch(versionActions.loadVersionInfo({ version: response }));
    }
```

willdurand:
> How about taking a baby step towards that approach like this?

💯 👍 

I'll update this patch with that, and I'll try to think more about the other patch with the `notifications` reducer.

willdurand:
@kumar303 so I think this is good for another look. I had to introduce a `updateWithProps()` helper because Enzyme does not support `async componentDidUpdate()`... (it does support `async componentDidMount()` though). I added tests for everything I touched.

src/components/App/index.spec.tsx
```diff
@@ -35,71 +40,125 @@ describe(__filename, () => {
...
       shallowOptions: { ...context },
     });
-
-    return root;
   };
 
-  it('renders without an authentication token', () => {
-    const root = render({ authToken: null });
+  it('renders without an authentication token', async () => {
+    const root = await render({ authToken: null });
```
kumar303: Why do you need `await` here? In the component itself, `componentDidMount` is not async, only `componentDidUpdate`.

src/components/App/index.spec.tsx
```diff
@@ -35,71 +40,125 @@ describe(__filename, () => {
...
+    const root = await render({ _log, authToken: null, store });
+
+    // Set the authentication token into the state.
+    store.dispatch(apiActions.setAuthToken({ authToken }));
+    const { api: apiState } = store.getState();
+    dispatch.mockReset();
+
+    // Wait until componentDidUpdate() logic has been executed with the new
+    // props.
+    await updateWithProps(root, { apiState });
```
kumar303: I'd rather not go down this route because once we have https://github.com/mozilla/addons-code-manager/issues/12 then we won't need it. Everything can be tested synchronously. However, I guess it's OK as a first take. 

As for side effects, I'm going to suggest using thunks because I don't have a better idea. I will be cleaning up https://github.com/mozilla/addons-code-manager/pull/176 asap.

src/components/App/index.spec.tsx
willdurand: ahm, correct. This is not needed 😞

src/components/App/index.spec.tsx
kumar303: I cleaned up the patch for thunks (https://github.com/mozilla/addons-code-manager/pull/176). Maybe you can build on top of that? If you convert to thunks, I don't think you'll need any async handling in the component or the tests.

willdurand: I can certainly build on top of the thunk patch, I'll do a final review first.

willdurand:
@kumar303 this should be ready for another look! It's full of thunks 🌟

src/reducers/users.spec.tsx
```diff
@@ -77,4 +78,55 @@ describe(__filename, () => {
       expect(dispatch).toHaveBeenCalledWith(actions.logOut());
     });
   });
+
+  describe('fetchCurrentUserProfile', () => {
+    it('calls getCurrentuserProfile', async () => {
```
kumar303: typo: getCurrentUserProfile

src/components/App/index.tsx
```diff
@@ -51,18 +61,13 @@ export class AppBase extends React.Component<Props> {
     }
   }
 
-  async componentDidUpdate(prevProps: Props) {
-    const { apiState, dispatch, profile } = this.props;
+  componentDidUpdate(prevProps: Props) {
+    const { apiState, profile } = this.props;
 
     if (!profile && prevProps.apiState.authToken !== apiState.authToken) {
```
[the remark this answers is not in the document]
willdurand: Agreed. Should we file an issue? This patch gets too big.

src/components/App/index.tsx
```diff
@@ -38,11 +42,17 @@ type PropsFromState = {
 /* eslint-disable @typescript-eslint/indent */
 type Props = PublicProps &
   PropsFromState &
+  DefaultProps &
```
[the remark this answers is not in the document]
willdurand: 🤷‍♂️ I am fine with merging `DefaultProps` into `PublicProps`. I expect `PublicProps` to be the public API for developers and they should not inject testing-related props.

src/pages/Browse/index.tsx
```diff
@@ -27,23 +27,26 @@ type PropsFromState = {
...
   ConnectedReduxProps;
 /* eslint-enable @typescript-eslint/indent */
 
 export class BrowseBase extends React.Component<Props> {
+  static defaultProps = {
+    _fetchVersion: fetchVersion,
+    _log: log,
+  };
+
   async componentDidMount() {
```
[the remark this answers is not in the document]
willdurand: D'oh! I am sorry.

willdurand:
> I miss `sinon.spy(store, 'dispatch')` a little. I tried wrapping up the jest methods in a helper function but couldn't figure out how to mirror the function parameters of `jest.spyOn`. TypeScript!
> 
> My failed attempt 😞

I added this function. I tweaked it a bit to make it TS-compliant.

src/components/App/index.spec.tsx
```diff
@@ -102,4 +113,70 @@ describe(__filename, () => {
...
+    expect(dispatch).not.toHaveBeenCalled();
+  });
+
+  it('does not fetch the current user profile on update when the profile has been already loaded', () => {
+    const store = configureStore();
+    store.dispatch(userActions.loadCurrentUser({ user: fakeUser }));
+    const fakeThunk = createFakeThunk();
+
+    const root = render({
+      _fetchCurrentUserProfile: fakeThunk.createThunk,
```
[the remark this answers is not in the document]
willdurand: Thanks for this suggestion, that's much better!

willdurand:
@kumar303 thanks a lot for the last review, I hope this is good to go now.

src/components/App/index.tsx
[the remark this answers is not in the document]
kumar303: Definitely. Not high priority: https://github.com/mozilla/addons-code-manager/issues/246

PR #220, opened 2019-02-15
Update code to reflect the latest API changes

### LINKED ISSUE #219: The Browse API has changed
The browse API has changed, let's update our code to reflect that: https://addons-server.readthedocs.io/en/latest/topics/api/reviewers.html#browse.

willdurand:
Fixes #219

### DIFF
--- a/src/components/App/index.tsx
+++ b/src/components/App/index.tsx
@@ -75,7 +75,7 @@ export class AppBase extends React.Component<Props> {
         <Route
           component={Browse}
           exact
-          path="/:lang/:application/files/browse/:versionId/"
+          path="/:lang/browse/:addonId/versions/:versionId/"
         />
         <Route component={NotFound} />
       </Switch>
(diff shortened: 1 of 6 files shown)


src/components/App/index.tsx
```diff
@@ -75,7 +75,7 @@ export class AppBase extends React.Component<Props> {
         <Route
           component={Browse}
           exact
-          path="/:lang/:application/files/browse/:versionId/"
+          path="/:lang/browse/:addonId/versions/:versionId/"
```
kumar303: Is the existing DevHub going to switch to this URL? It still looks like the old URL on dev. I believe we need to keep the URLs in sync.

willdurand: We discussed this during a call. We can use any URL because it is internal only, and we will never share this URL publicly. We agreed on changing the URL, removing the `application` parameter because it was not needed.

That being said, we did not change the URL in DevHub because we did not deploy this project on an hosted env yet. The old/current code review tool is served and linked in the -dev DevHub.

Last but not least, the API expects an `addonId` now so we have to have one in the URL, but this is a new behavior, different than the DevHub (which does not rely on the API). That's why I went ahead and updated this URL. We'll have to use this URL in DevHub when this project will be deployed on an hosted env and we won't be shy to actually link to it in DevHub.

PR #225, opened 2019-02-18
Ignore src/index.tsx from coverage

willdurand:
I don't think we can cover this file with tests, so let's ignore it to have a more accurate coverage report.

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,7 +118,8 @@
     "collectCoverageFrom": [
       "src/**/*.{js,jsx,ts,tsx}",
       "!<rootDir>/node_modules/",
-      "!src/jest-reporters/eslint-check.js"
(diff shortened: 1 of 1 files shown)


willdurand:
The suggestion in https://github.com/mozilla/addons-code-manager/pull/214#discussion_r258109772 + this patch would be a wonderful combination 😛

kumar303:
I think we won't have to ignore this file after the content is more testable (via https://github.com/mozilla/addons-code-manager/pull/228).

willdurand:
> Based on my last comment, I don't think we'll need to do this. I meant to submit the comment as a review to clear the request from my queue.

@kumar303 there is still code that won't be covered because it binds our app with the DOM on pageload.

PR #228, opened 2019-02-19
Extract app setup in a function

### LINKED ISSUE #227: Extract app setup into a `configureApplication` function
By extracting some code from `src/index.tsx`, we can test it and reuse it elsewhere, e.g., in the storybook setup. Let's do this!

willdurand:
Fixes #227

### DIFF
--- a/src/configureApplication.tsx
+++ b/src/configureApplication.tsx
@@ -1,0 +1,42 @@
+import log from 'loglevel';
+import Raven from 'raven-js';
+import { library } from '@fortawesome/fontawesome-svg-core';
+import { fas } from '@fortawesome/free-solid-svg-icons';
+
+export type ClientEnvVars = {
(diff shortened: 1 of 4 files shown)


src/configureApplication.tsx
```diff
@@ -0,0 +1,42 @@
...
+type ConfigureApplicationParams = {
+  _Raven?: typeof Raven;
+  _log?: typeof log;
+  env: ClientEnvVars;
+};
+
+const configureApplication = ({
+  _Raven = Raven,
+  _log = log,
+  env,
```
kumar303: This looks like it can get a default, like `env = (process.env as any) as ClientEnvVars`, right?

willdurand: ah correct, I don't know why I did not do that in the first place 🤦‍♂️ . Fixed, thanks!

PR #229, opened 2019-02-19
[do not merge] attempt at using redux-loop

kumar303:
I tried to assess redux-loop for https://github.com/mozilla/addons-code-manager/issues/12 but I'm about to give up. Here's my progress so far if anyone wants to see it.

The aspect I liked about redux-loop is we can theoretically reconstruct the whole flow of an app just with a stack of actions. 

However, I can't get things to work with TypeScript and I'm not convinced that it's worth battling this much further. It seems that redux-loop wants to make a fundamental change to how Redux reducers work. This causes a lot of problems, especially in type definitions. The type definitions for redux-loop do not seem well designed -- they redefine a lot of Redux types. They have [bugs](https://github.com/redux-loop/redux-loop/commits?author=kumar303) in them and are out of sync with the current Redux library. Even if we get it to work, the future might be plagued by maintenance.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "redux": "4.0.1",
     "redux-devtools-extension": "2.13.8",
     "redux-logger": "3.0.6",
+    "redux-loop": "4.5.4",
     "ts-node": "8.0.2",
     "typesafe-actions": "3.1.0",
     "typescript": "3.3.3",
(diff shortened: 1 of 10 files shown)


bobsilverberg:
I think you might as well close this RP @kumar303, as it seems we've decided to proceed with redux-thunk.

kumar303:
Yes, that's true. I'll add a more detailed write-up of choosing redux-thunk in the issue.

PR #232, opened 2019-02-20
Add selected_file to versions reducer (as selectedPath)

### LINKED ISSUE #231: Store `selected_file` in `versions` reducer
The Browse API now returns the `selected_file`: https://addons-server.readthedocs.io/en/latest/topics/api/reviewers.html#browse. This is a useful value, so let's store it.

willdurand:
Fixes #231

---

This PR adds a new field named `selectedPath` to the `versions` reducer. It stores the `file.selected_file` value. Given that retrieving a version without a `path` returns a default file, I also decided to store it. Last, I added a `getVersionFiles()` selector because we'll need it in the near future.

In the near future, we'll likely have a `selectPath` action.

### DIFF
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -191,15 +191,20 @@ describe(__filename, () => {
 
   describe('getVersionFiles', () => {
     it('returns all the files for a given version', () => {
-      const path = 'test.js';
+      const path1 = 'test.js';
+      const path2 = 'function.js';
       const version = fakeVersion;
-      const state = reducer(
-        undefined,
-        actions.loadVersionFile({ path, version }),
-      );
 
+      let state = reducer(
+        undefined,
+        actions.loadVersionFile({ path: path1, version }),
+      );
+      state = reducer(state, actions.loadVersionFile({ path: path2, version }));
+
+      const internalVersionFile = createInternalVersionFile(version.file);
       expect(getVersionFiles(state, version.id)).toEqual({
-        [path]: createInternalVersionFile(version.file),
+        [path1]: internalVersionFile,
+        [path2]: internalVersionFile,
       });
     });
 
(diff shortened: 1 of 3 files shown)


src/reducers/versions.spec.tsx
```diff
@@ -179,6 +189,25 @@ describe(__filename, () => {
     });
   });
 
+  describe('getVersionFiles', () => {
+    it('returns all the files for a given version', () => {
```
bobsilverberg: Should we store multiple files for a version in order to be able to assert that this does in fact return *all files* for a version?

src/reducers/versions.tsx
```diff
@@ -194,16 +205,26 @@ const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
...
     case getType(actions.loadVersionInfo): {
       const { version } = action.payload;
+
       return {
         ...state,
         versionInfo: {
           ...state.versionInfo,
           [version.id]: createInternalVersion(version),
         },
+        versionFiles: {
```
bobsilverberg: I wasn't really sure that we needed to do this, which is one of the reasons I didn't do it initially, because when we load a version with the default file we probably won't be displaying it. But I guess there's nothing really wrong with doing this, and it does give us the option to display it if we want to.

src/reducers/versions.tsx
```diff
@@ -96,6 +97,7 @@ export type Version = {
   id: VersionId;
   reviewed: string;
   version: string;
+  selectedPath: string;
```
bobsilverberg: Nit: Can we keep these in alphabetical order?

src/reducers/versions.tsx
```diff
@@ -168,15 +170,24 @@ export const createInternalVersion = (version: ExternalVersion): Version => {
     id: version.id,
     reviewed: version.reviewed,
     version: version.version,
+    selectedPath: version.file.selected_file,
```
bobsilverberg: Nit: Can we keep these in alphabetical order?

src/reducers/versions.tsx
willdurand: I think we could display it as it is the default file.

Issue #185, opened 2019-02-11
Connect the `FileTree` component?


willdurand:
The `FileTree` component uses https://github.com/gnapse/react-treefold in an uncontrolled way under the hood. I have the feeling that we could connect the tree by using the "controlled way".

I am wondering if we really want to do that though. "Connecting the tree" means putting the file tree into Redux and recording all the clicks on each node, i.e. all the state updates (opening a directory, clicking a file, etc.)

bobsilverberg:
While I do think we should store as much as possible in the redux store, I don't think storing the state of the tree (in terms of what's collapsed and what's expanded) is going to buy us much, if anything. It just seems like it will result in a lot of unnecessary dispatches.

kumar303:
Yes, we should definitely control as much of the app as possible with Redux. Something like file navigation is absolutely crucial to control with Redux. If we don't, we lose the ability to replay state from actions.

Additionally, Redux is the one true flow of data and when we violate that principle, subtle bugs creep into the app. See the [`<LinterProvider>` change](https://github.com/mozilla/addons-code-manager/pull/455#discussion_r268870489) as an example. In this case, the `FileTreeNode` component ceases to re-render because Redux is not in control of the flow of state changes. We have a hack in place that forces a re-render all the time but this may lead to performance problems at some point.

bobsilverberg:
One of the features needed for keyboard shortcuts (#199) is the ability to expand the entire tree (and, I would think, collapse it as well). Looking at the code for `react-treefold`, it looks like it might make sense to convert our tree to controlled version prior to attempting to write code that will do the expand/collapse.

@willdurand | @kumar303 Do you agree?

bobsilverberg:
Another feature required by shortcuts is the ability to select the next and previous files in the tree, and I believe that connecting the tree will also make this simpler.

kumar303:
Yeah, it sounds like that's a good reason to connect expansion to Redux

bobsilverberg:
Ok, I'm going to take this one on. I've started looking at how to make `react-treefold` controlled. It looks like we'll need to store all of the nodes of the tree in the redux store along with their expanded state. Then, when a user clicks on a node we would dispatch an action which updates the state, and that in turn would cause the FileTree to be updated accordingly.

Do we think we need to store a separate tree for each version, or will there always just be one active tree, and therefore we can just store one tree (i.e., the active one)?

willdurand:
[...]

PR #234, opened 2019-02-20
[WIP] Controle FileTree expanded/collapsed state

willdurand:
I am not quite sure that the tree should live in the component's state and not in redux (cf. #185), but in any case, if we want to have some control on this tree, we need to used `Treefold` in a "controlled" way. That's the aim of this PR.

Now, the first level of the tree (a.k.a. the root node) is automatically expanded.

screen shot 2019-02-20 at 12 28 28

### DIFF
--- a/src/@types/react-treefold/index.d.ts
+++ b/src/@types/react-treefold/index.d.ts
@@ -20,6 +20,8 @@ declare module 'react-treefold' {
   type TreefoldProps<NodeType> = {
     nodes: NodeType[];
     render: (props: TreefoldRenderProps<NodeType>) => React.Node;
+    isNodeExpanded?: (node: NodeType) => boolean;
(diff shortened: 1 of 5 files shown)


bobsilverberg:
I'm not sure I understand what this is doing. Prior to this, the FileTree seemed to be working, in that one could expand and collapse nodes, and it even seems to remember which nodes were expanded when you collapse at a higher level. So why do we need this change at all?

willdurand:
> I'm not sure I understand what this is doing. Prior to this, the FileTree seemed to be working, in that one could expand and collapse nodes, and it even seems to remember which nodes were expanded when you collapse at a higher level. So why do we need this change at all?

@bobsilverberg this is needed to expand the first level of the tree by default.

PR #241, opened 2019-02-21
Configure `@typescript-eslint/no-explicit-any` ESLint rule to throw an error

### LINKED ISSUE #238: ESLint `@typescript-eslint/no-explicit-any` rule should throw an error
The `@typescript-eslint/no-explicit-any` rule only outputs a warning currently, which makes it easy to miss. Let's configure this rule to throw an error instead so that CI and other tools will catch it.

willdurand:
Fixes #238

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -36,8 +36,7 @@
     // The beauty of TS is that it infers types quite well, so let's not write
     // too much code.
     "@typescript-eslint/explicit-function-return-type": "off",
-    // TODO: re-enable this rule once https://github.com/mozilla/addons-code-manager/issues/85 lands.
-    "amo/only-log-strings": "off",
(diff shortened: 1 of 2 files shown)


.eslintrc
```diff
@@ -36,8 +36,7 @@
     // The beauty of TS is that it infers types quite well, so let's not write
     // too much code.
     "@typescript-eslint/explicit-function-return-type": "off",
-    // TODO: re-enable this rule once https://github.com/mozilla/addons-code-manager/issues/85 lands.
-    "amo/only-log-strings": "off",
```
bobsilverberg: I take it that re-enabling this means removing it from this file, as opposed to changing the value `off` to something different (like `error`)?

willdurand: yep, it is enabled by default in `eslint-plugin-amo`.

Issue #108, opened 2019-02-01
View the contents of a text file


kumar303:
This supports the user story:

> As a code reviewer, I want to be able to inspect the entire source code and image assets of any version of an add-on I’m reviewing.

After selecting a file (https://github.com/mozilla/addons-code-manager/issues/107), see its contents.

The right side of this screen is an example of viewing `manifest.json`:

![51308318-21475f00-1a42-11e9-9d60-01d005a6ffe6](https://user-images.githubusercontent.com/55398/52148858-9913ad00-2630-11e9-9564-a83262f6f0c8.png)

Issue #109, opened 2019-02-01
View an image file


kumar303:
This supports the user story:

> As a code reviewer, I want to be able to inspect the entire source code and image assets of any version of an add-on I’m reviewing.

After selecting an image file (https://github.com/mozilla/addons-code-manager/issues/107), see the image. This is similar to viewing a text file (https://github.com/mozilla/addons-code-manager/issues/108).

bobsilverberg:
Just a note that this also came up in our walkthrough of the current reviewer tools, as a feature that is currently missing but desired.

AlexandraMoga:
Verified on -dev with Fx66, Win10x64

I've submitted an add-on with various image formats and then checked if and how they were rendered in the file viewer. I've tried to use most of the accepted images listed in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Firefox

Here is the example: https://code.addons-dev.allizom.org/en-US/browse/616837/versions/1688164/

 - I haven't noticed issues with `.jpg`, `.png` `.gif` or `.webp` 

@bobsilverberg I have some potential issues:
 - `bmp`s are not rendered - see [this](https://code.addons-dev.allizom.org/en-US/browse/616837/versions/1688164/?path=Img%2Fmydearbonnie.bmp) and [this](https://code.addons-dev.allizom.org/en-US/browse/616837/versions/1688164/?path=Img%2F640x400%20%28png%29.bmp)
 - `svg`s are not rendered - see [this](https://code.addons-dev.allizom.org/en-US/browse/616837/versions/1688164/?path=Img%2Flisting-small.svg) and [this](https://code.addons-dev.allizom.org/en-US/browse/385621/versions/1655456/?path=img%2Fhero-warning-major-networks.svg)
- invalid images (for example a non-image file that has an image extension type) are decoded as text - see [this example](https://code.addons-dev.allizom.org/en-US/browse/616837/versions/1688164/?path=Img%2Finvalid%20image%20type.png); Is this the expected outcome?

bobsilverberg:
Thanks for the testing, @AlexandraMoga!

> - `bmp`s are not rendered - see [this](https://code.addons-dev.allizom.org/en-US/browse/616837/versions/1688164/?path=Img%2Fmydearbonnie.bmp) and [this](https://code.addons-dev.allizom.org/en-US/browse/616837/versions/1688164/?path=Img%2F640x400%20%28png%29.bmp)

[...]

PR #243, opened 2019-02-21
Show the default file in the browse view

### LINKED ISSUE #110: Show code files with syntax highlighting
User story:

> As a code reviewer, I want the file viewer and diff viewer to provide code syntax highlighting when possible, so I can read the code more easily.

When viewing text files (https://github.com/mozilla/addons-code-manager/issues/108), the code should have syntax highlighting. This will probably happen by default but we should verify that all common add-on files will get syntax highlighting.

### LINKED ISSUE #240: Show a loading state when fetching a version

* Log in
* Go something like http://localhost:3000/en-US/browse/502955/versions/1541798/ to browse a version

### What happened?

There is no indication that data is loading

For slow connections, the screen should render a loading state

Showing skeleton text will make the app feel snappy and we could show that for any type of connection, not just slow connections.

willdurand:
Fixes #240 
Fixes #110 
Supports #108
~~⚠️ Depends on #218~~

---

Only the last commit is relevant here. The 3 first commits are #218.

This is a baby step to land #108.


bobsilverberg:
I understand this is a baby step, so I guess it's okay for now, but it is confusing that the default file is displayed on the right, but it does not appear to be the selected file in the tree on the left. That is something we definitely need to fix, or we shouldn't display the default file (imo).

src/pages/Browse/index.spec.tsx
```diff
@@ -74,32 +92,55 @@ describe(__filename, () => {
...
+  it('renders the content of the default file when a version has been loaded', () => {
+    const version = fakeVersion;
+
+    const store = configureStore();
+    store.dispatch(versionActions.loadVersionInfo({ version }));
+
+    const root = render({ store, versionId: String(version.id) });
+
+    expect(root.find(Highlight)).toHaveLength(1);
+    expect(root.find(Highlight)).toHaveProp('className', 'auto');
```
bobsilverberg: This is just asserting a static prop, which I think we don't normally do.

src/pages/Browse/index.spec.tsx
```diff
@@ -74,32 +92,55 @@ describe(__filename, () => {
...
+    expect(root.find(Highlight)).toHaveProp(
+      'children',
+      ((root.instance().props as Props).file as VersionFile).content,
+    );
+  });
 
-    const mockApi = jest.spyOn(api, 'getVersion');
-    mockApi.mockReturnValue(Promise.resolve(version));
+  it('does not render the content of another file when a version has been loaded', () => {
```
bobsilverberg: Nit: I find the description of this test to be confusing. The comment in the test helps to explain what it's testing, but I don't think this description captures it accurately.

src/pages/Browse/index.spec.tsx
willdurand: This one is important because it configures the syntax HL feature.

src/pages/Browse/index.spec.tsx
willdurand: ah, correct. It's confusing indeed. I changed it.

src/pages/Browse/index.tsx
```diff
@@ -21,39 +32,61 @@ type PropsFromRouter = {
...
       <React.Fragment>
-        <Col md="3">{version && <FileTree version={version} />}</Col>
-        <Col>
-          <p>Version ID: {this.props.match.params.versionId}</p>
+        <Col md="3">
+          <FileTree version={version} />
+        </Col>
+        <Col md="9">
+          {file ? (
+            <Highlight className="auto">{file.content}</Highlight>
```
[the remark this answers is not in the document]
willdurand: > This assumes a text file, but this could be an image. I was looking at #109, and I guess that will need to build on what you're doing here.

Yep, there is no support for that at all.

> Do you think it's worth extracting this out into a component at this point, or would you prefer to make that a separate issue as well?

I believe we can do that later when we know what we want to do. This patch only enables new features/issues.

PR #245, opened 2019-02-21
Select a file and fetch its content

### LINKED ISSUE #108: View the contents of a text file

willdurand:
~~⚠️ depends on #243~~
We will probably iterate over it but this PR fixes #108.

---

### DIFF
--- a/src/components/FileTreeNode/index.tsx
+++ b/src/components/FileTreeNode/index.tsx
@@ -8,7 +8,7 @@ import { gettext } from '../../utils';
 import { TreeNode } from '../FileTree';
 import styles from './styles.module.scss';
 
-type PublicProps = TreefoldRenderProps<TreeNode> & {
+export type PublicProps = TreefoldRenderProps<TreeNode> & {
   onSelect: (id: string) => void;
 };
 
--- a/src/components/FileTreeNode/index.tsx
+++ b/src/components/FileTreeNode/index.tsx
@@ -8,7 +8,9 @@ import { gettext } from '../../utils';
 import { TreeNode } from '../FileTree';
 import styles from './styles.module.scss';
 
-type PublicProps = TreefoldRenderProps<TreeNode>;
+type PublicProps = TreefoldRenderProps<TreeNode> & {
+  onSelect: (id: string) => void;
+};
 
 class FileTreeNodeBase<TreeNodeType> extends React.Component<PublicProps> {
   render() {
@@ -20,12 +22,14 @@ class FileTreeNodeBase<TreeNodeType> extends React.Component<PublicProps> {
       getToggleProps,
       hasChildNodes,
       renderChildNodes,
+      onSelect,
     } = this.props;
 
     let listGroupItemProps = {
       className: makeClassName(styles.node, {
         [styles.directoryNode]: isFolder,
       }),
+      onClick: () => onSelect(node.id),
     };
 
     if (isFolder) {
--- a/src/pages/Browse/index.spec.tsx
+++ b/src/pages/Browse/index.spec.tsx
@@ -42,6 +42,7 @@ describe(__filename, () => {
 
   type RenderParams = {
     _fetchVersion?: PublicProps['_fetchVersion'];
+    _fetchVersionFile?: PublicProps['_fetchVersionFile'];
     _log?: PublicProps['_log'];
     addonId?: string;
     versionId?: string;
@@ -50,6 +51,7 @@ describe(__filename, () => {
 
   const render = ({
     _fetchVersion,
+    _fetchVersionFile,
     _log,
     addonId = '999',
     versionId = '123',
@@ -58,6 +60,7 @@ describe(__filename, () => {
     const props = {
       ...createFakeRouteComponentProps({ params: { addonId, versionId } }),
       _fetchVersion,
+      _fetchVersionFile,
       _log,
     };
 
@@ -90,6 +93,10 @@ describe(__filename, () => {
       'version',
       createInternalVersion(version),
     );
+    expect(root.find(FileTree)).toHaveProp(
+      'onSelect',
+      (root.instance() as BrowseBase).onSelectFile,
+    );
   });
 
   it('renders the content of the default file when a version has been loaded', () => {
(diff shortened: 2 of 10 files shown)


bobsilverberg:
I think it makes sense to wait for the other patches to land before attempting to review this, but I'm happy to do so once they land.

willdurand:
> I think it makes sense to wait for the other patches to land before attempting to review this, but I'm happy to do so once they land.

@bobsilverberg I think this one is ready now, thank you!

willdurand:
I just rebased this PR.

src/components/FileTreeNode/index.tsx
```diff
@@ -8,7 +8,9 @@ import { gettext } from '../../utils';
 import { TreeNode } from '../FileTree';
 import styles from './styles.module.scss';
 
-type PublicProps = TreefoldRenderProps<TreeNode>;
+type PublicProps = TreefoldRenderProps<TreeNode> & {
+  onSelect: (id: string) => void;
```
bobsilverberg: The type for this same function (although the argument is named `id` here and `path` above) is also declared in `FileTree/index.tsx` (at line 129). I wonder if it would be worth defining its own type (e.g., `TreeNodeOnSelect`), that could be used in both places?

src/reducers/versions.tsx
```diff
@@ -213,6 +217,41 @@ export const fetchVersion = ({
   };
 };
 
+type FetchVersionFileParams = {
+  _log?: typeof log;
+  _getVersion?: typeof getVersion;
+  addonId: number;
+  versionId: number;
+  path: string;
```
bobsilverberg: Nit: Can we keep these in alphabetical order?

src/reducers/versions.tsx
```diff
@@ -213,6 +217,41 @@ export const fetchVersion = ({
...
+};
+
+export const fetchVersionFile = ({
+  _getVersion = getVersion,
+  _log = log,
+  addonId,
+  versionId,
+  path,
```
bobsilverberg: Nit: Can we keep these in alphabetical order?

src/pages/Browse/index.spec.tsx
```diff
@@ -114,14 +121,15 @@ describe(__filename, () => {
     const store = configureStore();
     store.dispatch(versionActions.loadVersionInfo({ version }));
 
+    // The user clicks a different file to view.
+    store.dispatch(
+      versionActions.updateSelectedPath({
```
bobsilverberg: I'm a little unsure about this, but I might be wrong. Looking at the code for `Browse`, it never directly dispatches `updateSelectedPath`, but rather that is dispatched by the `fetchVersionFile` thunk. I suppose this may be the only way to test this case, but it just feels a bit odd to be dispatching this in a test for a component that never actually dispatches this. Is there a better way to do this, or do you think my concern is misplaced?

src/pages/Browse/index.tsx
```diff
@@ -61,6 +64,19 @@ export class BrowseBase extends React.Component<Props> {
     );
   }
 
+  onSelectFile = (path: string) => {
+    const { _fetchVersionFile, dispatch, match } = this.props;
+    const { addonId, versionId } = match.params;
+
+    dispatch(
+      _fetchVersionFile({
```
bobsilverberg: We don't seem to have a test for this. We do have a test that *something* is dispatched (using `fakeThunk` in the test), but we don't seem to have a test that verifies that `addonId`, `versionId` and `path` are passed into the dispatch.

src/reducers/versions.spec.tsx
```diff
@@ -77,6 +78,27 @@ describe(__filename, () => {
         },
       });
     });
+
+    it('updates a selected path for a given version', () => {
+      const version = fakeVersion;
+      let state = reducer(undefined, actions.loadVersionInfo({ version }));
+
+      expect(state).toHaveProperty(
```
bobsilverberg: I like this `toHaveProperty` approach.

src/reducers/versions.spec.tsx
```diff
@@ -268,4 +290,111 @@ describe(__filename, () => {
...
+      const version = fakeVersion;
+      const _getVersion = jest.fn().mockReturnValue(Promise.resolve(version));
+
+      const addonId = 123;
+      const versionId = version.id;
+      const path = 'some/path.js';
+
+      const { store, thunk } = thunkTester({
+        createThunk: () =>
+          fetchVersionFile({ _getVersion, addonId, versionId, path }),
```
bobsilverberg: You're calling `fetchVersionFile` multiple times in this test suite. It seems like a candidate for a helper function.

src/reducers/versions.spec.tsx
```diff
@@ -268,4 +290,111 @@ describe(__filename, () => {
...
+
+      expect(dispatch).toHaveBeenCalledWith(
+        actions.loadVersionFile({
+          path,
+          version,
+        }),
+      );
+    });
+
+    it('logs an error when API response is not successful', async () => {
```
bobsilverberg: Perhaps we should also test that `loadVersionFile` is not dispatched when there is an error, either by adding an assertion into this test case, or creating an additional test case?

src/components/FileTreeNode/index.tsx
willdurand: That's a good idea.

src/pages/Browse/index.spec.tsx
willdurand: I don't think there is because we want to test the case where the file is selected but there is no content for it yet, so we have to dispatch this action only.

src/pages/Browse/index.tsx
willdurand: Indeed, there is no way to test that yet. I will file an issue.

willdurand: See: #262

src/reducers/versions.spec.tsx
willdurand: I added an assertion.

willdurand:
@bobsilverberg updated!

willdurand:
ah! 😂

PR #250, opened 2019-02-21
 Use REACT_APP_DEFAULT_API_VERSION in `LoginButton` component

### LINKED ISSUE #249: Use REACT_APP_DEFAULT_API_VERSION in `LoginButton` component
Instead of hardcoding the API version, we can now use the `REACT_APP_DEFAULT_API_VERSION` env var in the `LoginButton` component. 🎉

willdurand:
Fixes #249

---

Small code quality issue.

### DIFF
--- a/src/components/LoginButton/index.spec.tsx
+++ b/src/components/LoginButton/index.spec.tsx
@@ -5,7 +5,7 @@ import { Button } from 'react-bootstrap';
 import LoginButton from '.';
 
 describe(__filename, () => {
-  it('renders a page', () => {
+  it('renders a login button', () => {
     const apiVersion = 'api-version';
     const fxaConfig = 'some-fxa-config';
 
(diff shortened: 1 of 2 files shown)


src/components/LoginButton/index.spec.tsx
```diff
@@ -6,14 +6,17 @@ import LoginButton from '.';
 
 describe(__filename, () => {
   it('renders a page', () => {
```
kumar303: This spec is a little confusing

src/components/LoginButton/index.tsx
```diff
@@ -5,18 +5,20 @@ import { gettext } from '../../utils';
 import styles from './styles.module.scss';
 
 type PublicProps = {
+  apiVersion: string;
```
kumar303: I guess this is only a prop for testability? It feels like it doesn't need to be a prop but the alternative of injecting `process.env` as a prop might be cumbersome.

willdurand: Exactly, I don't want to hard-code `process.env` and injecting `process.env` implies more "src" code to write (because of the TS type and `((x as any) as Env)` trick).

src/components/LoginButton/index.spec.tsx
willdurand: ah correct, thanks!

Issue #239, opened 2019-02-20
Dispatch an actual loading state in fetchCurrentUserProfile


kumar303:
The approach in https://github.com/mozilla/addons-code-manager/pull/143 inferred a loading state from missing parameters. Now that we have a thunk in https://github.com/mozilla/addons-code-manager/pull/218 we can dispatch an actual loading state from `fetchCurrentUserProfile` while the user profile gets fetched.

This will make the loading state more robust and predictable. For example, you currently see a loading state after logging out which was not intended.

willdurand:
Note: the loader should be displayed as early as possible so that there is no visual glitch when the page loads, i.e. it should not be enabled before fetching the current user profile, it should be enabled by default and turned off when our code starts doing something.

kumar303:
I don't fully understand your comment but let me know if the patch in https://github.com/mozilla/addons-code-manager/pull/252 is missing something related to this.

willdurand:
> I don't fully understand your comment but let me know if the patch in #252 is missing something related to this.

ah well, me neither. What I meant is that, on page refresh, we should not have this flow:

```
content --> loading page --> content
```

It should be:

```
loading page --> content
```

No matter the authentication state.

Looks like this is covered by your patch, though.

PR #252, opened 2019-02-21
Make user loading state more robust

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/239

Here are the new UI states:

<details>
<summary>screencast</summary>

addons-code-manager-loading mov

</details>

### DIFF
--- a/src/api/index.spec.tsx
+++ b/src/api/index.spec.tsx
@@ -8,7 +8,7 @@ import {
 import {
   HttpMethod,
   callApi,
-  getCurrentUserProfile,
+  getCurrentUser,
   getVersion,
   isErrorResponse,
   logOutFromServer,
@@ -254,9 +254,9 @@ describe(__filename, () => {
     });
   });
 
-  describe('getCurrentUserProfile', () => {
+  describe('getCurrentUser', () => {
     it('calls the API to retrieve the current logged-in user profile', async () => {
-      await getCurrentUserProfile(defaultApiState);
+      await getCurrentUser(defaultApiState);
 
       expect(fetch).toHaveBeenCalledWith(
         expect.stringMatching(`/api/${defaultVersion}/accounts/profile/`),
--- a/src/components/App/index.tsx
+++ b/src/components/App/index.tsx
@@ -34,7 +34,7 @@ export type PublicProps = {
 type PropsFromState = {
   apiState: ApiState;
   loading: boolean;
-  profile: User | null;
+  user: User | null;
 };
 
 /* eslint-disable @typescript-eslint/indent */
@@ -59,9 +59,9 @@ export class AppBase extends React.Component<Props> {
   }
 
   componentDidUpdate(prevProps: Props) {
-    const { apiState, profile } = this.props;
+    const { apiState, user } = this.props;
 
-    if (!profile && prevProps.apiState.authToken !== apiState.authToken) {
+    if (!user && prevProps.apiState.authToken !== apiState.authToken) {
       const { _fetchCurrentUser, dispatch } = this.props;
 
       dispatch(_fetchCurrentUser());
@@ -75,7 +75,7 @@ export class AppBase extends React.Component<Props> {
   }
 
   renderContent() {
-    const { loading, profile } = this.props;
+    const { loading, user } = this.props;
 
     if (loading) {
       return this.renderRow(
@@ -89,7 +89,7 @@ export class AppBase extends React.Component<Props> {
       );
     }
 
-    if (profile) {
+    if (user) {
       return this.renderRow(
         <Switch>
           <Route exact path="/" component={Index} />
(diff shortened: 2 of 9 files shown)


src/components/App/index.tsx
```diff
@@ -67,61 +68,65 @@ export class AppBase extends React.Component<Props> {
...
+  render() {
+    const { loading } = this.props;
 
     return (
       <React.Fragment>
         {!loading && <Navbar />}
-
         <Container className={styles.container} fluid>
-          <Row
-            className={makeClassName(styles.content, styles.isoading, {
```
kumar303: Bummer that TS didn't catch this typo

src/components/App/index.tsx
```diff
@@ -67,61 +68,65 @@ export class AppBase extends React.Component<Props> {
...
-      </Switch>
-    ) : (
-      <p className={styles.loginMessage}>
-        {gettext('Please log in to continue.')}
-      </p>
-    );
+    if (loading) {
+      return this.renderRow(
+        <React.Fragment>
+          <p>{gettext('Getting your workspace ready')}</p>
```
willdurand: Could we add `<p><FontAwesomeIcon icon="spinner" size="3x" spin /></p>` above the text to have the final Windows™ Easter-egg? 🙏🤗

src/components/App/index.tsx
```diff
@@ -67,61 +68,65 @@ export class AppBase extends React.Component<Props> {
...
+          <p>{gettext(`Don't turn off your computer`)}</p>
```
willdurand: This string does not need back-ticks.

src/reducers/users.tsx
```diff
@@ -49,6 +49,8 @@ export type User = {
 };
 
 export const actions = {
+  abortLoadCurrentUser: createAction('ABORT_LOAD_CURRENT_USER'),
+  beginLoadCurrentUser: createAction('BEGIN_LOAD_CURRENT_USER'),
```
willdurand: The thunk function starts with `fetch` so maybe we could use `beginFetchCurrentUser` and `abortFetchCurrentUser`?

We could also use similar function names for the API and redux functions, by choosing either `xxCurrentUser` or `xxCurrentUserProfile`. Not necessarily here, though.

src/reducers/users.tsx
```diff
@@ -65,7 +67,7 @@ export const requestLogOut = ({
 };
 
 export type UsersState = {
-  currentUser: User | null;
+  currentUser: User | null | undefined;
```
willdurand: Why not keeping `User | null` and have a boolean flag to determine the loading status instead?

kumar303: I'd want to change the return of `getCurrentUser` to `User | null` as well but I guess that would make sense

kumar303: Yes, this does make more sense since `undefined` is an implementation detail of the state

src/reducers/users.tsx
kumar303: Sure. I like shortening everything to `xxCurrentUser` since the final result is `User`.

src/components/App/index.tsx
kumar303: whoa, it is totally the Windows 10 spinner, too

kumar303:
@willdurand thanks, it's ready for another look

src/api/index.spec.tsx
```diff
@@ -254,9 +254,9 @@ describe(__filename, () => {
     });
   });
 
-  describe('getCurrentUserProfile', () => {
+  describe('getCurrentUser', () => {
     it('calls the API to retrieve the current logged-in user profile', async () => {
-      await getCurrentUserProfile(defaultApiState);
+      await getCurrentUser(defaultApiState);
```
willdurand: 🎉

src/reducers/users.tsx
```diff
@@ -81,21 +79,27 @@ export const createInternalUser = (user: ExternalUser): User => {
   };
 };
 
-export const getCurrentUser = (users: UsersState) => {
-  return users.currentUser;
+export const selectCurrentUser = (users: UsersState) => {
```
willdurand: 👍

src/reducers/users.tsx
willdurand: So, we keep `| undefined` here to avoid the loading flag in the state, right?

kumar303: Whoops, when you first commented I mistook the context for the `profile: User | null` prop on `App` and `Navbar`.

> So, we keep `| undefined` here to avoid the loading flag in the state, right?

Yeah, I think it's OK to use `undefined` here for a loading state. When something is `null` it says to me that "it exists but it's empty." As you eluded to, storing a separate loading flag would make things more complex.

willdurand: thanks!

Issue #111, opened 2019-02-01
View diff between two versions


kumar303:
User story:

> As a code reviewer, I want to be able to view the source code diff between any two versions of an add-on I’m reviewing.

Here's an example of looking at the diff between version 1.69.0 and 1.69.1 of LastPass. URL: https://reviewers.addons-dev.allizom.org/en-US/firefox/files/compare/93833...93882/

<img width="1518" alt="screenshot 2019-02-01 15 31 09" src="https://user-images.githubusercontent.com/55398/52150767-95832480-2636-11e9-85c6-81a8cbb1e3ab.png">

EnTeQuAk:
Not necessarily blocked by, but the version-chooser component needs https://github.com/mozilla/addons-server/issues/10432 to be done first. (working on it).

PR #253, opened 2019-02-22
Introduce a Compare page

willdurand:
This patch creates a new page for the "compare" (or diff) view to start this new page soon-ish (hello #111). Of course, it is incomplete as there is no API endpoint yet, but it shows the `DiffView` (and fixes a visual issue). API responses will be similar but not exactly the same so I can't find a way to DRY better than what's here.

I created a new `Loading` component to display a loading state (message for now, but could maybe be reused for other loading states).

Given that validator results will be displayed in both views, I think we need this view right now so that we can think about both use cases.

## Screenshot

screen shot 2019-02-22 at 10 36 37

### DIFF
--- a/src/pages/Compare/index.spec.tsx
+++ b/src/pages/Compare/index.spec.tsx
@@ -77,36 +77,38 @@ describe(__filename, () => {
   });
 
   it('renders a FileTree component when a version has been loaded', () => {
-    const version = fakeVersion;
+    const baseVersion = fakeVersion;
+    const headVersion = { ...fakeVersion, id: baseVersion.id + 1 };
 
     const store = configureStore();
-    store.dispatch(versionActions.loadVersionInfo({ version }));
+    store.dispatch(versionActions.loadVersionInfo({ version: baseVersion }));
+    store.dispatch(versionActions.loadVersionInfo({ version: headVersion }));
 
     const root = render({
       store,
-      baseVersionId: String(version.id),
-      // TODO: ideally, we should have different version IDs.
-      headVersionId: String(version.id),
+      baseVersionId: String(baseVersion.id),
+      headVersionId: String(headVersion.id),
     });
 
     expect(root.find(FileTree)).toHaveLength(1);
     expect(root.find(FileTree)).toHaveProp(
       'version',
-      createInternalVersion(version),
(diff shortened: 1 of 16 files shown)


willdurand:
> When I load [http://localhost:3000/en-US/compare/502955/versions/1541798...1541798](http://localhost:3000/en-US/compare/502955/versions/1541798...1541798/) I get the following error. Does your local version work without the error?

It was definitely working, yes. That's probably related to https://github.com/mozilla/addons-code-manager/issues/255, which is fixed by #256.

willdurand:
I merged the fix and rebased, it should work again.

src/pages/Compare/index.spec.tsx
```diff
@@ -0,0 +1,132 @@
...
+  it('renders a FileTree component when a version has been loaded', () => {
+    const version = fakeVersion;
+
+    const store = configureStore();
+    store.dispatch(versionActions.loadVersionInfo({ version }));
+
+    const root = render({
+      store,
+      baseVersionId: String(version.id),
+      // TODO: ideally, we should have different version IDs.
```
kumar303: Can you just dispatch two of them?

```js
const version = fakeVersion;
const version2 = { ...fakeVersion, id: version.id + 1 };
```

src/pages/Compare/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  baseVersionId: string;
+  headVersionId: string;
+};
+
+type PropsFromState = {
+  apiState: ApiState;
+  version: Version;
+};
+
+/* eslint-disable @typescript-eslint/indent */
```
kumar303: Can't we just disable this rule globally since we use Prettier?

src/pages/Index/index.spec.tsx
```diff
@@ -8,6 +8,6 @@ describe(__filename, () => {
   it('renders a page', () => {
     const root = shallow(<Index />);
 
-    expect(root.find(Link)).toHaveLength(1);
+    expect(root.find(Link)).toHaveLength(2);
```
kumar303: This test also doesn't seem helpful. I would say `expect(root.find(Link)).toExist()` is good enough.

PR #257, opened 2019-02-22
Add FileTree story

### LINKED ISSUE #190: Add story for the FileTree component
We should add a storybook story for the `FileTree` component.

willdurand:
Fixes #190 

---

screen shot 2019-02-25 at 11 06 49

### DIFF
--- a/stories/FileTree.stories.tsx
+++ b/stories/FileTree.stories.tsx
@@ -1,0 +1,62 @@
+/* eslint @typescript-eslint/camelcase: 0 */
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+import FileTree from '../src/components/FileTree';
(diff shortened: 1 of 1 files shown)


willdurand:
> How about show casing some tree nesting? That's an important visual part of the component. I don't know how to make files within subdirectories but here's a start:

mmh, sure. I thought the `FileTreeNode` story could have been used for that, which would have left this `FileTree` story simple but we could have more data here.

kumar303:
> I thought the `FileTreeNode` story could have been used for that, which would have left this `FileTree` story simple but we could have more data here.

I see that the current `FileTreeNode` story is focused more on simple cases: a file type, a directory type, and one level of nesting. I think it makes sense to have the `FileTree` story be about complex nesting, even though the actual styling happens in `FileTreeNode`.

PR #259, opened 2019-02-22
Fix empty folder node height

### LINKED ISSUE #258: Empty node height is wrong according to Storybook
The height of an "empty folder" node is not the same as the other nodes, see:

screen shot 2019-02-22 at 14 47 34

It should be:

screen shot 2019-02-22 at 14 47 25

willdurand:
Fixes #258 

---

This is a small fix for a small issue I noticed in Storybook 😎 

Before:

screen shot 2019-02-22 at 14 47 34

After:

screen shot 2019-02-22 at 14 47 25

### DIFF
--- a/src/components/FileTreeNode/index.tsx
+++ b/src/components/FileTreeNode/index.tsx
@@ -58,7 +58,7 @@ class FileTreeNodeBase<TreeNodeType> extends React.Component<PublicProps> {
             renderChildNodes()
           ) : (
             <ListGroup.Item
-              className={styles.emptyNodeDirectory}
+              className={makeClassName(styles.node, styles.emptyNodeDirectory)}
               style={{ paddingLeft: `${(level + 2) * 20}px` }}
             >
               {gettext('This folder is empty')}


src/components/FileTreeNode/index.tsx
```diff
@@ -58,7 +58,7 @@ class FileTreeNodeBase<TreeNodeType> extends React.Component<PublicProps> {
             renderChildNodes()
           ) : (
             <ListGroup.Item
-              className={styles.emptyNodeDirectory}
+              className={makeClassName(styles.node, styles.emptyNodeDirectory)}
```
kumar303: Where does `emptyNodeDirectory` come from? I can't find it.

willdurand: It's just a CSS class, mostly used in unit tests so far. It's possible to declare CSS class names without any CSS in `styles.module.scss`. I used to write an empty definition (`.cls {}`) but a review comment asked me to get rid of that if there was no definition, so 🤷‍♂️

kumar303: > It's possible to declare CSS class names without any CSS in `styles.module.scss`.

Huh, I guess that's why TypeScript didn't catch the typo'd class name in that other patch.

PR #260, opened 2019-02-22
Display add-on's name in the FileTree

### LINKED ISSUE #261: Display the add-on's name in the FileTree
Now that we have `addon` data in the `versions` reducer, we can display the name of the add-on in the `FileTree`, instead of displaying the version ID.

willdurand:
Fixes #261

This patch was an excuse to see how to consume localized content :)

## Screenshot

screen shot 2019-02-22 at 17 38 21

### DIFF
--- a/src/utils.spec.tsx
+++ b/src/utils.spec.tsx
@@ -6,6 +6,7 @@ describe(__filename, () => {
       const lang = 'fr';
       const value = 'un peu de contenu';
       const localizedStringMap = {
+        en: 'some content',
         [lang]: value,
       };
 
--- a/src/utils.spec.tsx
+++ b/src/utils.spec.tsx
@@ -1,0 +1,32 @@
+import { getLocalizedString } from './utils';
+
+describe(__filename, () => {
+  describe('getLocalizedString', () => {
+    it('returns a localized string for a given language', () => {
+      const lang = 'fr';
+      const value = 'un peu de contenu';
+      const localizedStringMap = {
+        [lang]: value,
+      };
+
+      expect(getLocalizedString(localizedStringMap, lang)).toEqual(value);
+    });
+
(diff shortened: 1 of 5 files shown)


src/utils.spec.tsx
```diff
@@ -0,0 +1,32 @@
+import { getLocalizedString } from './utils';
+
+describe(__filename, () => {
+  describe('getLocalizedString', () => {
```
bobsilverberg: Would it be worth adding a test with multiple languages in the map, to verify that we are returning the *correct* one?

src/utils.tsx
```diff
@@ -1 +1,12 @@
+export type LocalizedStringMap = {
+  [lang: string]: string;
+};
+
 export const gettext = (text: string) => text;
+
+export const getLocalizedString = (
+  localizedStringMap: LocalizedStringMap,
+  lang = process.env.REACT_APP_DEFAULT_API_LANG as string,
```
bobsilverberg: Does TS understand from this that the type of `lang` is `string` but it is also optional (e.g., like `lang?: string`)?

willdurand: It understands that there is a default value, making the argument optional. Because it cannot infer what's in `process.env`, we have to hint it by adding `as string` (it considers everything into `process.env` to be optional strings). If we were to write `'en-US'`, it would have inferred the correct type.

src/utils.spec.tsx
willdurand: Maybe, although it is a map so there is no risk of collision.

willdurand: I added a second item in one of the maps.

src/utils.tsx
bobsilverberg: So then TS doesn't necessarily know that we want to restrict `lang` to only accept strings? I'm just trying to understand if this argument is typed or not.

PR #270, opened 2019-02-25
Use `Link` component in Navbar title

### LINKED ISSUE #254: Navbar "title" should use react-router `push()` instead of reloading the page

1. Load the application
2. Click the top-left title (`addons-code-manager`)

### What happened?

The page reloads.

The page does not reload. It uses client-side navigation to go to the index page.

This is not a new issue but it is time to fix it now :)

willdurand:
Fixes #254

### DIFF
--- a/src/components/Navbar/index.spec.tsx
+++ b/src/components/Navbar/index.spec.tsx
@@ -1,8 +1,6 @@
 import { shallow } from 'enzyme';
 import React from 'react';
 import { Store } from 'redux';
-import { NavbarBrand } from 'react-bootstrap';
-import { Link } from 'react-router-dom';
 
 import configureStore from '../../configureStore';
 import LoginButton from '../LoginButton';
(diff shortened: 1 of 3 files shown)


src/components/Navbar/index.spec.tsx
```diff
@@ -36,6 +38,14 @@ describe(__filename, () => {
     return store;
   };
 
+  it('renders a brand link', () => {
```
kumar303: I don't see anything in this test that will be helpful so you could remove it. Otherwise it will just be one extra thing you'll have to update when the JSX changes. 

All the test does is duplicate the static configuration of `<Navbar.Brand>` and `<Link>`. Static configuration doesn't changed based on any inputs so it has no potential for breakage. Also, the test is not executing any code that isn't already executed so it's not providing additional coverage. Finally, none of the JSX tested gets rendered conditionally so asserting its presence is not helpful.

src/components/Navbar/styles.module.scss
```diff
@@ -3,6 +3,10 @@
   justify-content: space-between;
 }
 
+.brand a {
+  color: #fff;
```
kumar303: Hmm, it would be nice to reference color variables in the theme. I'm not worried about it, though. It's just a thought I had.

src/components/Navbar/index.spec.tsx
willdurand: That's right.

Issue #275, opened 2019-02-25
Get hot reloading to work for development


kumar303:
### Describe the problem and steps to reproduce it:

* Start all development processes with `yarn dev`
* Open Redux Developer Tools
* Sign in to the app
* Sign out
* Open `components/App/index.tsx` and edit the text saying "Please log in..."

### What happened?

The page does a full refresh to show the new text. You lose all of your Redux state.

### What did you expect to happen?

The `App` component should be hot reloaded to show the new text, preserving Redux state.

### Anything else we should know?

Upstream issue: https://github.com/facebook/create-react-app/issues/6503

PR #276, opened 2019-02-25
Get hot reloading to work for development

kumar303:
This attempts to work around hot reloading issues: https://github.com/mozilla/addons-code-manager/issues/275


kumar303:
@bobsilverberg @willdurand What do you think? Is this too crazy? I dislike straying from `react-scripts`. This might be OK just as a temporary solution until the upstream issues is resolved, though?

It fixes the issue for me using the STR in https://github.com/mozilla/addons-code-manager/issues/275

willdurand:
If this patch fixes the issue, that's a good think because we'll likely need hot reload soon (that was not super useful until then). Yet, this will make upgrades more complicated (or even maybe impossible...) and temporary workarounds usually last for a very long time (not to say forever).

The README of this rewire project says: "As of Create React App 2.0 this repo is "lightly" maintained mostly by the community at this point."

What are our other options? Could we try to fix CRA instead?

kumar303:
> Could we try to fix CRA instead?

I guess it just involves [updating the loaders](https://github.com/cdharris/react-app-rewire-hot-loader/blob/331a8dd34cc9c4af50e37d362d6b8e66ef62d0fe/index.js).

bobsilverberg:
Should this be closed? Do we now have hot reloading working?

willdurand:
ah, yes.

Issue #282, opened 2019-02-26
Add syntax highlighting to the `DiffView` component


willdurand:
This is related to #110 and #111.

This is feasible in the `DiffView` component thanks to the [token system of the underlying lib](https://github.com/otakustay/react-diff-view/tree/4bca86cec1168b2ca05c16dd6d202053a727931a#token-system), but that looks complicated.

willdurand:
I have a working patch in https://github.com/mozilla/addons-code-manager/pull/296, but it depends on other patches so I am putting it on hold for now.

Issue #281, opened 2019-02-26
Add story for the DiffView component


willdurand:
We need a story for the `DiffView` component. Writing one will help us style this component.

Issue #248, opened 2019-02-21
VersionFile should include file type


bobsilverberg:
As discussed at https://github.com/mozilla/addons-code-manager/pull/243#discussion_r259045468, we probably want a `VersionFile` to know what file type it is, so that we know whether we need to display a text file, an image, or whether it is undisplayable. 

I originally thought we could use `VersionEntryType`, but that is calculated based on the data we have in the file tree, and not on the data returned by the API for the file itself, so we probably need another way to calculate the type of a file.

Looking further, it seems that we don't get any of the information we need in the API response for the file itself, other than the file name/path which includes an extension so we could possibly use that.

The file tree does contain "file type" information for each of the entries, which we are currently using to calculate `VersionEntryType`, and it also includes a `mimetype` which might also be useful for this use case.

So, looking at all of that, it seems like the app should make assumptions about the type of a file based on the data in the file tree (which is part of `Version`) and not based on the data of the file itself (which is part of `VersionFile`). Given that, I wonder if it would make sense to change the shape of the reducer so that all data about a file, including its content, is stored in `VersionEntry`, which is stored in the `entries` property of a `Version`. This would mean the reducer state would no longer have properties for both `versionInfo` and `versionFiles`, and would only have a property for `versions`, which would hold `Version`s, which would in turn hold `VersionEntry`s inside its `entries` property.

Am I making sense? I can put forward a POC if it would make it clearer.

@willdurand / @kumar303 what do you think?

willdurand:
We know the `path` and we have the `state`, so we can add `type` to a `VersionFile`.

bobsilverberg:
@willdurand We could, but do you think my suggestion makes sense to not store the `VersionFile`s separately from the `Version`?

willdurand:
`file.entries` is very useful for building the tree (which could go to redux too, maybe, #185), so we need to keep it, but then.. I don't know.

Maybe all we need is, indeed, a map of `Version`, indexed by `versionId`. Each `Version` has the props we are interested in coming from the API, and optionally, a `content`. No content = an API call is needed, otherwise we're good.

It's not clear to me yet.

bobsilverberg:
Ok, maybe I'll put together a POC for this idea and see what people think.

kumar303:
> Maybe all we need is, indeed, a map of `Version`, indexed by `versionId`.

Yeah, I was thinking maybe we only need a mapping. It would be easier to see a patch for it before advising.

[...]

PR #285, opened 2019-02-26
Introduce a new CodeView component

### LINKED ISSUE #283: Use a different lib for syntax highlighting (browse page)
Instead of Highlight.js, we should use https://github.com/conorhastings/react-syntax-highlighter. This lib uses Prism, which is also used by `refractor`, the lib used by https://github.com/otakustay/react-diff-view too. It makes sense to use the same syntax highlighting lib in both contexts.

In addition, `react-syntax-highlighter` can be customized to add additional components to the view, so we could inline comments and lint errors.

### LINKED ISSUE #284: Add line numbers in browse page
We should show the line numbers in the browse page.

willdurand:
Fixes #283
Fixes #284

---

This PR adds a new component and get rid of the previous syntax highlighting component. It is needed because the previous solution was not super open to modifications. ~~With the new lib, it seems a bit better. For example, the concept of "markers" (https://github.com/rexxars/react-refractor#markers) will be useful to add linter errors/warnings.~~ The updated version of this patch uses `refractor` directly, instead of the react wrapper mentioned before. This was needed to inline comments and have more control on the rendering. See: https://github.com/mozilla/addons-code-manager/pull/285#issuecomment-467844614

This PR also adds line numbers to this new component. I added a temporary workaround for #248 so that things continue to work. It will be better once https://github.com/mozilla/addons-server/issues/10796 lands (addons-server). Selecting code should ignore the line numbers.

Note: once #281 and maybe #282 are resolved, we can unify the styles of the compare and browse views. This is already closer but we can do better, especially when both pages show "code".

## Screenshots

screen shot 2019-02-26 at 17 31 53
screen shot 2019-02-26 at 17 31 32
screen shot 2019-02-26 at 17 30 57
screen shot 2019-02-26 at 17 33 22
screen shot 2019-02-26 at 17 33 20


src/components/CodeView/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  return (
+    <div className={styles.CodeView}>
+      <pre>
+        <code className={styles.lineNumbers}>
+          {getLines(content).map((_, i) => {
+            const lineNumber = i + 1;
+
+            return (
+              <span key={`line-number-${lineNumber}`}>{`${lineNumber}`}</span>
```
kumar303: I'm not totally sure how a separate column of line numbers will let us do what we want for review forms but if you already have a prototype then I believe you 😎

src/components/CodeView/index.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+    mimeType = 'mime/type',
+  } = {}) => {
+    return shallow(<CodeView mimeType={mimeType} content={content} />);
+  };
+
+  it('renders a CodeView component', () => {
+    const root = render();
+
+    expect(root.find(`.${styles.CodeView}`)).toHaveLength(1);
+    expect(root.find(`.${styles.lineNumbers}`)).toHaveLength(1);
```
kumar303: What about adding another test that makes sure multi-line content produces multiple span tags? Like: 
```
root.find(`.${styles.lineNumbers} span`)
``` 

This would test the integration of `CodeView` and `getLines()`.

src/components/CodeView/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+        {language ? (
+          <Refractor
+            className={styles.content}
+            language={language}
+            value={content}
+            inline
+          />
+        ) : (
+          <code className={styles.content}>{content}</code>
```
kumar303: It's a bummer that you can't render `<Refractor>` with a plain text language. I dislike seeing us reimplement the innards of `<Refractor>` right here.

src/components/CodeView/styles.module.scss
```diff
@@ -0,0 +1,18 @@
...
+  border: 1px solid $light-gray;
+  padding: 10px;
+}
+
+.lineNumbers {
+  display: grid;
+  float: left;
+  padding-right: 10px;
+
+  * {
```
kumar303: Why not just put a class on all the spans like `.lineNumber` and select that?

kumar303: github does `text-align: right` on line numbers which I think would be a good idea for when there are thousands of lines.

src/scss/variables.scss
```diff
@@ -0,0 +1,4 @@
+@import '~bootstrap/scss/functions';
+@import '~bootstrap/scss/variables';
+
+$light-gray: $gray-200;
```
kumar303: 👍

src/components/CodeView/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+      return 'css';
+    case 'text/html':
+      return 'html';
+    default:
+      return null;
+  }
+};
+
+export const getLines = (content: string) => {
+  return content.replace(/\n$/, '').split('\n');
```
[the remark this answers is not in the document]
willdurand: That would make sense indeed. We could replace `\r\n`.

src/components/CodeView/index.tsx
willdurand: Ok so that's what I wanted to double-check today and you are right, that won't be easy. I switched to a better version and I have a proof that it works as intended now:

screen shot 2019-02-27 at 11 15 06

src/components/CodeView/index.tsx
willdurand: Agreed.

src/components/CodeView/styles.module.scss
willdurand: Good idea, thanks!

src/components/CodeView/index.tsx
willdurand: I fixed that. Actually, I remove the `react-refractor` lib 😅

src/components/CodeView/styles.module.scss
willdurand: :+1:

willdurand:
I refactored the component to have full control over the rendering. With this "simple" patch below, we can inline comments:

```diff
diff --git a/src/components/CodeView/index.tsx b/src/components/CodeView/index.tsx
index a307c5c..a3c8fbf 100644
--- a/src/components/CodeView/index.tsx
+++ b/src/components/CodeView/index.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import refractor from 'refractor';
 
+import Comment from '../Comment';
 import styles from './styles.module.scss';
 import text from './text-language';
 import { getLanguage, getLines, mapWithDepth } from './utils';
@@ -34,19 +35,31 @@ const CodeViewBase = ({ mimeType, content }: PublicProps) => {
 
   return (
     <div className={styles.CodeView}>
-      <table>
-        <tbody>
+      <table className={styles.table}>
+        <tbody className={styles.tableBody}>
           {getLines(content).map((code, i) => {
             const line = i + 1;
 
             return (
-              <tr key={`row-${line}`}>
-                <td className={styles.lineNumber}>{`${line}`}</td>
+              <React.Fragment>
+                <tr key={`row-${line}`}>
+                  <td className={styles.lineNumber}>{`${line}`}</td>
 
-                <td className={styles.code}>
-                  {renderHighlightedCode(code, language)}
-                </td>
-              </tr>
+                  <td className={styles.code}>
+                    {renderHighlightedCode(code, language)}
+                  </td>
+                </tr>
+                {line === 3 && (
+                  <tr
+                    key={`row-comment-${line}`}
+                    style={{ display: 'inline-table', width: '100%' }}
+                  >
+                    <td colSpan={2}>
+                      <Comment comment={null} />
+                    </td>
+                  </tr>
+                )}
+              </React.Fragment>
             );
           })}
         </tbody>
```

screen shot 2019-02-27 at 13 29 32

kumar303:
> I refactored the component to have full control over the rendering.

Cool. Send me a re-review request when it's ready for another look.

willdurand:
> Cool. Send me a re-review request when it's ready for another look.

D'oh! I thought I did it because it was ready :/ Thanks!

willdurand:
I made a few minor changes to ease the PRs that depend on this patch. (more scss variables and extracted refractor logic to reuse it in the `DiffView`)

src/components/CodeView/index.spec.tsx
```diff
@@ -0,0 +1,49 @@
...
+describe(__filename, () => {
+  const render = ({
+    content = 'some content',
+    mimeType = 'mime/type',
+  } = {}) => {
+    return shallow(<CodeView mimeType={mimeType} content={content} />);
+  };
+
+  it('renders plain text code when mime type is not supported', () => {
+    const root = render();
```
kumar303: Instead of rely on the hidden default, this should explicitly set an unsupported mime type.

src/components/CodeView/index.spec.tsx
```diff
@@ -0,0 +1,49 @@
...
+    expect(root.find(`.${styles.lineNumber}`)).toHaveLength(1);
+    expect(root.find(`.${styles.code}`)).toHaveLength(1);
+    expect(root.find(`.${styles.highlightedCode}`)).toHaveLength(1);
+    expect(root.find('.language-text')).toHaveLength(1);
+  });
+
+  it('renders highlighted code when language is supported', () => {
+    const content = '{ "foo": "bar" }';
+    const mimeType = 'application/json';
+    const root = render({ mimeType, content });
```
kumar303: This could be more concise:

```js
const root = render({ mimeType: 'application/json', content: '{ "foo": "bar" }' });
```

src/components/CodeView/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+// This function mimics what https://github.com/rexxars/react-refractor does,
+// but we need a different layout to inline comments so we cannot use this
+// component.
+const renderHighlightedCode = (code: string, language: string) => {
+  const ast = refractor.highlight(code, language);
+  const value = ast.length === 0 ? code : ast.map(mapWithDepth(0));
+
+  return (
+    <pre className={styles.highlightedCode}>
+      <code className={`language-${language}`}>{value}</code>
```
kumar303: `value` doesn't seem to be tested at all. All the tests pass for me with this:

```diff
diff --git a/src/components/CodeView/index.tsx b/src/components/CodeView/index.tsx
index 1b4bbad..cbf0d90 100644
--- a/src/components/CodeView/index.tsx
+++ b/src/components/CodeView/index.tsx
@@ -18,7 +18,7 @@ const renderHighlightedCode = (code: string, language: string) => {
 
   return (
     <pre className={styles.highlightedCode}>
-      <code className={`language-${language}`}>{value}</code>
+      <code className={`language-${language}`}></code>
     </pre>
   );
 };
```

Can you add tests for both of `ast` cases?

src/components/CodeView/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { getLanguage, getLines, mapWithDepth } from './utils';
+import refractor from '../../refractor';
+
+type PublicProps = {
+  mimeType: string;
+  content: string;
+};
+
+// This function mimics what https://github.com/rexxars/react-refractor does,
+// but we need a different layout to inline comments so we cannot use this
```
kumar303: Why couldn't you use [markers](https://github.com/rexxars/react-refractor/#markers)?

The tooltip example seemed like it would allow us to do it, no?

```jsx
<Refractor
  language="js"
  value={source}
  markers={[
    {line: 1, className: 'no-not-use-foo-in-examples'},
    {line: 3, component: props => (
      <TooltipedLine tooltipText="Prefer template for string concatenation">
        {props.children}
      </TooltipedLine>
    )}
  ]}
/>
```

src/components/CodeView/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+    </pre>
+  );
+};
+
+const CodeViewBase = ({ mimeType, content }: PublicProps) => {
+  const language = getLanguage(mimeType);
+
+  return (
+    <div className={styles.CodeView}>
+      <table className={styles.table}>
```
kumar303: Why did you use a table? To match `react-diff-view`? A css grid would probably be easier to deal with.

src/components/CodeView/styles.module.scss
```diff
@@ -0,0 +1,36 @@
...
+
+.lineNumber {
+  padding-right: $default-padding;
+  text-align: right;
+  user-select: none;
+}
+
+.code,
+.lineNumber {
+  font-family: Consolas, Monaco, 'Andale Mono', 'Ubuntu Mono', monospace;
```
kumar303: Can we sync this and other styles with `DiffView`? It could be done in a follow-up, of course.

src/components/CodeView/utils.spec.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  describe('getLines', () => {
+    it('splits a content into lines', () => {
+      const lines = ['foo', 'bar'];
+      const content = lines.join('\n');
+
+      expect(getLines(content)).toEqual(lines);
+    });
+
+    it('supports Windnows line endings', () => {
```
kumar303: typo

src/components/CodeView/utils.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { RefractorNode } from 'refractor';
+import makeClassName from 'classnames';
+
+const mapChild = (
+  child: RefractorNode,
+  i: number,
+  depth: number,
+): React.ReactElement | string => {
+  if (child.tagName) {
+    // @ts-ignore
```
kumar303: It looks like `properties` was duplicated, right? This seems to fix it:

```diff
diff --git a/src/@types/refractor/index.d.ts b/src/@types/refractor/index.d.ts
index b6a6f3f..f5b809e 100644
--- a/src/@types/refractor/index.d.ts
+++ b/src/@types/refractor/index.d.ts
@@ -3,7 +3,6 @@ declare module 'refractor' {
 
   type RefractorNode = {
     children: RefractorNode[];
-    properties: object;
     tagName: string;
     type: string;
     value: string;
diff --git a/src/components/CodeView/utils.tsx b/src/components/CodeView/utils.tsx
index 546e1ae..64c8f4b 100644
--- a/src/components/CodeView/utils.tsx
+++ b/src/components/CodeView/utils.tsx
@@ -8,8 +8,9 @@ const mapChild = (
   depth: number,
 ): React.ReactElement | string => {
   if (child.tagName) {
-    // @ts-ignore
-    const className = makeClassName(child.properties.className);
+    const className = makeClassName(
+      (child.properties && child.properties.className) || '',
+    );
 
     return React.createElement(
       child.tagName,
```

src/components/CodeView/utils.tsx
```diff
@@ -0,0 +1,53 @@
...
+      }),
+      // eslint-disable-next-line no-use-before-define, @typescript-eslint/no-use-before-define
+      child.children && child.children.map(mapWithDepth(depth + 1)),
+    );
+  }
+
+  return child.value;
+};
+
+export const mapWithDepth = (depth: number) => {
```
kumar303: This and `mapChild` doesn't appear to have test coverage

src/components/CodeView/utils.tsx
```diff
@@ -0,0 +1,53 @@
...
+  i: number,
+  depth: number,
+): React.ReactElement | string => {
+  if (child.tagName) {
+    // @ts-ignore
+    const className = makeClassName(child.properties.className);
+
+    return React.createElement(
+      child.tagName,
+      Object.assign({ key: `code-child-${depth}-${i}` }, child.properties, {
```
kumar303: I know this is copied code but you can use spreads here to make it more readable.

src/components/CodeView/index.tsx
willdurand: We can't have markers and line numbers.

src/components/CodeView/index.tsx
willdurand: Since we render data, `table` does make sense. That's what GitHub uses too (and yes, react-diff view also uses table).

src/components/CodeView/styles.module.scss
willdurand: Yep, I think this actually comes from `react-diff-view`. I'll file an issue to sync both.

src/components/CodeView/utils.tsx
willdurand: oops, thanks

src/components/CodeView/utils.tsx
willdurand: According to codecov, this is tested a it's used by the component. Can I add specific test cases in a follow-up?

src/components/CodeView/utils.tsx
willdurand: OMG, I actually never realized `properties` was duplicated. I am so stupid... I just realized now while trying to fix the issue with your patch 😥

src/components/CodeView/utils.spec.tsx
willdurand: Thanks!

src/components/CodeView/styles.module.scss
willdurand: I filed: https://github.com/mozilla/addons-code-manager/issues/298

src/components/CodeView/utils.tsx
willdurand: I filed: https://github.com/mozilla/addons-code-manager/issues/299

src/components/CodeView/index.spec.tsx
willdurand: those variables will actually be useful to fix your comment below.

src/components/CodeView/index.tsx
willdurand: yep, thanks

src/components/CodeView/index.tsx
kumar303: Well, I imagine github only uses tables for historic reasons and maybe to support IE. I guess syncing with `react-diff` is a good goal but I find that CSS grid is more flexible for these types of layouts.

willdurand: I can file an issue to switch to a grid but I really think it is simpler as a table, and it works.

We will have to inline fullwidth rows for the comments, and doing this with a grid, when you don't know the number of rows in the css, seems though. I initially tried the grid approach, then I looked at github and react-diff-view and convinced myself that table was the right approach.

Issue #112, opened 2019-02-01
Comment on a specific line of a file


kumar303:
User story:

> As a code reviewer, I want to be able to enter custom comments for specific code lines in the file and diff viewer.

When viewing a file (https://github.com/mozilla/addons-code-manager/issues/108), the reviewer can enter a comment on a specific line.

The suggested UI flow per [meeting notes](https://docs.google.com/document/d/1y4crrHxhGejb3eZ6HpjtBdDqONJakfwtfjaOqwq-qZo/edit) is roughly this:
- click on a line of a file to begin entering a comment
- ~~when presented with a list of canned responses, click *other*~~ (this will be added in https://github.com/mozilla/addons-code-manager/issues/780)
- begin entering text
- press save

**For QA**: note that refreshing the page won't show the comment yet (see https://github.com/mozilla/addons-code-manager/issues/713)

EnTeQuAk:
API-depencency: https://github.com/mozilla/addons-server/issues/11277

kumar303:
It looks like we still need a POST endpoint https://github.com/mozilla/addons-server/issues/11378

EnTeQuAk:
With mozilla/addons-server#11378 being fixed, I'm going to remove the `needs: api` label for now. Let me know if you need anything specific on top of that.

kumar303:
I still need to add an env var for this on the hosted dev site

kumar303:
Commenting is enabled on dev 🎉 It is disabled in stage and prod.

PR #290, opened 2019-02-27
Add a new basic Comment component

### LINKED ISSUE #289: Create a Comment component
We need a `Comment` component that will be inlined in the code of the browse and compare pages.

willdurand:
Fixes #289

---

This patch adds a new `Comment` component (shown in #285). It is basic but I believe it is a good start to prototype the UI and the commenting flow (hello #112).

screen shot 2019-02-27 at 14 03 41
screen shot 2019-03-01 at 09 39 17


src/components/Comment/styles.module.scss
```diff
@@ -0,0 +1,38 @@
...
+  // FontAwesome icon is 32px.
+  min-height: 32px + $padding + $padding;
+  // We add a padding-left to position the content after the icon.
+  padding-left: 32px + $padding + $padding;
+  position: relative;
+  width: 100%;
+}
+
+.commentIcon {
+  position: absolute;
```
kumar303: Why do you need absolute positioning for this? Were you trying to keep the icon at the top for a multi-line comment? If so, flex or grid should work for that.

src/components/Comment/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+      </Form>
+    );
+  }
+
+  render() {
+    const { comment } = this.props;
+
+    return (
+      <div className={styles.container}>
+        {comment ? this.renderComment() : this.renderForm()}
```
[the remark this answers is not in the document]
willdurand: Yeah, I thought about the `readOnly` prop too.

src/components/Comment/styles.module.scss
```diff
@@ -0,0 +1,38 @@
+@import '../../scss/variables.scss';
+
+$padding: 10px;
+
+.container {
+  padding: 10px;
```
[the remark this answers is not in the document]
willdurand: Not really but I started to use `10px` everywhere, I was thinking about creating a variable for that.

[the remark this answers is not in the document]
willdurand: haha, here is the variable. Yes, I guess it would make sense to use a variable, especially since that one is defined above.

src/components/Comment/styles.module.scss
willdurand: Exactly, it is to keep the comment in the top-left corner.

src/components/Comment/styles.module.scss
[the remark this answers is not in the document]
willdurand: I added this scss variable in another patch.

willdurand:
I'll rebase this patch once other patches will land so that I can leverage the shared SCSS variables.

willdurand:
I added the `readOnly` prop, the form renders the content of a comment (if any) for editing and I use a CSS grid.

src/components/Comment/styles.module.scss
```diff
@@ -0,0 +1,26 @@
...
+.comment,
+.form {
+  background-color: $light-gray;
+  border: 1px solid darken($light-gray, 10);
+  padding: $default-padding;
+}
+
+.textarea {
+  display: block;
+  margin-bottom: $default-padding;
```
kumar303: A bit of padding might be nice, too

src/components/Comment/styles.module.scss
```diff
@@ -0,0 +1,26 @@
...
+  width: 100%;
+}
+
+.comment {
+  display: grid;
```
kumar303: The defaults for flex seem to accomplish the same layout:

```diff
diff --git a/src/components/Comment/styles.module.scss b/src/components/Comment/styles.module.scss
index 42d5a91..04ac44f 100644
--- a/src/components/Comment/styles.module.scss
+++ b/src/components/Comment/styles.module.scss
@@ -19,8 +19,6 @@
 }
 
 .comment {
-  display: grid;
-  // The FontAwesome icon size is 32px.
-  grid-template-columns: 32px + $default-padding auto;
+  display: flex;
   width: 100%;
 }
```

Am I missing something? I checked the storybook and I think this does what we want.

src/components/Comment/index.spec.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const fakeComment = {
+    content: 'This is a review comment',
+  };
+
+  type RenderParams = {
+    comment?: PublicProps['comment'];
+    readOnly?: PublicProps['readOnly'];
+  };
+
+  const render = ({ comment = null, readOnly = true }: RenderParams = {}) => {
```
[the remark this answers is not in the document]
willdurand: Ah, good finding. I didn't know this meta type.

src/components/Comment/styles.module.scss
willdurand: ah, right..

src/components/Comment/styles.module.scss
willdurand: ah yes, thanks

willdurand:
Padding added in the `textarea`:


screen shot 2019-03-04 at 12 40 19

Issue #280, opened 2019-02-26
The `DiffView` component should only display the diff of a single file


willdurand:
The current `DiffView` component renders the diff of more than one file, which is not something we want. In #111, we want to show the diff between two add-on version but for a given file. Let's simplify this component!

willdurand:
I believe we can live with that for now, given that if it works for many files, it should work for a single file :)

PR #291, opened 2019-02-27
Add story for DiffView and make it pretty

### LINKED ISSUE #281: Add story for the DiffView component

willdurand:
Fixes #281 

---

This PR improves the UI of the `DiffView` component and adds a story for it.

## Screenshots

diff with deletions:

screen shot 2019-02-27 at 16 42 42

basic diff:

screen shot 2019-02-27 at 16 43 24

multiple diffs (which should not be supported, cf. #280):

screen shot 2019-02-27 at 16 43 33

### DIFF
--- a/src/@types/react-diff-view/index.d.ts
+++ b/src/@types/react-diff-view/index.d.ts
@@ -1,3 +1,5 @@
+/* eslint react/no-multi-comp: 0 */
+
 declare module 'react-diff-view' {
   type ChangeType = 'delete' | 'insert' | 'normal';
 
@@ -12,7 +14,7 @@ declare module 'react-diff-view' {
     type: ChangeType;
   };
 
-  type HunkInfo = {
+  export type HunkInfo = {
     changes: ChangeInfo[];
     content: string;
     isPlain: boolean;
@@ -24,7 +26,7 @@ declare module 'react-diff-view' {
 
   type Hunks = HunkInfo[];
 
-  type DiffInfo = {
+  export type DiffInfo = {
     oldPath: string;
     newPath: string;
     hunks: Hunks;
@@ -37,11 +39,12 @@ declare module 'react-diff-view' {
     type: string;
   };
 
-  function parseDiff(text: string, options?: object): DiffInfo[];
+  declare function parseDiff(text: string, options?: object): DiffInfo[];
 
   type ViewType = 'split' | 'unified';
 
   type DiffProps = {
+    className?: string;
     diffType: string;
     hunks: Hunks;
     viewType: ViewType;
@@ -49,4 +52,19 @@ declare module 'react-diff-view' {
 
   // eslint-disable-next-line no-undef
   export class Diff extends React.Component<DiffProps, {}> {}
+
+  type DecorationProps = {
+    className?: string;
+  };
+
+  // eslint-disable-next-line no-undef
+  export class Decoration extends React.Component<DecorationProps, {}> {}
+
+  type HunkProps = {
+    className?: string;
+    hunk: HunkInfo;
+  };
+
+  // eslint-disable-next-line no-undef
+  export class Hunk extends React.Component<HunkProps, {}> {}
 }
--- a/src/@types/react-diff-view/index.d.ts
+++ b/src/@types/react-diff-view/index.d.ts
@@ -44,6 +44,9 @@ declare module 'react-diff-view' {
   type ViewType = 'split' | 'unified';
 
   type DiffProps = {
+    // Don't ask why `children` is a function...
+    // https://github.com/otakustay/react-diff-view/blob/882f85adc46431f065a3458104e13f6233cdfcfd/src/Diff/__test__/Diff.test.jsx#L18
+    children: (hunks: HunkInfo[]) => ReactNode;
     className?: string;
     diffType: string;
     hunks: Hunks;
--- a/src/@types/react-diff-view/index.d.ts
+++ b/src/@types/react-diff-view/index.d.ts
@@ -44,8 +44,6 @@ declare module 'react-diff-view' {
   type ViewType = 'split' | 'unified';
 
   type DiffProps = {
-    // Don't ask why `children` is a function...
-    // https://github.com/otakustay/react-diff-view/blob/882f85adc46431f065a3458104e13f6233cdfcfd/src/Diff/__test__/Diff.test.jsx#L18
     children: (hunks: HunkInfo[]) => ReactNode;
     className?: string;
     diffType: string;
--- a/src/components/DiffView/index.spec.tsx
+++ b/src/components/DiffView/index.spec.tsx
@@ -4,6 +4,8 @@ import { Diff, parseDiff } from 'react-diff-view';
 
 import basicDiff from './fixtures/basicDiff';
 import multipleDiff from './fixtures/multipleDiff';
+import diffWithDeletions from './fixtures/diffWithDeletions';
+import styles from './styles.module.scss';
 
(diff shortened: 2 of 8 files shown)


src/components/DiffView/index.spec.tsx
```diff
@@ -43,4 +45,25 @@ describe(__filename, () => {
       expect(root.find(Diff).at(index)).toHaveProp('hunks', diff.hunks);
     });
   });
+
+  it('renders a header with diff stats', () => {
+    const root = render();
+
+    expect(root.find(`.${styles.header}`)).toHaveLength(1);
+    expect(root.find(`.${styles.stats}`)).toHaveLength(1);
+    expect(root.find(`.${styles.stats}`)).toIncludeText('+++ 2--- 0');
```
kumar303: This is good but it only covers one hunk. Can you add another test for a diff with more than one hunk?

src/components/DiffView/index.tsx
```diff
@@ -13,24 +22,81 @@ class DiffView extends React.Component<Props> {
...
+      <div className={styles.header}>
+        <div className={styles.stats}>
+          <span className={styles.statsAdditions}>{`+++ ${additions}`}</span>
+          <span className={styles.statsDeletions}>{`--- ${deletions}`}</span>
+        </div>
+      </div>
+    );
+  }
+
+  renderHunk(hunk: HunkInfo, index: number) {
```
kumar303: Since this is used as a callback to `.map()` it might make more sense to define it as a bound function to avoid future surprises.

src/components/DiffView/index.tsx
```diff
@@ -13,24 +22,81 @@ class DiffView extends React.Component<Props> {
...
+    const components = [
+      <Hunk key={`hunk-${hunk.content}`} hunk={hunk} className={styles.hunk} />,
```
kumar303: Sticking the entire content into `key` seems a bit brutal. Is there an easier key value? I mean, I guess it works.

src/components/DiffView/styles.module.scss
```diff
@@ -0,0 +1,42 @@
+@import '../../scss/variables';
+
+.diff {
+  border-bottom-left-radius: $border-radius;
+  border-bottom-right-radius: $border-radius;
+  border-collapse: separate !important;
```
kumar303: Can you use specificity instead of `!important`?

src/components/DiffView/styles.module.scss
```diff
@@ -0,0 +1,42 @@
...
+  padding: $default-padding;
+}
+
+.stats {
+  font-size: 0.8em;
+  font-weight: 600;
+}
+
+.statsAdditions {
+  color: green;
```
kumar303: What about pulling this and `red` into variable so we know to change them later?

src/components/DiffView/index.spec.tsx
```diff
@@ -43,4 +45,25 @@ describe(__filename, () => {
...
+  });
+
+  it('renders hunks with separators', () => {
+    const parsedDiff = parseDiff(diffWithDeletions)[0];
+
+    const root = shallow(<DiffView diff={diffWithDeletions} />);
```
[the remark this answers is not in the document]
willdurand: Correct, I think I made mistakes in the file between the different branches.

src/@types/react-diff-view/index.d.ts
```diff
@@ -37,16 +39,32 @@ declare module 'react-diff-view' {
     type: string;
   };
 
-  function parseDiff(text: string, options?: object): DiffInfo[];
+  declare function parseDiff(text: string, options?: object): DiffInfo[];
 
   type ViewType = 'split' | 'unified';
 
   type DiffProps = {
```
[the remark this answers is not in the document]
willdurand: Good idea, I'll move the comment in the definition file.

src/components/DiffView/index.tsx
willdurand: Why not, although `renderHunks` is bound already, no?

src/components/DiffView/index.tsx
willdurand: That's what the `react-diff-view` lib does.

src/components/DiffView/styles.module.scss
willdurand: I could add a comment at least, because it overrides the `react-diff-view` style. But maybe I could be more specific.

src/components/DiffView/styles.module.scss
willdurand: ah, that's a good idea.

src/components/DiffView/index.spec.tsx
```diff
@@ -43,4 +45,25 @@ describe(__filename, () => {
...
+    const diff = root.find(Diff).shallow();
```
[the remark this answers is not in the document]
willdurand: This works for me, although if the lib changes the interface, we won't notice it, right? I am thinking about `children` being a function. It's weird and maybe it will evolve?

src/components/DiffView/index.spec.tsx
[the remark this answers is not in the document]
willdurand: Yeah, I have this fixed somewhere else, sorry :(

src/components/DiffView/index.spec.tsx
willdurand: ah yes, thanks for that!

src/components/DiffView/index.spec.tsx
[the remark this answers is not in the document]
kumar303: Expecting `children` to be a function is a common [render prop technique](https://reactjs.org/docs/render-props.html#using-props-other-than-render) so it doesn't seem weird to me and I'm not too worried about it changing.

src/components/DiffView/index.tsx
kumar303: I think accessing `this` wouldn't work in the way you last had it.

src/components/DiffView/styles.module.scss
```diff
@@ -0,0 +1,43 @@
+@import '../../scss/variables';
+
+// We add `.diff` to override the react-diff-view default style.
```
kumar303: yeah, I think that's better 👍

src/@types/react-diff-view/index.d.ts
```diff
@@ -37,16 +39,35 @@ declare module 'react-diff-view' {
...
+    // Don't ask why `children` is a function...
```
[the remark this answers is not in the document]
willdurand: ah ok, I did not know `children` was used for that. It feels a bit weird to me because we can handle React `children` in a different way (using `React.Children.map`) but it is in the official docs so 🤷‍♂️

src/components/DiffView/index.spec.tsx
```diff
@@ -27,20 +29,51 @@ describe(__filename, () => {
...
+    expect(root.find(`.${styles.stats}`)).toIncludeText('+++ 24--- 4');
+  });
+
+  it('renders hunks with separators', () => {
+    const parsedDiff = parseDiff(diffWithDeletions)[0];
+    const root = render({ diff: diffWithDeletions });
+
+    // Simulate the interface of <Diff />
+    const children = root.find(Diff).prop('children');
+    const diff = shallow(<div>{children(parsedDiff.hunks)}</div>);
```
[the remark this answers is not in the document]
willdurand: That makes sense, I didn't think about accessing the `children` function directly, that's a good idea. Thanks!

PR #296, opened 2019-02-28
Syntax HL in diff view

### LINKED ISSUE #282: Add syntax highlighting to the `DiffView` component

willdurand:
Fixes #282 
~~⚠️ depends on #285 and #291~~

---

screen shot 2019-02-28 at 16 08 39

### DIFF
--- a/src/scss/variables.scss
+++ b/src/scss/variables.scss
@@ -4,5 +4,3 @@
 $border-radius: 0.25rem;
 $default-padding: 10px;
 $light-gray: $gray-200;
-$text-green: $green;
-$text-red: $red;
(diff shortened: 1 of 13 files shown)


src/scss/variables.scss
```diff
@@ -4,5 +4,5 @@
 $border-radius: 0.25rem;
 $default-padding: 10px;
 $light-gray: $gray-200;
-$text-green: green;
-$text-red: red;
+$text-green: $green;
```
kumar303: Could we just use `$green` instead of `$text-green` everywhere now?

willdurand: I suppose we could yes

PR #301, opened 2019-03-01
Unify styles between CodeView and DiffView

### LINKED ISSUE #298: Unify CodeView and DiffView styles
This is mainly about having the same values for `font-*`, `color`, etc.

willdurand:
Fixes #298

---

This patch adds common styles for the `CodeView` and `DiffView` components.

screen shot 2019-03-01 at 16 17 58
screen shot 2019-03-01 at 16 17 55


kumar303:
Thanks for jumping on this issue so quickly. I will take a look. 

Please try to only work on issues in the project [todo list](https://github.com/orgs/mozilla/projects/53) that come before the *Deploy dev, stage, prod* issue. I just want to be sure we're only working on what's necessary to launch a first version as it's getting very close to the end of the quarter. 

Unifying the styles between `CodeView` and `DiffView` is definitely nice but we probably could have deferred it until later.

kumar303:
> Please try to only work on...

I meant for future patches. Since you've already started this one, we should land it.

willdurand:
Roger that. It was "papercut Friday" today (and I was on a half-day PTO).

src/scss/variables.scss
```diff
@@ -4,3 +4,8 @@
 $border-radius: 0.25rem;
 $default-padding: 10px;
 $light-gray: $gray-200;
+
+// Shared style for the CodeView and DiffView components.
+$code-font-family: Consolas, Courier, monospace;
```
bobsilverberg: This is not the same as what was in `CodeView/styles.module.scss`. Is this an improvement on that?

src/components/DiffView/styles.module.scss
```diff
@@ -41,3 +41,12 @@
   line-height: 1;
   text-align: center;
 }
+
+:global {
```
kumar303: Sorry that I didn't have time to review this patch. I think we should avoid `:global` (there really [isn't a strong use case for it](https://github.com/gajus/react-css-modules#global-css)) and I also see other issues with this patch. I filed https://github.com/mozilla/addons-code-manager/issues/346 (low priority) as a follow-up.

Issue #294, opened 2019-02-27
Extract notices from linter result


kumar303:
For https://github.com/mozilla/addons-code-manager/issues/41 we need a function to extract linter messages data for two purposes:

1. Showing a message at the top of a file
2. Inserting a message at a specific line of a file

See https://github.com/mozilla/addons-code-manager/issues/41 for an example of the linter result format.

kumar303:
Note to self: I think I can grab some data fixtures from these old test cases: https://github.com/mozilla/addons-server/blob/1afea1c731ab450725ee5ee5eb2c2a218f54837d/media/js/zamboni/tests/validator_tests.js

PR #302, opened 2019-03-01
Extract messages from linter result

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/294

It's a little risky to try and build a data structure before you use it but when I look at `<CodeView>`, I think this is the way we'll need to insert messages.

### DIFF
--- a/src/linter/index.spec.tsx
+++ b/src/linter/index.spec.tsx
@@ -1,6 +1,6 @@
 import {
-  externalLinterResultSample,
-  externalLinterMessageSample,
+  fakeExternalLinterResult,
+  fakeExternalLinterMessage,
 } from '../test-helpers';
 
 import { createInternalMessage, ExternalLinterMessage, getMessageMap } from '.';
@@ -8,11 +8,11 @@ import { createInternalMessage, ExternalLinterMessage, getMessageMap } from '.';
 describe(__filename, () => {
   const _getMessageMap = (messages: Partial<ExternalLinterMessage>[]) => {
     return getMessageMap({
-      ...externalLinterResultSample,
+      ...fakeExternalLinterResult,
       validation: {
-        ...externalLinterResultSample.validation,
+        ...fakeExternalLinterResult.validation,
         messages: messages.map((msg) => {
-          return { ...externalLinterMessageSample, ...msg };
+          return { ...fakeExternalLinterMessage, ...msg };
         }),
       },
     });
@@ -158,7 +158,7 @@ describe(__filename, () => {
       const description = 'This is a detailed message about a code problem';
       expect(
         createInternalMessage({
-          ...externalLinterMessageSample,
+          ...fakeExternalLinterMessage,
           description,
         }),
       ).toMatchObject({
@@ -170,7 +170,7 @@ describe(__filename, () => {
       const description = ['This is a detailed message about a code problem'];
       expect(
         createInternalMessage({
-          ...externalLinterMessageSample,
+          ...fakeExternalLinterMessage,
           description,
         }),
       ).toMatchObject({
@@ -189,7 +189,7 @@ describe(__filename, () => {
 
       expect(
         createInternalMessage({
-          ...externalLinterMessageSample,
+          ...fakeExternalLinterMessage,
           column,
           description,
           file,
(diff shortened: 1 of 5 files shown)


src/linter/index.spec.tsx
```diff
@@ -0,0 +1,186 @@
...
+    for_appversions: {
+      '{ec8030f7-c20a-464f-9b0e-13a3a9e97384}': ['4.0b1'],
+    },
+    id: [],
+    line: 226,
+    message: 'on* attribute being set using setAttribute',
+    tier: 3,
+    type: 'notice',
+    uid: '9a07163bb74e476c96a2bd467a2bbe52',
+  };
```
willdurand: Could we move this to `src/test-helpers.tsx`? It might be useful in the near future (thinking about the reducer test file).

src/linter/index.spec.tsx
```diff
@@ -0,0 +1,186 @@
...
+            line,
+            message,
+            type,
+            uid,
+          },
+        ]),
+      ).toMatchObject({
+        [file]: {
+          byLine: {
+            [line]: [
```
willdurand: Yep, that sounds like a good way to store the linter results 👍

src/linter/index.tsx
```diff
@@ -0,0 +1,78 @@
+type Message = {
+  column: number | null;
+  description: string | string[];
```
willdurand: It'd be easier to use if we had only `string[]` (the `string |` could be replaced with a single-element array).

src/linter/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+    notices: number;
+    success: boolean;
+    warnings: number;
+  };
+};
+
+const createInternalMessage = (message: ExternalMessage): Message => {
+  return {
+    column: message.column,
+    description: message.description,
```
willdurand: Related to my comment above, a trick could be to `[].concat(message.description)` to return an array in any case.

src/linter/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+    file: message.file,
+    line: message.line,
+    message: message.message,
+    type: message.type,
+    uid: message.uid,
+  };
+};
+
+type MessageMap = {
+  [relativeFileName: string]: {
```
willdurand: This prop named `relativeFileName` is called `path` everywhere else (because the API names it that way). I'd suggest to use `path` here or to file a new issue to use an explicit and consistent name everywhere.

src/linter/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+    global: Message[];
+    byLine: { [line: number]: Message[] };
```
willdurand: I wonder how we'll consume this map in the UI 🤔  Either we'll test `byLine` for each line or do an `Object.keys()` before the loop over all the lines of the file. I am just thinking out loud.

kumar303: I have a feeling we will adjust the map shape once we start using it but that's OK. It's harder to guess but I didn't want to wait on the API.

src/linter/index.tsx
kumar303: Good idea, I like it

kumar303:
> I wonder why you decided to not put this logic into `src/reducers/`? 

It's a kind of a big chunk of code so it may be easier to deal with in its own module but I see what you're saying. I'm not totally sure what reducer it would live in yet. Maybe versions? We can always move it later.

src/linter/index.tsx
kumar303: TypeScript says no: `[].concat()` only accepts an array value (which MDN seems to confirm).

willdurand: mmh, MDN says `Arrays and/or values to concatenate into a new array`. I think TS is too picky here because this is one of the main purposes of `concat()`.

src/linter/index.tsx
kumar303: Ah, yes, this should be easier to work with

src/linter/index.tsx
kumar303: Oh, I see, mdn does say non-array values are allowed. That's what I thought at first when I started to get lost in a rabbit hole of TypeScript issues. Oh, well, this reads perfectly fine:

```js
    description: Array.isArray(message.description)
      ? message.description
      : [message.description],
```

kumar303:
Thanks for the review ⚡️

PR #303, opened 2019-03-01
Add tests for the map* functions

### LINKED ISSUE #299: Add more direct coverage to map* functions in CodeView/utils.tsx
Some map functions in `CodeView/utils.tsx` don't have test cases. These functions are covered only because they are used in the `CodeView` component.

willdurand:
Fixes #299

---

This patch adds test cases for the `map*` functions, used in the `CodeView` component.

### DIFF
--- a/src/@types/refractor/index.d.ts
+++ b/src/@types/refractor/index.d.ts
@@ -8,6 +8,9 @@ declare module 'refractor' {
     value: string;
     properties?: {
       className?: string | string[];
+      // A node might have arbitrary properties.
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      [key: string]: any;
     };
   };
 
--- a/src/@types/refractor/index.d.ts
+++ b/src/@types/refractor/index.d.ts
@@ -2,8 +2,8 @@ declare module 'refractor' {
   declare function register(language: object): void;
 
   type RefractorNode = {
-    children: RefractorNode[];
-    tagName: string;
+    children?: RefractorNode[];
+    tagName?: string;
     type: string;
     value: string;
     properties?: {
--- a/src/components/CodeView/utils.spec.tsx
+++ b/src/components/CodeView/utils.spec.tsx
@@ -136,6 +136,20 @@ describe(__filename, () => {
 
         expect(element).toEqual(value);
       });
+
+      it('passes all child properties to the element', () => {
+        const target = '_blank';
+        const title = 'link title';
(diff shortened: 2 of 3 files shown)


src/components/CodeView/utils.spec.tsx
```diff
@@ -16,4 +19,138 @@ describe(__filename, () => {
...
+        children,
+        properties,
+        tagName,
+        type,
+        value,
+      };
+    };
+
+    describe('mapChild', () => {
+      it('creates a React element from a child with a tagName and a className', () => {
```
kumar303: In addition to `className`, I'd like to see coverage of arbitrary properties. This test should do it but the TS types need adjusting somehow.

```diff
diff --git a/src/components/CodeView/utils.spec.tsx b/src/components/CodeView/utils.spec.tsx
index 9a3a839..1023a8e 100644
--- a/src/components/CodeView/utils.spec.tsx
+++ b/src/components/CodeView/utils.spec.tsx
@@ -52,6 +52,17 @@ describe(__filename, () => {
         expect(element).toHaveProperty('props.className', className);
       });
 
+      it('passes all child properties to the element', () => {
+        const target = '_blank';
+        const title = 'link title';
+        const child = createChild({ tagName: 'a', properties: { target, title } });
+
+        const element = mapChild(child, 0, 0);
+
+        expect(element).toHaveProperty('props.target', target);
+        expect(element).toHaveProperty('props.title', title);
+      });
+
       it('creates a React element with strictly positive depth and index', () => {
         const depth = 2;
         const i = 4;
```

willdurand: Added, thanks!

src/@types/refractor/index.d.ts
```diff
@@ -2,12 +2,15 @@ declare module 'refractor' {
...
-    tagName: string;
+    children?: RefractorNode[];
+    tagName?: string;
     type: string;
     value: string;
     properties?: {
       className?: string | string[];
+      // A node might have arbitrary properties.
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      [key: string]: any;
```
kumar303: Ooh, I like you can make this a fallback for any keys. I could be wrong, but I don't think Flow let you do this.

willdurand: Yes, and it does not alter the previous keys. I was afraid that `className` type would be overridden, but that's not the case, so that's super nice!

PR #305, opened 2019-03-04
Version chooser

### LINKED ISSUE #293: Create a `VersionChooser` component for the compare page
In order to ship #111, we need a version chooser that relies on this API patch: https://github.com/mozilla/addons-server/issues/10432.

See screenshot below:

willdurand:
Fixes #293

---

This patch adds a new `VersionChooser` component (that relies on a `VersionSelect` component for DRY). It will be used to select different versions of an add-on in order to see a `diff` in the compare page.

This is only the first step. Once we get and store data from the API, we can render real data and start the work to select a version. I envision this user action as a simple click into the list of versions that updates the URL and the diff view. That's why there is no "Go" button. [GitHub does that](https://github.com/mozilla/addons-code-manager/compare), too. We will probably need an exclusion list (or at least one version) on the `VersionSelect` component too, so that we can disable the version that has been selected in the other `select` (because we can't diff the same version).

## Screenshots

**New UI:**

screen shot 2019-03-05 at 19 23 28

Old UI:

screen shot 2019-03-04 at 11 43 11

### DIFF
--- a/src/components/VersionChooser/index.tsx
+++ b/src/components/VersionChooser/index.tsx
@@ -1,0 +1,56 @@
+import * as React from 'react';
+import { Col, Form, Row } from 'react-bootstrap';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+import VersionSelect, { Version } from '../VersionSelect';
+import { gettext } from '../../utils';
+import styles from './styles.module.scss';
+
+type PublicProps = {
+  versions: Version[];
+};
+
+class VersionChooserBase extends React.Component<PublicProps> {
+  render() {
+    const { versions } = this.props;
+
+    const listedVersions = versions.filter(
+      (version) => version.channel === 'listed',
+    );
+    const unlistedVersions = versions.filter(
+      (version) => version.channel === 'unlisted',
+    );
+
+    return (
+      <div className={styles.VersionChooser}>
+        <Form>
+          <Row className={styles.heading}>
+            <Col>
+              <h3>{gettext('Compare changes between two versions')}</h3>
+            </Col>
+          </Row>
+
+          <Form.Row>
+            <VersionSelect
+              label={gettext('Choose a base version')}
+              listedVersions={listedVersions}
+              unlistedVersions={unlistedVersions}
+            />
+
+            <div className={styles.arrow}>
+              <FontAwesomeIcon icon="long-arrow-alt-left" />
+            </div>
+
+            <VersionSelect
+              label={gettext('Choose a head version')}
+              listedVersions={listedVersions}
+              unlistedVersions={unlistedVersions}
+            />
+          </Form.Row>
+        </Form>
+      </div>
+    );
+  }
+}
+
+export default VersionChooserBase;
--- a/src/components/VersionChooser/index.tsx
+++ b/src/components/VersionChooser/index.tsx
@@ -32,7 +32,7 @@ class VersionChooserBase extends React.Component<PublicProps> {
 
           <Form.Row>
             <VersionSelect
-              label={gettext('Choose a base version')}
+              label={gettext('Choose an old version')}
               listedVersions={listedVersions}
               unlistedVersions={unlistedVersions}
             />
@@ -42,7 +42,7 @@ class VersionChooserBase extends React.Component<PublicProps> {
             </div>
 
             <VersionSelect
-              label={gettext('Choose a head version')}
+              label={gettext('Choose a new version')}
               listedVersions={listedVersions}
               unlistedVersions={unlistedVersions}
             />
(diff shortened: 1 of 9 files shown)


src/components/VersionChooser/index.tsx
```diff
@@ -0,0 +1,56 @@
+import * as React from 'react';
+import { Col, Form, Row } from 'react-bootstrap';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+import VersionSelect, { Version } from '../VersionSelect';
+import { gettext } from '../../utils';
+import styles from './styles.module.scss';
+
+type PublicProps = {
+  versions: Version[];
```
willdurand: The `Version` type (defined in the `VersionSelect` component) has a bad name but it is confusing because everything is called "version" now (the version number is `version`, a list of versions contains "version" objects), an add-on has versions, which have lots of data...

src/components/VersionChooser/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+class VersionChooserBase extends React.Component<PublicProps> {
+  render() {
+    const { versions } = this.props;
+
+    const listedVersions = versions.filter(
+      (version) => version.channel === 'listed',
+    );
+    const unlistedVersions = versions.filter(
+      (version) => version.channel === 'unlisted',
+    );
```
willdurand: The two `filter()` calls might go to the reducer in the near future and from there, we could have an enum or constants to represent the `channel` of a version. It's either `listed` or `unlisted` AFAIK.

willdurand:
> I don't think we should base this entirely on the github UI because that's based on git and git is confusing. The domain of add-on versions is a lot simpler than git.

I only tried to keep the existing behavior, without the confusing button. The names come from the API, so it isn/t really github. I would also argue that they probably had UX input on their component, so reusing their UI/UX pattern sounds like it would be an improvement already.

> What about a chooser that just lets you select an old and a new version? You could, of course, compare a new to an old version but no one really wants to do that.
> 
> This would introduce a work flow: you'd start by selecting the old version then move on to select the new version. This would allow us to filter the values in the second select box. As soon as you select the new version, it would show the compare view, just like on github.

We can, of course, revisit the work flow entirely but I thought we wanted to ship things (that reproduce what is already available in the existing tool), and then iterate?

IMO, we should rather fill the two select with the N-1 and N versions by default (N = latest version) because that makes sense from a reviewer POV. So the workflow I had in mind was rather:

- you go to the compare page of an add-on
- you see the diff between the N-1 and N versions of this add-on (N = latest version)
- you can choose a different base/head version, a.k.a. old/new (but the API says base/head)

It sounds to me that what's confusing for you here are the names "head" and "base". If we change that in the current patch, that would allow us to select an old version and a new version. Would that work for you? We can add labels if you want too, but I don't think we need to disable the second select until the first one has been updated.

> compare a new to an old version but no one really wants to do that.

For that, either the API supports this and we can do it *or* my idea was to detect that and invert the two parameters (with a flash message to inform the end user). This is, of course, a feature for after the MVP.

kumar303:
> It sounds to me that what's confusing for you here are the names "head" and "base". If we change that in the current patch, that would allow us to select an old version and a new version. Would that work for you? 

Yeah, the names are confusing to me and I don't think they capture the intent of the reviewer. If you can change them in the dropdowns then that would help.


> We can add labels if you want too...

I've been reading *Form Design Patterns* and it says you pretty much [always need a label](https://medium.com/simple-human/always-use-a-label-a39ceab554e6) (according to accessibility research), even for dropdowns. It's not super important for `code-manager` but it's a good habit to get into.

Your idea of limiting the options sounds good, too. I'll do a proper review of what's here so we can land something simple and iterate.

willdurand:
@kumar303 I quickly updated the UI:

screen shot 2019-03-05 at 19 23 28

Let me know if that works for you so that we can move forward (#310 adds the code on top of that to populate this component with real data).

src/components/VersionSelect/index.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+      {
+        id: 456,
+        channel: 'unlisted',
+        version: 'v2',
+      },
+    ];
+
+    const root = render({ listedVersions, unlistedVersions });
+
+    expect(root.find('optgroup').at(0)).toHaveProp('label', 'Listed');
```
kumar303: Instead of `.at(0)` and `.at(1)`, I suggest adding classes for each (like `.listedContro`) and using those in your query.

src/components/VersionChooser/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    return (
+      <div className={styles.VersionChooser}>
+        <Form>
+          <Row className={styles.heading}>
+            <Col>
+              <h3>{gettext('Compare changes between two versions')}</h3>
```
kumar303: You could simplify this prompt to *Compare changes* since the labels you added, *Choose an old version* and *Choose a new version*, imply enough context.

PR #308, opened 2019-03-04
Whitelist  the addons-linter validation report URL

### LINKED ISSUE #307: Whitelist the addons-linter validation report URL in the proxy
The Browse API returns a `validation_url_json` field that contains the URL to the linter results in JSON: https://addons-server.readthedocs.io/en/latest/topics/api/reviewers.html#get--api-v4-reviewers-addon-(int-addon_id)-versions-(int-version_id)-. This URL can be used to fetch the validation data (cf. #41) but because it is not an API endpoint, our server code fails to intercept the request and the client-side code tries to load a page for this URL.

Since the URL pattern is always the same, let's assume `/.../validation.json` is an API endpoint, similar to all URLs starting with `/api`.

willdurand:
Fixes #307

### DIFF
--- a/src/server/index.tsx
+++ b/src/server/index.tsx
@@ -65,7 +65,7 @@ export const createServer = ({
     }
 
     app.use(
-      proxy('/api', {
+      proxy(['/api/**', '/**/validation.json'], {
         target: env.REACT_APP_API_HOST,
         autoRewrite: true,
         changeOrigin: true,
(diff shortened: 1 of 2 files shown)


src/server/index.tsx
```diff
@@ -65,7 +65,7 @@ export const createServer = ({
     }
 
     app.use(
-      proxy('/api', {
+      proxy(['/api/**', '/**/validation.json'], {
```
kumar303: What about a more specific pattern?

willdurand:
> ok, I guess we will need this. I suggested a more specific regex to get `validation.json` but that path is already weird enough that it's unlikely clash. Whatever works.

Yes, I think it is fine as is now, we don't have API endpoints ending with `.json` and we don't serve JSON content from the client-side app.

Issue #295, opened 2019-02-27
Linter message component


kumar303:
This component will be a simple wrapper around `<Alert>` that renders a linter message for use in https://github.com/mozilla/addons-code-manager/issues/41

The old ones looked like this:

<img width="928" alt="screen shot 2019-02-27 at 10 54 39 am" src="https://user-images.githubusercontent.com/55398/53507793-273b5180-3a7e-11e9-96a9-ff1cb1b60739.png">

Issue #41, opened 2019-01-17
Show linter validation warnings/errors/notices in file viewer


EnTeQuAk:
User story:

> As a code reviewer, I want to see linter messages shown in the file viewer to speed up the review process.

Reviewers need inline access to addons-linter validation warnings, notices and errors. For reviewers though, warnings will be the primary entity. Errors usually lead to rejection so that the upload can't be submitted.

![screenshot_2019-01-17 code review and search tools prd 1](https://user-images.githubusercontent.com/139033/51310447-f6abd500-1a46-11e9-8702-c254a44e31ea.png)

The relevant data can be retrieved from `validation_url_json` from https://addons-server.readthedocs.io/en/latest/topics/api/reviewers.html#browse (note https://github.com/mozilla/addons-server/pull/10433 to update the docs)

Example json object:

```javascript
{
"errors": 0, // number of fatal validation errors
"success": true, // ``true`` or ``false`` if the overall validation ends up being a success (equivalent to ``errors == 0``)
"warnings": 1, // number of warnings, warnings should be visible and can result in rejection depending on their nature and frequency
"notices": 2, // number of notices, more informal messages
"messages": [{
    "description": [
        "Something something, see https://bugzilla.mozilla.org/"],
    "column": 0,  // column where the message applies to (not all warnings/errors/notices have this set properly, most do though)
    "line": 1,  // line where the message applies to
    "file": "manifest.json", // the file this message applies to
    "tier": 2,  // can be ignored, amo-validator had layered validation, addons-linter doesn't implement that, will always be ``1`` for WebExtensions
    "message": "Some warning", // the actual message
    "type": "warning", // the type of this message, "warning", "notice" or "error"
    "id": ['validation', 'messages', 'some_message_id'], // internal message id, used for deduplication
    "uid": "bb9948b604b111e09dfdc42c0301fe38" // unique identifier for this message
}],
"metadata": { // various metadata information that is usually used for auto-aproval risk weighting
    "id": "@webextension-guid",
    "manifestVersion": 2,
    "name": "WebExtensions test",
    "type": 1,
    "version": "0.0.2",
    "totalScannedFileSize": 502,
    "emptyFiles": [],
    "jsLibs": {},
    "unknownMinifiedFiles": [
        "background.js",
        "constant.js"
    ]
}
```

That's the object that's currently used by the file-viewer, I'm happy to put this into a proper API endpoint and change formatting into whatever works best for you.

wagnerand:
[...]

PR #309, opened 2019-03-04
Adds a LinterMessage component

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/295

Here is a sampling of the variants:

screen shot 2019-03-04 at 9 37 22 am

### DIFF
--- a/src/components/LinterMessage/index.spec.tsx
+++ b/src/components/LinterMessage/index.spec.tsx
@@ -101,4 +101,44 @@ describe(__filename, () => {
     expect(link.at(1)).toHaveProp('href', urls[1]);
     expect(link.at(2)).toHaveProp('href', urls[2]);
   });
+
+  it('renders HTML entities in messages', () => {
+    const message = 'The value of &lt;em:id&gt; is invalid';
+    const root = renderMessage({ message });
+
+    expect(root.find(Alert.Heading).html()).toContain(message);
+  });
+
+  it('renders HTML entities in descriptions', () => {
+    const description = [
+      'The values supplied for &lt;em:id&gt; in the install.rdf file is not a valid UUID string.',
+    ];
+    const root = renderMessage({ description });
+
+    expect(root.find(`.${styles.description}`).html()).toContain(
+      description[0],
+    );
+  });
+
+  it('escapes HTML in messages', () => {
+    // This is not a realistic input value. In real life, the HTML
+    // would be escaped.
+    const message = 'The value of <em:id> is invalid';
+    const root = renderMessage({ message });
+
+    expect(root.find(Alert.Heading).html()).toContain('&lt;em:id&gt;');
+  });
+
+  it('escapes HTML in descriptions', () => {
+    // This is not a realistic input value. In real life, the HTML
+    // would be escaped.
+    const description = [
+      'The values supplied for <em:id> in the install.rdf file...',
+    ];
+    const root = renderMessage({ description });
+
+    expect(root.find(`.${styles.description}`).html()).toContain(
+      '&lt;em:id&gt;',
+    );
+  });
 });
--- a/src/components/LinterMessage/index.spec.tsx
+++ b/src/components/LinterMessage/index.spec.tsx
@@ -1,10 +1,97 @@
 import * as React from 'react';
 import { shallow } from 'enzyme';
+import { Alert } from 'react-bootstrap';
 
-import LinterMessage from '.';
+import styles from './styles.module.scss';
+
+import LinterMessage, { PublicProps } from '.';
 
 describe(__filename, () => {
-  it('is cool', () => {
-    shallow(<LinterMessage type="error" message="thing" description="yes" />);
+  const render = (moreProps = {}) => {
+    const props: PublicProps = {
+      description: 'Your add-on uses a JavaScript library we consider unsafe.',
+      message: 'Banned 3rd-party JS library',
+      type: 'error',
+      ...moreProps,
+    };
+    return shallow(<LinterMessage {...props} />);
+  };
+
+  it.each([
+    ['danger', 'error'],
+    ['warning', 'warning'],
+    ['secondary', 'notice'],
+  ])('renders the Alert variant "%s" for "%s"', (alertVariant, messageType) => {
+    expect(render({ type: messageType }).find(Alert)).toHaveProp(
+      'variant',
+      alertVariant,
+    );
+  });
+
+  it('renders a message and description', () => {
+    const message = 'Markup parsing error';
+    const description = 'There was an error parsing...';
+    const root = render({ message, description });
+
+    expect(root.find(Alert.Heading)).toHaveText(message);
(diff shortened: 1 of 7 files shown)


kumar303:
Consider this in a draft state (I guess you can't switch to draft state?). I will fix up https://github.com/mozilla/addons-code-manager/pull/302 and rebase on that first.

willdurand:
Looks like there is an escaping issue in the message:

screen shot 2019-03-05 at 16 04 11

src/components/LinterMessage/index.tsx
```diff
@@ -0,0 +1,66 @@
+import * as React from 'react';
+import { Alert } from 'react-bootstrap';
+
+import styles from './styles.module.scss';
+import { LinterMessage as LinterMessageType } from '../../linter';
+
+export type PublicProps = {
+  description: LinterMessageType['description'];
+  message: LinterMessageType['message'];
+  type: LinterMessageType['type'];
```
willdurand: Question: why not passing a `LinterMessage` directly?

src/components/LinterMessage/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+};
+
+const getAlertVariant = (type: PublicProps['type']) => {
+  switch (type) {
+    case 'error':
+      return 'danger';
+    case 'warning':
+      return 'warning';
+    default:
+    case 'notice':
```
willdurand: `default` is positioned before so I don't think this is needed.

src/components/LinterMessage/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+      return 'secondary';
+  }
+};
+
+const urlPattern = /^https?:\/\//;
```
willdurand: Why is this extracted outside the function? It's used once, so I think we can inline it.

src/components/LinterMessage/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+const renderDescriptionPart = (part: string) => {
+  // Split a description into words to intercept and replace
+  // URLs with JSX links.
+  return part
+    .trim()
+    .split(' ')
```
willdurand: Would it make sense to split on `\s` instead?

src/components/LinterMessage/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+    <Alert variant={getAlertVariant(type)}>
+      <Alert.Heading>{message}</Alert.Heading>
+      <p className={styles.description}>
+        {description.reduce((lines: React.ReactNode[], text, index) => {
+          if (index > 0) {
+            lines.push(<br key={text} />);
+          }
+          lines.push(renderDescriptionPart(text));
+          return lines;
+        }, [])}
```
willdurand: Why is `renderDescriptionPart()` outside the component but not this `reduce()`? It'd be easier to read if we could have a `renderDescription()` function (that could maybe be a single function, embedding the current `renderDescriptionPart()` impl).

src/components/LinterMessage/index.tsx
kumar303: Whoa, `.split(/(\s)/)` magically adds the space back! I didn't know that.

> If separator is a regular expression that contains capturing parentheses, then each time separator is matched, the results (including any undefined results) of the capturing parentheses are spliced into the output array.

src/components/LinterMessage/index.tsx
kumar303: It's an old Python habit of only compiling the regex once but who knows how JS works. I'll move it inline.

src/components/LinterMessage/index.tsx
kumar303: I moved the reducers into one function and that does read better. Thanks.

kumar303:
> Looks like there is an escaping issue in the message

D'oh. It worked at first. Surprisingly, React escapes when a prop has nested strings (like an array or object) but not when it's just a string 🤷‍♀️

kumar303:
OK, yeah, I'm renaming this to `src/reducers/linter` because I'll definitely need it there for https://github.com/mozilla/addons-code-manager/issues/41

kumar303:
> I'm renaming this to src/reducers/linter

omg, I'm losing my mind. I don't know why I did that in this branch. Reverted 😛 I'll do it in the next patch.

willdurand:
> omg, I'm losing my mind. I don't know why I did that in this branch. Reverted 😛 I'll do it in the next patch.

ok, you can also do that in a small patch. easy to review.

src/components/LinterMessage/index.spec.tsx
```diff
@@ -0,0 +1,153 @@
...
+    expect(root.find(Alert.Heading)).toHaveText(message);
+    expect(root.find(`.${styles.description}`)).toIncludeText(description[0]);
+  });
+
+  it('handles descriptions with extra whitespace', () => {
+    const description = ['  lots    of    space     between'];
+    const root = renderMessage({ description });
+
+    expect(root.find(`.${styles.description}`)).toIncludeText(
+      description[0].trim(),
```
willdurand: If I understand correctly, it only trims each description entry. The value of `description` in this test case makes me think that, maybe, the extra spaces between each words are also trimmed, but that's not the case, right?

src/components/LinterMessage/index.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react';
+import unescape from 'lodash.unescape';
+import { Alert } from 'react-bootstrap';
+
+import styles from './styles.module.scss';
+import { LinterMessage as LinterMessageType } from '../../linter';
```
willdurand: If you rename the component `LinterMessageBase`, you could avoid this alias.

src/components/LinterMessage/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    allLines.push(
+      // Intercept and replace URLs with JSX links.
+      line
+        .trim()
+        .split(/(\s+)/)
+        .reduce((allParts: React.ReactNode[], part) => {
+          if (urlPattern.test(part)) {
+            allParts.push(<Alert.Link href={part}>{part}</Alert.Link>);
```
willdurand: Don't we need a `key` here too? 🤔

willdurand: Will these links be internal or external? I am thinking about react-router and target=_blank.

kumar303: They will be external, yeah

src/components/LinterMessage/index.spec.tsx
kumar303: Actually, this doesn't really need trim at all

src/components/LinterMessage/index.tsx
kumar303: I will file an issue to make the links external because I don't see an obvious way to do it

PR #310, opened 2019-03-04
Fetch, store and render a list of versions for an add-on

### LINKED ISSUE #306: Store the versions of an add-on in redux
In order to ship #111, we need to store the data coming from the API. The endpoint is already available and documented at: https://addons-server.readthedocs.io/en/latest/topics/api/reviewers.html#list-versions

A "version" in this context refers to an object with 3 properties:

- `id`: the version ID
- `channel`: either `listed` or `unlisted`
- `version`: (again, heh) the version number

willdurand:
Fixes #306 
~~⚠️ depends on #305~~

---

This patch adds the logic to fetch, store and give the data to the `VersionChooser` component (cf. #305). We can select versions but it does nothing yet (see #321). The Storybook has been updated to show the two main states of the `VersionChooser` component.

## Screenshots

screen shot 2019-03-06 at 11 11 03

screen shot 2019-03-04 at 17 00 38

### DIFF
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -2,7 +2,7 @@ import log from 'loglevel';
 
 import { ApiState } from '../reducers/api';
 import { ExternalUser } from '../reducers/users';
-import { ExternalVersion } from '../reducers/versions';
+import { ExternalVersion, ExternalVersionsList } from '../reducers/versions';
 
 export enum HttpMethod {
   DELETE = 'DELETE',
@@ -119,6 +119,21 @@ export const getVersion = async ({
     apiState,
     endpoint: `reviewers/addon/${addonId}/versions/${versionId}`,
     query: path ? { file: path } : undefined,
+  });
+};
+
+type GetVersionsListParams = {
+  addonId: number;
+  apiState: ApiState;
+};
+
+export const getVersionsList = async ({
+  apiState,
+  addonId,
+}: GetVersionsListParams) => {
+  return callApi<ExternalVersionsList>({
+    apiState,
+    endpoint: `reviewers/addon/${addonId}/versions/`,
   });
 };
 
--- a/src/components/VersionChooser/index.spec.tsx
+++ b/src/components/VersionChooser/index.spec.tsx
@@ -1,18 +1,64 @@
 import * as React from 'react';
-import { shallow } from 'enzyme';
+import { Store } from 'redux';
 
+import configureStore from '../../configureStore';
 import VersionSelect from '../VersionSelect';
-import { fakeVersions } from '../../test-helpers';
+import Loading from '../Loading';
+import {
+  ExternalVersionsList,
+  actions as versionActions,
+} from '../../reducers/versions';
+import {
+  createFakeThunk,
+  fakeVersionsList,
+  shallowUntilTarget,
+  spyOn,
+} from '../../test-helpers';
 
-import VersionChooser from '.';
+import VersionChooser, { VersionChooserBase, PublicProps } from '.';
 
 describe(__filename, () => {
-  const render = ({ versions = fakeVersions } = {}) => {
-    return shallow(<VersionChooser versions={versions} />);
+  type RenderParams = Partial<PublicProps> & { store?: Store };
+
+  const render = ({
+    _fetchVersionsList,
+    addonId = 123,
+    store = configureStore(),
+  }: RenderParams = {}) => {
+    const props = { addonId, _fetchVersionsList };
+
+    return shallowUntilTarget(
+      <VersionChooser {...props} />,
+      VersionChooserBase,
+      {
+        shallowOptions: {
+          context: { store },
+        },
+      },
+    );
   };
 
-  it('renders a VersionChooser', () => {
+  const _loadVersionsList = (
+    store: Store,
+    addonId: number,
+    versions: ExternalVersionsList,
+  ) => {
+    store.dispatch(versionActions.loadVersionsList({ addonId, versions }));
+  };
+
+  it('renders a loading message when lists of versions are not loaded', () => {
     const root = render();
+
+    expect(root.find(VersionSelect)).toHaveLength(0);
+    expect(root.find(Loading)).toHaveLength(1);
+  });
+
+  it('renders two VersionSelect components when lists of versions are loaded', () => {
+    const addonId = 999;
+    const store = configureStore();
+    _loadVersionsList(store, addonId, fakeVersionsList);
+
+    const root = render({ addonId, store });
 
     expect(root.find(VersionSelect)).toHaveLength(2);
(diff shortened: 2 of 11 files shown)


stories/VersionChooser.stories.tsx
```diff
@@ -1,16 +1,43 @@
...
+            return renderWithStoreAndRouter(
+              <VersionChooser addonId={addonId} />,
+              store,
+            );
+          },
+        },
+        {
+          title: 'loading state',
+          sectionFn: () => {
+            const addonId = 124;
```
kumar303: This makes me think these stories could benefit from a common render function that sets some defaults.

src/api/index.tsx
```diff
@@ -122,6 +122,21 @@ export const getVersion = async ({
...
+type GetVersionsListParams = {
+  addonId: number;
+  apiState: ApiState;
+};
+
+export const getVersionsList = async ({
+  apiState,
+  addonId,
+}: GetVersionsListParams) => {
+  return callApi<ExternalVersionsList>({
```
kumar303: I'm beginning to wonder if we really need functions to wrap `callApi`. It's just one line of code and still mockable. This is fine for now, just thinking out loud.

src/components/VersionChooser/index.spec.tsx
```diff
@@ -27,24 +73,54 @@ describe(__filename, () => {
...
+    render({ _fetchVersionsList, addonId, store });
+
+    expect(dispatch).toHaveBeenCalledWith(fakeThunk.thunk);
+    expect(_fetchVersionsList).toHaveBeenCalledWith({ addonId });
+  });
+
+  it('does not dispatch fetchVersionsList() on mount when lists are already loaded', () => {
+    const addonId = 888;
+    const store = configureStore();
+    _loadVersionsList(store, addonId, []);
```
kumar303: It could be useful to add another test that loads versions for the wrong addonId to check that versions are re-fetched.

src/components/VersionSelect/index.spec.tsx
```diff
@@ -2,28 +2,32 @@ import * as React from 'react';
...
-  const render = (props = {}) => {
+  const render = ({
+    label = 'select a version',
+    versions = fakeVersionsList,
+    withLeftArrow = false,
+  } = {}) => {
+    const addonId = 123;
+    const store = configureStore();
+    store.dispatch(versionActions.loadVersionsList({ addonId, versions }));
+    const versionsLists = store.getState().versions.byAddonId[addonId];
```
kumar303: I don't think this is necessary. You can just do:

```js
const versionsLists = createVersionsLists(versions);
```

src/components/VersionSelect/index.spec.tsx
```diff
@@ -50,22 +54,22 @@ describe(__filename, () => {
   });
 
   it('renders two lists of versions', () => {
-    const listedVersions = [
+    const listedVersions: ExternalVersionsList = [
       {
```
kumar303: It would be more future proof to add `fakeVersionList` to `test-helpers` and extend it here, like:

```js
const listedVersions: ExternalVersionsList = [
  {
    ...fakeVersionList,
    id: 123,
    channel: 'listed',
    version: 'v1',
  },
]
```

src/api/index.tsx
willdurand: it makes sense because we can type the return value.

src/components/VersionChooser/index.tsx
```diff
@@ -29,24 +45,41 @@ class VersionChooserBase extends React.Component<PublicProps> {
...
+              <VersionSelect
+                label={gettext('Choose a new version')}
+                listedVersions={versionsLists.listed}
+                unlistedVersions={versionsLists.unlisted}
+                withLeftArrow
+              />
+            </Form.Row>
+          ) : (
+            <Loading
+              message={gettext('Retrieving all the versions of this add-on...')}
```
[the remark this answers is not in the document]
willdurand: I am fine with aborting the action and turning off the loading state, but then what? What do we do?

The only issue we could have is someone unplugging the ethernet cable at this point, which is something we did not handle anywhere yet (except for the user maybe). Would it make sense to file an issue for that, for all the api calls? 

The current patch is pretty big, so that might be easier as a follow-up.

src/reducers/versions.tsx
```diff
@@ -112,6 +112,23 @@ export type Version = {
...
+  version: string;
+};
+
+type ExternalVersionsListItem = VersionsListItem;
+
+export type ExternalVersionsList = ExternalVersionsListItem[];
+
+export type VersionsList = VersionsListItem[];
+
+export type VersionsLists = {
```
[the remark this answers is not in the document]
willdurand: ah, that sounds better indeed, thanks!

stories/VersionChooser.stories.tsx
willdurand: :+1:

src/components/VersionSelect/index.spec.tsx
willdurand: ah, yes that's much simpler!

willdurand:
I just rebased this PR to fix a conflict.

src/api/index.tsx
kumar303: But if the thunk executes `callApi<ExternalVersionsList>(...)` directly then it's typed just the same. If we were using that same call in multiple places then it would make sense to have a function but that's the only reason I can think of. That's still a pretty good reason, though. When adding the next call, it may not be obvious to check for existing calls.

src/components/VersionChooser/index.tsx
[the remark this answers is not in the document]
kumar303: > I am fine with aborting the action and turning off the loading state, but then what? What do we do?

Then we render the error. Until then, the console will show the error. But, yeah, it's not required right now, I guess. I find it's easier on the brain to look at the Redux action stack and see an explicit loading state rather than an implicit one. 

We can deal with it later. I won't file an issue because we could possibly still infer a loading state even when we have an error handler. We can revisit the design when adding error handling.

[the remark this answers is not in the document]
kumar303: I just think it's a better habit to use explicit loading states. It will make everything fall into place later on.

src/components/VersionSelect/index.spec.tsx
```diff
@@ -50,22 +50,22 @@ describe(__filename, () => {
...
```
kumar303: I don't see this last review comment fully applied. I still think it would be a helpful change.

<hr />

It would be more future proof to add `fakeVersionList` to `test-helpers` and extend it here, like:

```js
const listedVersions: ExternalVersionsList = [
  {
    ...fakeVersionList,
    id: 123,
    channel: 'listed',
    version: 'v1',
  },
]
```

PR #311, opened 2019-03-04
WIP: Store file contents alongside VersionEntries

### LINKED ISSUE #248: VersionFile should include file type

bobsilverberg:
Fixes #248 

This ended up being a pretty big patch, but the work is mostly done now, so if we like this approach I think we can move forward with it. I do like the new shape of the reducer, and I think it's a bit friendlier to work with from a consumer standpoint, but the tests have become a bit more complicated.

I can do some work to clean the tests up a bit, perhaps with some helpers if we do decide to go this route.

For now perhaps it's best to just focus on the changes to the reducer and the consumers of the reducer, and if everyone is happy with this approach then we can look at the tests as well.

### DIFF
--- a/src/reducers/versions.tsx
+++ b/src/reducers/versions.tsx
@@ -3,7 +3,7 @@ import { ActionType, createAction, getType } from 'typesafe-actions';
 import log from 'loglevel';
 
 import { ThunkActionCreator } from '../configureStore';
-import { getVersion, isErrorResponse } from '../api';
+import { getVersion as getExternalVersion, isErrorResponse } from '../api';
 import { LocalizedStringMap } from '../utils';
 
 type VersionId = number;
@@ -36,12 +36,14 @@ export type ExternalVersionEntry = {
   size: number | null;
 };
 
+export type ExternalVersionEntries = {
+  [nodeName: string]: ExternalVersionEntry;
+};
+
 export type ExternalVersionFile = {
   content: string;
   created: string;
-  entries: {
-    [nodeName: string]: ExternalVersionEntry;
-  };
+  entries: ExternalVersionEntries;
   hash: string;
   id: number;
   is_mozilla_signed_extension: boolean;
@@ -89,6 +91,7 @@ export type VersionFile = {
 
 type VersionEntry = {
   depth: number;
+  file: VersionFile | null;
   filename: string;
   mimeType: string;
   modified: string;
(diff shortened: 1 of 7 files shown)


src/reducers/versions.tsx
```diff
@@ -89,6 +91,7 @@ export type VersionFile = {
 
 type VersionEntry = {
   depth: number;
+  file: VersionFile | null;
```
kumar303: We've had a lot of trouble in the past dealing with fields like this that may or may not be populated with an object. This is an indication to me that `file` does not belong here. Can we use another mapping?

bobsilverberg:
Well, I did spend a lot of time on this, but it seems like we have higher priorities, and @willdurand's workaround seems fine for now, so I guess we can just shelve it and perhaps revisit it in the future. I'm fine with that.

bobsilverberg:
I'm going to take a look at just adding some code to the reducer so that we have access to the entry data whenever we request a file, which is what @willdurand's workaround does. Adding this code into the reducer, instead of into the consumers seems like a good idea as we may need it in more than one place.

That should drastically reduce the size of the patch and shouldn't affect tests, etc. as this current patch does.

kumar303:
> I'm going to take a look at just adding some code to the reducer so that we have access to the entry data whenever we request a file

Thanks. I think that will be a great place to start. We may come back to this patch here later.

bobsilverberg:
I'll close this PR and try out the smaller idea.

PR #313, opened 2019-03-04
Create an ImageView component

### LINKED ISSUE #312: Create an ImageView component
This is a spin-off from #109. In order to view an image, instead of a code listing, we should have an `ImageView` component that is similar to the `CodeView` component, but renders an `` tag.

bobsilverberg:
Fixes #312 

Note that I tested this with some data from the API, and using the `btoa` function seems to be what we need to do in order to render the images from the API to the browser.


bobsilverberg:
> It'd be nice to have a storybook file for it.

Sure I can add one. Would you like it to be part of this patch, or a follow-up?

willdurand:
I think it'd make more sense to have it here so that I can test the component without hacking the Browse page.

bobsilverberg:
Ok, I've added a story that uses the 4 most common image file formats.

src/components/ImageView/index.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+    expect(root.find('img')).toHaveLength(0);
+  });
+
+  it('calls btoa to convert image data', () => {
+    const _btoa = jest.fn();
+
+    render({ _btoa });
+
+    expect(_btoa).toHaveBeenCalled();
```
willdurand: Do we want to assert the arguments given to `btoa` too?

src/components/ImageView/index.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  it('logs a warning when btoa fails', () => {
```
willdurand: It's a debug message instead of a warning, no?

src/components/ImageView/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  } catch (error) {
+    _log.debug('Error converting image data to ascii:', error);
+    return null;
+  }
+};
+
+const ImageViewBase = ({ _btoa, _log, content, mimeType }: PublicProps) => {
+  const imageData = convertImageData({ _btoa, _log, content });
+  return imageData ? (
+    <img alt="" src={`data:${mimeType};base64,${imageData}`} />
```
willdurand: Can we remove the empty `alt`?

src/components/ImageView/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  ) : null;
+};
```
willdurand: The other `CodeView` component has a border and round corners: https://mozilla.github.io/addons-code-manager/?selectedKind=CodeView&selectedStory=all%20variants&full=0&addons=1&stories=1&panelRight=0. Maybe we should wrap the `img` tag inside a `div` and add a similar style?

stories/ImageView.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
...gwBR0AXFlYAYVjAP658C3uQEgCAXNZRAg0UALAFQN9ZwLphQIowAIfIIR7tA7NsxHnkA7oMA9OeBC7gABYkAQTAARm5QGpJgHeUwDF0A1ZAQ9TcQ/wwA3dMAyBgAE/MAYscAI3cAIUhok4UAJNgAUIsA3bgCVJeIR44g7GQARX0AQT8AUggIkokAIc4Ac6AADFsA0pUQ6jSEi84A286AIh51gjMALA9W4o4AUw8AROMADbQA3O2A9NiA5IOBL84A/3sA6wYAERFAFeoI2Z2AFVEAQZkAvcQCxIBm6+oA3EEAgaMAUs8ADt6GwpYHEsEAb0WI/bQHtMcYTncJLogA70cA/b8AoWEAUz0P4B2BiPIOAFHmACYrABhVAN81AOSLaS8vAN36AKbRCIJIgDzoYDXrCUXgCSlDMA1GAO7FCKHXEP7YAO2KAKLykDEvAD2egFDwAGKVAB3rME1QAOaAkO4cAMzbANu5AISlAEVVAGX3ADOHCXTBkGe0CPZjAAx8IOHPGER5iS5zANqpAEGdBLHwABIMRPJqBtG2AAtoAM4VCZ4bAMy6AMrAAITIAFWTADEQAEQSAEP0ByT0AFe/AGZmAEehANxMIR+nCE8oAOMoENoaABBoA1TdAEYeB2UmAFXcAETNAHnXAMlFmZy8AMpNAIaZAGbvBXW2AF0pl/W1CdlGMErBkNJv7BEftoFIWpCoCAAErgBkhQBG/QBEUABVAgnICwAA1QnMiAmc2gDKVwCAtwBmhgB2/QBq/3ek5QBEiABNhpBHjwCdgQK9w5mDLhDNggDI2gBmpwBgkwAEaABEegBBiqBEzgnvDpDJipDKPQCaJlBqs5AEQAXUQwoEZAop0FGe8gDwl6D0h2kueADbpQCo1gCHeAB3lwBz56B4AQpIDQCaSwDOIgDh6qDL5wo4xwNiQapED6o3fAGKEQC9jgDeqQDXiCEPeQkkahDs+gDMqQDMlQDdVQC7QQC7LACrrQpvGJDupADs4wp9gADclQC7aAp7YQC7qAGLTgMbRQC7WQDHfZUKjZMCcJmpJxSg7qQA+kqA+akg7YMKnPkA3q4Azp0KjZ8AyVmg3YYA6gGqqTyqDYMA3lMA3UoAzYIA/+UKiIGqP0QA70MKtIVg7gUA9dmqmW6g/O0KhgWqnqoA7tEJXVEKpHepLgUA7Kag/TgA35IA+FKg8BAQA7',
```
willdurand: Could we put the files into a `fixtures` folder and `fs.readFileSync` them instead of hardcoding a long string like this?

src/components/ImageView/index.spec.tsx
bobsilverberg: Sure, although on second thought maybe we don't need this test at all? We are not mocking the `btoa` function in the first test, and that is checking the output we expect, so maybe we don't need a test that simply checks that `btoa` was called, because if it isn't being called then the first test would fail. What do you think?

src/components/ImageView/index.tsx
bobsilverberg: Without it there is an `ESLint` failure:

`img elements must have an alt prop, either with meaningful text, or an empty string for decorative images  jsx-a11y/alt-text`

stories/ImageView.stories.tsx
bobsilverberg: Sure, that sounds like a good idea. Would those fixtures live next to the component, or with the stories?

src/components/ImageView/index.tsx
willdurand: ah ok, my bad.

stories/ImageView.stories.tsx
willdurand: Probably in `stories/` if you're not using any of these files in the test file of the `ImageView` component, WDYT?

src/components/ImageView/index.spec.tsx
willdurand: That's right. We can probably keep the first test case only.

stories/ImageView.stories.tsx
bobsilverberg: We discussed this in IRC and there doesn't seem to be a way to make this work.

PR #319, opened 2019-03-05
Return mimeType and type from getVersionFile

### LINKED ISSUE #248: VersionFile should include file type

bobsilverberg:
Fixes #248 

This one is a bit smaller than the previous patch 😉

### DIFF
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -253,7 +253,7 @@ describe(__filename, () => {
       // condition that checks for a file will return undefined before we reach
       // this test under normal circumstances.
       state.versionInfo[version.id].entries = [];
-      getVersionFile(state, version.id, version.file.selected_file, _log);
+      getVersionFile(state, version.id, version.file.selected_file, { _log });
 
       expect(_log.debug).toHaveBeenCalled();
     });
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -219,9 +219,9 @@ describe(__filename, () => {
         actions.loadVersionInfo({ version: fakeVersion }),
       );
 
-      expect(
-        getVersionFile(state, version.id, `${version.file.selected_file}-1`),
-      ).toEqual(undefined);
+      expect(getVersionFile(state, version.id, 'path-to-unknown-file')).toEqual(
+        undefined,
+      );
     });
 
     it('returns undefined if there is no entry for the path', () => {
@@ -238,6 +238,24 @@ describe(__filename, () => {
       expect(
         getVersionFile(state, version.id, version.file.selected_file),
       ).toEqual(undefined);
+    });
+
+    it('logs a debug message if there is no entry for the path', async () => {
+      const _log = getFakeLogger();
+
+      const version = fakeVersion;
+      const state = reducer(
+        undefined,
+        actions.loadVersionInfo({ version: fakeVersion }),
+      );
+
+      // We have to manually remove the entry to test this. This is because the
+      // condition that checks for a file will return undefined before we reach
+      // this test under normal circumstances.
+      state.versionInfo[version.id].entries = [];
+      getVersionFile(state, version.id, version.file.selected_file, _log);
+
+      expect(_log.debug).toHaveBeenCalled();
     });
   });
 
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -1,6 +1,7 @@
 import { getType } from 'typesafe-actions';
 
 import reducer, {
+  VersionEntryType,
   actions,
   createInternalVersion,
   createInternalVersionAddon,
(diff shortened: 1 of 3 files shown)


src/reducers/versions.spec.tsx
```diff
@@ -175,23 +176,68 @@ describe(__filename, () => {
...
         undefined,
-        actions.loadVersionFile({ path, version }),
+        actions.loadVersionInfo({ version: fakeVersion }),
       );
 
-      expect(getVersionFile(state, version.id, path)).toEqual(
-        createInternalVersionFile(version.file),
-      );
+      expect(
+        getVersionFile(state, version.id, `${version.file.selected_file}-1`),
```
willdurand: is `-1` used to pass an unknown file? I think passing `unknown-file` would be enough.

src/reducers/versions.tsx
```diff
@@ -197,21 +208,37 @@ export const getVersionFiles = (
...
+      return {
+        ...file,
+        mimeType: entry.mimeType,
+        type: entry.type,
+      };
+    }
+  }
+
+  // The version or an entry was not found.
+  return undefined;
```
willdurand: If `entry` is missing, we should log a debug message because that'd be abnormal.

src/reducers/versions.spec.tsx
bobsilverberg: Yes, the intention is to pass a path for a file that hasn't been loaded.

I was trying to be explicit in the test that the path we're requesting is definitely not the one that's already loaded. The test loads a version with default values, rather than with an explicit path, because the latter requires more set up. So while *we* know that `unknown-file` is a path that doesn't match what's loaded, that's not explicit in the test.

Does that all make sense? I just don't like to rely on a knowledge of what the default values are in a test. Using `${version.file.selected_file}-1` is explicitly using a value that does not match what has been loaded.

src/reducers/versions.tsx
bobsilverberg: Hmm, `getVersionFile` is a selector, and we usually don't give them objects as arguments, but if I want to add a `_log` argument in order to test the logging we should probably change the signature to accept an object.

I could either just use `log` in the selector and not add a test for it, or I could change the signature to allow us to pass a `_log` argument into `getVersionFile` in order to test the log part.

Which do you think is better, @willdurand?

src/reducers/versions.spec.tsx
willdurand: I find it more difficult to understand than hard-coding a value that is self-explanatory but 🤷‍♂️ . Maybe add another suffix than `-1` then.

src/reducers/versions.tsx
willdurand: I think you can simply add a 4th argument:

```js
export const getVersionFile = (
  versions: VersionsState,
  versionId: VersionId,
  path: string,
  _log = log,
): VersionFile | void => {
  // ...
};
```

kumar303: I prefer to see options as objects because it's easier to grow the function later:

```js
export const getVersionFile = (
  versions: VersionsState,
  versionId: VersionId,
  path: string,
  { _log = log } = {},
): VersionFile | void => {
  // ...
};
```

src/reducers/versions.spec.tsx
bobsilverberg: Fair enough. I'll just use a self-explanatory string, as you suggest.

bobsilverberg:
> That looks good to me. I suggest we file an issue on addons-server requesting the following changes:
> 
>     * `file.entries` should be `entries`, outside `file` because that's confusing
> 
>     * `file` should contain `mimetype` and `mime_category`, possibly `sha256`, `size` and `modified` too. Except `mime_category`, I don't think we need the other props in `entries`. The algorithm to build the file tree only relies on `depth`, `filename`, `path` and `mime_category`.

I agree that having `entries` inside `file` is a bit odd, but it works for now and we probably won't have to touch this code again, so I'm not convinced it's worth asking for changes on the server side which we will then have to incorporate into our code.

I suppose if we don't need anything in `entry` other than `mime_category`, then it makes sense to move those other fields out of `entry` and into `file`. Is that what you're suggesting?

Finally, as long as we're suggesting changes to this API to directly support what we need in the app, would it make sense to also consider asking for validation messages for the file to be contained in the `file` entry. Right now we're getting all validation messages via a single end-point, and I guess we'll need those to build the file tree, so maybe it doesn't make sense... but it occurred to me that it might be nice to have just the validation messages for the current file with the file data. What do you think?

bobsilverberg:
I've pushed a commit which addressed the raised issues, but I'll hold off merging it until we decide what to do about the follow-up server bug(s) being discussed above.

willdurand:
> > That looks good to me. I suggest we file an issue on addons-server requesting the following changes:
> > ```
> > * `file.entries` should be `entries`, outside `file` because that's confusing
> > 
> > * `file` should contain `mimetype` and `mime_category`, possibly `sha256`, `size` and `modified` too. Except `mime_category`, I don't think we need the other props in `entries`. The algorithm to build the file tree only relies on `depth`, `filename`, `path` and `mime_category`.
> > ```
> 
> I agree that having `entries` inside `file` is a bit odd, but it works for now and we probably won't have to touch this code again, so I'm not convinced it's worth asking for changes on the server side which we will then have to incorporate into our code.

The API response has been designed before any client started to consume the API, so we don't have to keep it because we found a way to make it work. The proposed changes to the API response are needed to improve code quality, we need an issue in any case but that can wait after the MVP. It is not needed right now so this patch can land.

> I suppose if we don't need anything in `entry` other than `mime_category`, then it makes sense to move those other fields out of `entry` and into `file`. Is that what you're suggesting?

Correct. Hence the need to move `entries` outside `file` because that's not related to the file (viewer) but to the file tree.

> Finally, as long as we're suggesting changes to this API to directly support what we need in the app, would it make sense to also consider asking for validation messages for the file to be contained in the `file` entry. Right now we're getting all validation messages via a single end-point, and I guess we'll need those to build the file tree, so maybe it doesn't make sense... but it occurred to me that it might be nice to have just the validation messages for the current file with the file data. What do you think?

I am not sure yet but in any case, that would be a different issue. I have the feeling that having a different endpoint is simpler because we can use the same endpoint (and code to fetch/store/load) for both the diff and code views. That being said, we'll see once we'll have done that and I do agree that we can request more API changes if we need them (at least, we can file issues and discuss them).

I talked to @EnTeQuAk about that already and we both agree that the API responses can change to ease the frontend work (and to have better API responses in general too). This does not mean such things will get implemented right away.

src/reducers/versions.tsx
bobsilverberg: Thanks @willdurand and @kumar303. The options object is a good idea. I have updated the patch to do that.

bobsilverberg:
Thanks @willdurand, that all makes sense. I will merge this PR once Travis passes with my latest commit.

Issue #200, opened 2019-02-13
Provide an outline image of the code of a file including where linter warnings appear


bobsilverberg:
This is a vertical bar that appears to the left of the code of a file, when viewing a file, and includes colouring to indicate where in the fil linter warnings exist. It could also potentially be used to show where in the file comments exist (with a different colour), once we implement inline comments.

Here's an example from the current reviewer tools:

<img width="981" alt="screenshot 2019-02-13 11 44 05" src="https://user-images.githubusercontent.com/142755/52728115-b0359180-2f84-11e9-9a6a-c612f19f6467.png">

bobsilverberg:
I don't recall if we decided to include this in the current MVP or not.

kumar303:
We do need to somehow indicate that a line of a file has a notice so I'll mark it as mvp. It might be easier just to show the notice inline rather than use a marker but we'd then need a button to hide the notice, too.

willdurand:
> We do need to somehow indicate that a line of a file has a notice so I'll mark it as mvp. It might be easier just to show the notice inline rather than use a marker but we'd then need a button to hide the notice, too.

Maybe we could inline the notices in a way that is similar to how we'll inline comments?

kumar303:
Yeah, we can experiment with it and get some feedback from reviewers once we have a screen. The old approach of hiding notices behind a marker seems like it would encourage reviewers to ignore notices but I'm not sure.

willdurand:
Quick prototype based on #285 and #290:

![screen shot 2019-02-27 at 18 20 34](https://user-images.githubusercontent.com/217628/53509602-64becf80-3abc-11e9-92b4-efd92a6d7e3b.png)

kumar303:
Yeah, I was thinking inline like that, too. We just need to make sure we are addressing the right user story for how reviewers want to see validation errors.

kumar303:
This issue was originally filed with the assumption that reviewers wanted to retain the current tool's feature of hiding inline linter messages by default. In the current code manager, inline linter messages are not hidden by default. Let's get some feedback on that from real world usage before implementing the old style.

In the meantime, I'm closing this issue. Once we get user feedback we can open new issues as needed.

The description of this issue suggests that we need a better way to navigate inline linter messages. I'm moving that to: https://github.com/mozilla/addons-code-manager/issues/523

wagnerand:
[...]

PR #322, opened 2019-03-06
Add anchors/links on each line of the `CodeView` component

### LINKED ISSUE #317: Add anchors on each line of the `CodeView` component
We should have anchors on each line of the `CodeView` component so that a user can link to a specific line of a file (this needs #286). This is also a prerequisite for #200.

willdurand:
Fixes #317

---

This PR adds links on each line number of the `CodeView` component so that we can link to a specific line. This is needed for the outline (#200).

## Screenshots

screen shot 2019-03-06 at 14 58 43
screen shot 2019-03-06 at 14 57 55

### DIFF
--- a/src/components/CodeView/index.spec.tsx
+++ b/src/components/CodeView/index.spec.tsx
@@ -1,19 +1,54 @@
 import * as React from 'react';
-import { shallow } from 'enzyme';
+import { Link } from 'react-router-dom';
+import { Location } from 'history';
+import { mount } from 'enzyme';
 
 import refractor from '../../refractor';
 import { mapWithDepth } from './utils';
 import { getLanguageFromMimeType } from '../../utils';
+import {
+  createContextWithFakeRouter,
+  createFakeLocation,
+  shallowUntilTarget,
+} from '../../test-helpers';
 import styles from './styles.module.scss';
 
-import CodeView from '.';
+import CodeView, { CodeViewBase, PublicProps, scrollToSelectedLine } from '.';
 
 describe(__filename, () => {
+  type RenderParams = Partial<PublicProps> & {
+    location?: Location<{}>;
+  };
+
   const render = ({
-    content = 'some content',
-    mimeType = 'mime/type',
-  } = {}) => {
-    return shallow(<CodeView mimeType={mimeType} content={content} />);
+    location = createFakeLocation(),
+    ...otherProps
+  }: RenderParams = {}) => {
+    const props = {
+      content: 'some content',
+      mimeType: 'mime/type',
+      ...otherProps,
+    };
+
+    return shallowUntilTarget(<CodeView {...props} />, CodeViewBase, {
+      shallowOptions: createContextWithFakeRouter({ location }),
+    });
+  };
+
+  const renderWithMount = ({
+    location = createFakeLocation(),
+    ...otherProps
+  }: RenderParams = {}) => {
+    const props = {
+      content: 'some content',
+      mimeType: 'mime/type',
+      ...otherProps,
+    };
+
+    return mount(
+      <CodeView {...props} />,
+      createContextWithFakeRouter({ location }),
+    );
   };
 
   it('renders plain text code when mime type is not supported', () => {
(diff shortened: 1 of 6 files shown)


src/components/CodeView/index.spec.tsx
```diff
@@ -1,19 +1,54 @@
...
+    return shallowUntilTarget(<CodeView {...props} />, CodeViewBase, {
+      shallowOptions: createContextWithFakeRouter({ location }),
+    });
+  };
+
+  const renderWithMount = ({
+    location = createFakeLocation(),
+    ...otherProps
+  }: RenderParams = {}) => {
+    const props = {
```
kumar303: This can be factored into a shared function like `getRenderProps()`

src/components/CodeView/index.spec.tsx
```diff
@@ -65,4 +100,72 @@ describe(__filename, () => {
...
+    expect(root.find(`.${styles.line}`).at(1)).toHaveProp('id', 'L2');
+  });
+
+  it('marks a row as selected', () => {
+    const selectedLine = 2;
+    const location = createFakeLocation({ hash: `#L${selectedLine}` });
+
+    const root = render({ content: 'line 1\nline 2', location });
+
+    expect(root.find(`.${styles.selectedLine}`)).toHaveLength(1);
```
kumar303: Can this check that exactly line 2 is selected?

src/components/CodeView/index.spec.tsx
```diff
@@ -65,4 +100,72 @@ describe(__filename, () => {
...
+    const selectedLine = 2;
+    const lines = ['first', 'second'];
+    const content = lines.join('\n');
+    const location = createFakeLocation({ hash: `#L${selectedLine}` });
+    const _scrollToSelectedLine = jest.fn();
+
+    // We need `mount()` because `ref` is only used in a DOM environment.
+    const root = renderWithMount({ _scrollToSelectedLine, content, location });
+
+    expect(_scrollToSelectedLine).toHaveBeenCalledWith(
```
kumar303: I know that this relies on an execution of `ref` but I don't think that's necessary. This `ref` execution behavior is part of the React interface and won't be breaking anytime soon. You could do a shallow render and simply ensure that the `ref` prop with the expected value is passed into the right element. The drawback of using `mount()` is that all leaf components are rendered; this could make for a brittle test.

src/test-helpers.tsx
```diff
@@ -195,7 +195,7 @@ export const createFakeHistory = ({
 
 export const createContextWithFakeRouter = ({
   history = createFakeHistory(),
-  location = null,
+  location = null as Location<{}> | null,
```
kumar303: Huh. This is hurting my brain.

src/components/CodeView/index.spec.tsx
willdurand: ref does not exist outside mount, it's a special prop and enzyme does not see it when using shallow :(

src/test-helpers.tsx
willdurand: It forces `location` to be a nullable `Location` object.

src/components/CodeView/index.spec.tsx
willdurand: yep

src/components/CodeView/index.spec.tsx
willdurand: right, thanks!

src/components/CodeView/index.spec.tsx
kumar303: Ah, bummer. It's weird that enzyme hides the ref configuration.

src/components/CodeView/index.spec.tsx
```diff
@@ -65,4 +100,76 @@ describe(__filename, () => {
...
+      root.find(`#L${selectedLine}`).getDOMNode(),
+    );
+  });
+
+  describe('scrollToSelectedLine', () => {
+    it('calls scrollIntoView() when the element is not null', () => {
+      const element = {
+        // Create a HTMLTableRowElement that we can override.
+        ...document.createElement('tr'),
+        id: 'L2',
```
kumar303: This id doesn't seem necessary.

src/test-helpers.tsx
```diff
@@ -195,7 +195,8 @@ export const createFakeHistory = ({
 
 export const createContextWithFakeRouter = ({
   history = createFakeHistory(),
-  location = null,
+  // We tell TS that `location` is a nullable Location object.
```
kumar303: Thanks for the comment. I still don't get it but... 🤷‍♂️

PR #323, opened 2019-03-06
Add link + anchor to each line of the DiffView component

### LINKED ISSUE #318: Add anchors on each line of the `DiffView` component
We should have anchors on each line of the `DiffView` component so that a user can link to a specific line of a file (this needs #286) in the compare page. This is also a prerequisite for #200.

willdurand:
Fixes #318

---

This PR adds anchors/links to each line of the `DiffView` component. I used the built-in feature for that. It is not exactly like #322 but this page is different too, so I guess that's fine for now. ~~Maybe we could change the highlight color though.~~ done

## Screenshot

screen shot 2019-03-06 at 18 28 09

## Gif

(highlight color is gray now, see above)


src/components/DiffView/index.tsx
```diff
@@ -6,28 +6,48 @@ import {
...
+    const { _document, location } = this.props;
+
+    if (!location.hash.length) {
+      return;
+    }
+
+    const element = _document.querySelector(location.hash);
+
+    if (element) {
+      element.scrollIntoView();
```
kumar303: This seems fine for the first mount but what about scrolling when you click on a line? Could you file a follow-up issue to address that? However, I'm not sure why `react-diff-view` doesn't already scroll the page whenever you click on a line. Maybe its anchor tag click handler is returning false.

src/components/DiffView/index.spec.tsx
```diff
@@ -97,4 +112,69 @@ describe(__filename, () => {
...
+
+    expect(root.find(Diff)).toHaveProp('selectedChanges', [selectedChange]);
+  });
+
+  it('does not pass any selected changes when location hash is empty', () => {
+    const location = createFakeLocation({ hash: '' });
+
+    const root = render({ location });
+
+    expect(root.find(Diff)).toHaveProp('selectedChanges');
```
kumar303: This spec suggests that you meant to check that the prop is empty, right? Something like this, maybe?

```diff
diff --git a/src/components/DiffView/index.spec.tsx b/src/components/DiffView/index.spec.tsx
index f2aba68..04a951a 100644
--- a/src/components/DiffView/index.spec.tsx
+++ b/src/components/DiffView/index.spec.tsx
@@ -136,7 +136,7 @@ describe(__filename, () => {
 
     const root = render({ location });
 
-    expect(root.find(Diff)).toHaveProp('selectedChanges');
+    expect(root.find(Diff)).toHaveProp('selectedChanges', undefined);
   });
 
   it('tries to find the selected line element on mount', () => {
```

src/components/DiffView/index.tsx
willdurand: o_O Do you mean scrolling does not work for you? I believed it scrolled unless  the bottom of the file reaches the bottom of the page.

src/components/DiffView/index.spec.tsx
willdurand: ah yes, it'll be more explicit indeed

src/components/DiffView/index.tsx
```diff
@@ -111,4 +138,7 @@ class DiffView extends React.Component<PublicProps> {
   }
 }
 
-export default DiffView;
+// TODO: I can't make it work, I don't understand what's going on between
+// public, default and router props. TS gets confused by `withRouter` HOC.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export default withRouter(DiffViewBase) as any;
```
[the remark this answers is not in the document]
willdurand: Yes, I will give it another try to see if I can do anything, but this diff sounds like a better hack in the meantime.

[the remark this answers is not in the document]
willdurand: I think I understand what's going on now. The `withRouter` type is a bit weird but it does make sense: it takes a type `P` that extends the router props type and omits it when it returns the component so that the result of `withRouter<P>()` is of type `P`.

That makes sense but TS is confused because if `P` contains other props, e.g., from Redux, it can't tell the difference between required props from `PublicProps` and already provided props from Redux, hence the huge error with `InstrinsicAttributes`.

A better fix would be to be more descriptive about what the public API (i.e. public type) of our component is. I believe it is `PublicProps & Partial<DefaultProps>` so that we prevent rendering the component with unknown props and we also allow dependency injection in the test file.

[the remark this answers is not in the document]
willdurand: I think I understand what's going on now. The `withRouter` type is a bit weird but it does make sense: it takes a type `P` that extends the router props type and omits it when it returns the component so that the result of `withRouter<P>()` is of type `P`.

That makes sense but TS is confused because if `P` contains other props, e.g., from Redux, it can't tell the difference between required props from `PublicProps` and already provided props from Redux, hence the huge error with `InstrinsicAttributes`. This is not exactly what's happening here but I found a similar issue in another component. I believe here, TS gets confused because it looses track of the default props (TS support for that is very new).

A better fix would be to be more descriptive about what the public API (i.e. public type) of our component is. I believe it is `PublicProps & Partial<DefaultProps>` so that we prevent rendering the component with unknown props and we also allow dependency injection in the test file.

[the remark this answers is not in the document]
willdurand: I think I understand what's going on now. The `withRouter` type is a bit weird but it does make sense: it takes a type `P` that extends the router props type and omits it when it returns the component so that the result of `withRouter<P>()` is of type `P`.

That makes sense but TS is confused because if `P` contains other props, e.g., from Redux, it can't tell the difference between required props from `PublicProps` and already provided props from Redux, hence the huge error with `InstrinsicAttributes`. This is not exactly what's happening here but I found a similar issue in another component. I believe here, TS gets confused because it looses track of the default props (TS support for that is very new).

A better fix would be to be more descriptive about what the public API (i.e. public type) of our component is. I believe it is `PublicProps & Partial<DefaultProps>` so that we prevent rendering the component with unknown props and we also allow dependency injection in the test file.

src/components/DiffView/index.tsx
willdurand: It works for me, but because the diff is small, only the first lines scroll.

src/components/DiffView/index.spec.tsx
willdurand: I changed to empty array because passing `undefined` to `Diff` results in a prop set to `[]` (default value) so I set this default value in our code too.

src/components/DiffView/index.tsx
kumar303: > because the diff is small, only the first lines scroll

Ah, yes, this is what was leading me to believe it was not working. It is working 🎉

Issue #314, opened 2019-03-04
View linter messages in the compare view for versions


kumar303:
User story:

> As a code reviewer, I want to see linter messages shown in the version compare view to speed up the review process.

This will be similar to showing linter messages in the file viewer: https://github.com/mozilla/addons-code-manager/issues/41

Example from the current site:

<img width="1197" alt="screen shot 2019-03-04 at 3 00 03 pm" src="https://user-images.githubusercontent.com/55398/53762706-63abea80-3e8e-11e9-8ca7-8c6759dca071.png">

kumar303:
We will need a way to link into the `validation.json` URL: https://github.com/mozilla/addons-server/issues/10323#issuecomment-469415559

kumar303:
Linter messages returned by the API will be only for the current version

kumar303:
The linter messages shown will be for the newer version in the compare view

willdurand:
In the `DiffView` component, one can show inlined content using `getWidgets()`:

```diff
+  getWidgets = (hunks) => {
+    return (
+      this.state.changesWithComments
+        .reduce((widgets, change) => {
+          const changeKey = getChangeKey(change);
+
+          return {
+            ...widgets,
+            [changeKey]: (
+              <Comment
+                comment={
+                  change.lineNumber === 39
+                    ? null
+                    : { content: 'This is not needed.' }
+                }
+              />
+            ),
+          };
+        }, {})
+    );
+  };
```

It also needs `widgets={this.getWidgets(hunks)}` to each `Diff` in `render()`.

The diff above ha a list of comments stored in `this.state.changesWithComments` because a button was also added to add new comments. For linter results, we won't need the `state`, but it shows how to create a list of widgets. The `lineNumber` can be found via `change.lineNumber` but it could also be `change.oldLineNumber` or `change.newLineNumber`.

kumar303:
I think we added `needs: api` because we wanted the results included in the main API response (https://github.com/mozilla/addons-server/issues/10620) but we have a workaround in place.

PR #342, opened 2019-03-07
Add a quick check to ensure old version is older than the new version

willdurand:
This will be useful to make sure both #111 and #314 behave correctly. All the other components in `Compare` won't have to deal with the problem of knowing which version is the "old" one and which version is the "new" one. This check happens before any reviewer API calls.

It works because version IDs are integers and newer version IDs get higher numbers (there is some sort of auto-increment). I know we should avoid relying on the auto-increment property of an ID in general, but this is a very simple solution that works well for our purpose. I don't think we'll change the type of version IDs anytime soon (we'll likely always deal with integers) and I don't think we'll change the behavior that currently guarantees that new version IDs have higher numbers than older version IDs. Hence the proposed patch.

The longer version of this (mvp-polish, maybe?) would be to:

- fetch both versions separately (I don't think we need version data in the compare page so it would mean fetching the data only for the comparison)
- compare both versions, but based on what? There is currently no reliable value exposed to perform such a comparison. A `created_at` date could work, if this is something that we store in the database already
- do what's in this patch

If you browse: http://localhost:3000/en-US/compare/502955/versions/1541799...1541798/, this patch detects that the right version is lower than the left version and inverts them + update the URL.

Let me know what you think!

### DIFF
--- a/src/pages/Compare/index.spec.tsx
+++ b/src/pages/Compare/index.spec.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { Store } from 'redux';
+import { History } from 'history';
 
 import {
   createFakeHistory,
@@ -26,6 +27,7 @@ describe(__filename, () => {
       addonId: '999',
       baseVersionId: '1',
       headVersionId: '2',
+      lang: 'fr',
     },
   } = {}) => {
     return {
@@ -45,6 +47,8 @@ describe(__filename, () => {
     addonId?: string;
     baseVersionId?: string;
     headVersionId?: string;
+    history?: History;
+    lang?: string;
     store?: Store;
   };
 
@@ -53,11 +57,14 @@ describe(__filename, () => {
     addonId = '999',
     baseVersionId = '1',
     headVersionId = '2',
+    history = createFakeHistory(),
+    lang = 'fr',
     store = configureStore(),
   }: RenderParams = {}) => {
     const props = {
       ...createFakeRouteComponentProps({
-        params: { addonId, baseVersionId, headVersionId },
+        history,
+        params: { lang, addonId, baseVersionId, headVersionId },
       }),
       _fetchVersion,
     };
@@ -136,6 +143,41 @@ describe(__filename, () => {
     expect(_fetchVersion).toHaveBeenCalledWith({
       addonId,
       versionId: baseVersion.id,
+    });
+  });
+
+  it('makes sure the old version ID is older than the new version ID', () => {
+    const addonId = 123456;
+    const baseVersion = fakeVersion;
+    const headVersion = { ...fakeVersion, id: baseVersion.id - 1 };
+    const lang = 'es';
+
+    const store = configureStore();
+    const dispatch = spyOn(store, 'dispatch');
+    const fakeThunk = createFakeThunk();
+    const _fetchVersion = fakeThunk.createThunk;
+    const history = createFakeHistory();
+    const push = spyOn(history, 'push');
+
+    render({
+      _fetchVersion,
+      store,
+      addonId: String(addonId),
+      baseVersionId: String(baseVersion.id),
+      headVersionId: String(headVersion.id),
+      history,
+      lang,
+    });
+
+    expect(push).toHaveBeenCalledWith(
+      `/${lang}/compare/${addonId}/versions/${headVersion.id}...${
+        baseVersion.id
+      }/`,
+    );
+    expect(dispatch).toHaveBeenCalledWith(fakeThunk.thunk);
+    expect(_fetchVersion).toHaveBeenCalledWith({
+      addonId,
+      versionId: headVersion.id,
     });
   });
 });
--- a/src/pages/Compare/index.spec.tsx
+++ b/src/pages/Compare/index.spec.tsx
@@ -77,15 +77,15 @@ describe(__filename, () => {
   };
 
   type GetRouteParamsParams = {
-    addonId: number;
-    baseVersionId: number;
-    headVersionId: number;
+    addonId?: number;
+    baseVersionId?: number;
+    headVersionId?: number;
   };
 
   const getRouteParams = ({
-    addonId,
-    baseVersionId,
-    headVersionId,
+    addonId = 9999,
+    baseVersionId = 1,
+    headVersionId = 1000,
   }: GetRouteParamsParams) => {
     return {
       addonId: String(addonId),
@@ -142,7 +142,6 @@ describe(__filename, () => {
   it('dispatches fetchVersion() on mount', () => {
     const addonId = 123456;
     const baseVersion = fakeVersion;
-    const headVersion = { ...fakeVersion, id: baseVersion.id + 1 };
 
     const store = configureStore();
     const dispatch = spyOn(store, 'dispatch');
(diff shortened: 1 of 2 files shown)


src/pages/Compare/index.tsx
```diff
@@ -40,13 +41,25 @@ export class CompareBase extends React.Component<Props> {
...
+      );
+      oldVersionId = newVersionId;
+      newVersionId = parseInt(baseVersionId, 10);
+    }
 
     dispatch(
       _fetchVersion({
         addonId: parseInt(addonId, 10),
-        versionId: parseInt(baseVersionId, 10),
+        versionId: oldVersionId,
```
kumar303: This is a convenient way to do it but it's confusing because the component is no longer fetching the version requested in the URL. This makes the logic hard to follow and is going to make it hard to debug later on. 

To keep the logic nice and simple, we need to migrate to a `useEffect` hook but until then I would like to see the fetching logic changed to always use the URL parameter. I know it's a bit more complicated and will require more tests but this is more like what we need:

```diff
diff --git a/src/pages/Compare/index.tsx b/src/pages/Compare/index.tsx
index 7f8bc7f..764cca7 100644
--- a/src/pages/Compare/index.tsx
+++ b/src/pages/Compare/index.tsx
@@ -52,16 +52,33 @@ export class CompareBase extends React.Component<Props> {
       history.push(
         `/${lang}/compare/${addonId}/versions/${headVersionId}...${baseVersionId}/`,
       );
-      oldVersionId = newVersionId;
-      newVersionId = parseInt(baseVersionId, 10);
+      return;
     }
 
-    dispatch(
-      _fetchVersion({
-        addonId: parseInt(addonId, 10),
-        versionId: oldVersionId,
-      }),
-    );
+    this.loadData();
+  }
+
+  componentDidUpdate(oldProps: Props) {
+    this.loadData(oldProps);
+  }
+
+  loadData(oldProps?: Props) {
+    const { _fetchVersion, dispatch, match } = this.props;
+    const { addonId, baseVersionId } = match.params;
+    let oldVersionId = parseInt(baseVersionId, 10);
+
+    if (
+      !oldProps ||
+      addonId !== oldProps.match.params.addonId ||
+      baseVersionId !== oldProps.match.params.baseVersionId
+    ) {
+      dispatch(
+        _fetchVersion({
+          addonId: parseInt(addonId, 10),
+          versionId: oldVersionId,
+        }),
+      );
+    }
   }
 
   onSelectFile = () => {};
```

willdurand: I wasn't sure because these url params won't be used in the rest of this component and because there is a re-render, the children will receive the versions IDs that match the url (given that the url gets updated directly here) 0

kumar303: I'm not sure what you mean. The router prop `props.match.headVersionId` is passed to `_fetchVersion()` but your patch overloads that behavior. Why not always use the router prop and let `history.push()` be the one to change it? That's the patch I proposed.

willdurand: Sorry, let me rephrase: I was not sure implementing both `componentDidMount()` and `componentDidUpdate()` was useful because:

- the `baseVersionId` and `headVersionId` are only used in `componentDidMount()`, no need to invert them in `render()` for example (so the confusing behavior is small and scoped)
- as soon as `history.push()` is called, the URL is correct
- the `if` statement inverts the version IDs so that it matches the pushed URL, reflecting the URL seen by the user
- when children are rendered, they will receive the correct values in `baseVersionId` and `headVersionId` (if the children access the router props with `withRouter()`for example)

It's probably a good idea to early return after `history.push()` but this makes me cry already:

```
+    if (
+      !oldProps ||
+      addonId !== oldProps.match.params.addonId ||
+      baseVersionId !== oldProps.match.params.baseVersionId
+    ) {
```

src/pages/Compare/index.tsx
```diff
@@ -38,15 +39,47 @@ export class CompareBase extends React.Component<Props> {
...
+
+  loadData(prevProps?: Props) {
+    const { match } = this.props;
+    const { addonId, baseVersionId, headVersionId } = match.params;
+
+    if (
+      !prevProps ||
+      addonId !== prevProps.match.params.addonId ||
+      baseVersionId !== prevProps.match.params.baseVersionId ||
+      headVersionId !== prevProps.match.params.headVersionId
```
kumar303: `headerVersionId` isn't used in the fetch so you don't need to worry about this and you can delete the test.

src/pages/Compare/index.tsx
```diff
@@ -38,15 +39,47 @@ export class CompareBase extends React.Component<Props> {
...
+    ) {
+      const { dispatch, _fetchVersion } = this.props;
+      const oldVersionId = parseInt(baseVersionId, 10);
```
kumar303: For the purpose of `fetchVersion()`, I don't think it's necessary to rename this to `oldVersionId`. It doesn't add any helpful context.

src/pages/Compare/index.spec.tsx
```diff
@@ -127,9 +152,192 @@ describe(__filename, () => {
...
+      store,
+    });
+
+    dispatch.mockClear();
+    _fetchVersion.mockClear();
+    root.setProps({
+      match: {
+        params: getRouteParams({
+          addonId,
+          baseVersionId: baseVersion.id,
```
kumar303: Neither `baseVersionId` or `headVersionId` seem necessary for this test so if you can move them both to defaults in the helpers then that would make the test easier to read.

src/pages/Compare/index.spec.tsx
```diff
@@ -127,9 +152,192 @@ describe(__filename, () => {
...
+    });
+
+    expect(dispatch).toHaveBeenCalledWith(fakeThunk.thunk);
+    expect(_fetchVersion).toHaveBeenCalledWith({
+      addonId,
+      versionId: baseVersion.id,
+    });
+  });
+
+  it('dispatches fetchVersion() on update if head version is different', () => {
```
kumar303: This test looks like it can be deleted since it deals with `headVersionId`.

src/pages/Compare/index.spec.tsx
```diff
@@ -127,9 +152,192 @@ describe(__filename, () => {
...
+      },
+    });
+
+    expect(dispatch).not.toHaveBeenCalled();
+  });
+
+  it('dispatches fetchVersion() on update if base version is different', () => {
+    const addonId = 123456;
+    const baseVersion = fakeVersion;
+    const headVersion = { ...fakeVersion, id: baseVersion.id + 1 };
```
kumar303: `headVersionId` doesn't seem relevant to this test so it can be moved to a default in the helpers (if possible).

src/pages/Compare/index.spec.tsx
```diff
@@ -127,9 +152,192 @@ describe(__filename, () => {
     render({
       _fetchVersion,
       store,
+      ...getRouteParams({
+        addonId,
+        baseVersionId: baseVersion.id,
+        headVersionId: headVersion.id,
```
kumar303: `headVersionId` doesn't seem relevant to this test

src/pages/Compare/index.tsx
willdurand: ah sorry I thought I left a comment about it, this will be needed very soon because the compare api requires both ids.

src/pages/Compare/index.spec.tsx
willdurand: it is actually tricky to remove because the values of `addonId` and `baseVersionId` are used in the `expect()` call right after. `headVersionId` could be omitted though.

PR #343, opened 2019-03-07
Add Dockerfile and docker-compose files

willdurand:
This patch adds a production-ready `Dockerfile` as well as a `docker-compose.yml` file that shows how to use this `Dockerfile`.

Running `docker-compose up` should start code-manager in a local -dev mode (that is production mode but with -dev API and FxA). As you can see in the `docker-compose.yml` file, in order to run the -dev env, we only need to override a few settings. Because `docker-compose` is used locally, we also need the local settings (from `.env.common-local`) but it won't be needed for the hosted envs.

I believe ops will only have to `docker run` the image built with this `Dockerfile` and, for non-production environments, the following environment variable should be passed to `docker`: `REACT_APP_API_HOST`. Environment variables that are overridden for non-production envs are documented in `.env.stage` and `.env.dev`. 

We'll have to pass `REACT_APP_SENTRY_DSN` to each environment in addition to that. This should also be done in the config maintained by QA.

:warning: One thing that's missing is the `version.json` file.


Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+WORKDIR /app
+
+# Install the project's dependencies
+COPY package.json yarn.lock ./
+RUN yarn install --pure-lockfile --production=true
+
+# Add the compiled client-side application
+COPY --from=build /app/build/ ./build
+# Add the server code
+COPY --from=build /app/src/server/ ./src/server
```
kumar303: I'm slightly concerned that this will be hard to maintain: if we add new files outside of this path for any reason, we'll probably forget to update the Docker image.

An alternative would be to copy everything from `build` and just remove what we don't need, like `node_modules`. Did you consider that?

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '3'
```
kumar303: Would it be better to name this file `docker-compose.dev.yml` to indicate that it's not intended for production?

Dockerfile
willdurand: I like to bundle only what's used so I did not consider copying all the files. I don't really buy the maintainability issue because if, for some reason, we need more files, it means we need a build step for the server code as well.

docker-compose.yml
willdurand: ah, that would make sense indeed.

willdurand:
We could probably add some docs, although it does not seem super useful to run `docker-compose` with a production build targeting the -dev env.

I believe the `version.json` file can be generated in a separate patch (the one where we build + push the image to the Docker Hub). We can add `COPY version.json ./` to the second stage of the `Dockerfile`.

Dockerfile
kumar303: Yes, well, if we were to adjust the build step for the server then we'd have to remember to update this Dockerfile, too. That won't be obvious and nothing will fail. I guess we just need automated functional tests that run from the docker container to make it fail. 

I think the fragility of explicitly allowing select files isn't worth the security benefit but I don't want to hold up the patch. Ops may have more change requests when they look into this, too.

willdurand:
I heard you. I changed to copy all the source files. That should be less fragile.

Issue #348, opened 2019-03-07
Return a download link to the file from getVersionFile 


bobsilverberg:
Depends on https://github.com/mozilla/addons-server/issues/10877.

This is a follow-up to #347. The API does not provide us with a download link yet, so it is not being returned, but once it does we need to update the selector to return that data as well.

PR #349, opened 2019-03-07
Return data required by the file footer from getVersionFile

### LINKED ISSUE #347: Return data required by the file footer from getVersionFile
This supports #345.

To render the footer, in addition to the information already in `VersionFile` we need:
- ~~download link (which is currently not available via the API response)~~ (this will be done in #348)
- filename
- version
- SHA256 hash

bobsilverberg:
Fixes #347 

Note that this does not implement the download link. I opened #348 as a follow-up.


bobsilverberg:
@willdurand I filed https://github.com/mozilla/addons-server/issues/10891. Can you please take a look when you get a chance to see if I've missed anything?

PR #352, opened 2019-03-08
Create FileMetadata component

### LINKED ISSUE #345: Create FileMetadata component
This is a step towards #326. As this will need to be used in both the browse and compare views, it makes sense to make it a separate component.

Here's an screenshot from the current reviewer tools showing the footer. It currently includes:
- ~~a download link~~ (this is being deferred to #353)
- the version "number"
- the file size (Note that this currently just displays a number which represents the size of the file in bytes. I opened #355 as a follow-up to address this)
- the SHA256 hash
- the mime type

screenshot 2019-02-13 11 44 05

bobsilverberg:
Fixes #345 

Note: some follow-ups are needed, which I fill file and link back to the original issue.

### DIFF
--- a/src/components/FileMetadata/index.spec.tsx
+++ b/src/components/FileMetadata/index.spec.tsx
@@ -1,14 +1,28 @@
 import * as React from 'react';
 import { shallow } from 'enzyme';
 
-import { testVersionFile } from '../../test-helpers';
+import configureStore from '../../configureStore';
+import {
+  actions as versionActions,
+  getVersionFile,
+  VersionFile,
+} from '../../reducers/versions';
+import { fakeVersion } from '../../test-helpers';
 import styles from './styles.module.scss';
 
 import FileMetadata from '.';
 
 describe(__filename, () => {
   it('renders metadata for a file', () => {
-    const file = testVersionFile;
+    const store = configureStore();
+    const version = fakeVersion;
+    store.dispatch(versionActions.loadVersionInfo({ version }));
+
+    const file = getVersionFile(
+      store.getState().versions,
+      version.id,
+      version.file.selected_file,
+    ) as VersionFile;
 
     const root = shallow(<FileMetadata file={file} />);
 
--- a/src/components/FileMetadata/index.spec.tsx
+++ b/src/components/FileMetadata/index.spec.tsx
@@ -1,0 +1,20 @@
+import * as React from 'react';
+import { shallow } from 'enzyme';
+
+import { testVersionFile } from '../../test-helpers';
+import styles from './styles.module.scss';
+
+import FileMetadata from '.';
+
+describe(__filename, () => {
+  it('renders metadata for a file', () => {
+    const file = testVersionFile;
+
+    const root = shallow(<FileMetadata file={file} />);
+
+    expect(root.find(`.${styles.version}`)).toHaveText(file.version);
+    expect(root.find(`.${styles.size}`)).toHaveText(file.size.toString());
+    expect(root.find(`.${styles.sha256}`)).toHaveText(file.sha256);
+    expect(root.find(`.${styles.mimeType}`)).toHaveText(file.mimeType);
+  });
+});
--- a/src/components/FileMetadata/index.tsx
+++ b/src/components/FileMetadata/index.tsx
@@ -1,0 +1,28 @@
+import * as React from 'react';
+
+import { VersionFile } from '../../reducers/versions';
+import styles from './styles.module.scss';
+import { gettext } from '../../utils';
+
+type PublicProps = {
+  file: VersionFile;
+};
+
+const FileMetadataBase = ({ file }: PublicProps) => {
+  return (
+    <div className={styles.FileMetadata}>
+      <dl>
+        <dt>{gettext('Version')}</dt>
+        <dd className={styles.version}>{file.version}</dd>
+        <dt>{gettext('Size')}</dt>
+        <dd className={styles.size}>{file.size}</dd>
+        <dt>{gettext('SHA256 hash')}</dt>
+        <dd className={styles.sha256}>{file.sha256}</dd>
+        <dt>{gettext('Mimetype')}</dt>
+        <dd className={styles.mimeType}>{file.mimeType}</dd>
+      </dl>
+    </div>
+  );
+};
+
+export default FileMetadataBase;
--- a/src/components/FileMetadata/index.tsx
+++ b/src/components/FileMetadata/index.tsx
@@ -17,8 +17,10 @@ const FileMetadataBase = ({ file }: PublicProps) => {
         <dt>{gettext('Size')}</dt>
         <dd className={styles.size}>{file.size}</dd>
         <dt>{gettext('SHA256 hash')}</dt>
-        <dd className={styles.sha256}>{file.sha256}</dd>
-        <dt>{gettext('Mimetype')}</dt>
+        <dd className={styles.sha256}>
+          <code>{file.sha256}</code>
+        </dd>
+        <dt>{gettext('MIME type')}</dt>
         <dd className={styles.mimeType}>{file.mimeType}</dd>
       </dl>
     </div>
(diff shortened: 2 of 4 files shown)


src/test-helpers.tsx
```diff
@@ -88,6 +90,15 @@ export const fakeVersion: ExternalVersion = Object.freeze({
   version: '1.0',
 });
 
+export const testVersionFile: VersionFile = Object.freeze({
```
bobsilverberg: Note: I know this is ugly/confusing. The problem is that we now have three different "types" of VersionFile. 

1. There is [`ExternalVersionFile`](https://github.com/mozilla/addons-code-manager/blob/1c2910b410402786a1986851106b27757ff59a82/src/reducers/versions.tsx#L39), which describes the structure of file information returned by the API. [`fakeVersionFile`](https://github.com/mozilla/addons-code-manager/blob/1c2910b410402786a1986851106b27757ff59a82/src/test-helpers.tsx#L36) in this `test-helpers.tsx` file represents an instance of that - the external representation of the data, which is what all of the `fakeXxx` objects represent.
2. There is [`InternalVersionFile`](https://github.com/mozilla/addons-code-manager/blob/1c2910b410402786a1986851106b27757ff59a82/src/reducers/versions.tsx#L85), which describes the internal representation of a file that we store in the redux store.
3. There is [`VersionFile`](https://github.com/mozilla/addons-code-manager/blob/1c2910b410402786a1986851106b27757ff59a82/src/reducers/versions.tsx#L92), which describes the structure of a file that the `versions` reducer returns to us. This is what the app uses when rendering anything. I introduced this awhile back and I was worried at the time that it was going to cause confusion, but it seemed to be the least evil way of doing it.

This means that we now really need to be able to have an object which is of type `VersionFile` for both unit tests and storybook, which is why I created this thing called `testVersionFile`. I didn't want to use `fakeXxx`, because our convention thus far has been that those represent external objects returned by the API.

I think a better solution would be to introduce a new function into the reducer for `createVersionFile`, which accepts data from both [`versionInfo`](https://github.com/mozilla/addons-code-manager/blob/1c2910b410402786a1986851106b27757ff59a82/src/reducers/versions.tsx#L169) and [`versionFiles`](https://github.com/mozilla/addons-code-manager/blob/1c2910b410402786a1986851106b27757ff59a82/src/reducers/versions.tsx#L172), and then constructs an object of type `VersionFile`. I started down that path, but felt like the patch was getting too big, so I figured maybe it was best to submit this for now, and do that refactor in a follow-up, if it is deemed valuable.

willdurand: Would https://github.com/mozilla/addons-server/issues/10891 fix those issues?

I have started to play with the Compare API and that's a mess, now `version.file` can have a `diff` instead of `content`, and the rest is unchanged (and not very useful). I believe we should fix the API responses now, because that will never scale.

bobsilverberg: Yes, I think that https://github.com/mozilla/addons-server/issues/10891 coule potentially resolve this entire issue. I agree with bumping up its priority. I still think we can try to land things (especially as I now only have 2 days left before I leave for nearly 2 weeks), so maybe we can land this and live with it for awhile and fix it once https://github.com/mozilla/addons-server/issues/10891 lands?

willdurand: > so maybe we can land this and live with it for awhile and fix it once [mozilla/addons-server#10891](https://github.com/mozilla/addons-server/issues/10891) lands?

Yes!

src/components/FileMetadata/index.spec.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { shallow } from 'enzyme';
+
+import { testVersionFile } from '../../test-helpers';
+import styles from './styles.module.scss';
+
+import FileMetadata from '.';
+
+describe(__filename, () => {
+  it('renders metadata for a file', () => {
+    const file = testVersionFile;
```
willdurand: We could probably use a `store` and dispatch an action to retrieve this `file`, instead of having another object in `test-helpers`. This should be slightly more robust, and less weird.

I like that our `fake*` objects are API representations and the rest requires either `createInternal*` or an action when we need a selector.

src/components/FileMetadata/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+const FileMetadataBase = ({ file }: PublicProps) => {
+  return (
+    <div className={styles.FileMetadata}>
+      <dl>
+        <dt>{gettext('Version')}</dt>
+        <dd className={styles.version}>{file.version}</dd>
+        <dt>{gettext('Size')}</dt>
+        <dd className={styles.size}>{file.size}</dd>
+        <dt>{gettext('SHA256 hash')}</dt>
+        <dd className={styles.sha256}>{file.sha256}</dd>
```
willdurand: Maybe we could wrap the `sha256` into `<code />`, wdyt?

src/components/FileMetadata/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+        <dt>{gettext('Mimetype')}</dt>
```
willdurand: Is it `Mimetype` or `MIME type`?

src/components/FileMetadata/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+        <dd className={styles.mimeType}>{file.mimeType}</dd>
```
willdurand: Here is a patch for the suggestions above:

```diff
diff --git a/src/components/FileMetadata/index.tsx b/src/components/FileMetadata/index.tsx
index cb8b06a..b46afb3 100644
--- a/src/components/FileMetadata/index.tsx
+++ b/src/components/FileMetadata/index.tsx
@@ -17,8 +17,8 @@ const FileMetadataBase = ({ file }: PublicProps) => {
         <dt>{gettext('Size')}</dt>
         <dd className={styles.size}>{file.size}</dd>
         <dt>{gettext('SHA256 hash')}</dt>
-        <dd className={styles.sha256}>{file.sha256}</dd>
-        <dt>{gettext('Mimetype')}</dt>
+        <dd className={styles.sha256}><code>{file.sha256}</code></dd>
+        <dt>{gettext('MIME type')}</dt>
         <dd className={styles.mimeType}>{file.mimeType}</dd>
       </dl>
     </div>
```

src/components/FileMetadata/index.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+
+import { VersionFile } from '../../reducers/versions';
+import styles from './styles.module.scss';
```
willdurand: If this file does not contain anything, we can omit it I believe.

stories/FileMetadata.stories.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+import FileMetadata from '../src/components/FileMetadata';
+import { testVersionFile } from '../src/test-helpers';
+
+storiesOf('FileMetadata', module).add('default', () => (
+  <FileMetadata file={testVersionFile} />
```
willdurand: We should rather use a `store`. There is a `renderWithStoreAndRouter` helper in `./utils` (in `stories/`).

bobsilverberg: I didn't realize we had the option of using a `store` in a story, which is the main reason I created the `testVersionFile`. I'll change it to use a `store`.

src/components/FileMetadata/index.tsx
bobsilverberg: I used "Mimetype" as that's what's used in the current app:

Screenshot 2019-03-11 14 29 56

I will change it to "MIME type", but maybe we can get @wagnerand's opinion on what he things the label for this data should be?

src/components/FileMetadata/index.tsx
bobsilverberg: How can we omit it and also use things like `className={styles.FileMetadata}`? I can hardcode the classNames instead, but you had requested I not do that in an earlier patch.

willdurand: heh, nevermind.

PR #356, opened 2019-03-08
Add API code and types for the compare API

### LINKED ISSUE #358: Add API logic to fetch a diff
Supports #111

willdurand:
Fixes #358

---

This patch adds some code to call the Compare API (not released yet) and some types.

### DIFF
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -3,7 +3,7 @@ import log from 'loglevel';
 import { ApiState } from '../reducers/api';
 import { ExternalUser } from '../reducers/users';
 import {
-  ExternalVersion,
+  ExternalVersionWithContent,
   ExternalVersionWithDiff,
   ExternalVersionsList,
 } from '../reducers/versions';
@@ -119,7 +119,7 @@ export const getVersion = async ({
   addonId,
   versionId,
 }: GetVersionParams) => {
-  return callApi<ExternalVersion>({
+  return callApi<ExternalVersionWithContent>({
     apiState,
     endpoint: `reviewers/addon/${addonId}/versions/${versionId}`,
     query: path ? { file: path } : undefined,
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -2,7 +2,11 @@ import log from 'loglevel';
 
 import { ApiState } from '../reducers/api';
 import { ExternalUser } from '../reducers/users';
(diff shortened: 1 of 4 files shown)


willdurand:
I updated a type and added some more documentation for the TS synrax.

src/api/index.tsx
```diff
@@ -151,3 +155,25 @@ export const getCurrentUser = async (apiState: ApiState) => {
...
+  addonId,
+  apiState,
+  baseVersionId,
+  headVersionId,
+  path,
+}: GetDiffParams) => {
+  return callApi<ExternalVersionWithDiff>({
+    apiState,
+    endpoint: `reviewers/addon/${addonId}/versions/${baseVersionId}/compare_to/${headVersionId}`,
+    query: path ? { file: path } : undefined,
```
bobsilverberg: When would we not have a path? I guess this is the default screen, where no file has been chosen yet (like with the Browse page)?

src/reducers/versions.tsx
```diff
@@ -80,6 +80,55 @@ export type ExternalVersion = {
...
+  old_path: string;
+  parent: string;
+  path: string;
+  size: number;
+};
+
+// An `ExternalVersionFileWithDiff` object is very similar to an
+// `ExternalVersionFile` object: instead of having a `content` field, it has a
+// `diff` field.
+export type ExternalVersionFileWithDiff = Pick<
```
bobsilverberg: I've not seen this `Pick` syntax before. It's pretty cool. 👍

bobsilverberg: I wonder about the naming, though. These new types are called `ExternalVersionFileWithDiff` and `ExternalVersionWithDiff`, which means that the existing types are just called `ExternalVersionFile` and `ExternalVersion`. That means we are making an assumption about what these "default" types contain. Should those be renamed to `ExternalVersionFileWithFile` and `ExternalVersionWithFile` to make it more explicit what each type contains?

src/api/index.tsx
willdurand: Yep, exactly. This is very similar to the Browse API endpoint. There is a default file.

src/reducers/versions.tsx
willdurand: That would make all the version types more consistent.

willdurand: I believe it should be something `ExternalVersionFileWithContent` and `ExternalVersionWithContent` though.

PR #361, opened 2019-03-08
Add `createInternalDiffs` function to `versions` reducer

### LINKED ISSUE #359: Map API diff response to an internal representation needed for `react-diff-view`
Supports #111

willdurand:
Fixes #359 
~~⚠️ Depends on #356~~

---

This patch adds a function to map the diff part of the Compare API response to the internal representation needed for `react-diff-view`. It needs #356 which adds types.

Only the second commit is relevant here.

### DIFF
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -698,7 +698,7 @@ describe(__filename, () => {
           externalDiff.new_ending_new_line,
         );
 
-        // These props will be tested later.
+        // These props will be tested in a different test case below.
         expect(diff).toHaveProperty('type');
         expect(diff).toHaveProperty('hunks');
       });
@@ -748,7 +748,7 @@ describe(__filename, () => {
         expect(hunk).toHaveProperty('oldStart', externalHunk.old_start);
         expect(hunk).toHaveProperty('newStart', externalHunk.new_start);
 
-        // This prop will be tested later.
+        // This prop will be tested in a different test case below.
         expect(hunk).toHaveProperty('changes');
       });
     });
@@ -782,7 +782,7 @@ describe(__filename, () => {
           externalChange.new_line_number,
         );
 
-        // These props will be tested later.
+        // These props will be tested in a different test case below.
         expect(change).toHaveProperty('lineNumber');
         expect(change).toHaveProperty('isDelete');
         expect(change).toHaveProperty('isInsert');
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -1,9 +1,11 @@
 import { getType } from 'typesafe-actions';
 
 import reducer, {
+  ExternalVersionWithDiff,
   ExternalVersionsList,
   VersionEntryType,
   actions,
+  createInternalDiffs,
   createInternalVersion,
   createInternalVersionAddon,
   createInternalVersionEntry,
@@ -18,6 +20,7 @@ import reducer, {
   initialState,
 } from './versions';
 import {
+  fakeExternalDiff,
   fakeVersion,
   fakeVersionAddon,
   fakeVersionEntry,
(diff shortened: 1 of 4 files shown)


willdurand:
Rebased 😎

src/reducers/versions.tsx
```diff
@@ -428,6 +431,64 @@ export const fetchVersionsList = ({
...
+  headVersionId: number;
+};
+
+// This function returns an array of objects compatible with `react-diff-view`.
+export const createInternalDiffs = ({
+  baseVersionId,
+  headVersionId,
+  version,
+}: CreateInternalDiffParams) => {
+  const GIT_STATUS_TO_TYPE: { [mode: string]: string } = {
```
bobsilverberg: Should this be an `enum` somewhere to enforce the fact that `type` needs to be one of these strings? Currently `type` in `DiffInfo` is just defined as `string`.

src/reducers/versions.tsx
```diff
@@ -428,6 +431,64 @@ export const fetchVersionsList = ({
   };
 };
 
+type CreateInternalDiffParams = {
```
bobsilverberg: This type is called `CreateInternalDiffParams` but the function is called `createInternalDiffs`. Note the `diff` in the former but `diffs` in the latter.

src/reducers/versions.tsx
```diff
@@ -428,6 +431,64 @@ export const fetchVersionsList = ({
...
+    R: 'rename',
+  };
+
+  return version.file.diff.map(
+    (diff: ExternalDiff): DiffInfo => {
+      return {
+        newRevision: String(headVersionId),
+        oldRevision: String(baseVersionId),
+        hunks: diff.hunks.map((hunk: ExternalHunk) => ({
+          changes: hunk.changes.map((change: ExternalChange) => ({
```
bobsilverberg: Observation: Sometimes we extract functions for these, such as `createInternalChange` and `createInternalHunk`, but I guess if you don't need to call those anywhere else (we often need them in tests), then there's no reason to extract them into their own functions.

src/reducers/versions.spec.tsx
```diff
@@ -616,4 +619,237 @@ describe(__filename, () => {
...
+        );
+
+        // These props will be tested later.
+        expect(diff).toHaveProperty('type');
+        expect(diff).toHaveProperty('hunks');
+      });
+    });
+
+    it.each([
+      ['add', 'A'],
```
bobsilverberg: It seems like if you made these into an `enum` that could also be used in the test.

src/reducers/versions.spec.tsx
```diff
@@ -616,4 +619,237 @@ describe(__filename, () => {
       });
     });
   });
+
+  describe('createInternalDiffs', () => {
```
bobsilverberg: I like the approach you took with this, deferring the tests for more "nested" data for a test to be executed later. I think what we've done in the past is created helper methods like `createInternalChange` and `createInternalHunk` and then used those in the tests for assertions, and then we've also had to include tests for the helper methods themselves. This (what you've done here) seems like a reasonable approach to me, but I thought I'd point out that it's different from what we've done in the past.

src/reducers/versions.spec.tsx
```diff
@@ -616,4 +619,237 @@ describe(__filename, () => {
...
+        expect(diff).toHaveProperty(
+          'oldEndingNewLine',
+          externalDiff.old_ending_new_line,
+        );
+        expect(diff).toHaveProperty(
+          'newEndingNewLine',
+          externalDiff.new_ending_new_line,
+        );
+
+        // These props will be tested later.
```
bobsilverberg: Nit: When I first encountered this comment I thought you meant this was being deferred to a follow-up, so maybe saying something like "These props will be tested in a different test case below." would make it a bit clearer?

src/reducers/versions.spec.tsx
```diff
@@ -616,4 +619,237 @@ describe(__filename, () => {
...
+      const change = diff.hunks[0].changes[0];
+
+      expect(change).toHaveProperty('type', type);
+      expect(change).toHaveProperty('isDelete', false);
+      expect(change).toHaveProperty('isInsert', true);
+      expect(change).toHaveProperty('isNormal', false);
+      expect(change).toHaveProperty('lineNumber', newLineNumber);
+    });
+
+    it('creates "normal" changes', () => {
```
bobsilverberg: Nit: You have individual tests for each of the change types, but there is no test that deals with a diff with multiple changes, and verifies that each of them are properly translated and the resulting InternalDiff is accurate. Maybe we don't need that, but it feels to me like something is missing.

src/reducers/versions.tsx
willdurand: If we can do the exact same thing (i.e. not using a switch/case), then why not.

src/reducers/versions.tsx
willdurand: ah oops, typo

src/reducers/versions.tsx
willdurand: A `hunk` or a `change` per se does not have much value.

src/reducers/versions.spec.tsx
willdurand: :+1:

src/reducers/versions.spec.tsx
willdurand: In `creates changes from external diff changes`, we iterate over a set of `changes` so we have that. Given that the test case highlighted above (and the similar ones) always assert the first `change`, we can conclude that if it works for the first one and we have multiple changes, the same logic will be applied to all changes.

src/reducers/versions.tsx
willdurand: Ah! There are string enums: https://www.typescriptlang.org/docs/handbook/enums.html 🏆

willdurand: So... enums don't work in declaration files apparently: https://lukasbehal.com/2017-05-22-enums-in-declaration-files/.

I tried and failed. `enum` is never exported. I introduced a union type instead and we get what we want in the end: type safety on `type`.

src/reducers/versions.spec.tsx
willdurand: As explained above, `enum` does not work.

src/reducers/versions.tsx
bobsilverberg: > enums don't work in declaration files 

That's too bad. Oh well.

PR #364, opened 2019-03-08
Display a friendly file size in the FileMetadata component

### LINKED ISSUE #355: Display a friendly file size in the FileMetadata component
Depends on #345

Currently, the size of a file is displayed in bytes, which is the number that is returned to us by the API. The old version of reviewer tools displays this in a friendlier format (e.g., in "KiB").

We should decide if we need to do that for code-manager, and if so, include code to do the conversion/translation. There is existing code in addons-frontend that I wrote for the versions page which does this, so this can be as simple as porting that code over to code-manager.

I imagine this should be an "mvp" issue, as opposed to "mvp-polish", but we can discuss that. @wagnerand What is your opinion on that?

Note that I am formatting the sizes using [`iec` notation](https://en.wikipedia.org/wiki/Binary_prefix#IEC_prefixes), which appears to be what the current review tools use.

bobsilverberg:
Fixes #355


bobsilverberg:
> I'd rather see the formatting logic in the UI because that's part of the UI. We should keep the raw data as is IMO (or add a distinct formatted_size prop).

Fair enough. I'll make this depend on #345 again and make the changes in that component.

bobsilverberg:
Ok, this is now built on top of #345, so only the second commit in here is relevant.

bobsilverberg:
Thanks @willdurand. It's now been rebased.

bobsilverberg:
Oops, I see that CI is failing. I didn't test after rebasing (silly me). I'm about to go to training, but I will address these failures when I get back.

bobsilverberg:
CI seems to be happy with this now.

src/components/FileMetadata/index.spec.tsx
```diff
@@ -1,34 +1,44 @@
-import * as React from 'react';
 import { shallow } from 'enzyme';
+import * as React from 'react';
 
 import configureStore from '../../configureStore';
 import {
   actions as versionActions,
-  getVersionFile,
+  getVersionFile as _getVersionFile,
```
willdurand: We tend to the do the opposite, i.e. prefixing with `_` in the code and not using import aliases.

src/components/FileMetadata/index.spec.tsx
```diff
@@ -1,34 +1,44 @@
-import * as React from 'react';
 import { shallow } from 'enzyme';
+import * as React from 'react';
```
willdurand: This does not seem like a useful change.

src/components/FileMetadata/index.spec.tsx
```diff
@@ -1,34 +1,44 @@
...
     const root = shallow(<FileMetadata file={file} />);
 
     expect(root.find(`.${styles.version}`)).toHaveText(file.version);
-    expect(root.find(`.${styles.size}`)).toHaveText(file.size.toString());
     expect(root.find(`.${styles.sha256}`)).toHaveText(file.sha256);
     expect(root.find(`.${styles.mimeType}`)).toHaveText(file.mimeType);
   });
+
+  it('renders a formatted filesize', () => {
+    const file = { ...getVersionFile(), size: 12345 };
```
willdurand: I believe `getVersionFile` should take `size` as input.

src/components/FileMetadata/index.spec.tsx
bobsilverberg: I had originally added another import, at which point I realized these were not in alphabetical order. I can revert it though.

Issue #370, opened 2019-03-08
Add reducer / thunk for fetching and storing linter messages


kumar303:
This will add the thunk and reducer to fetch and store linter messages. This is needed to display linter messages in https://github.com/mozilla/addons-code-manager/issues/41

PR #371, opened 2019-03-08
Add reducer / thunk for fetching and storing linter messages

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/370

I have another patch that applies this to the UI so I think it's a good structure to start with.

### DIFF
--- a/src/reducers/linter.tsx
+++ b/src/reducers/linter.tsx
@@ -107,7 +107,7 @@ export const getMessageMap = (
 };
 
 export type LinterState = {
-  forVersionId: undefined | number;
+  forVersionId: void | number;
   isLoading: boolean;
   messageMap: undefined | LinterMessageMap;
 };
--- a/src/reducers/linter.tsx
+++ b/src/reducers/linter.tsx
@@ -1,6 +1,12 @@
+import log from 'loglevel';
+import { Reducer } from 'redux';
+import { ActionType, createAction, getType } from 'typesafe-actions';
+
+import { ThunkActionCreator } from '../configureStore';
+
 type LinterMessageBase = {
   column: number | null;
-  file: string;
+  file: string | null;
   line: number | null;
   message: string;
   type: 'notice' | 'error' | 'warning';
@@ -54,7 +60,7 @@ export const createInternalMessage = (
   };
 };
 
-type LinterMessageMap = {
+export type LinterMessageMap = {
   // The 'path' key matches the key of ExternalVersionFile['entries']
   [path: string]: {
     global: LinterMessage[];
@@ -62,10 +68,24 @@ type LinterMessageMap = {
   };
 };
 
-export const getMessageMap = (result: ExternalLinterResult) => {
+export const getMessageMap = (
+  result: ExternalLinterResult,
+  { _log = log }: { _log?: typeof log } = {},
+) => {
   const msgMap: LinterMessageMap = {};
 
   result.validation.messages.forEach((message) => {
+    if (!message.file) {
+      // In reality, this will probably never happen but it's
+      // possible since errors like this do exist. Let's log an
+      // error to Sentry to alert us about it.
+      _log.error(
+        `Unexpectedly received a message not mapped to a file: ${
+          message.message
+        }`,
+      );
+      return;
+    }
     if (!msgMap[message.file]) {
       msgMap[message.file] = { global: [], byLine: {} };
     }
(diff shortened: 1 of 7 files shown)


src/reducers/linter.tsx
```diff
@@ -85,3 +105,108 @@ export const getMessageMap = (result: ExternalLinterResult) => {
 
   return msgMap;
 };
+
+export type LinterState = {
+  forVersionId: undefined | number;
```
willdurand: In TS, we should use `void` instead of `undefined`.

src/reducers/linter.tsx
```diff
@@ -85,3 +105,108 @@ export const getMessageMap = (result: ExternalLinterResult) => {
...
+    dispatch(actions.beginFetchLinterResult({ versionId }));
+
+    // This is a special URL and returns a non-standard JSON response.
+    const response = await fetch(url);
+
+    try {
+      if (!response.ok) {
+        throw new Error(`Got status ${response.status} for URL ${url}`);
+      }
+      const data = await response.json();
```
willdurand: You could probably name this variable `result`.

src/test-helpers.tsx
```diff
@@ -112,6 +112,22 @@ export const fakeUser: ExternalUser = Object.freeze({
...
+  for_appversions: {
+    '{ec8030f7-c20a-464f-9b0e-13a3a9e97384}': ['4.0b1'],
+  },
+  id: [],
+  line: 226,
+  message: 'on* attribute being set using setAttribute',
+  tier: 3,
+  type: 'notice',
+  uid: '9a07163bb74e476c96a2bd467a2bbe52',
+}) as ExternalLinterMessage;
```
willdurand: Could you type the variable instead?

```
export const fakeExternalLinterMessage: ExternalLinterMessage = ...
```

TS might complain that `type` has the wrong value (because `'"notice"' !== string`) but maybe not.

src/reducers/linter.tsx
```diff
@@ -85,3 +105,108 @@ export const getMessageMap = (result: ExternalLinterResult) => {
...
+        ...state,
+        forVersionId: action.payload.versionId,
+        messageMap: undefined,
+        isLoading: true,
+      };
+    case getType(actions.abortFetchLinterResult):
+      return {
+        ...state,
+        forVersionId: action.payload.versionId,
+        // TODO: when we have proper error handling, this can
```
willdurand: Is there an issue for that? (or a comment in an existing issue)

src/reducers/linter.tsx
```diff
@@ -85,3 +105,108 @@ export const getMessageMap = (result: ExternalLinterResult) => {
...
+export type LinterState = {
+  forVersionId: undefined | number;
+  isLoading: boolean;
+  messageMap: undefined | LinterMessageMap;
+};
+
+export const initialState: LinterState = {
+  forVersionId: undefined,
+  isLoading: false,
+  messageMap: undefined,
```
willdurand: Why is this reducer storing data for at most one version?

src/reducers/linter.tsx
```diff
@@ -85,3 +105,108 @@ export const getMessageMap = (result: ExternalLinterResult) => {
...
+};
+
+export const actions = {
+  abortFetchLinterResult: createAction(
+    'ABORT_FETCH_LINTER_RESULT',
+    (resolve) => {
+      return (payload: { versionId: number }) => resolve(payload);
```
willdurand: We do have a `VersionId` type in `src/reducers/versions.tsx`.

kumar303: I can switch to that but what's the point? What benefit does it add? A `number` is a `number`. If we need to change the type then we can just change it. The compiler will tell us if we forget to change one of the types in a call that uses `versionId`.

src/reducers/linter.tsx
kumar303: I couldn't think of any realistic situation where we'll need to load multiple versions on the same page. Limiting the reducer to one version keeps the structure simple and easy to work with. The selector will return undefined for version mismatches which will trigger a refresh of the data.

willdurand: Like everything else: we load the data only once. If the user goes back and forth on different files, we don't call the API all the time and it loads faster.

willdurand: > we don't call the API all the time

I discovered (~2h ago) that this got lost somewhere... on the browse page. We re-fetch/load the data all the time, but because data are stored already, it feels very fast and I suppose that's why no one noticed it before. We just do an extra API call every time /o\

src/reducers/linter.tsx
willdurand: I suggest we remove `VersionId` entirely then.

kumar303: I just don't see how it's beneficial. Maybe I'm missing something? If not then, yeah, we should remove it (in a future issue).

willdurand: I don't think you're missing anything 😃 Your first reply made me realize that I don't think it is very useful either and I believe it causes more issues than anything else.

src/reducers/linter.tsx
kumar303: Re-fetching data isn't a bad thing because the browser will serve a cached API response and that is very fast. But, sure, I guess I could change it to preserve all fetched versions to accommodate quick-switching between versions.

src/reducers/linter.tsx
kumar303: OK, I filed it for a rainy day, very low priority: https://github.com/mozilla/addons-code-manager/issues/377

src/reducers/linter.tsx
kumar303: > maybe we could start with your single-version state and see how it goes in real life?

As discussed on IRC ^ we're going to try it as is. It will probably be fast enough because of how the browser will cache the API response.

willdurand: Given that this works already, caching other linter results could be done as a follow-up (if we needed it).

src/test-helpers.tsx
kumar303: > TS might complain that type has the wrong value (because `'"notice"' !== string`)

It does, yeah. That's why I used the type assertion. I considered adding a comment but didn't think it was so useful since TS will complain if you try to change it anyway.

src/reducers/linter.tsx
kumar303: I can't really think of a helpful issue to file because it all depends on how we implement an error handler. I only added the comment here because it's kind of wrong to load an empty map. It means that the component couldn't re-try an error. But, anyway, retrying an error is already complicated since it wouldn't be an error directly caused by a user action.

kumar303:
Thanks for the review. There are a few unresolved comments where I replied with more information. I'm going to merge it as is but I can follow up if needed. I just need to merge it so I can put up another PR that depends on it.

Issue #376, opened 2019-03-11
Show global linter messages in file viewer


kumar303:
As a step towards showing all linter messages in the file view (https://github.com/mozilla/addons-code-manager/issues/41), show just the global messages at the top. Global messages are those that do not relate to a specific line number.

STR:
* Browse a version that has a global linter message, such as http://localhost:3000/en-US/browse/491343/versions/1527716/
* Select the file which has the message, such as `chrome/content/jquery-1.6.4.min.js`

Issue #368, opened 2019-03-08
Do not execute thunks in tests


kumar303:
### Describe the problem and steps to reproduce it:

* Set up a test for a component that dispatches a thunk on mount. For example, [`Compare`](https://github.com/mozilla/addons-code-manager/blob/da388493553b5a460889689d4d99683ca1c0a01f/src/pages/Compare/index.tsx) dispatches `_fetchVersion` on mount which executes async loading code.

### What happened?

By default (i.e. without stubbing `store.dispatch`), the thunk will be dispatched and executed

### What did you expect to happen?

The thunk should not be executed.

### Anything else we should know?

Since we use [createFakeThunk()](https://github.com/mozilla/addons-code-manager/blob/da388493553b5a460889689d4d99683ca1c0a01f/src/test-helpers.tsx#L305), we never need to execute a thunk in a component test. We *infer* its execution by asserting how / when it gets dispatched.

Executing thunks in tests might create confusing side effects, especially when we add proper error handling (https://github.com/mozilla/addons-code-manager/issues/182).

Maybe we can create a special test store with the thunk middleware removed? Or stub it by default somehow?

bobsilverberg:
This would be good to address. I'd prefer to keep this code quality issue open.

PR #378, opened 2019-03-11
Show global linter messages in file viewer

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/376

Screenshot 2019-03-11 11 02 21
Screenshot 2019-03-11 11 01 30

### DIFF
--- a/src/pages/Browse/index.spec.tsx
+++ b/src/pages/Browse/index.spec.tsx
@@ -191,7 +191,6 @@ describe(__filename, () => {
 
     const store = configureStore();
     store.dispatch(versionActions.loadVersionInfo({ version }));
-    spyOn(store, 'dispatch');
 
     const root = render({ store, versionId: String(version.id) });
 
--- a/src/pages/Browse/index.spec.tsx
+++ b/src/pages/Browse/index.spec.tsx
@@ -14,6 +14,7 @@ import {
 import configureStore from '../../configureStore';
 import {
   ExternalLinterMessage,
+  ExternalLinterResult,
   actions as linterActions,
   createInternalMessage,
 } from '../../reducers/linter';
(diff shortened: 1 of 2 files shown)


src/pages/Browse/index.spec.tsx
```diff
@@ -84,6 +196,7 @@ describe(__filename, () => {
 
     const store = configureStore();
     store.dispatch(versionActions.loadVersionInfo({ version }));
+    spyOn(store, 'dispatch');
```
bobsilverberg: This is being done to prevent thunks from actually being dispatched?

src/pages/Browse/index.spec.tsx
```diff
@@ -181,4 +353,57 @@ describe(__filename, () => {
       path,
     });
   });
+
+  it('renders a global LinterMessage', () => {
+    const file = 'lib/react.js';
+    const externalMessage = globalExternalMessage({
+      file,
+      message: 'This is a known third party library',
```
bobsilverberg: This doesn't seem strictly needed, as we're never asserting anything about it. Could this not just be the default message that comes with `fakeExternalLinterMessage`?

src/pages/Browse/index.spec.tsx
```diff
@@ -181,4 +353,57 @@ describe(__filename, () => {
...
+      ],
+    });
+
+    const message = root.find(LinterMessage);
+    expect(message).toHaveLength(2);
+    expect(message.at(0).prop('message')).toMatchObject({ uid: firstUid });
+    expect(message.at(1).prop('message')).toMatchObject({ uid: secondUid });
+  });
+
+  it('ignores global messages for the wrong version', () => {
```
bobsilverberg: Would it be worth also having a test where the version matches but we load external messages for `pathA`, and then render with `pathB`, and we make sure the messages are not rendered?

src/pages/Browse/index.spec.tsx
```diff
@@ -49,27 +70,118 @@ describe(__filename, () => {
...
     _fetchVersion,
     _fetchVersionFile,
     _log,
     addonId = '999',
     versionId = '123',
-    store = configureStore(),
+    store,
   }: RenderParams = {}) => {
+    let safeStore = store;
+    if (!safeStore) {
```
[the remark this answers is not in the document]
kumar303: True, these changes are a bit confusing. I can remove them. I'm not really sure why I added them for this patch. I think I was getting confused by the console errors during development so I just wanted to get rid of them. I'll save it all for https://github.com/mozilla/addons-code-manager/issues/368

src/pages/Browse/index.spec.tsx
kumar303: Yes, good idea. That's important logic to cover.

kumar303:
Thanks for the review

PR #384, opened 2019-03-12
Support empty list of diffs in `DiffView`

willdurand:
Supports #111 
~~⚠️ depends on #375~~

---

The API returns an empty list of differences sometimes. We should support this case in `DiffView`. This patch has been extracted from #374. There is an empty header for now so that it is similar to when there are differences. In the future, we might want to add the filename (path) to this header, in both cases.

## Screenshot

Screen Shot 2019-03-12 at 11 17 44

In context:

Screen Shot 2019-03-12 at 11 27 43

### DIFF
--- a/src/components/DiffView/index.tsx
+++ b/src/components/DiffView/index.tsx
@@ -10,9 +10,10 @@ import {
   tokenize,
 } from 'react-diff-view';
 import { withRouter, RouteComponentProps } from 'react-router-dom';
+import makeClassName from 'classnames';
 
 import refractor from '../../refractor';
-import { getLanguageFromMimeType } from '../../utils';
+import { getLanguageFromMimeType, gettext } from '../../utils';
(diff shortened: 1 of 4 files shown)


src/components/DiffView/index.tsx
```diff
@@ -114,6 +115,15 @@ export class DiffViewBase extends React.Component<Props> {
 
     return (
       <div className={styles.DiffView}>
+        {diffs.length === 0 && (
+          <React.Fragment>
+            <div className={styles.header} />
+            <div className={makeClassName(styles.diff, styles.noDiffs)}>
+              {gettext('no differences')}
```
bobsilverberg: Nit: Should we capitalize the "N", as in "No differences"?

willdurand: Yes, thanks!

Issue #363, opened 2019-03-08
Port Compare page to the useEffect hook


kumar303:
The `Compare` page grew some gnarly load/update logic in https://github.com/mozilla/addons-code-manager/pull/342

This component can be ported to [useEffect](https://reactjs.org/docs/hooks-reference.html#useeffect) which will simplify the loading logic and let us delete a bunch of the tests.

willdurand:
This could be done together with #111.

willdurand:
I guess this is blocked by upstream for now because Enzyme (shallow) does not support Hooks: https://github.com/airbnb/enzyme/issues/2011.

kumar303:
If we only want `useEffect`, we might only need to wait on https://github.com/facebook/react/issues/15275

Issue #286, opened 2019-02-26
Link to a URL for viewing a single file (browse page)


kumar303:
### Describe the problem and steps to reproduce it:

- Go to a version such as http://localhost:3000/en-US/browse/502955/versions/1541798/
- Click on a file, such as `background.js` to review it
- Share the link with another reviewer

### What happened?

The URL does not point to the `background.js` file so you can't share a direct link

### What did you expect to happen?

The URL should point directly to `background.js` so you can share it.

This is a follow-up to https://github.com/mozilla/addons-code-manager/issues/108

willdurand:
> The URL should point directly to `manifest.json` so you can share it.

I don't think this is really relevant here because this is the default file, so sharing the default url should always load the default file. I'll update the issue description to reflect that.

willdurand:
I also changed the title of this issue to target the browse page only. I'll file a similar issue for the compare page.

AlexandraMoga:
Verified fixed on -dev with FF66, WIn10x64

When selecting a specific file in the file tree (by clicking on it or by keyboard navigation) the `path` to the file location is added to the URL => link to a specific file can be shared 

Examples:

1. https://code.addons-dev.allizom.org/en-US/browse/8542/versions/1655535/?path=content-script%2Fviews%2Fbasic-auth%2Fbasic-auth.html
2. https://code.addons-dev.allizom.org/en-US/browse/616653/versions/1655542/?path=options.js

![2019-04-16_13-36-07](https://user-images.githubusercontent.com/31961530/56203586-3ab27400-604e-11e9-94b8-d0cb11ecf71f.gif)

PR #385, opened 2019-03-12
Render diffs in Compare page

### LINKED ISSUE #292: `DiffView` component should show the whole source code
The `DiffView` only shows hunks but the existing tool always shows the whole source code. We should do that too, **by default**.

### LINKED ISSUE #362: Fetch the diff of a file when it is clicked in the file tree
Supports #111

willdurand:
Fixes #292 
Fixes #362 
~~⚠️ depends on #374~~

---

This patch adds the logic for rendering diffs between two versions of an add-on. It handles errors and loading state.

### DIFF
--- a/src/pages/Compare/index.spec.tsx
+++ b/src/pages/Compare/index.spec.tsx
@@ -5,30 +5,49 @@ import { History } from 'history';
 import {
   createFakeHistory,
   createFakeThunk,
-  fakeVersion,
+  fakeVersionWithDiff,
   shallowUntilTarget,
   spyOn,
 } from '../../test-helpers';
 import configureStore from '../../configureStore';
 import {
-  actions as versionActions,
+  actions as versionsActions,
+  createInternalDiffs,
   createInternalVersion,
 } from '../../reducers/versions';
 import FileTree from '../../components/FileTree';
 import DiffView from '../../components/DiffView';
 import Loading from '../../components/Loading';
+import VersionChooser from '../../components/VersionChooser';
+import styles from './styles.module.scss';
 
 import Compare, { CompareBase, PublicProps } from '.';
 
 describe(__filename, () => {
+  type GetRouteParamsParams = {
+    addonId?: number;
+    baseVersionId?: number;
+    headVersionId?: number;
+    lang?: string;
+  };
+
+  const getRouteParams = ({
+    addonId = 9999,
+    baseVersionId = 1,
+    headVersionId = 1000,
+    lang = 'fr',
+  }: GetRouteParamsParams = {}) => {
+    return {
+      addonId: String(addonId),
+      baseVersionId: String(baseVersionId),
+      headVersionId: String(headVersionId),
+      lang,
+    };
+  };
+
   const createFakeRouteComponentProps = ({
     history = createFakeHistory(),
-    params = {
-      addonId: '999',
-      baseVersionId: '1',
-      headVersionId: '2',
-      lang: 'fr',
-    },
+    params = getRouteParams(),
   } = {}) => {
     return {
       history,
@@ -43,7 +62,7 @@ describe(__filename, () => {
   };
 
   type RenderParams = {
-    _fetchVersion?: PublicProps['_fetchVersion'];
+    _fetchDiff?: PublicProps['_fetchDiff'];
     addonId?: string;
     baseVersionId?: string;
     headVersionId?: string;
@@ -53,7 +72,7 @@ describe(__filename, () => {
   };
 
   const render = ({
-    _fetchVersion,
+    _fetchDiff,
     addonId = '999',
     baseVersionId = '1',
     headVersionId = '2',
@@ -66,7 +85,7 @@ describe(__filename, () => {
         history,
         params: { lang, addonId, baseVersionId, headVersionId },
       }),
-      _fetchVersion,
+      _fetchDiff,
     };
 
     return shallowUntilTarget(<Compare {...props} />, CompareBase, {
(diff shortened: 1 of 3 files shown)


willdurand:
This is also ready :)

willdurand:
rebased!

src/pages/Compare/index.tsx
```diff
@@ -71,34 +77,57 @@ export class CompareBase extends React.Component<Props> {
...
 
-  onSelectFile = () => {};
+  onSelectFile = (path: string) => {
+    const { _fetchDiff, dispatch, match } = this.props;
+    const { addonId, baseVersionId, headVersionId } = match.params;
 
-  render() {
-    const { addonId, version } = this.props;
+    dispatch(
+      _fetchDiff({
```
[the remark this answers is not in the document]
willdurand: Yep, it is much better. Thanks a bunch!

src/pages/Compare/index.tsx
```diff
@@ -62,43 +70,66 @@ export class CompareBase extends React.Component<Props> {
   }
 
   loadData(prevProps?: Props) {
-    const { match } = this.props;
+    const { match, path } = this.props;
     const { addonId, baseVersionId, headVersionId } = match.params;
 
     if (
       !prevProps ||
+      path !== prevProps.path ||
```
kumar303: In your current patch at http://localhost:3000/en-US/compare/502955/versions/1541786...1541798/, the network request to fetch the diff happens twice which isn't great. The first fetch happens without a path and the second one happens with a path. 

I think, also, the fetch should not be happening while one is in progress. Below is a patch that only does the fetch whe `isLoading` is `false`. This stops the double requesting from happening but it might not be the right thing to do. In any case, it would mean that `null` can't be used to signify an error state.

```diff
diff --git a/src/pages/Compare/index.tsx b/src/pages/Compare/index.tsx
index 3b8ae14..704e092 100644
--- a/src/pages/Compare/index.tsx
+++ b/src/pages/Compare/index.tsx
@@ -70,15 +70,16 @@ export class CompareBase extends React.Component<Props> {
   }
 
   loadData(prevProps?: Props) {
-    const { match, path } = this.props;
+    const { isLoading, match, path } = this.props;
     const { addonId, baseVersionId, headVersionId } = match.params;
 
     if (
-      !prevProps ||
-      path !== prevProps.path ||
-      addonId !== prevProps.match.params.addonId ||
-      baseVersionId !== prevProps.match.params.baseVersionId ||
-      headVersionId !== prevProps.match.params.headVersionId
+      !isLoading &&
+      (!prevProps ||
+        path !== prevProps.path ||
+        addonId !== prevProps.match.params.addonId ||
+        baseVersionId !== prevProps.match.params.baseVersionId ||
+        headVersionId !== prevProps.match.params.headVersionId)
     ) {
       const { dispatch, _fetchDiff } = this.props;
 
diff --git a/src/reducers/versions.tsx b/src/reducers/versions.tsx
index 6d3ea48..adeea9e 100644
--- a/src/reducers/versions.tsx
+++ b/src/reducers/versions.tsx
@@ -233,8 +233,8 @@ export type VersionsState = {
   };
   compareInfo:
     | CompareInfo // data successfully loaded
-    | null // an error has occured
-    | undefined; // data not fetched yet
+    | null // data fetching has not begun or an error occurred
+    | undefined; // data fetching is in progress
   versionInfo: {
     [versionId: number]: Version;
   };
@@ -247,7 +247,7 @@ export type VersionsState = {
 
 export const initialState: VersionsState = {
   byAddonId: {},
-  compareInfo: undefined,
+  compareInfo: null,
   versionFiles: {},
   versionInfo: {},
 };
```

willdurand: @kumar303 what about this:

```diff
diff --git a/src/pages/Compare/index.tsx b/src/pages/Compare/index.tsx
index 3b8ae14..0e33d7a 100644
--- a/src/pages/Compare/index.tsx
+++ b/src/pages/Compare/index.tsx
@@ -75,7 +75,7 @@ export class CompareBase extends React.Component<Props> {
 
     if (
       !prevProps ||
-      path !== prevProps.path ||
+      (prevProps.path && path !== prevProps.path) ||
       addonId !== prevProps.match.params.addonId ||
       baseVersionId !== prevProps.match.params.baseVersionId ||
       headVersionId !== prevProps.match.params.headVersionId
```

This fixes the issue on the first load.

willdurand: I pushed a new commit with a test case.

src/pages/Compare/index.tsx
```diff
@@ -62,43 +70,66 @@ export class CompareBase extends React.Component<Props> {
   }
 
   loadData(prevProps?: Props) {
-    const { match } = this.props;
+    const { match, path } = this.props;
     const { addonId, baseVersionId, headVersionId } = match.params;
 
     if (
       !prevProps ||
+      (prevProps.path && path !== prevProps.path) ||
```
kumar303: I guess this is fine but relying on previous props still feels a little fragile to me.

I wanted to present an example of relying on Redux state instead of previous props but you don't necessarily need to make this change. I want to show this because I don't think it's any more or less complex than what you already have. In other words, relying on previous props may not be simpler than just relying on Redux state.

This patch mostly works but it doesn't handle error states since there's no loading flag in the state. It will probably infinitely fetch data on error but that could be fixed by storing a loading flag in the state.

```diff
diff --git a/src/pages/Compare/index.tsx b/src/pages/Compare/index.tsx
index 0e33d7a..c3e8944 100644
--- a/src/pages/Compare/index.tsx
+++ b/src/pages/Compare/index.tsx
@@ -70,23 +70,27 @@ export class CompareBase extends React.Component<Props> {
   }
 
   loadData(prevProps?: Props) {
-    const { match, path } = this.props;
-    const { addonId, baseVersionId, headVersionId } = match.params;
+    const { compareInfo, match, path } = this.props;
+
+    const addonId = parseInt(match.params.addonId, 10);
+    const baseVersionId = parseInt(match.params.baseVersionId, 10);
+    const headVersionId = parseInt(match.params.headVersionId, 10);
 
     if (
-      !prevProps ||
-      (prevProps.path && path !== prevProps.path) ||
-      addonId !== prevProps.match.params.addonId ||
-      baseVersionId !== prevProps.match.params.baseVersionId ||
-      headVersionId !== prevProps.match.params.headVersionId
+      compareInfo === null ||
+      (compareInfo &&
+        (compareInfo.path !== path ||
+          compareInfo.addonId !== addonId ||
+          compareInfo.baseVersionId !== baseVersionId ||
+          compareInfo.headVersionId !== headVersionId))
     ) {
       const { dispatch, _fetchDiff } = this.props;
 
       dispatch(
         _fetchDiff({
-          addonId: parseInt(addonId, 10),
-          baseVersionId: parseInt(baseVersionId, 10),
-          headVersionId: parseInt(headVersionId, 10),
+          addonId,
+          baseVersionId,
+          headVersionId,
           path: path || undefined,
         }),
       );
diff --git a/src/reducers/versions.tsx b/src/reducers/versions.tsx
index fcf8e64..a776909 100644
--- a/src/reducers/versions.tsx
+++ b/src/reducers/versions.tsx
@@ -190,7 +190,11 @@ export type VersionsMap = {
 };
 
 export type CompareInfo = {
+  addonId: number;
   diffs: DiffInfo[];
+  path: string;
+  baseVersionId: number;
+  headVersionId: number;
   mimeType: string;
 };
 
@@ -220,7 +224,7 @@ export const actions = {
       baseVersionId: number;
       headVersionId: number;
       version: ExternalVersionWithDiff;
-      path?: string;
+      path: string;
     }) => resolve(payload);
   }),
 };
@@ -231,7 +235,7 @@ export type VersionsState = {
   };
   compareInfo:
     | CompareInfo // data successfully loaded
-    | null // an error has occured
+    | null // data has not yet loaded or an error occurred
     | undefined; // data not fetched yet
   versionInfo: {
     [versionId: number]: Version;
@@ -245,7 +249,7 @@ export type VersionsState = {
 
 export const initialState: VersionsState = {
   byAddonId: {},
-  compareInfo: undefined,
+  compareInfo: null,
   versionFiles: {},
   versionInfo: {},
 };
@@ -535,7 +539,7 @@ export const fetchDiff = ({
 
     dispatch(actions.beginFetchDiff());
 
-    const response = await _getDiff({
+    const externalVersion = await _getDiff({
       addonId,
       apiState,
       baseVersionId,
@@ -543,18 +547,18 @@ export const fetchDiff = ({
       path,
     });
 
-    if (isErrorResponse(response)) {
-      _log.error(`TODO: handle this error response: ${response.error}`);
+    if (isErrorResponse(externalVersion)) {
+      _log.error(`TODO: handle this error response: ${externalVersion.error}`);
       dispatch(actions.abortFetchDiff());
     } else {
-      dispatch(actions.loadVersionInfo({ version: response }));
+      dispatch(actions.loadVersionInfo({ version: externalVersion }));
       dispatch(
         actions.loadDiff({
           addonId,
           baseVersionId,
           headVersionId,
-          path,
-          version: response,
+          path: path || externalVersion.file.selected_file,
+          version: externalVersion,
         }),
       );
     }
@@ -630,7 +634,13 @@ const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
       };
     }
     case getType(actions.loadDiff): {
-      const { baseVersionId, headVersionId, version } = action.payload;
+      const {
+        addonId,
+        baseVersionId,
+        headVersionId,
+        path,
+        version,
+      } = action.payload;
 
       const { entries, selectedPath } = getVersionInfo(state, headVersionId);
       const entry = entries.find((e) => e.path === selectedPath);
@@ -643,12 +653,16 @@ const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
       return {
         ...state,
         compareInfo: {
+          addonId,
+          baseVersionId,
+          headVersionId,
           diffs: createInternalDiffs({
             baseVersionId,
             headVersionId,
             version,
           }),
           mimeType: entry.mimeType,
+          path,
         },
       };
     }
```

src/pages/Compare/index.spec.tsx
```diff
@@ -76,266 +95,349 @@ describe(__filename, () => {
...
+
+    expect(dispatch).toHaveBeenCalledWith(
+      versionsActions.updateSelectedPath({
+        selectedPath: path,
+        versionId: headVersionId,
+      }),
+    );
+  });
+
+  it('does not dispatch fetchDiff() twice for the default file', () => {
```
kumar303: This test is good but I don't see coverage of the path equality check. All tests pass when I make this change but at least one test should fail:

```diff
diff --git a/src/pages/Compare/index.tsx b/src/pages/Compare/index.tsx
index 0e33d7a..d519776 100644
--- a/src/pages/Compare/index.tsx
+++ b/src/pages/Compare/index.tsx
@@ -75,7 +75,7 @@ export class CompareBase extends React.Component<Props> {
 
     if (
       !prevProps ||
-      (prevProps.path && path !== prevProps.path) ||
+      (prevProps.path) ||
       addonId !== prevProps.match.params.addonId ||
       baseVersionId !== prevProps.match.params.baseVersionId ||
       headVersionId !== prevProps.match.params.headVersionId
```

src/pages/Compare/index.tsx
```diff
@@ -62,43 +70,66 @@ export class CompareBase extends React.Component<Props> {
...
       addonId !== prevProps.match.params.addonId ||
```
kumar303: This doesn't seem like it ever got test coverage:

```diff
diff --git a/src/pages/Compare/index.tsx b/src/pages/Compare/index.tsx
index 0e33d7a..e9926a1 100644
--- a/src/pages/Compare/index.tsx
+++ b/src/pages/Compare/index.tsx
@@ -76,7 +76,6 @@ export class CompareBase extends React.Component<Props> {
     if (
       !prevProps ||
       (prevProps.path && path !== prevProps.path) ||
-      addonId !== prevProps.match.params.addonId ||
       baseVersionId !== prevProps.match.params.baseVersionId ||
       headVersionId !== prevProps.match.params.headVersionId
     ) {
```

All tests pass

willdurand: ah oops, there is a test case for it but one parameter was missing :/

src/pages/Compare/index.tsx
willdurand: > I guess this is fine but relying on previous props still feels a little fragile to me.

I do agree. I believe this could be fixed by #363, which I am willing to take later this quarter.

kumar303: `useEffect` will help cut down on the tests but you'll still need to make comparisons to something to decide when to fetch data. In `useEffect`, the approach of comparing to previous props may not be possible. 

But, yeah, we can revisit this later. I'm slightly worried that https://github.com/mozilla/addons-code-manager/issues/286 won't work with this implementation but we'll find out soon enough!

PR #389, opened 2019-03-12
Show inline linter messages in code files

kumar303:
This is the last puzzle piece to fix https://github.com/mozilla/addons-code-manager/issues/41

The UX isn't great. It accomplishes the bare minimum of showing an inline linter message. I have some ideas to make it better but I'll defer those to follow-ups.

Examples:

<hr>

Screenshot 2019-03-12 14 15 05

<hr>

Screenshot 2019-03-12 14 13 06

### DIFF
--- a/src/pages/Browse/index.spec.tsx
+++ b/src/pages/Browse/index.spec.tsx
@@ -2,11 +2,13 @@ import * as React from 'react';
 import { Store } from 'redux';
 
 import {
+  createFakeExternalLinterResult,
   createFakeHistory,
   createFakeThunk,
   fakeExternalLinterMessage,
   fakeExternalLinterResult,
   fakeVersion,
+  fakeVersionFile,
   fakeVersionEntry,
   shallowUntilTarget,
   spyOn,
@@ -17,6 +19,7 @@ import {
   ExternalLinterResult,
   actions as linterActions,
   createInternalMessage,
+  getMessageMap,
 } from '../../reducers/linter';
 import {
   ExternalVersionWithContent,
@@ -135,10 +138,9 @@ describe(__filename, () => {
     const store = configureStore();
 
     // Prepare a result with all messages for the selected file.
-    const externalResult = result || {
-      ...fakeExternalLinterResult,
-      validation: {
-        ...fakeExternalLinterResult.validation,
+    const externalResult =
+      result ||
+      createFakeExternalLinterResult({
         messages: externalMessages.map((msg) => {
           if (msg.file !== file) {
             throw new Error(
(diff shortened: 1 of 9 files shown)


stories/CodeView.stories.tsx
```diff
@@ -34,30 +38,129 @@ const CSS = `html, body {
...
+          },
+          {
+            title: 'text/css',
+            sectionFn: () => render({ mimeType: 'text/css', content: CSS }),
+          },
+        ],
+      },
+    ],
+  })
+  .addWithChapters('linter messages', {
```
willdurand: That's super cool! Why did you split the stories into two chapters though? I thought we wanted to have everything on a single page so that we see when UI things break? (this will be important for mvp-polish). I am fine with two chapters but we should know when to use 2 chapters.

src/pages/Browse/index.tsx
```diff
@@ -113,11 +113,14 @@ export class BrowseBase extends React.Component<Props> {
...
     if (linterMessages && version) {
-      globalMessages =
-        linterMessages[version.selectedPath] &&
-        linterMessages[version.selectedPath].global;
+      messageMap = linterMessages[version.selectedPath];
+    }
+
+    let globalMessages;
+    if (messageMap) {
+      globalMessages = messageMap.global;
```
willdurand: Why not using `messageMap && messageMap.global` below? (since you do that for the `CodeView` prop)

src/pages/Browse/index.spec.tsx
```diff
@@ -149,16 +151,23 @@ describe(__filename, () => {
...
       ...fakeVersion,
-      entries: {
-        [file]: fakeVersionEntry,
+      file: {
+        ...fakeVersionFile,
+        entries: {
+          [file]: {
+            ...fakeVersionEntry,
+            filename: file,
+            path: file,
```
willdurand: This should be different than `filename`: `filename` should be `test.js` and `path` should be `src/test.js`. I don't know if it is relevant here, but we should not assume both are equal.

Instead of using `file` here, it should be `path` (which is used for the key in `entries` as well as for the `path` field. I would either leave `filename` unchanged or use a `does-not-matter.js` filename.

stories/CodeView.stories.tsx
kumar303: The initial reason for why I wanted stories on one page was to empower this kind of development flow:

1. Make a change to the code
2. Let storybook auto-reload
3. View a single storybook page to check if you broke anything

This doesn't really fit with the `CodeView` stories because the mime-type stories take up so much of the screen. You already have to scroll to see the stories so adding more to the bottom doesn't seem useful. It would be easier to click a new chapter link.

So, yeah, I think a rule of thumb would be to ask, do these screens all roughly fit on the page in full screen mode? If the answer is no then it makes sense to break them into new chapters.

willdurand: That works for me, thanks!

src/pages/Browse/index.spec.tsx
kumar303: I was confused about that, thanks. I also renamed that variable to `path` which is what it is.

Issue #48, opened 2019-01-18
Implement connected-react-router


bobsilverberg:
Depends on #6 
Depends on #24 

Once we have redux in the app, we'll likely want to upgrade the vanilla [react-router](https://reacttraining.com/react-router) to [connected-react-router](https://github.com/supasate/connected-react-router).

bobsilverberg:
Version 6 (the current version) of `connected-react-router` requires `React` 16.4 and `React Redux` 6.0 or later. We're okay for the `React` dependency but not the `Redux` dependency. As I understand it there's an issue with `Enzyme` that is preventing us from using `Redux` 6.

So, we can either implement version 5 of `connected-react-router` now, and plan to upgrade later, or we can wait until we are able to use `Redux` 6 and then implement version 6 of `connected-react-router`.

@kumar303 / @willdurand what do you think we should do?

kumar303:
Hmm, tough call. I believe that https://github.com/airbnb/enzyme/pull/1960 will fix the issue in enzyme and that looks close to landing. Maybe we should wait a bit?

bobsilverberg:
I agree that it's worth waiting a bit. I guess if we get to the point where we actually need it, and we still can't upgrade Redux, then we can implement version 5 for now.

willdurand:
I'll use [v5](https://github.com/supasate/connected-react-router/releases/tag/v5.0.1) because we cannot upgrade Redux to v6 or v7.

PR #391, opened 2019-03-13
VersionChooser can now select versions

### LINKED ISSUE #111: View diff between two versions

### LINKED ISSUE #321: Fetch a diff of two versions when a version is selected in the VersionChooser
In order to ship #111, we want to call the API when a version is selected in the `VersionChooser` component. This action will also update the URL.

willdurand:
Fixes #321 
~~⚠️ depends on #385~~

I believe this also fixes #111 🏆

---

This PR adds the logic to change versions in `VersionChooser`. It also ensures version order is correct (old < new). I did it in the `VersionChooser` component because it would add too much logic into the `Compare` component, but I can move everything in one place (I would understand that we want to keep the same logic in a unique file).

Only the last commit is relevant.

### DIFF
--- a/src/components/VersionChooser/index.tsx
+++ b/src/components/VersionChooser/index.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { Col, Form, Row } from 'react-bootstrap';
 import { connect } from 'react-redux';
+import { withRouter, RouteComponentProps } from 'react-router-dom';
 
 import Loading from '../Loading';
 import { ApplicationState, ConnectedReduxProps } from '../../configureStore';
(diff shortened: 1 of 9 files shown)


willdurand:
This is working but the storybook has an issue and I don't really know how to fix it yet: the component receives router props with empty `match.params`. I tried to pass the params directly but it does not work. It's the first time we have such an issue because it's the first time we render a component that depends on the routing layer in Storybook. 😿

willdurand:
rebased!

stories/VersionChooser.stories.tsx
```diff
@@ -8,7 +8,14 @@ import { fakeVersionsList } from '../src/test-helpers';
 import { renderWithStoreAndRouter } from './utils';
 
 const render = ({ addonId = 124, store = configureStore() } = {}) => {
-  return renderWithStoreAndRouter(<VersionChooser addonId={addonId} />, store);
+  const match = {
```
kumar303: Could we just dispatch a real `push` action similar to how we [did it on the frontend](https://github.com/mozilla/addons-frontend/blob/0c5b2ecb3440303bcece1919305b07cf82ba391f/tests/unit/amo/sagas/test_collections.js#L617-L620)?

I don't know if we're using a Redux-aware router yet, though.

src/components/VersionChooser/index.tsx
```diff
@@ -73,13 +113,16 @@ export class VersionChooserBase extends React.Component<Props> {
...
+  const { addonId } = ownProps.match.params;
 
   return {
-    versionsMap,
+    versionsMap: byAddonId[parseInt(addonId, 10)],
   };
 };
 
-export default connect(mapStateToProps)(VersionChooserBase);
+export default withRouter<Partial<DefaultProps> & RouterProps>(
```
kumar303: I would feel better if this started with `PublicProps` (which would be empty) just because we will most definitely forget to add `PublicProps` to this type in case we need to add some later.

src/components/VersionSelect/index.tsx
```diff
@@ -18,11 +20,18 @@ class VersionSelectBase extends React.Component<PublicProps> {
     withLeftArrow: false,
   };
 
+  onChange = (event: React.FormEvent) => {
+    const { value } = event.currentTarget as HTMLSelectElement;
```
kumar303: Huh. We used to do it in Flow like this (which doesn't work):

```js
  onChange = (event: React.FormEvent<HTMLSelectElement>) => {
    const { value } = event.currentTarget;

    this.props.onChange(value);
  };
```

stories/VersionChooser.stories.tsx
willdurand: Ah, that could be an idea but we don't have a redux-aware router and I don't know if that'll be easy (see #48).

src/components/VersionChooser/index.tsx
willdurand: Oh right!

src/components/VersionSelect/index.tsx
willdurand: Yeah, looks like react-bootstrap types are.. not easy to work with.

stories/VersionChooser.stories.tsx
willdurand: I fixed this issue by using a `Route`.

willdurand:
This is ready for a review :)

src/components/VersionSelect/index.tsx
willdurand: I found a slightly better type.

willdurand:
I know it looks broken but it actually works :D Circle-CI isn't configured yet.

willdurand:
Rebased!

src/components/VersionSelect/index.tsx
```diff
@@ -1,15 +1,20 @@
 import * as React from 'react';
 import { Col, Form } from 'react-bootstrap';
+// eslint-disable-next-line import/no-unresolved
```
kumar303: Why is this needed?

stories/VersionChooser.stories.tsx
```diff
@@ -8,7 +9,15 @@ import { fakeVersionsList } from '../src/test-helpers';
 import { renderWithStoreAndRouter } from './utils';
 
 const render = ({ addonId = 124, store = configureStore() } = {}) => {
-  return renderWithStoreAndRouter(<VersionChooser addonId={addonId} />, store);
+  return renderWithStoreAndRouter(
+    // We wrap the `VersionChooser` component into a `Route` because it reads
+    // data from the URL (via `match.params`).
+    <Route path="/:addonId" component={VersionChooser} />,
```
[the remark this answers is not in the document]
willdurand: Yep, that works for me.

src/components/VersionChooser/index.tsx
```diff
@@ -9,32 +10,85 @@ import { VersionsMap, fetchVersionsList } from '../../reducers/versions';
...
+  }) => {
+    const { history, match } = this.props;
+    const { addonId, lang } = match.params;
+
+    const oldVersionId = parseInt(baseVersionId, 10);
+    const newVersionId = parseInt(headVersionId, 10);
+
+    // We make sure old version is older than the new version when user changes
+    // any of the two versions.
+    if (oldVersionId > newVersionId) {
```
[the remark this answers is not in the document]
willdurand: Ok, I'll revert that and push a follow-up :)

src/components/VersionSelect/index.tsx
willdurand: `FormControlProps` is required by TS and only available from `react-bootstrap/lib/FormControl`, but ESLint cannot find it that way (it does not resolve the `import` correctly apparently). I could not find a better solution and did not find any useful issue about that on the ESLint plugin repo :(

src/components/VersionChooser/index.tsx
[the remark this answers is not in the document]
willdurand: I filed: https://github.com/mozilla/addons-code-manager/pull/391

PR #392, opened 2019-03-14
Fix changes content

willdurand:
This is a fix for https://github.com/mozilla/addons-server/issues/10932 so that we can have correct data in the frontend (until the API gets fixed).

## Screenshots

The diff is on the `version` of the `manifest.json`.

Before:

Screen Shot 2019-03-14 at 15 28 05

After:

Screen Shot 2019-03-14 at 15 27 54

### DIFF
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -892,7 +892,10 @@ describe(__filename, () => {
       firstHunk.changes.forEach((change, index) => {
         const externalChange = firstExternalHunk.changes[index];
 
-        expect(change).toHaveProperty('content', externalChange.content);
+        expect(change).toHaveProperty(
+          'content',
+          externalChange.content.replace(/\n$/, ''),
+        );
         expect(change).toHaveProperty('type', externalChange.type);
         expect(change).toHaveProperty(
           'oldLineNumber',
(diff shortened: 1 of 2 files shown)


src/reducers/versions.spec.tsx
```diff
@@ -892,7 +892,10 @@ describe(__filename, () => {
       firstHunk.changes.forEach((change, index) => {
         const externalChange = firstExternalHunk.changes[index];
 
-        expect(change).toHaveProperty('content', externalChange.content);
+        expect(change).toHaveProperty(
+          'content',
+          externalChange.content.replace(/\n$/, ''),
```
kumar303: Do we need to worry about Windows line endings?

src/reducers/versions.tsx
```diff
@@ -488,7 +488,9 @@ export const createInternalDiffs = ({
         oldRevision: String(baseVersionId),
         hunks: diff.hunks.map((hunk: ExternalHunk) => ({
           changes: hunk.changes.map((change: ExternalChange) => ({
-            content: change.content,
+            // TODO: remove the call to `replace()` once the API is fixed.
+            // See: https://github.com/mozilla/addons-server/issues/10932
+            content: change.content.replace(/\n$/, ''),
```
kumar303: For paranoia, it might be helpful if you added a test to make sure that new lines in the middle of content do not get removed.

src/reducers/versions.spec.tsx
willdurand: I am not sure. The parser used by `react-diff-view` only splits on `\n` (DOS is `\r\n` anyway, no?)

src/reducers/versions.tsx
willdurand: :+1:

PR #393, opened 2019-03-14
Add CSP and other security headers

### LINKED ISSUE #316: Configure a content security policy (CSP)
Before deploying live, we need to configure a content security policy (CSP). This will probably look similar to that of https://github.com/mozilla/addons-frontend/

Some notes:
- The frontend talks to https://addons-dev.allizom.org/api/ which might be on a *different* domain, depending on https://github.com/mozilla/addons-code-manager/issues/25
- The frontend is configured to [send errors to Sentry](https://github.com/mozilla/addons-code-manager/blob/9d4aa9d9d757ea4021f898b3db9ecff6f1c4a26a/src/configureApplication.tsx#L32-L36)
- The app will probably use the same CDN as `addons-frontend`
- Development is, of course, complicated by webpack, hot reloading, and our [api proxy thing](https://github.com/mozilla/addons-code-manager/blob/9d4aa9d9d757ea4021f898b3db9ecff6f1c4a26a/src/server/index.tsx#L116-L120). We could turn off CSP for development if that gets tricky.

muffinresearch:
Fixes #316

* Adds a CSP for statics and document requests along with other typical security headers.
* We'll need to configure sentry later when the endpoints are configured for deploys.
* Updates will be required once we know the CDN endpoints for statics.
* The report URI url is configured to be a no-op response locally. We'll need a ticket to have ops capture these above the app server when it's deployed.


package.json
```diff
@@ -89,7 +92,7 @@
     "type-coverage": "^1.8.0"
   },
   "scripts": {
-    "build": "react-scripts build",
+    "build": "INLINE_RUNTIME_CHUNK=false react-scripts build",
```
muffinresearch: This var is needed to disable a chunk of js that CRA injects by default. From their [docs](https://facebook.github.io/create-react-app/docs/advanced-configuration):

> By default, Create React App will embed the runtime script into index.html during the production build. When set to false, the script will not be embedded and will be imported as usual. This is normally required when dealing with CSP.

src/configureApplication.tsx
```diff
@@ -1,8 +1,12 @@
 import log from 'loglevel';
 import Raven from 'raven-js';
-import { library } from '@fortawesome/fontawesome-svg-core';
+import { config, library } from '@fortawesome/fontawesome-svg-core';
 import { fas } from '@fortawesome/free-solid-svg-icons';
 import { far } from '@fortawesome/free-regular-svg-icons';
+import '@fortawesome/fontawesome-svg-core/styles.css';
+
+// Turn off inline CSS injection for CSP.
+config.autoAddCss = false;
```
kumar303: To keep the logic of this module consistent, this should move inside the body of `configureApplication()`.

src/server/index.tsx
```diff
@@ -48,9 +49,73 @@ export const createServer = ({
...
+    mediaSrc: ["'none'"],
+    objectSrc: ["'none'"],
+    scriptSrc: ["'self'"],
+    styleSrc: ["'self'"],
+    workerSrc: ["'none'"],
+    reportUri: '/__cspreport__',
+  };
+
+  // This CSP is the one above with overrides for local development.
+  const localCSP = {
```
kumar303: Could this definition move to the exact place it's needed? which I think is here: https://github.com/mozilla/addons-code-manager/blob/9ca65f81b8daadf29dd6f643ac8418ac32ac8413/src/server/index.tsx#L203

Since it's only used in that place, I think it would help keep things concise.

src/configureApplication.tsx
muffinresearch: The docs say this needs to be called before any other fontawesome APIs. If this is moved into a function its it guaranteed to be run early enough both now and in the future?

package.json
willdurand: Can we move this variable into `.env` instead? (with a comment ideally)

src/@types/content-security-policy-parser/index.d.ts
```diff
@@ -0,0 +1,8 @@
+declare module 'content-security-policy-parser' {
+  function fn(
+    policy: string,
+  ): {
+    [directive: string]: string;
+  };
+  export = fn;
```
willdurand: We should probably have a comment in this file to explain that the module exposes a default anonymous function.

src/configureApplication.tsx
willdurand: > If this is moved into a function is it guaranteed to be run early enough both now and in the future?

The `configureApplication()` function already contains some FontAwesome code to initialize the lib, so it'd say yes.

yarn.lock
```diff
@@ -4469,6 +4481,16 @@ content-disposition@0.5.2:
   resolved "https://registry.yarnpkg.com/content-disposition/-/content-disposition-0.5.2.tgz#0cf68bb9ddf5f2be7961c3a85178cb85dba78cb4"
   integrity sha1-DPaLud318r55YcOoUXjLhdunjLQ=
 
+content-security-policy-builder@2.0.0:
+  version "2.0.0"
+  resolved "https://registry.yarnpkg.com/content-security-policy-builder/-/content-security-policy-builder-2.0.0.tgz#8749a1d542fcbe82237281ea9f716ce68b394dd2"
+  integrity sha512-j+Nhmj1yfZAikJLImCvPJFE29x/UuBi+/MWqggGGc515JKaZrjuei2RhULJmy0MsstW3E3htl002bwmBNMKr7w==
+
+content-security-policy-parser@0.1.1:
```
willdurand: There should be a `^` here. Could you push an updated `yarn.lock` file?

src/server/index.tsx
```diff
@@ -48,9 +49,66 @@ export const createServer = ({
...
+    defaultSrc: ["'none'"],
+    baseUri: ["'none'"],
+    formAction: ["'none'"],
+    objectSrc: ["'none'"],
+    reportUri: '/__cspreport__',
+  };
+
+  // This config sets the non-static CSP for deployed instances.
+  // TODO: A separate CDN should be configured for statics. When that happens
+  // the 'self' should be removed and replaced with the CDN host + path for statics.
```
willdurand: Could you file an issue for that please?

src/server/index.tsx
```diff
@@ -48,9 +49,66 @@ export const createServer = ({
...
+  const prodCSP = {
+    defaultSrc: ["'none'"],
+    childSrc: ["'none'"],
+    connectSrc: [
+      // Relax the connect-src if using the proxy otherwise
+      // Use the env var or 'none' if the API host isn't set.
+      env.REACT_APP_USE_INSECURE_PROXY === 'true'
+        ? "'self'"
+        : env.REACT_APP_API_HOST || "'none'",
```
willdurand: It looks like the `'none'` value is not covered by a test case.

src/server/index.spec.tsx
```diff
@@ -88,6 +89,79 @@ describe(__filename, () => {
...
+
+        const response = await server.get('/');
+
+        expect(response.status).toEqual(200);
+        expect(response.header).toHaveProperty('content-security-policy');
+
+        const policy = cspParser(response.header['content-security-policy']);
+        expect(policy['default-src']).toEqual(["'none'"]);
+        expect(policy['connect-src']).toEqual([
+          'https://code-manager.addons.cdn.mozilla.net',
```
willdurand: It should probably be `env.REACT_APP_API_HOST` here (`env` needs to be extracted out).

src/server/index.tsx
muffinresearch: Yep I'll file any follow-up issues after this lands.

package.json
muffinresearch: Ah good idea 👍

Issue #379, opened 2019-03-11
Render loading states with skeleton text, not spinners


kumar303:
We currently use spinners to render loading states. These are problematic because the page flickers when it loads quickly and looks jumpy when it loads less quickly.

Let's switch to skeleton text for loading states. This approach would render a full layout as much as possible, using placeholders for any text that has not yet loaded.

* It will address the flicker problem since the layout will be close to final. Transitioning to a fully loaded layout should trigger only a minimal re-rendering.
* It should mostly address the jumpiness problem because each component will be fully laid out. There will still be some jumpiness for things like file content when the final content only includes one or two lines of code (it will jump from many lines to a few).
* It will provide an illusion of speed. When you click on something, it will render immediately before the data has loaded.
* We used this technique on `addons-frontend` and it worked well

Some considerations:
* It's important to render a skeleton screen as close to the final screen as possible. This means everything: line-height, margins, headers. Any deviation creates jumpiness.
* On `addons-frontend` we used an [inline-block](https://github.com/mozilla/addons-frontend/blob/2860acacb5d662ff359f76cda86d68eadf470c44/src/ui/components/LoadingText/styles.scss#L27-L28) with a background color as the skeleton text. Because a div is not text, this sometimes caused a major difference in layout.
* The width of skeleton text needs to be realistic for what the final text will look like.

kumar303:
I played around with this but it needs work. The idea is to save an imprint of the current file and use that as a skeleton when transitioning to the next file. For the very first file load, it has to generate a random skeleton.
 
The spacing of the imprint lines doesn't match the code view exactly so it would be nice to fix that. 


<details>
<summary>Screencast (with simulated slowness)</summary>

![2019-12-03 22 20 38](https://user-images.githubusercontent.com/55398/70112550-4e314000-161b-11ea-8c43-227082cbdeb8.gif)


</details>

This was an older patch but I just got it working again on top of cb35fdfbca9619236e2f988178c17826750dce36 :

<details>
<summary>Patch</summary>

```diff
diff --git a/src/components/CodeLineShapes/index.tsx b/src/components/CodeLineShapes/index.tsx
index 7f0d1c6..933fa9a 100644
--- a/src/components/CodeLineShapes/index.tsx
+++ b/src/components/CodeLineShapes/index.tsx
@@ -1,23 +1,33 @@
 import * as React from 'react';
+import makeClassName from 'classnames';
 
+import { AnyReactNode } from '../../typeUtils';
[...]

PR #398, opened 2019-03-15
Add FileMetadata in browse page

### LINKED ISSUE #326: Implement footer for file viewer
This is an issue that was discussed in our review of the current reviewer tools and I neglected to file.

The file viewer should contain a footer which should include at least the file hash and a ~~download link~~ (this will be added in https://github.com/mozilla/addons-code-manager/issues/353) for the file.

We should add a screenshot of the current footer to this issue.

willdurand:
Fixes #326

---

Screen Shot 2019-03-15 at 11 27 53

### DIFF
--- a/src/components/FileMetadata/index.tsx
+++ b/src/components/FileMetadata/index.tsx
@@ -11,6 +11,8 @@ type PublicProps = {
 const FileMetadataBase = ({ file }: PublicProps) => {
   return (
     <div className={styles.FileMetadata}>
+      <h4>{gettext('Information about this file')}</h4>
+
       <dl>
         <dt>{gettext('Version')}</dt>
         <dd className={styles.version}>{file.version}</dd>
--- a/src/components/FileMetadata/styles.module.scss
+++ b/src/components/FileMetadata/styles.module.scss
@@ -1,0 +1,7 @@
+@import '../../scss/variables';
+
+.FileMetadata {
+  border-radius: $border-radius;
+  border: 1px solid $light-gray;
+  padding: $default-padding;
+}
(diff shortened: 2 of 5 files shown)


willdurand:
Rebased!

kumar303:
Hmm. The screenshot you put in the description suggests that the hash overflow is fixed but I still see it after building your branch just now. Did something regress?

Screenshot 2019-03-19 11 18 54

willdurand:
@kumar303 mmh 🤔  I just rebased and `yarn && yarn dev`:

Screen Shot 2019-03-19 at 16 10 34

I pushed the rebased version of this patch, just in case.

kumar303:
I develop with Firefox release (currently 66) and `break-word` is not supported there. From [this thread](https://stackoverflow.com/questions/47413677/alternative-for-word-break-break-word) it looks like you can use `break-all`. 

I guess `break-word` was added in an older patch?

willdurand:
> I develop with Firefox release (currently 66) and `break-word` is not supported there. From [this thread](https://stackoverflow.com/questions/47413677/alternative-for-word-break-break-word) it looks like you can use `break-all`.

Oh, ok.
 
> I guess `break-word` was added in an older patch?

Isn't it part of Bootstrap? I don't see any occurrence of `break-word` in the code.

kumar303:
Ah, probably. We can probably override it for just this component. Feel free to file it as a separate issue since it's not related to your patch.

willdurand:
> Ah, probably. We can probably override it for just this component. Feel free to file it as a separate issue since it's not related to your patch.

I filed https://github.com/mozilla/addons-code-manager/issues/424

willdurand:
> This looks good to ship. I dislike the UI positioning of it, though. When there's a big list of files, it will get pushed all the way to the bottom. One alternative would be to make it an absolute positioned panel that always floats in the lower left corner. I haven't fully thought that through but it's just something that came to mind. We can revisit this later.

Once we have loading indicators (#379), I'd like to move this component above the file tree.

PR #400, opened 2019-03-15
Mark a file node as selected

### LINKED ISSUE #399: Mark a (file) node as selected in the FileTree
Let's mark a file as "selected" in the `FileTree` component when it is clicked (or loaded by default). This is needed for #329 because we will need access to the store and we don't want to have styling issues.

willdurand:
Fixes #399

---

### DIFF
--- a/src/components/FileTree/index.spec.tsx
+++ b/src/components/FileTree/index.spec.tsx
@@ -1,7 +1,7 @@
 /* eslint @typescript-eslint/camelcase: 0 */
 import * as React from 'react';
 import { shallow } from 'enzyme';
-import Treefold, { TreefoldRenderProps } from 'react-treefold';
+import Treefold from 'react-treefold';
 import { ListGroup } from 'react-bootstrap';
 
 import configureStore from '../../configureStore';
@@ -13,8 +13,10 @@ import {
 } from '../../reducers/versions';
 import { fakeVersion, fakeVersionEntry } from '../../test-helpers';
 import { getLocalizedString } from '../../utils';
+import { getTreefoldRenderProps } from '../FileTreeNode/index.spec';
+import FileTreeNode from '../FileTreeNode';
 
-import FileTree, { buildFileTree, TreeNode, DirectoryNode } from '.';
+import FileTree, { buildFileTree, DirectoryNode } from '.';
 
 describe(__filename, () => {
   describe('buildFileTree', () => {
@@ -382,12 +384,28 @@ describe(__filename, () => {
       const root = render({ version, onSelect });
 
       const node = (root.instance() as FileTree).renderNode(
-        // It does not really matter which props are given here.
-        // eslint-disable-next-line @typescript-eslint/no-object-literal-type-assertion
-        {} as TreefoldRenderProps<TreeNode>,
+        getTreefoldRenderProps(),
       );
 
-      expect(node.props.onSelect).toEqual(onSelect);
+      expect(shallow(<div>{node}</div>).find(FileTreeNode)).toHaveProp(
+        'onSelect',
+        onSelect,
+      );
+    });
+
+    it('passes the version prop to FileTreeNode', () => {
+      const version = getVersion({ ...fakeVersion, id: 777 });
+
+      const root = render({ version });
+
+      const node = (root.instance() as FileTree).renderNode(
+        getTreefoldRenderProps(),
+      );
+
+      expect(shallow(<div>{node}</div>).find(FileTreeNode)).toHaveProp(
+        'version',
+        version,
+      );
     });
   });
 });
--- a/src/components/FileTreeNode/index.spec.tsx
+++ b/src/components/FileTreeNode/index.spec.tsx
@@ -2,45 +2,53 @@ import * as React from 'react';
 import { shallow } from 'enzyme';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
+import { createInternalVersion } from '../../reducers/versions';
+import { fakeVersion } from '../../test-helpers';
 import styles from './styles.module.scss';
 
-import FileTreeNode from '.';
+import FileTreeNode, { PublicProps } from '.';
+
+const fakeGetToggleProps = () => ({
+  onClick: jest.fn(),
+  onKeyDown: jest.fn(),
+  role: 'button',
+  tabIndex: 0,
+});
+
+export const getTreefoldRenderProps = ({
+  getToggleProps = fakeGetToggleProps,
+  hasChildNodes = false,
+  id = 'root',
+  isExpanded = false,
+  isFolder = true,
+  name = 'root',
+  renderChildNodes = jest.fn(),
+  onSelect = jest.fn(),
+} = {}) => {
+  return {
+    node: {
+      id,
+      name,
+    },
+    getToggleProps,
+    hasChildNodes,
+    isExpanded,
+    isFolder,
+    level: 0,
+    renderChildNodes,
+    onSelect,
+  };
+};
 
 describe(__filename, () => {
-  const fakeGetToggleProps = () => ({
-    onClick: jest.fn(),
-    onKeyDown: jest.fn(),
-    role: 'button',
-    tabIndex: 0,
-  });
+  type RenderParams = Partial<PublicProps>;
 
-  const getTreefoldRenderProps = ({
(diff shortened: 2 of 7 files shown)


willdurand:
Rebased!

src/components/FileTreeNode/index.tsx
```diff
@@ -73,4 +85,15 @@ class FileTreeNodeBase<TreeNodeType> extends React.Component<PublicProps> {
   }
 }
 
-export default FileTreeNodeBase;
+const mapStateToProps = (state: ApplicationState, ownProps: PublicProps) => {
+  const { versionId } = ownProps;
+
+  const version = getVersionInfo(state.versions, versionId);
+  const selectedPath = version ? version.selectedPath : undefined;
```
kumar303: If `version` is not loaded, `selectedPath` would be undefined and the component cannot repair itself because it does not know how to fetch versions. When I see this pattern, it tells me that the component should not be mapping its props from Redux state. 

I suggest accepting `version` or `selectedPath` as a public prop instead. This would put the burden of configuration on the code that calls `FileTreeNode` but that seems fine to me. It's already configuring it with `versionId` anyway. Switching to that approach will address the problem up above and will make testing easier.

src/components/FileTree/index.spec.tsx
```diff
@@ -389,5 +389,19 @@ describe(__filename, () => {
...
+
+    it('passes the version ID to FileTreeNode', () => {
+      const version = getVersion({ ...fakeVersion, id: 777 });
+
+      const root = render({ version });
+
+      const node = (root.instance() as FileTree).renderNode(
+        // It does not really matter which props are given here.
+        // eslint-disable-next-line @typescript-eslint/no-object-literal-type-assertion
+        {} as TreefoldRenderProps<TreeNode>,
```
kumar303: Why not just get the props using the helper that already exists?

```diff
diff --git a/src/components/FileTree/index.spec.tsx b/src/components/FileTree/index.spec.tsx
index faae880..6ccce01 100644
--- a/src/components/FileTree/index.spec.tsx
+++ b/src/components/FileTree/index.spec.tsx
@@ -13,6 +13,7 @@ import {
 } from '../../reducers/versions';
 import { fakeVersion, fakeVersionEntry } from '../../test-helpers';
 import { getLocalizedString } from '../../utils';
+import { getTreefoldRenderProps } from '../FileTreeNode/index.spec';
 
 import FileTree, { buildFileTree, TreeNode, DirectoryNode } from '.';
 
@@ -396,9 +397,7 @@ describe(__filename, () => {
       const root = render({ version });
 
       const node = (root.instance() as FileTree).renderNode(
-        // It does not really matter which props are given here.
-        // eslint-disable-next-line @typescript-eslint/no-object-literal-type-assertion
-        {} as TreefoldRenderProps<TreeNode>,
+        getTreefoldRenderProps(),
       );
 
       expect(node.props.versionId).toEqual(version.id);
diff --git a/src/components/FileTreeNode/index.spec.tsx b/src/components/FileTreeNode/index.spec.tsx
index ef47651..996697a 100644
--- a/src/components/FileTreeNode/index.spec.tsx
+++ b/src/components/FileTreeNode/index.spec.tsx
@@ -9,39 +9,39 @@ import styles from './styles.module.scss';
 
 import FileTreeNode, { FileTreeNodeBase, PublicProps } from '.';
 
-describe(__filename, () => {
-  const fakeGetToggleProps = () => ({
-    onClick: jest.fn(),
-    onKeyDown: jest.fn(),
-    role: 'button',
-    tabIndex: 0,
-  });
+const fakeGetToggleProps = () => ({
+  onClick: jest.fn(),
+  onKeyDown: jest.fn(),
+  role: 'button',
+  tabIndex: 0,
+});
 
-  const getTreefoldRenderProps = ({
-    getToggleProps = fakeGetToggleProps,
-    hasChildNodes = false,
-    id = 'root',
-    isExpanded = false,
-    isFolder = true,
-    name = 'root',
-    renderChildNodes = jest.fn(),
-    onSelect = jest.fn(),
-  } = {}) => {
-    return {
-      node: {
-        id,
-        name,
-      },
-      getToggleProps,
-      hasChildNodes,
-      isExpanded,
-      isFolder,
-      level: 0,
-      renderChildNodes,
-      onSelect,
-    };
+export const getTreefoldRenderProps = ({
+  getToggleProps = fakeGetToggleProps,
+  hasChildNodes = false,
+  id = 'root',
+  isExpanded = false,
+  isFolder = true,
+  name = 'root',
+  renderChildNodes = jest.fn(),
+  onSelect = jest.fn(),
+} = {}) => {
+  return {
+    node: {
+      id,
+      name,
+    },
+    getToggleProps,
+    hasChildNodes,
+    isExpanded,
+    isFolder,
+    level: 0,
+    renderChildNodes,
+    onSelect,
   };
+};
 
+describe(__filename, () => {
   type RenderParams = Partial<PublicProps> & { store?: Store };
 
   const render = ({
```

src/components/FileTree/index.spec.tsx
```diff
@@ -389,5 +389,19 @@ describe(__filename, () => {
...
+      );
+
+      expect(node.props.versionId).toEqual(version.id);
```
kumar303: Hmm, I've never seen tests work directly with JSX objects like this. Is this a stable API? I would feel better about doing shallow rendering which we know is a reliable API:

```diff
diff --git a/src/components/FileTree/index.spec.tsx b/src/components/FileTree/index.spec.tsx
index faae880..44e9cb4 100644
--- a/src/components/FileTree/index.spec.tsx
+++ b/src/components/FileTree/index.spec.tsx
@@ -13,6 +13,7 @@ import {
 } from '../../reducers/versions';
 import { fakeVersion, fakeVersionEntry } from '../../test-helpers';
 import { getLocalizedString } from '../../utils';
+import FileTreeNode from '../FileTreeNode';
 
 import FileTree, { buildFileTree, TreeNode, DirectoryNode } from '.';
 
@@ -401,7 +402,10 @@ describe(__filename, () => {
         {} as TreefoldRenderProps<TreeNode>,
       );
 
-      expect(node.props.versionId).toEqual(version.id);
+      expect(shallow(<div>{node}</div>).find(FileTreeNode)).toHaveProp(
+        'versionId',
+        version.id,
+      );
     });
   });
 });
```

src/components/FileTreeNode/index.spec.tsx
```diff
@@ -178,4 +196,28 @@ describe(__filename, () => {
...
+      id: version.file.selected_file,
+    });
+
+    const root = render({
+      ...renderProps,
+      store,
+      versionId: version.id,
+    });
+
+    expect(root.find(`.${styles.selected}`)).toHaveLength(1);
```
kumar303: for paranoia, it would be nice to have a test that makes sure this style isn't applied when the file is not selected.

src/components/FileTree/index.spec.tsx
willdurand: ah, because I am stupid I suppose 🤦‍♂️

src/components/FileTree/index.spec.tsx
willdurand: That works for me, although we get a react component in `node`, and `props` is definitely part of the public/stable API.

willdurand:
@kumar303 thanks for your feedback. I rewrote this patch.

src/components/FileTreeNode/index.spec.tsx
```diff
@@ -178,4 +186,34 @@ describe(__filename, () => {
...
+  });
+
+  it('does not mark a file node as selected when it is not selected', () => {
+    const version = createInternalVersion(fakeVersion);
+
+    const renderProps = getTreefoldRenderProps({
+      id: `src/${version.selectedPath}`,
```
kumar303: This is a little confusing. The test is supposed to set `id` to an unrelated path, right? I think it would be less confusing if it had no relation at all.

kumar303: Otherwise, you could add a comment saying that it shouldn't be selected since the parent directory differs.

willdurand: I changed to an `id` that is totally unrelated.

willdurand: I changed to an id that is totally unrelated.

Issue #115, opened 2019-02-01
Mark files / code that can be skipped during review


kumar303:
User story:

> As a code reviewer, I want the file and diff viewer to mark files or code fragments that can be skipped during review, like known library and framework code.

The validator data (https://github.com/mozilla/addons-code-manager/issues/41) will provide info about files that can be skipped during review. The UX for this should not completely hide the files that can be skipped. They should be grayed out or something like that.

wagnerand:
Example of what it looks like in the current file viewer:

<img width="230" alt="screenshot 2019-03-06 at 19 32 52" src="https://user-images.githubusercontent.com/873258/53904637-d8298980-4046-11e9-9a3a-291a2022f54d.png">

wagnerand:
Hovering over the blue icon:

<img width="363" alt="screenshot 2019-03-06 at 19 35 49" src="https://user-images.githubusercontent.com/873258/53904818-41a99800-4047-11e9-90c8-5173aff7631a.png">

willdurand:
We could build this feature on-top of #401:

![Screen Shot 2019-03-19 at 15 36 52](https://user-images.githubusercontent.com/217628/54614568-ede27a00-4a5c-11e9-8d87-bb5c44b37eba.png)

PR #401, opened 2019-03-15
Add icons related to linter messages in FileTreeNode

### LINKED ISSUE #329: Indicate that a file has linter messages in the file tree
If a file has linter messages associated with it, some indication should be given in the file tree, such as coloring or an icon.

Here's an example from the current app, note the orange triangle icon.

screenshot 2019-03-06 at 19 32 52

### LINKED ISSUE #330: Indicate that a folder has files with linter messages in the file tree
If a file has linter messages associated with it, the folder that contains it should have an indication in the file tree, such as coloring or an icon. This would be true of folders that contain folders that contain files with messages, etc. The indication should bubble up all the way to the top of the tree.

Here's an example from the current app, note the orange triangle icon.

screenshot 2019-03-06 at 19 32 52

willdurand:
Fixes #329 
Fixes #330 

~~⚠️ depends on #400~~

---

This patch adds icons/colors to indicate when a file has linter messages. Directories containing such files are also marked with icons/colors.

Additional notes:

- only the most severe linter message type is shown
- HTML titles are used to give more information about the icons
- the "error" type is usually not used because errors block the submission of a version (admins can override this behavior, though)
- the first two screenshots show a yellow color for the warnings but it has been change to orange to be more readable
- I imagine that #115 could also be implemented with an icon (and some more style probably) **UPDATE:** see https://github.com/mozilla/addons-code-manager/pull/418/commits/04cac0a2ab0f23e38540787f7cdfc44a87a82a18

## Screenshot

With warning and notice messages shown. Because warnings are more severe than notices, the `chrome` folder has the "warning" icon:

Screen Shot 2019-03-15 at 13 34 17

With error messages (please see the additional notes above):

Screen Shot 2019-03-19 at 11 47 50

Updated color for the warnings:

Screen Shot 2019-03-19 at 13 42 10

### DIFF
--- a/src/components/FileTreeNode/index.spec.tsx
+++ b/src/components/FileTreeNode/index.spec.tsx
@@ -284,13 +284,14 @@ describe(__filename, () => {
     (type, className, icon) => {
       const message = {
         ...fakeExternalLinterMessage,
-        file: 'manifest.json',
+        file: 'src',
         type: type as ExternalLinterMessage['type'],
       };
 
       const root = renderWithLinterMessage({
         message,
         treefoldRenderProps: {
+          id: message.file,
           isFolder: true,
         },
       });
--- a/src/components/FileTreeNode/index.spec.tsx
+++ b/src/components/FileTreeNode/index.spec.tsx
@@ -3,10 +3,15 @@ import { shallow } from 'enzyme';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 import { createInternalVersion } from '../../reducers/versions';
-import { fakeVersion } from '../../test-helpers';
+import { ExternalLinterMessage, getMessageMap } from '../../reducers/linter';
+import {
+  createFakeExternalLinterResult,
+  fakeExternalLinterMessage,
+  fakeVersion,
+} from '../../test-helpers';
 import styles from './styles.module.scss';
 
-import FileTreeNode, { PublicProps } from '.';
+import FileTreeNode, { PublicProps, findMostSevereTypeForPath } from '.';
 
 const fakeGetToggleProps = () => ({
   onClick: jest.fn(),
@@ -45,15 +50,40 @@ describe(__filename, () => {
 
   const render = ({
     version = createInternalVersion(fakeVersion),
+    linterMessages,
     ...props
   }: RenderParams = {}) => {
     const allProps = {
       version,
+      linterMessages,
       ...getTreefoldRenderProps(),
       ...props,
     };
 
     return shallow(<FileTreeNode {...allProps} />);
+  };
+
+  const _getMessageMap = (
+    messages: Partial<ExternalLinterMessage>[] = [fakeExternalLinterMessage],
+  ) => {
+    return getMessageMap(createFakeExternalLinterResult({ messages }));
+  };
+
+  const renderWithLinterMessage = ({
+    version = createInternalVersion(fakeVersion),
+    message = fakeExternalLinterMessage,
+    treefoldRenderProps = {},
+  }) => {
+    const renderProps = getTreefoldRenderProps({
+      id: version.selectedPath,
+      ...treefoldRenderProps,
+    });
+
+    return render({
+      ...renderProps,
+      linterMessages: _getMessageMap([message]),
+      version,
+    });
   };
 
   it('renders a simple directory node', () => {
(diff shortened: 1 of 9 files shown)


src/components/FileTreeNode/index.spec.tsx
```diff
@@ -178,4 +247,177 @@ describe(__filename, () => {
...
+        },
+      ] as ExternalLinterMessage[];
+      const map = _getMessageMap(messages);
+
+      const type = findMostSevereTypeForPath(map, path);
+
+      expect(type).toEqual(messages[1].type);
+    });
+
+    it('returns the most severe type given a map of messags and a full qualified file path', () => {
```
kumar303: As far as I can tell there isn't a test to cover multiple messages at the same exact path

src/components/FileTreeNode/index.tsx
```diff
@@ -3,31 +3,135 @@ import { TreefoldRenderProps } from 'react-treefold';
...
+      linterMessages,
+      node,
       onSelect,
+      renderChildNodes,
+      selectedPath,
     } = this.props;
 
+    const hasLinterMessages =
+      linterMessages &&
+      Object.keys(linterMessages).some((path) => path.startsWith(node.id));
```
kumar303: This doesn't seem to have test coverage. The tests still pass:

```diff
diff --git a/src/components/FileTreeNode/index.tsx b/src/components/FileTreeNode/index.tsx
index 092f273..8e5d41e 100644
--- a/src/components/FileTreeNode/index.tsx
+++ b/src/components/FileTreeNode/index.tsx
@@ -106,9 +106,7 @@ export class FileTreeNodeBase<TreeNodeType> extends React.Component<Props> {
       selectedPath,
     } = this.props;
 
-    const hasLinterMessages =
-      linterMessages &&
-      Object.keys(linterMessages).some((path) => path.startsWith(node.id));
+    const hasLinterMessages = !!linterMessages;
 
     let nodeIcons = null;
     let linterType = null;
```

src/components/FileTreeNode/index.tsx
```diff
@@ -73,4 +179,19 @@ class FileTreeNodeBase<TreeNodeType> extends React.Component<PublicProps> {
   }
 }
 
-export default FileTreeNodeBase;
+const mapStateToProps = (state: ApplicationState, ownProps: PublicProps) => {
+  const { versionId } = ownProps;
+
+  const version = getVersionInfo(state.versions, versionId);
+  const selectedPath = version ? version.selectedPath : undefined;
+  const linterMessages = version
```
kumar303: This patch introduces the same problem as https://github.com/mozilla/addons-code-manager/pull/400 but for `linterMessages`. The problem is that the component maps data from Redux but it doesn't know how to load that data when it doesn't exist. 

If this component can't load its own data then it has an implicit dependency on loader components such as [`Browse`](https://github.com/mozilla/addons-code-manager/blob/cd51cf5cea56138961e9eec7ff7d9ac193320369/src/pages/Browse/index.tsx#L86) or [`DiffView`](https://github.com/mozilla/addons-code-manager/blob/cd51cf5cea56138961e9eec7ff7d9ac193320369/src/components/DiffView/index.tsx#L104). The dependency is not explicit which makes it a fragile relationship. Let's say I move the loading logic *out* of `Browse` (which I plan to do); I would probably forget to check that this component still works.

In https://github.com/mozilla/addons-code-manager/pull/400 I suggested passing a direct public prop for what you need but if you did that here, you'd have to prop-drill `linterMessages` through `FileTree` which is not great. 

The best solution would be to implement [`<LinterProvider>`](https://github.com/mozilla/addons-code-manager/issues/414) because that's exactly what you need here. 

<details>
<summary>example patch</summary>

```diff
diff --git a/src/components/FileTreeNode/index.tsx b/src/components/FileTreeNode/index.tsx
index 092f273..dd976fc 100644
--- a/src/components/FileTreeNode/index.tsx
+++ b/src/components/FileTreeNode/index.tsx
@@ -92,14 +92,13 @@ const getIconForType = (type: string | null) => {
 };
 
 export class FileTreeNodeBase<TreeNodeType> extends React.Component<Props> {
-  render() {
+  renderWithMessages(linterMessages) {
     const {
       getToggleProps,
       hasChildNodes,
       isExpanded,
       isFolder,
       level,
-      linterMessages,
       node,
       onSelect,
       renderChildNodes,
@@ -177,6 +176,16 @@ export class FileTreeNodeBase<TreeNodeType> extends React.Component<Props> {
       </React.Fragment>
     );
   }
+
+  render() {
+    const { version } = this.props;
+
+    return (
+      <LinterProvider version={version}>
+        {(linterMessages) => this.renderWithMessages(linterMessages)}
+      </LinterProvider>
+    );
+  }
 }
 
 const mapStateToProps = (state: ApplicationState, ownProps: PublicProps) => {
```

</details>

It would introduce two linter message loaders on the same page but that's fine -- the first one to begin rendering will be the one that loads the data. The others will block on the `isLoading` flag.

The quickest solution would be to just use prop drilling and make a TODO comment to remove it once we have `<LinterProvider>`. I'm open to other ideas, though. Let me know if you have questions about why this is a problem.

stories/FileTreeNode.stories.tsx
```diff
@@ -23,48 +28,47 @@ const getProps = ({
...
+      id: nodeId,
       name: nodeName,
     },
     renderChildNodes,
     onSelect: () => {},
   };
 };
 
+const render = ({ versionId = 123, store = configureStore(), ...props }) => {
+  return renderWithStoreAndRouter(
```
kumar303: The `FileTree` stories also need to be wrapped in a store (depending on how you address previous comments). Right now they are broken for me with a red screen. Hmm. The tests should have broken for that; I'll file an issue.

kumar303:
The bottom icon looks a little broken:

Screenshot 2019-03-19 16 26 21

willdurand:
I believe I addressed most of the comments. I also made sure this view can be rendered on the Apple watch too 😁:

Screen Shot 2019-03-20 at 13 18 46
Screen Shot 2019-03-20 at 13 18 40

Note that I did not add new stories yet, I'd like to do that in a follow-up if possible.

**Important:** prop-drilling did not work well for the "Compare" page because, for some reasons, I did not notice that `DiffView` was actually fetching/loading linter messages 😞 This does not look like a great idea and I am not sure connecting `FileTree` would be smart, unless we add `LinterProvider` everywhere, which does not sound like a great idea either.

I'd suggest we add a `linterMessagesByLine` prop in the `DiffView` component (like the `CodeView`, these two component APIs should be similar IMO) and we make `Compare` responsible for fetching/loading the linter messages as well as rendering the global messages (this code should be the same code as in `Browse`, so it means we could have a `GlobalLinterMessages` component). The `LinterProvider` could be used in both `Compare` and `Browse`.

willdurand:
rebased to reduce the diff size :)

src/components/FileTreeNode/styles.module.scss
```diff
@@ -2,9 +2,32 @@
...
   cursor: pointer;
+  display: flex;
+  flex-direction: row;
+  flex-wrap: nowrap;
+  justify-content: space-between;
   padding: 0.5rem 1.25rem;
 }
 
+.nodeName {
+  word-break: break-all;
```
kumar303: These styles don't look so nice on laptop-sized windows. How about this patch? It's a bit of an improvement.

<details>
<summary>patch</summary>

```diff
diff --git a/src/components/FileTreeNode/index.tsx b/src/components/FileTreeNode/index.tsx
index d02a372..7067676 100644
--- a/src/components/FileTreeNode/index.tsx
+++ b/src/components/FileTreeNode/index.tsx
@@ -134,23 +134,17 @@ class FileTreeNodeBase<TreeNodeType> extends React.Component<PublicProps> {
       <React.Fragment>
         <ListGroup.Item {...listGroupItemProps}>
           <span
-            className={styles.nodeName}
-            style={{ paddingLeft: `${level * 20}px` }}
+            className={styles.nodeItem}
+            style={{ paddingLeft: `${level * 12}px` }}
           >
             {isFolder ? (
-              <React.Fragment>
-                <FontAwesomeIcon icon={isExpanded ? 'folder-open' : 'folder'} />
-                &nbsp;{node.name}
-              </React.Fragment>
+              <FontAwesomeIcon icon={isExpanded ? 'folder-open' : 'folder'} />
             ) : (
-              <React.Fragment>
-                <FontAwesomeIcon icon="file" />
-                &nbsp;{node.name}
-              </React.Fragment>
+              <FontAwesomeIcon icon="file" />
             )}
+            <span className={styles.nodeName}>{node.name}</span>
+            {nodeIcons}
           </span>
-
-          {nodeIcons}
         </ListGroup.Item>
 
         {isExpanded &&
diff --git a/src/components/FileTreeNode/styles.module.scss b/src/components/FileTreeNode/styles.module.scss
index 54d97e8..e0dd03b 100644
--- a/src/components/FileTreeNode/styles.module.scss
+++ b/src/components/FileTreeNode/styles.module.scss
@@ -9,8 +9,29 @@
   padding: 0.5rem 1.25rem;
 }
 
+.nodeItem {
+  align-items: flex-start;
+  display: flex;
+  line-height: 1;
+  width: 100%;
+}
+
+.nodeName,
+.nodeIcons {
+  margin-left: 0.5rem;
+  width: 100%;
+}
+
 .nodeName {
-  word-break: break-all;
+  overflow: hidden;
+  padding-bottom: 0.2rem;
+  text-overflow: ellipsis;
+  white-space: wrap;
+}
+
+.nodeIcons {
+  display: flex;
+  justify-content: flex-end;
 }
 
 .selected {
@@ -28,6 +49,3 @@
     color: $orange;
   }
 }
-
-.nodeIcons {
-}
```

</details>

Screenshot 2019-03-20 12 57 59
Screenshot 2019-03-20 12 57 44


I filed https://github.com/mozilla/addons-code-manager/issues/431 to make a storybook out of this.

src/components/FileTreeNode/index.spec.tsx
```diff
@@ -45,17 +50,42 @@ describe(__filename, () => {
...
 
+  const _getMessageMap = (
+    messages: Partial<ExternalLinterMessage>[] = [fakeExternalLinterMessage],
+  ) => {
+    return getMessageMap(createFakeExternalLinterResult({ messages }));
+  };
+
+  const renderWithLinterMessage = ({
+    version = createInternalVersion(fakeVersion),
+    message = fakeExternalLinterMessage,
```
kumar303: This seems like it would be better as an array but I guess if you don't need it then it's fine.

src/components/FileTreeNode/index.spec.tsx
```diff
@@ -216,4 +246,200 @@ describe(__filename, () => {
...
+      message,
+      treefoldRenderProps: {
+        isFolder: false,
+      },
+    });
+
+    expect(root.find(className)).toHaveLength(1);
+
+    const nodeIcons = root.find(`.${styles.nodeIcons}`);
+    expect(nodeIcons).toHaveLength(1);
```
kumar303: You could use `.toHaveClassName()`

src/components/FileTreeNode/index.spec.tsx
```diff
@@ -216,4 +246,200 @@ describe(__filename, () => {
...
+      };
+
+      const root = renderWithLinterMessage({
+        message,
+        treefoldRenderProps: {
+          isFolder: true,
+        },
+      });
+
+      expect(root.find(className)).toHaveLength(1);
```
kumar303: `.toHaveClassName()` would give a better message here

src/components/FileTreeNode/index.spec.tsx
```diff
@@ -216,4 +246,200 @@ describe(__filename, () => {
...
+  it.each([
+    ['error', `.${styles.hasLinterErrors}`, 'times-circle'],
+    ['notice', `.${styles.hasLinterMessages}`, 'info-circle'],
+    ['warning', `.${styles.hasLinterWarnings}`, 'exclamation-triangle'],
+  ])(
+    'renders a directory node that has linter %ss',
+    (type, className, icon) => {
+      const message = {
+        ...fakeExternalLinterMessage,
+        file: 'manifest.json',
```
kumar303: Could you make this more realistic for the test? Was this path supposed to be a directory?

src/components/FileTreeNode/index.spec.tsx
```diff
@@ -216,4 +246,200 @@ describe(__filename, () => {
...
+    };
+
+    const root = renderWithLinterMessage({
+      message,
+      treefoldRenderProps: {
+        id: 'some/other/file.js',
+      },
+    });
+
+    expect(root.find(`.${styles.hasLinterMessages}`)).toHaveLength(0);
```
kumar303: I won't comment on them all but `.toHaveClassName()` would produce better messages in all of the assertions I'm seeing.

src/components/FileTreeNode/index.spec.tsx
```diff
@@ -216,4 +246,200 @@ describe(__filename, () => {
...
+          ...fakeExternalLinterMessage,
+          file: `${path}/file-1.js`,
+          type: 'warning',
+        },
+        {
+          ...fakeExternalLinterMessage,
+          file: `${path}/file-2.js`,
+          type: 'error',
+        },
+      ] as ExternalLinterMessage[];
```
kumar303: This is not a change request but since this has come up a lot, maybe we should switch to string enums for `type`.

src/pages/Compare/index.tsx
```diff
@@ -31,7 +31,7 @@ type PropsFromRouter = {
 
 type PropsFromState = {
   addonId: number;
-  compareInfo: CompareInfo | null | void;
+  compareInfo: CompareInfo | null | undefined;
```
kumar303: I thought `void` was preferred, no? If it's trying to match another type, maybe change that one?

src/components/FileTreeNode/index.spec.tsx
willdurand: it's a bummer that it does not understand this notation...

src/components/FileTreeNode/styles.module.scss
willdurand: oh yes, I love it! Thanks ❤️

src/components/FileTreeNode/index.spec.tsx
willdurand: `toHaveClassName()` and `.find()` are different. One is looking for a class name on the current wrapper, the other is used to make sure a component identified by its class name exits.

src/pages/Compare/index.tsx
willdurand: ahm, this is not needed actually

src/components/FileTreeNode/index.spec.tsx
willdurand: yep, thanks

src/components/FileTreeNode/index.spec.tsx
willdurand: yep, I'll do it in the next PR

Issue #409, opened 2019-03-18
View global linter messages in version compare view


kumar303:
As a step towards https://github.com/mozilla/addons-code-manager/issues/314, show just the global linter messages for the new version of the compare view. These are linter messages that apply to the whole file, not just a specific line.

PR #410, opened 2019-03-18
Show global linter messages in compare view

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/409

**NOTE**: There is some duplication here between how `pages/Browse` and `components/CodeView` handle [linter messages](https://github.com/mozilla/addons-code-manager/blob/ed2f053714c09230b55ff10eb93de98d9240b1e5/src/pages/Browse/index.tsx#L82-L89). To keep this patch small-ish, I'm going to address that in a follow-up issue (https://github.com/mozilla/addons-code-manager/issues/414). The `Browse` component was doing it in an awkward way so the fixup will be slightly more involved.

Screenshot 2019-03-18 09 54 08

### DIFF
--- a/src/components/DiffView/index.tsx
+++ b/src/components/DiffView/index.tsx
@@ -220,7 +220,7 @@ export class DiffViewBase extends React.Component<Props> {
 
 const mapStateToProps = (
   state: ApplicationState,
-  ownProps: Props,
+  ownProps: PublicProps,
 ): PropsFromState => {
   const { version } = ownProps;
 
--- a/src/components/DiffView/index.tsx
+++ b/src/components/DiffView/index.tsx
@@ -67,6 +67,7 @@ export class DiffViewBase extends React.Component<Props> {
 
   constructor(props: Props) {
     super(props);
+    // Allow dependency injection to test all the ways loadData() gets executed.
     this.loadData = props._loadData || this._loadData;
   }
 
--- a/src/components/DiffView/index.tsx
+++ b/src/components/DiffView/index.tsx
@@ -98,7 +98,7 @@ export class DiffViewBase extends React.Component<Props> {
       linterMessagesAreLoading,
     } = this.props;
 
-    if (version && linterMessages === undefined && !linterMessagesAreLoading) {
+    if (linterMessages === undefined && !linterMessagesAreLoading) {
       dispatch(
         _fetchLinterMessages({
           versionId: version.id,
--- a/src/components/DiffView/index.tsx
+++ b/src/components/DiffView/index.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { connect } from 'react-redux';
 import {
   Decoration,
   Diff,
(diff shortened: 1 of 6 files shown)


src/components/DiffView/index.tsx
```diff
@@ -12,35 +13,66 @@ import {
...
   static defaultProps: DefaultProps = {
     _document: document,
+    _fetchLinterMessages: fetchLinterMessages,
     _tokenize: tokenize,
     viewType: 'unified',
   };
 
+  constructor(props: Props) {
+    super(props);
+    this.loadData = props._loadData || this._loadData;
```
willdurand: Is all of this used to "fix" https://github.com/mozilla/addons-code-manager/issues/368? I am not a huge fan of polluting the public API that way. Could we try to stub `dispatch` instead?

stories/DiffView.stories.tsx
```diff
@@ -2,34 +2,144 @@ import React from 'react';
...
-          title: 'diff with additions and deletions',
-          sectionFn: () => {
-            return renderWithStoreAndRouter(
-              <DiffView
-                diffs={parseDiff(diffWithDeletions)}
-                mimeType="application/javascript"
-              />,
-            );
+// eslint-disable-next-line no-var
+var uid = 0;
```
willdurand: `let`?

src/components/DiffView/index.tsx
```diff
@@ -53,6 +85,29 @@ export class DiffViewBase extends React.Component<Props> {
...
+  _loadData = () => {
+    const {
+      _fetchLinterMessages,
+      dispatch,
+      version,
+      linterMessages,
+      linterMessagesAreLoading,
+    } = this.props;
+
+    if (version && linterMessages === undefined && !linterMessagesAreLoading) {
```
willdurand: It seems the `version &&` is not tested.

```diff
diff --git a/src/components/DiffView/index.tsx b/src/components/DiffView/index.tsx
index ff5e4c1..25ee49c 100644
--- a/src/components/DiffView/index.tsx
+++ b/src/components/DiffView/index.tsx
@@ -98,7 +98,7 @@ export class DiffViewBase extends React.Component<Props> {
       linterMessagesAreLoading,
     } = this.props;
 
-    if (version && linterMessages === undefined && !linterMessagesAreLoading) {
+    if (/*version && */linterMessages === undefined && !linterMessagesAreLoading) {
       dispatch(
         _fetchLinterMessages({
           versionId: version.id,
```

kumar303: oh, whoops, it is not possible to render `DiffView` without a version now.

src/components/DiffView/index.tsx
kumar303: My intention was to use dependency injection to make an assertion that `componentDidMount()` and `componentDidUpdate()` will run the exact same loader function. The assertions are made [here](https://github.com/mozilla/addons-code-manager/blob/00b49f9cb33a650f01536aa925b281aac4012a9b/src/components/DiffView/index.spec.tsx#L263-L281). This approach explicitly checks that the same loader is used whereas stubbing `dispatch()` would only implicitly check it. 

An alternative would be to duplicate all load logic tests, the first set using `componentDidMount()` and the second using `componentDidUpdate()`. This seems like an unecessary burden.

It's a bit clunky to do dependency injection like this (although we do it for thunks, too). Once we port this component to `useEffect()`, it can go away.

willdurand: ah, I see. Could you please add a comment above the `loadData` variable declaration?

kumar303:
Since the build is only blocked on Synk, I'm going to merge it (I need to post another patch on top of it).

PR #417, opened 2019-03-19
Show inline linter messages in the compare view

kumar303:
This implements inline linter messages which is the final part to fix https://github.com/mozilla/addons-code-manager/issues/314

In http://localhost:3000/en-US/compare/494431/versions/1532144...1532144/ you can click on `common.js` to see an inline message.

Example from storybook:

Screenshot 2019-03-19 10 05 36

### DIFF
--- a/src/components/DiffView/index.spec.tsx
+++ b/src/components/DiffView/index.spec.tsx
@@ -482,7 +482,7 @@ describe(__filename, () => {
     );
   });
 
-  it('only renders messages per line', () => {
+  it('renders just the right amount of widgets', () => {
     const store = configureStore();
 
     const externalMessages = [
@@ -500,8 +500,10 @@ describe(__filename, () => {
       store,
       version,
     });
-    // Make sure we have truthy React nodes for exactly the same
-    // number of messages.
+
+    // As a sanity check on how the widgets are mapped,
+    // make sure we have truthy React nodes for exactly the same
+    // number of lines containing messages.
     expect(getWidgetNodes(widgets).length).toEqual(externalMessages.length);
   });
 
--- a/src/components/DiffView/index.spec.tsx
+++ b/src/components/DiffView/index.spec.tsx
@@ -1,6 +1,12 @@
 /* eslint-disable @typescript-eslint/camelcase */
 import React from 'react';
-import { Diff, parseDiff, getChangeKey } from 'react-diff-view';
+import {
+  Diff,
+  Hunks,
+  WidgetMap,
+  parseDiff,
+  getChangeKey,
+} from 'react-diff-view';
 import { Store } from 'redux';
 import { shallow } from 'enzyme';
 import { Location } from 'history';
@@ -29,7 +35,12 @@ import {
 } from '../../test-helpers';
 import styles from './styles.module.scss';
 
-import DiffView, { DiffViewBase, DefaultProps, PublicProps } from '.';
+import DiffView, {
+  DiffViewBase,
+  DefaultProps,
+  PublicProps,
+  getAllHunkChanges,
+} from '.';
 
 describe(__filename, () => {
   type RenderParams = { store?: Store; location?: Location } & Partial<
@@ -82,12 +93,14 @@ describe(__filename, () => {
     return { _fetchLinterMessages, dispatch, root, fakeThunk };
   };
 
-  const setUpLinterMessagesForPath = ({
+  const dispatchLinterMessages = ({
     messages,
     path = 'lib/react.js',
+    store,
   }: {
     messages: Partial<ExternalLinterMessage>[];
     path?: string;
+    store: Store;
   }) => {
     const version = createInternalVersion({
       ...fakeVersion,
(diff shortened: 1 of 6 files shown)


src/components/DiffView/styles.module.scss
```diff
@@ -60,3 +60,7 @@
     background-color: $light-gray;
   }
 }
+
+.inlineLinterMessages {
+  padding: 16px 16px 0 16px;
```
willdurand: Bootstrap sets `margin-botom: 1rem` to the alert component, so I guess we should also use `1rem` here:

```suggestion
  padding: 1rem 1rem 0 1rem;
```

src/components/DiffView/index.spec.tsx
```diff
@@ -82,12 +93,14 @@ describe(__filename, () => {
     return { _fetchLinterMessages, dispatch, root, fakeThunk };
   };
 
-  const setUpLinterMessagesForPath = ({
+  const dispatchLinterMessages = ({
```
willdurand: In the past, we used the name of the action prefixed with an underscore (`_loadLinterResult`) for such functions.

kumar303:
This is the issue to fix the `parseDiff()` types: https://github.com/mozilla/addons-code-manager/issues/420

src/components/DiffView/index.spec.tsx
```diff
@@ -97,15 +110,54 @@ describe(__filename, () => {
...
+      throw new Error(`Could not find a change at line ${line}`);
+    }
+
+    const key = getChangeKey(result[0]);
+    return shallow(<div>{widgets[key]}</div>);
+  };
+
+  const getWidgetNodes = (widgets: WidgetMap) => {
+    // Return a list of truthy React nodes in arbitrary order.
+    return Object.values(widgets).filter(Boolean);
```
willdurand: I believe `.filter((widgeet) => widget !== null)` would be slightly more readable. Could we have `widget` assigned to any other value?

src/components/DiffView/index.spec.tsx
```diff
@@ -371,35 +417,153 @@ describe(__filename, () => {
...
+      store,
+      messages: externalMessages,
+    });
+
+    const diffs = parseDiff(diffWithDeletions);
+    const widgets = renderAndGetWidgets({ diffs, store, version });
+
+    const { hunks } = diffs[0];
+
+    const firstRoot = renderWidget(hunks, widgets, externalMessages[0].line);
```
willdurand: shouldn't this be `firstWidget`?

src/components/DiffView/index.spec.tsx
```diff
@@ -371,35 +417,153 @@ describe(__filename, () => {
...
+    expect(firstRoot.find(LinterMessage)).toHaveLength(1);
+    expect(firstRoot.find(LinterMessage)).toHaveProp(
+      'message',
+      expect.objectContaining({
+        uid: externalMessages[0].uid,
+      }),
+    );
+
+    const secondRoot = renderWidget(hunks, widgets, externalMessages[1].line);
```
willdurand: `secondWidget`, maybe?

src/components/DiffView/index.spec.tsx
```diff
@@ -371,35 +417,153 @@ describe(__filename, () => {
...
+    expect(secondRoot.find(LinterMessage)).toHaveLength(1);
+    expect(secondRoot.find(LinterMessage)).toHaveProp(
+      'message',
+      expect.objectContaining({
+        uid: externalMessages[1].uid,
+      }),
+    );
+  });
+
+  it('only renders messages per line', () => {
```
willdurand: I don't really understand this test case description.

src/components/DiffView/index.tsx
```diff
@@ -109,6 +118,33 @@ export class DiffViewBase extends React.Component<Props> {
...
   };
 
+  getWidgets = (hunks: Hunks) => {
+    const { linterMessages } = this.props;
+
+    return getAllHunkChanges(hunks).reduce((widgets, change) => {
+      const changeKey = getChangeKey(change);
+      const line = change.lineNumber;
+
+      let widget = null;
```
willdurand: I'd move this declaration down, closer to where it gets assigned, but that's a personal taste I guess.

src/components/DiffView/index.spec.tsx
kumar303: Since this is a direct wrapper around `loadLinterResult` then, yeah, I guess that would make sense.

src/components/DiffView/index.spec.tsx
kumar303: Heh. I had that at first but it felt a bit fragile since react would also not render `undefined`. I was really just trying to mimic what react would do with a list of `React.ReactNode` types.

willdurand: ah ok, it took me some time to understand what this filter was doing and the comment did not help much. That being said, now that I understand what the filter does, the comment is relevant.

src/components/DiffView/styles.module.scss
kumar303: Thanks

PR #418, opened 2019-03-19
Add icon/title for known JS files

### LINKED ISSUE #115: Mark files / code that can be skipped during review

willdurand:
Fixes #115
~~:warning: depends on #401~~

---

This patch adds style/visual indication for file that are known (js) libraries.

You can also try this addon/version: http://localhost:3000/en-US/browse/502914/versions/1541734/

## Screenshot

Screen Shot 2019-03-25 at 12 06 43

Screen Shot 2019-03-25 at 12 15 50

Screen Shot 2019-03-25 at 12 15 42

Screen Shot 2019-03-25 at 12 06 43

### DIFF
--- a/src/components/FileTreeNode/index.spec.tsx
+++ b/src/components/FileTreeNode/index.spec.tsx
@@ -612,6 +612,22 @@ describe(__filename, () => {
       expect(isKnownLibrary(map, path)).toEqual(false);
     });
 
+    it('returns false when another path is a known library', () => {
+      const path = 'jquery.js';
+      const messages = [
+        {
+          ...fakeExternalLinterMessage,
+          file: path,
+          id: [LINTER_KNOWN_LIBRARY_CODE],
+          line: null,
+          type: 'notice',
+        },
+      ] as ExternalLinterMessage[];
+      const map = _getMessageMap(messages);
+
+      expect(isKnownLibrary(map, 'not-the-same-path')).toEqual(false);
+    });
+
     it('returns true when a path is a known library', () => {
       const path = 'jquery.js';
       const messages = [
@@ -626,6 +642,29 @@ describe(__filename, () => {
       const map = _getMessageMap(messages);
 
       expect(isKnownLibrary(map, path)).toEqual(true);
+    });
+
+    it('throws an error if an extra key is found in the linter message map', () => {
+      const path = 'jquery.js';
+      const messages = [
+        {
+          ...fakeExternalLinterMessage,
+          file: path,
+          id: [LINTER_KNOWN_LIBRARY_CODE],
+          line: null,
+          type: 'notice',
+        },
+      ] as ExternalLinterMessage[];
+      const map = _getMessageMap(messages);
+
+      const unexpectedKey = 'future';
+      // Artifically inject a new key in the message map.
+      // @ts-ignore
+      map[path][unexpectedKey] = {};
+
+      expect(() => {
+        isKnownLibrary(map, path);
+      }).toThrow(new RegExp(`Unexpected key "${unexpectedKey}" found`));
     });
   });
 });
--- a/src/components/FileTreeNode/index.spec.tsx
+++ b/src/components/FileTreeNode/index.spec.tsx
@@ -14,7 +14,12 @@ import {
 } from '../../test-helpers';
 import styles from './styles.module.scss';
 
-import FileTreeNode, { PublicProps, findMostSevereTypeForPath } from '.';
+import FileTreeNode, {
+  LINTER_KNOWN_LIBRARY_CODE,
+  PublicProps,
+  findMostSevereTypeForPath,
+  isKnownLibrary,
+} from '.';
 
 const fakeGetToggleProps = () => ({
   onClick: jest.fn(),
@@ -97,9 +102,9 @@ describe(__filename, () => {
     });
   };
 
-  const renderWithLinterMessage = ({
+  const renderWithLinterMessages = ({
     version = createInternalVersion(fakeVersion),
-    message = fakeExternalLinterMessage,
+    messages = [fakeExternalLinterMessage],
     treefoldRenderProps = {},
   }) => {
     const renderProps = getTreefoldRenderProps({
@@ -108,7 +113,7 @@ describe(__filename, () => {
     });
 
     return renderWithLinterProvider({
-      messageMap: _getMessageMap([message]),
+      messageMap: _getMessageMap(messages),
       version,
       ...renderProps,
     });
@@ -274,9 +279,11 @@ describe(__filename, () => {
       type: type as ExternalLinterMessage['type'],
     };
 
-    const root = renderWithLinterMessage({
-      message,
-      treefoldRenderProps: { isFolder: false },
+    const root = renderWithLinterMessages({
+      messages: [message],
+      treefoldRenderProps: {
+        isFolder: false,
+      },
     });
 
     expect(root.find(className)).toHaveLength(1);
@@ -302,8 +309,8 @@ describe(__filename, () => {
         type: type as ExternalLinterMessage['type'],
       };
 
-      const root = renderWithLinterMessage({
-        message,
+      const root = renderWithLinterMessages({
+        messages: [message],
         treefoldRenderProps: {
           id: message.file,
           isFolder: true,
@@ -327,8 +334,8 @@ describe(__filename, () => {
       file: 'manifest.json',
     };
 
-    const root = renderWithLinterMessage({
-      message,
+    const root = renderWithLinterMessages({
+      messages: [message],
       treefoldRenderProps: {
         id: 'some/other/file.js',
       },
@@ -343,8 +350,8 @@ describe(__filename, () => {
       file: 'src/manifest.json',
     };
 
-    const root = renderWithLinterMessage({
-      message,
+    const root = renderWithLinterMessages({
+      messages: [message],
       treefoldRenderProps: {
         id: 'src/',
         isFolder: true,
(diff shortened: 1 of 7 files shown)


src/components/FileTreeNode/index.tsx
```diff
@@ -86,11 +86,15 @@ const getIconForType = (type: string | null) => {
...
     case 'warning':
       return 'exclamation-triangle';
+    case 'known-library':
+      return 'check-circle';
     default:
       return 'info-circle';
   }
 };
 
+export const LINTER_MESSAGE_KNOWN_LIBRARY = 'JS Library Detected';
```
kumar303: Let's use the code instead. Here's a view of the API response:

Screenshot 2019-03-19 16 28 29

More info: https://github.com/mozilla/addons-linter/blob/master/docs/rules.md

src/components/FileTreeNode/index.tsx
```diff
@@ -114,13 +118,21 @@ export class FileTreeNodeBase<TreeNodeType> extends React.Component<Props> {
...
       linterType = findMostSevereTypeForPath(linterMessages, node.id);
+      let title = getTitleForType(linterType, isFolder);
+
+      if (
+        linterMessages[node.id] &&
+        linterMessages[node.id].global.some(
+          ({ message }) => message === LINTER_MESSAGE_KNOWN_LIBRARY,
+        )
+      ) {
+        linterType = 'known-library';
```
kumar303: In the case that `linterType` is already defined, I don't think it should be overridden. Consider this case: jquery was detected but the version of jquery they used has been banned. That should show up as a red error since it would lead to rejection.

Screenshot 2019-03-19 16 37 08

src/components/FileTreeNode/index.tsx
willdurand: oh, I could not find this constant. I found it in the linter but did not find it in the JSON file. I believe it is not stored in redux yet?

src/components/FileTreeNode/index.tsx
willdurand: Good call, thanks

willdurand:
@kumar303 could you share the addon id/version id related to the validation data and feedback above? looks like I only have old code (that are not even part of addons-linter o_O)

willdurand:
> @kumar303 could you share the addon id/version id related to the validation data and feedback above? looks like I only have old code (that are not even part of addons-linter o_O)

nevermind, finally found it :D

src/components/FileTreeNode/index.tsx
```diff
@@ -101,13 +105,22 @@ class FileTreeNodeBase<TreeNodeType> extends React.Component<PublicProps> {
     let linterType = null;
     if (linterMessages && hasLinterMessages) {
       linterType = findMostSevereTypeForPath(linterMessages, node.id);
+      let title = getTitleForType(linterType, isFolder);
+
+      if (
+        linterType === 'notice' &&
```
willdurand: I believe if the most severe type is `notice`, we can semi-hide the file, no? Or should we make sure that there is only/exactly one message?

src/components/FileTreeNode/index.tsx
```diff
@@ -73,11 +73,15 @@ const getIconForType = (type: string | null) => {
...
     case 'warning':
       return 'exclamation-triangle';
+    case 'known-library':
+      return 'check-circle';
     default:
       return 'info-circle';
   }
 };
 
+export const LINTER_KNOW_LIBRARY_CODE = 'KNOWN_LIBRARY';
```
kumar303: typo: this should be `LINTER_KNOWN_LIBRARY_CODE`

src/components/FileTreeNode/index.spec.tsx
```diff
@@ -340,6 +344,62 @@ describe(__filename, () => {
...
+      },
+    });
+
+    expect(root.find(`.${styles.isKnownLibrary}`)).toHaveLength(1);
+
+    const nodeIcons = root.find(`.${styles.nodeIcons}`);
+    expect(nodeIcons).toHaveLength(1);
+    expect(nodeIcons.find(FontAwesomeIcon)).toHaveProp('icon', 'check-circle');
+    expect(nodeIcons.find(FontAwesomeIcon).prop('title')).toMatch(
+      new RegExp('known library'),
```
kumar303: this would be slightly more concise as `/known library/`

src/components/FileTreeNode/index.tsx
kumar303: I'm not convinced that known library messages will always have the type `notice` (I can't be sure from looking at addons-linter). I'd rather not see our code make that assumption. Can you find another way to allow more severe messages take precedence? To be clear, I do agree that a known library message should override any other message of `type=notice`.

src/reducers/linter.spec.tsx
```diff
@@ -378,6 +379,7 @@ describe(__filename, () => {
       expect(
         createInternalMessage({
           ...fakeExternalLinterMessage,
+          id,
```
kumar303: Could you also fill in some value for `id` in the [helper](https://github.com/mozilla/addons-code-manager/blob/9457fa8069298c76d0f32dbe288dc7f257b2ea1e/src/test-helpers.tsx#L128)?

It looks like pretty much everything will have a code: https://github.com/mozilla/addons-linter/blob/master/src/messages/javascript.js

src/pages/Index/index.tsx
```diff
@@ -10,7 +10,7 @@ export class IndexBase extends React.Component<PublicProps> {
       <Col>
         <p>
           There is nothing you can do here, but try{' '}
-          <Link to="/en-US/browse/502955/versions/1541798/">
+          <Link to="/en-US/browse/494431/versions/1532144/">
```
kumar303: Maybe add links to both?

src/components/FileTreeNode/index.tsx
willdurand: @kumar303 heh. the docs says it is a `notice`: https://github.com/mozilla/addons-linter/blob/5df2b4ecce7d9c9ad6abc0e6c7d1283a5b0e249c/docs/rules.md#javascript

src/pages/Index/index.tsx
willdurand: one is enough I guess, and this one is better because it has more data

src/components/FileTreeNode/index.tsx
```diff
@@ -101,13 +105,22 @@ class FileTreeNodeBase<TreeNodeType> extends React.Component<PublicProps> {
     let linterType = null;
     if (linterMessages && hasLinterMessages) {
       linterType = findMostSevereTypeForPath(linterMessages, node.id);
+      let title = getTitleForType(linterType, isFolder);
+
+      if (
+        linterType === 'notice' &&
+        linterMessages[node.id] &&
+        linterMessages[node.id].global.some(({ code }) =>
```
kumar303: I've considered this more and I think we should *only* mark files as known libraries if there is a *single* message (global or not) that has the matching code. For all other cases, let's highlight the file like we normally would.

The current patch could encourage the reviewer to ignore `FLAGGED_FILE_TYPE` (a `notice`) which might mean the file is binary, if I'm reading these docs right: https://github.com/mozilla/addons-linter/blob/5df2b4ecce7d9c9ad6abc0e6c7d1283a5b0e249c/docs/rules.md#package-layout (I could be wrong)

More importantly, the current patch seems like it would ignore any errors or warnings that are in the `byLine` mapping.

willdurand: > More importantly, the current patch seems like it would ignore any errors or warnings that are in the `byLine` mapping.

I don't think this is correct. the function checks for linter messages everywhere.

> I've considered this more and I think we should _only_ mark files as known libraries if there is a _single_ message (global or not) that has the matching code. For all other cases, let's highlight the file like we normally would.

well, if all messages are notice, it certainly means the messages are not super important. If `FLAGGED_*` is important, it should not be a `notice`. More importantly, if `KNOWN_LIBRARY` is less important than other notices, it should not have a `notice` level.

> Hiding known libraries has severe security implications so I want to be cautious with how we do it.

We are not hiding files.

willdurand: > I've considered this more and I think we should _only_ mark files as known libraries if there is a _single_ message (global or not) that has the matching code. For all other cases, let's highlight the file like we normally would.

I had this question too: https://github.com/mozilla/addons-code-manager/pull/418#discussion_r267805279

@kumar303 so, what should we do here in the end?

kumar303: > should we make sure that there is only/exactly one [known_library] message?

Yes, I think that is the safest way to do it. We should only consider it a known library file if there is exactly one message having the known library code across `global` and `byLine` mappings and any *future* mappings.

> We are not hiding files.

The purpose of dimming / graying the file is to let reviewers safely ignore the file so I'd say it is the equivalent of hiding the file from review.

>> More importantly, the current patch seems like it would ignore any errors or warnings that are in the byLine mapping.

> I don't think this is correct. the function checks for linter messages everywhere.

ok, good. Can you add an explicit test for that?

willdurand: > Yes, I think that is the safest way to do it.

Alrighty, let's do this! 💪

src/components/FileTreeNode/index.spec.tsx
```diff
@@ -443,4 +504,105 @@ describe(__filename, () => {
...
+          ...fakeExternalLinterMessage,
+          file: path,
+          id: [LINTER_KNOWN_LIBRARY_CODE],
+          line: null,
+          type: 'notice',
+        },
+      ] as ExternalLinterMessage[];
+      const map = _getMessageMap(messages);
+
+      expect(isKnownLibrary(map, path)).toEqual(true);
```
kumar303: It would be nice to see an additional test for when a known library code exists but for the wrong path

src/components/FileTreeNode/index.tsx
```diff
@@ -50,6 +50,24 @@ export const findMostSevereTypeForPath = (
...
+export const isKnownLibrary = (
+  linterMessageMap: LinterMessageMap,
+  path: string,
+): boolean => {
+  if (!linterMessageMap[path]) {
+    return false;
+  }
+
+  const m = linterMessageMap[path];
+  if (Object.keys(m.global).length > 1 || Object.keys(m.byLine).length > 0) {
```
kumar303: I would like to see this account for future keys that we might add to `linterMessageMap`. You could do it a couple ways. Ideas:

- Assume that all keys on `linterMessageMap` will be arrays and concat them all together
- Iterate through a list of known keys (`global`, `byLine`) and throw an error if a new key is present

It's a bit paranoid but, as it stands, the known library code would win against messages in a new mapping key since there are no failing tests or type checks to remind us to update this part of the code. For something as sensitive as hiding files, I want to be sure that this will do the right thing.

willdurand: I see, thanks!

willdurand: > Assume that all keys on `linterMessageMap` will be arrays and concat them all together

that won't work because `byLine` is a map too.

willdurand: I went for the error thrown if a key is unexpected.

PR #429, opened 2019-03-20
Add code/id linter message in redux

willdurand:
I went for `code` instead of `id` in the internal message type because
`id` is very close to `uid` but not unique, and `code` is also used in
the addons-linter.


willdurand:
mmh, this is a very small patch actually so it will be part of #418

PR #430, opened 2019-03-20
Expose download URL

### LINKED ISSUE #348: Return a download link to the file from getVersionFile

### LINKED ISSUE #353: Add download link to FileMetadata component
Depends on #345 
Depends on #348
Depends on https://github.com/mozilla/addons-server/issues/10877

Once we are receiving the download link from the APIs, and once we have updated the `getVersionFile` selector to return that information, we should update the `FileMetadata` component to display the download link.

willdurand:
Fixes #353
Fixes #348 

---

This is not working because the API returns a link that requires authentication, but the UI is there. I believe the API will be fixed to return a link that works. If not, we could add some server code to fix the links in this project, but that can be done in a follow-up.

## Screenshot

Screen Shot 2019-03-21 at 16 30 29

### DIFF
--- a/src/components/FileMetadata/index.spec.tsx
+++ b/src/components/FileMetadata/index.spec.tsx
@@ -42,6 +42,9 @@ describe(__filename, () => {
     expect(root.find(`.${styles.version}`)).toHaveText(file.version);
     expect(root.find(`.${styles.sha256}`)).toHaveText(file.sha256);
     expect(root.find(`.${styles.mimeType}`)).toHaveText(file.mimeType);
+    expect(root.find(`.${styles.downloadURL}`).html()).toContain(
+      `<a href="${file.downloadURL}">${file.filename}</a>`,
+    );
   });
 
   it('renders a formatted filesize', () => {
--- a/src/components/FileMetadata/index.spec.tsx
+++ b/src/components/FileMetadata/index.spec.tsx
@@ -41,10 +41,11 @@ describe(__filename, () => {
 
     expect(root.find(`.${styles.version}`)).toHaveText(file.version);
     expect(root.find(`.${styles.sha256}`)).toHaveText(file.sha256);
-    expect(root.find(`.${styles.mimeType}`)).toHaveText(file.mimeType);
-    expect(root.find(`.${styles.downloadURL}`).html()).toContain(
-      `<a href="${file.downloadURL}">${file.filename}</a>`,
-    );
+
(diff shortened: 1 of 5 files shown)


kumar303:
> This is not working because the API returns a link that requires authentication

Is there an issue filed for this? I can't find one.

src/components/FileMetadata/index.spec.tsx
```diff
@@ -42,6 +42,9 @@ describe(__filename, () => {
     expect(root.find(`.${styles.version}`)).toHaveText(file.version);
     expect(root.find(`.${styles.sha256}`)).toHaveText(file.sha256);
     expect(root.find(`.${styles.mimeType}`)).toHaveText(file.mimeType);
+    expect(root.find(`.${styles.downloadURL}`).html()).toContain(
```
kumar303: `html()` does a render so we should use it sparingly. This seems to work fine:

```diff
diff --git a/src/components/FileMetadata/index.spec.tsx b/src/components/FileMetadata/index.spec.tsx
index c8cc14b..35e2e9d 100644
--- a/src/components/FileMetadata/index.spec.tsx
+++ b/src/components/FileMetadata/index.spec.tsx
@@ -42,9 +42,11 @@ describe(__filename, () => {
     expect(root.find(`.${styles.version}`)).toHaveText(file.version);
     expect(root.find(`.${styles.sha256}`)).toHaveText(file.sha256);
     expect(root.find(`.${styles.mimeType}`)).toHaveText(file.mimeType);
-    expect(root.find(`.${styles.downloadURL}`).html()).toContain(
-      `<a href="${file.downloadURL}">${file.filename}</a>`,
-    );
+
+    const downloadLink = root.find(`.${styles.downloadURL}`).find('a');
+    expect(downloadLink).toHaveLength(1);
+    expect(downloadLink).toHaveText(file.filename);
+    expect(downloadLink).toHaveProp('href', file.downloadURL);
   });
 
   it('renders a formatted filesize', () => {
```

willdurand:
> Is there an issue filed for this? I can't find one.

I don't think @EnTeQuAk created one yet, so I can do that.

willdurand:
> > Is there an issue filed for this? I can't find one.
> 
> I don't think @EnTeQuAk created one yet, so I can do that.

I filed: https://github.com/mozilla/addons-server/issues/10994

Issue #383, opened 2019-03-11
MDN links in linter messages are not rendered


kumar303:
### Describe the problem and steps to reproduce it:

View a file that has a message with an anchor link to an MDN URL. Example: http://localhost:3000/en-US/browse/494431/versions/1532144/ (`manifest.json`)

### What happened?

The link is not rendered correctly.

<img width="668" alt="Screenshot 2019-03-11 17 00 18" src="https://user-images.githubusercontent.com/55398/54161571-94f16100-4420-11e9-9c6a-96ef39664d36.png">

### What did you expect to happen?

The link should be rendered correctly.

kumar303:
@EnTeQuAk do all outgoing links appear as complete anchor tags in linter messages, like this screenshot? I saw an older test fixture that showed links rendered as plain URLs without anchor tags.

EnTeQuAk:
> @EnTeQuAk do all outgoing links appear as complete anchor tags in linter messages, like this screenshot? I saw an older test fixture that showed links rendered as plain URLs without anchor tags.

All new links these days should be escaped, yes.

willdurand:
Given that we have a `sanitizeHTML()` function (in `src/utils.tsx`), we could maybe remove HTML in the linter messages completely and then the existing logic would linkify links correctly.

PR #433, opened 2019-03-20
Handle MDN links correctly

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/383

Example:

Screenshot 2019-03-20 14 50 44

### DIFF
--- a/src/components/LinterMessage/index.tsx
+++ b/src/components/LinterMessage/index.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import purify from 'dompurify';
 import he from 'he';
 import { Alert } from 'react-bootstrap';
 
@@ -24,7 +25,6 @@ export const decodeHtmlEntities = (text: string) => {
 };
 
 const renderDescription = (description: LinterMessage['description']) => {
-  const linkPattern = /<a href="([^"]+)">[^<]+<\/a>/g;
   const urlPattern = /^https?:\/\//;
 
   return description.reduce((allLines: React.ReactNode[], line, lineIndex) => {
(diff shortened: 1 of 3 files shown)


src/components/LinterMessage/index.tsx
```diff
@@ -32,20 +33,24 @@ const renderDescription = (description: LinterMessage['description']) => {
...
-            <Alert.Link key={`link:${part}`} href={part}>
-              {part}
-            </Alert.Link>,
-          );
-        } else {
-          allParts.push(decodeHtmlEntities(part));
-        }
+      line
+        // Replace anchor tags with just their URLs.
+        .replace(linkPattern, '$1')
```
willdurand: Why not using DOMPurify (already used in this project) to remove all HTML tags (given that we don't intend to render HTML)? The regexp seems fragile if, for instance, new attributes are added (e.g., `target` and `rel`).

kumar303: It seems unlikely to me that this html would change but I could try to see if dompurify is easy enough to work with. It probably won't give me the `href` but maybe that's OK.

willdurand: oh wait, right. it won't work well if the `href` value is different than the link content.

src/components/LinterMessage/index.tsx
```diff
@@ -24,6 +24,7 @@ export const decodeHtmlEntities = (text: string) => {
 };
 
 const renderDescription = (description: LinterMessage['description']) => {
+  const linkPattern = /<a href="([^"]+)">[^<]+<\/a>/g;
```
willdurand: should we make it case insensitive too?

willdurand: maybe you could try to match a link with attributes, just in case. WDYT?

kumar303:
Good call on dompurify. I think that's a safe route to go. The anchor text will always be a URL (as far as I know) so it should work fine. If a reviewer notices a problem, they will tell us.

src/components/LinterMessage/index.tsx
```diff
@@ -32,20 +33,24 @@ const renderDescription = (description: LinterMessage['description']) => {
...
-            <Alert.Link key={`link:${part}`} href={part}>
-              {part}
-            </Alert.Link>,
-          );
-        } else {
-          allParts.push(decodeHtmlEntities(part));
-        }
+      // Replace anchor tags with just their text, which are the URLs.
+      purify
+        .sanitize(line, { ALLOWED_TAGS: [] })
```
willdurand: 🏆

Issue #441, opened 2019-03-21
Apply LinterProvider to DiffView


kumar303:
As a small step towards https://github.com/mozilla/addons-code-manager/issues/438 , refactor `DiffView` to use the `LinterProvider` component added in https://github.com/mozilla/addons-code-manager/issues/414

PR #443, opened 2019-03-21
Apply LinterProvider to DiffView

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/441

The next patches are going to be a bit big so maybe it's easier if I break them up.

### DIFF
--- a/src/components/DiffView/index.spec.tsx
+++ b/src/components/DiffView/index.spec.tsx
@@ -21,17 +21,16 @@ import {
   ExternalLinterMessage,
   actions as linterActions,
 } from '../../reducers/linter';
-import { Version, createInternalVersion } from '../../reducers/versions';
+import { createInternalVersion } from '../../reducers/versions';
 import {
   createContextWithFakeRouter,
   createFakeExternalLinterResult,
   createFakeLocation,
-  createFakeThunk,
   fakeVersion,
   fakeVersionEntry,
   fakeVersionFile,
   shallowUntilTarget,
-  spyOn,
+  simulateLinterProvider,
 } from '../../test-helpers';
 import styles from './styles.module.scss';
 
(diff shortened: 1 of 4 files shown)


kumar303:
I had an afterthought about the `DiffView` testing strategy. I think it will be simpler without requiring a full store. Stay tuned...

kumar303:
@willdurand it's ready for a look now

willdurand:
>  The diff coverage is `90.47%`.

Do you know why the diff coverage is not 100%?

kumar303:
It seems to be saying that the initialization of parameters to `simulateLinterProvider()` are not covered: https://codecov.io/gh/mozilla/addons-code-manager/pull/443/diff

This is odd because the function is definitely executed but only in tests (maybe it ignores tests?). When I run `yarn test --coverage` locally it says that `test-helpers.tsx` has 100% line coverage.

kumar303:
I'm not worried about this sort of coverage because TypeScript covers it better. I'll see if I can just ignore it with a comment.

src/components/DiffView/index.spec.tsx
```diff
@@ -62,73 +53,32 @@ describe(__filename, () => {
...
+    messagesAreLoading = false,
+    selectedMessageMap = undefined,
+    ...renderParams
+  }: RenderWithLinterProviderParams = {}): ShallowWrapper => {
+    const root = render(renderParams);
+
+    return simulateLinterProvider(root, {
+      messageMap,
+      messagesAreLoading,
+      selectedMessageMap,
```
willdurand: 😍

src/components/DiffView/index.tsx
```diff
@@ -253,30 +195,19 @@ export class DiffViewBase extends React.Component<Props> {
...
+    );
+  }
+}
 
-export default withRouter<PublicProps & Partial<DefaultProps> & RouterProps>(
-  connect(mapStateToProps)(DiffViewBase),
-) as React.ComponentType<PublicProps & Partial<DefaultProps>>;
+export default withRouter(DiffViewBase) as React.ComponentType<
+  PublicProps & Partial<DefaultProps>
+>;
```
willdurand: Is this still needed here? Can't we type the `withRouter` HOC instead?

```
withRouter<PublicProps & Partial<DefaultProps> & RouterProps>(DiffViewBase);
```

I believe if `PublicProps` and `DefaultProps` were the same type, we would not even have to add types to `withRouter` (it works in `CodeView`):

```
withRouter(DiffViewBase);
```

PR #446, opened 2019-03-22
Add a simple error handling

### LINKED ISSUE #182: Handle API errors in App, Browse, and Compare components

willdurand:
Fixes #182

---

The idea is to have a reducer responsible for storing errors. Given that we focus on API errors, the action currently receives an `Error` object and the reducer extracts this message (could be done in a `createInternalError()` function, that's an implementation detail).

Each error gets an ID assigned to it, unique and predictable given that the next ID is stored into the state. Having such IDs and the increment logic into the reducer avoids complicated unique ID generation and multiple errors for the same component. We never decrement the ID to avoid collisions.

Having an ID on each error allows to "hide" them when the user clicks the "close" button. We do not have an automated way of closing such error messages yet. In the future, we could maybe add a `context` prop to an `ApplicationError`, which would group errors from the same component/page and when trying to reload a page, the `fetch` action would send a `clearErrorsForContext()` action that would automatically remove all the errors for a given context.

**Note:** I applied the logic to the `fetchVersion` thunk only. I'll replace the other `TODO` in a follow-up patch to keep this one concise.

### DIFF
--- a/src/components/App/index.spec.tsx
+++ b/src/components/App/index.spec.tsx
@@ -185,22 +185,27 @@ describe(__filename, () => {
   });
 
   it('dispatches removeError() when dismissing an error', () => {
-    const error = new Error('version not found');
+    const error1 = new Error('first error');
+    const error2 = new Error('second error');
     const store = configureStore();
-    store.dispatch(errorsActions.addError({ error }));
+    store.dispatch(errorsActions.addError({ error: error1 }));
+    store.dispatch(errorsActions.addError({ error: error2 }));
     const dispatch = spyOn(store, 'dispatch');
 
     const root = render({ store });
 
-    const onClose = root.find(Alert).prop('onClose') as Function;
-    // User clicks the "close" button (it is a cross).
+    // User clicks the "close" button of the first error shown.
+    const onClose = root
+      .find(Alert)
+      .at(0)
+      .prop('onClose') as Function;
     onClose();
 
     const { errors } = store.getState().errors;
-    const lastError = errors[errors.length - 1];
+    const firstError = errors[0];
     expect(dispatch).toHaveBeenCalledWith(
       errorsActions.removeError({
-        id: lastError.id,
+        id: firstError.id,
       }),
     );
   });
--- a/src/components/App/index.spec.tsx
+++ b/src/components/App/index.spec.tsx
@@ -173,8 +173,8 @@ describe(__filename, () => {
     const error1 = new Error('error 1');
     const error2 = new Error('error 2');
     const store = configureStore();
-    store.dispatch(errorsActions.showError({ error: error1 }));
-    store.dispatch(errorsActions.showError({ error: error2 }));
+    store.dispatch(errorsActions.addError({ error: error1 }));
+    store.dispatch(errorsActions.addError({ error: error2 }));
 
     const root = render({ store });
 
@@ -184,10 +184,10 @@ describe(__filename, () => {
     expect(root.find(Alert).at(1)).toIncludeText(error2.message);
   });
 
-  it('dispatches dismissError() when dismissing an error', () => {
+  it('dispatches removeError() when dismissing an error', () => {
     const error = new Error('version not found');
     const store = configureStore();
-    store.dispatch(errorsActions.showError({ error }));
+    store.dispatch(errorsActions.addError({ error }));
     const dispatch = spyOn(store, 'dispatch');
 
     const root = render({ store });
@@ -199,7 +199,7 @@ describe(__filename, () => {
     const { errors } = store.getState().errors;
     const lastError = errors[errors.length - 1];
     expect(dispatch).toHaveBeenCalledWith(
-      errorsActions.dismissError({
+      errorsActions.removeError({
         id: lastError.id,
       }),
     );
--- a/src/components/App/index.spec.tsx
+++ b/src/components/App/index.spec.tsx
@@ -1,11 +1,12 @@
 import React from 'react';
-import { Container } from 'react-bootstrap';
+import { Alert, Container } from 'react-bootstrap';
 import { Store } from 'redux';
 import { Route } from 'react-router-dom';
 
 import styles from './styles.module.scss';
 import configureStore from '../../configureStore';
 import { actions as apiActions } from '../../reducers/api';
+import { actions as errorsActions } from '../../reducers/errors';
 import { actions as userActions } from '../../reducers/users';
 import {
   createContextWithFakeRouter,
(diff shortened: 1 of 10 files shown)


willdurand:
@kumar303 before I go ahead and write tests and so on, how do you feel about us taking this direction?

willdurand:
Thanks for the quick feedback!

kumar303:
I see that you already read my tweet 😆 but I was going to mention that it helps me a lot when there is a 5 second (or more pause) between actions in screencasts. I should mention this to QA, too. I usually end up slowing down their gifs through https://ezgif.com/speed

src/reducers/errors.tsx
```diff
@@ -0,0 +1,53 @@
...
+};
+
+export const initialState: ErrorsState = {
+  errors: [],
+  // This is a sequence to give each error a unique ID.
+  nextErrorId: 1,
+};
+
+export const actions = {
+  showError: createAction('SHOW_ERROR', (resolve) => {
```
bobsilverberg: I wonder if `showError` is the best name, or if it should be something like `recordError`? I suppose with this implementation `showError` is accurate, as it will cause the error to appear, but if/when we enhance it to only show errors that belong to a particular component, then `showError` may no longer be accurate.

src/reducers/errors.spec.tsx
```diff
@@ -0,0 +1,63 @@
+import reducer, { actions, initialState } from './errors';
+
+describe(__filename, () => {
+  describe('reducer', () => {
+    it('stores an error to show', () => {
```
bobsilverberg: Further to my naming comment below, the test says "stores an error", but the action is called `showError`.

src/reducers/errors.spec.tsx
```diff
@@ -0,0 +1,63 @@
...
+      state = reducer(
+        state,
+        actions.showError({ error: new Error('Bad Request, again') }),
+      );
+      expect(state).toMatchObject({
+        nextErrorId: initialState.nextErrorId + 2,
+      });
+    });
+
+    it('removes an error when dismissed', () => {
```
bobsilverberg: Would it be worth having a test where there are multiple errors in state and we remove one that isn't the last one?

src/reducers/errors.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { ActionType, createAction, getType } from 'typesafe-actions';
+
+export type ApplicationError = {
+  id: number;
+  message: string;
+};
+
+export type ErrorsState = {
+  errors: ApplicationError[];
+  nextErrorId: number;
```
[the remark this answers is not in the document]
willdurand: no, I don't think it is possible.

src/reducers/errors.tsx
willdurand: yep, I am open to anything else. Could also be `add`/`remove`, I don't know. I like the high level naming but maybe we should have something more generic.

src/reducers/errors.spec.tsx
willdurand: yep, that's a good idea

src/reducers/errors.spec.tsx
willdurand: hehe yes... so add/remove?

src/components/App/index.spec.tsx
```diff
@@ -161,4 +162,46 @@ describe(__filename, () => {
...
+
+    const root = render({ store });
+
+    expect(root.find(`.${styles.errors}`)).toHaveLength(1);
+    expect(root.find(Alert)).toHaveLength(2);
+    expect(root.find(Alert).at(0)).toIncludeText(error1.message);
+    expect(root.find(Alert).at(1)).toIncludeText(error2.message);
+  });
+
+  it('dispatches dismissError() when dismissing an error', () => {
```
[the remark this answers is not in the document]
willdurand: I don't know. It is dismissed by ID, which is checked in this test case. I can add one more test case if you feel that would be better.

willdurand:
Updated with new names and one more test case :)

src/components/App/index.tsx
```diff
@@ -122,13 +127,46 @@ export class AppBase extends React.Component<Props> {
     });
   }
 
+  removeError = (errorId: number) => {
```
bobsilverberg: Nit: I don't think this function needs to be renamed from `dismissError` to `removeError`, because it's the handler for the `onClose` event and therefore the user actually is dismissing the error. I'm fine with the change, but I think `dismissError` is a better name in this instance.

src/components/App/index.spec.tsx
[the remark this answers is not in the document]
bobsilverberg: My concern is that the `id` being checked is using information from the state (`errors[errors.length - 1]`), rather than information from what was rendered on the page, so an error in the logic of the page might not be caught. Does that make sense?

src/components/App/index.tsx
willdurand: ah, I tried to come up with names that match together. Can "dismiss" be used as the opposite of "add"? (because we actually remove the error from the state).

willdurand:
> My concern is that the id being checked is using information from the state (errors[errors.length - 1]), rather than information from what was rendered on the page, so an error in the logic of the page might not be caught. Does that make sense?

Oh totally!

src/components/App/index.tsx
bobsilverberg: I marked this as a nit because it's not really that important to me. `removeError` is fine, I just think `dismissError` is better in this specific case. I still think the action should be called `removeError`, it's just the name of the function in `App` that I think could be improved. If you'd prefer to have everything match up and use `remove` that's fine with me.

willdurand: ah I see, that makes sense.

kumar303:
I can take a quick look as well

src/components/App/index.tsx
```diff
@@ -122,13 +127,46 @@ export class AppBase extends React.Component<Props> {
...
+    const { errors } = this.props;
+
+    if (errors.length === 0) {
+      return null;
+    }
+
+    return this.renderRow(
+      <Col>
+        {errors.map((error) => (
+          <Alert
```
kumar303: We'll probably need to move this and the `onClose` logic to its own component but I suppose we could do that later

src/components/App/index.spec.tsx
```diff
@@ -161,4 +162,51 @@ describe(__filename, () => {
...
+  it('renders application errors', () => {
+    const error1 = new Error('error 1');
+    const error2 = new Error('error 2');
+    const store = configureStore();
+    store.dispatch(errorsActions.addError({ error: error1 }));
+    store.dispatch(errorsActions.addError({ error: error2 }));
+
+    const root = render({ store });
+
+    expect(root.find(`.${styles.errors}`)).toHaveLength(1);
```
kumar303: Since this style is not used visually, I don't think you need it. The call to `root.find(Alert)` is enough.

src/reducers/errors.spec.tsx
```diff
@@ -0,0 +1,86 @@
...
+      state = reducer(state, actions.removeError({ id: firstError.id }));
+      expect(state.errors).toEqual([
+        {
+          id: expect.any(Number),
+          message: error2.message,
+        },
+      ]);
+    });
+
+    it('does not update the next error ID when removing an error', () => {
```
kumar303: This test doesn't seem very useful to me. I don't see any risk of accidentally leaking the behavior that you're testing for.

src/components/App/index.tsx
willdurand: I agree, thanks!

PR #454, opened 2019-03-25
Improve inlined linter message style

### LINKED ISSUE #436: Make it more obvious that linter messages belong to the line above
Feedback after the add-ons demo: it is not obvious  that linter messages are related to the lines above them. We should either reduce the `margin-top` or increase the `margin-bottom` of the messages so that it becomes obvious that messages belong to the line above.

willdurand:
Fixes #436 

---

This is another proposal to fix #436.

Unfortunately, I don't know how to reuse the Bootstrap `background-color` without using its HEX code.

## Screenshots

Screen Shot 2019-03-26 at 17 09 18
Screen Shot 2019-03-26 at 17 09 15
Screen Shot 2019-03-26 at 17 08 25
Screen Shot 2019-03-26 at 17 07 58
Screen Shot 2019-03-26 at 17 07 47
Screen Shot 2019-03-26 at 17 07 35

### DIFF
--- a/src/components/LinterMessage/styles.module.scss
+++ b/src/components/LinterMessage/styles.module.scss
@@ -1,3 +1,43 @@
+$top-arrow-width: 10px;
+$top-arrow-height: 20px;
+
+.inline:first-child {
+  &:after,
+  &:before {
+    border-color: transparent;
+    border-style: solid;
+    content: '';
+    display: block;
+    height: 0;
+    left: 10px;
+    position: absolute;
+    width: 0;
+  }
+
+  &:after {
+    border-width: $top-arrow-width;
+    top: -$top-arrow-height + 2;
+  }
+
+  &:before {
+    border-bottom-color: inherit;
+    border-width: $top-arrow-width;
+    top: -$top-arrow-height;
+  }
+}
+
+.danger.inline:after {
+  border-bottom-color: #f8d7da;
+}
+
+.secondary.inline:after {
+  border-bottom-color: #e2e3e5;
+}
+
+.warning.inline:after {
+  border-bottom-color: #fff3cd;
+}
+
 .description {
   margin-bottom: 0;
 }
--- a/src/components/LinterMessage/styles.module.scss
+++ b/src/components/LinterMessage/styles.module.scss
@@ -1,7 +1,12 @@
+@import '~bootstrap/scss/_functions.scss';
+@import '~bootstrap/scss/_variables.scss';
+
 $top-arrow-width: 10px;
 $top-arrow-height: 20px;
 
 .inline:first-child {
+  position: relative;
+
   &:after,
   &:before {
     border-color: transparent;
@@ -9,7 +14,7 @@ $top-arrow-height: 20px;
     content: '';
     display: block;
     height: 0;
-    left: 10px;
+    left: $top-arrow-height;
     position: absolute;
     width: 0;
   }
@@ -27,15 +32,15 @@ $top-arrow-height: 20px;
 }
 
 .danger.inline:after {
-  border-bottom-color: #f8d7da;
+  border-bottom-color: theme-color-level('danger', $alert-bg-level);
 }
 
 .secondary.inline:after {
-  border-bottom-color: #e2e3e5;
+  border-bottom-color: theme-color-level('secondary', $alert-bg-level);
 }
 
 .warning.inline:after {
-  border-bottom-color: #fff3cd;
+  border-bottom-color: theme-color-level('warning', $alert-bg-level);
 }
 
 .description {
--- a/stories/LinterMessage.stories.tsx
+++ b/stories/LinterMessage.stories.tsx
@@ -139,6 +139,35 @@ storiesOf('LinterMessage', module).addWithChapters('all variants', {
             />
           ),
         },
+        {
+          title: 'Two inline messages stacked',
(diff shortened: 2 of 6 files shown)


willdurand:
> I agree with the premise of the bug and I realize that your patch is a quick and simple fix but I don't think it adds much value. To my eyes, it just makes the layout look lopsided for no apparent reason.

Ah. Asymmetry is the problem here, especially in the `CodeView`.

> I'd rather we do something that highlights the line with the error on it. I realize that highlighting a line involves some widget mapping which is a bit complicated so it would be fine to defer that until later. I don't want to spend a lot of time on patches until after the high priority issues are out of the way.

I don't think highlighting the line would be wise, especially since lines can be selected (and therefore highlighted too). Everything will end up being highlighted. We should ask for UX help I believe.

I also think this is an important issue. We don't want reviewers to not use the tool because they don't understand it, right?

Because I really think this should be done sooner rather than later, what about this proposal?

Screen Shot 2019-03-25 at 16 13 39
Screen Shot 2019-03-25 at 16 12 45

kumar303:
> what about this proposal? (small border arrow added on the first message)

Yeah, I like that. Let's do it.

> I don't think highlighting the line would be wise, especially since lines can be selected (and therefore highlighted too).

I just meant some kind of visual indication, not necessarily a highlight that would conflict with the selection style. We also need to point to the exact column in the code. We have the column number in the message.

willdurand:
> Yeah, I like that. Let's do it.

Thanks. Done.

willdurand:
Just rebased.

src/components/LinterMessage/styles.module.scss
```diff
@@ -1,3 +1,43 @@
...
+.inline:first-child {
+  &:after,
+  &:before {
+    border-color: transparent;
+    border-style: solid;
+    content: '';
+    display: block;
+    height: 0;
+    left: 10px;
+    position: absolute;
```
kumar303: This relies on `position: relative` within `.alert` which is upstream in react-bootstrap as far as I can tell. I suggest making this explicit in our own styles.

src/components/LinterMessage/styles.module.scss
```diff
@@ -1,3 +1,43 @@
+$top-arrow-width: 10px;
+$top-arrow-height: 20px;
+
+.inline:first-child {
+  &:after,
+  &:before {
```
kumar303: Just out of curiosity, what makes this render as a half triangle? Typically, a `transform` would be applied to a square to rotate it but I don't see that happening.

src/components/LinterMessage/styles.module.scss
```diff
@@ -1,3 +1,43 @@
...
+    border-color: transparent;
+    border-style: solid;
+    content: '';
+    display: block;
+    height: 0;
+    left: 10px;
```
kumar303: Could this be `$top-arrow-width`? It's the same value.

src/components/LinterMessage/styles.module.scss
```diff
@@ -1,3 +1,43 @@
...
+
+  &:before {
+    border-bottom-color: inherit;
+    border-width: $top-arrow-width;
+    top: -$top-arrow-height;
+  }
+}
+
+.danger.inline:after {
+  border-bottom-color: #f8d7da;
```
kumar303: Is there a way to import the colors directly? Otherwise, I think it would help to move these to color variables just so that we can see them all in one place.

stories/LinterMessage.stories.tsx
```diff
@@ -92,6 +92,53 @@ storiesOf('LinterMessage', module).addWithChapters('all variants', {
...
+                  'The manifest contains a dictionary but no id property.',
+                description: [
+                  'A dictionary was found in the manifest, but there was no id set.',
+                ],
+              })}
+            />
+          ),
+        },
+        {
+          title: 'Inlined notice',
```
kumar303: It would help to see one more story that stacks two inline messages so you can see how the triangle should only apply to the first one.

src/components/LinterMessage/styles.module.scss
willdurand: Here is a nice explanation: https://codepen.io/chriscoyier/pen/lotjh

src/components/LinterMessage/styles.module.scss
willdurand: I did not use this variable because this is the number of pixels to put the arrow slightly on the right (from the left border).

src/components/LinterMessage/styles.module.scss
willdurand: It is a mess. It uses two or three different SCSS mixins, and the final color is derived from another set of colors that is different :/

stories/LinterMessage.stories.tsx
willdurand: :+1: Good idea, we have that in the `CodeView` and `DiffView` components but that would make sense here too

src/components/LinterMessage/styles.module.scss
willdurand: oh! I was more successful this evening, and I found how to use bootstrap code (by reading the sources, again)

src/components/LinterMessage/styles.module.scss
kumar303: Cool.

Issue #451, opened 2019-03-22
Apply LinterProvider to FileTreeNode


kumar303:
As a small step towards https://github.com/mozilla/addons-code-manager/issues/438 , refactor `FileTreeNode` to use the `LinterProvider` component added in https://github.com/mozilla/addons-code-manager/issues/414

PR #455, opened 2019-03-25
Adds LinterProvider to FileTreeNode and fixes the compare view

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/451

This change was needed to make linter icons show up in the compare view

### DIFF
--- a/src/components/FileTreeNode/index.tsx
+++ b/src/components/FileTreeNode/index.tsx
@@ -1,17 +1,16 @@
 import * as React from 'react';
-import { TreefoldRenderProps } from 'react-treefold';
 import { ListGroup } from 'react-bootstrap';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import makeClassName from 'classnames';
 
 import { gettext } from '../../utils';
-import { TreeNode } from '../FileTree';
+import { TreefoldRenderPropsForFileTree } from '../FileTree';
+import LinterProvider, { LinterProviderInfo } from '../LinterProvider';
 import { Version } from '../../reducers/versions';
 import { LinterMessage, LinterMessageMap } from '../../reducers/linter';
 import styles from './styles.module.scss';
 
-export type PublicProps = TreefoldRenderProps<TreeNode> & {
-  linterMessages: LinterMessageMap | void;
+export type PublicProps = TreefoldRenderPropsForFileTree & {
   onSelect: (id: string) => void;
   version: Version;
 };
@@ -79,14 +78,13 @@ const getIconForType = (type: string | null) => {
 };
 
 class FileTreeNodeBase<TreeNodeType> extends React.Component<PublicProps> {
-  render() {
+  renderWithLinterInfo = ({ messageMap }: LinterProviderInfo) => {
     const {
       getToggleProps,
       hasChildNodes,
       isExpanded,
       isFolder,
       level,
-      linterMessages,
       node,
       onSelect,
       renderChildNodes,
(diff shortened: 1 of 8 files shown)


stories/FileTree.stories.tsx
```diff
@@ -1,72 +1,85 @@
...
-    version={version}
-  />
-));
+const render = ({
+  store = configureStore(),
+  ...moreProps
+}: { store?: Store } & Partial<FileTreeProps> = {}) => {
+  const props: FileTreeProps = {
+    onSelect: onSelectFile,
+    version: getVersion(),
```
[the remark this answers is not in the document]
kumar303: It just made me feel twitchy because code execution in modules leads to slow startup times as the app grows. It's hard to track down this type of slowness.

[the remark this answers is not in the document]
bobsilverberg: I'm not sure I follow. This is in a storybook story. Do you mean slow startup for the storybook?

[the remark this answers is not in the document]
willdurand: mmh, I am confused here too.

src/components/FileTreeNode/index.tsx
```diff
@@ -160,6 +158,26 @@ class FileTreeNodeBase<TreeNodeType> extends React.Component<PublicProps> {
...
+        {/*
+          props.children needs be an anonymous function so that the
+          shallow prop check in connect(LinterProvider) will always think
+          there is a new value for props.children.
+
+          Without a new value, LinterProvider will not re-render its
+          children often enough. For example, the uncontrolled ListGroup
+          components (which could be nested recursively) need to
+          re-render when their internal state changes.
+        */}
```
willdurand: hu oh, is that because of `Treefold`?

stories/FileTree.stories.tsx
```diff
@@ -1,72 +1,85 @@
...
-import FileTree from '../src/components/FileTree';
+import configureStore from '../src/configureStore';
+import FileTree, {
+  PublicProps as FileTreeProps,
+} from '../src/components/FileTree';
 import { createInternalVersion } from '../src/reducers/versions';
 import { fakeVersion, fakeVersionEntry } from '../src/test-helpers';
+import { renderWithStoreAndRouter } from './utils';
 
-const version = createInternalVersion({
```
willdurand: That seems unnecessary to me. Storybook outputs static, compiled files for prod and, in dev, there is hot reload with webpack.

kumar303: Well, it would still slow down the startup of all stories but, of course, it's most likely YAGNI. It's just a bad habit for us to get into. No worries, I can change it back 😛

src/components/FileTreeNode/index.tsx
kumar303: Yes. We should be controlling it with Redux and that would probably solve it. I'm not totally sure, though. I tried to do `<LinterProvider key={isExpanded}` ...>` but that doesn't account for the expansion of nested nodes.

willdurand: I was wondering if we want to connect this tree in #185, maybe we should reopen the issue? Or maybe not if the problem only occurs in the test suite 🤷‍♂️

kumar303: This issue doesn't affect the test suite at all. Clicking on the menu had no affect before I switched to using an anonymous function. It appeared broken because `connect()` was preventing the component from re-rendering.

> I was wondering if we want to connect this tree in #185

Yes, we should. We moved quickly on getting the prototype put together (which is fine) but I feel strongly that all parts of the app should be controlled by Redux. If we don't control *all* of the app with Redux, we shouldn't even bother to use Redux.

stories/FileTree.stories.tsx
kumar303: In the current code, the `version` creation code executed at [startup time](https://github.com/mozilla/addons-code-manager/blob/fc93cfa4b1592ba8f2cddff3c5a1f0532bfdc479/stories/setup/config.tsx#L27). In my patch, it only executed in [the handler function](https://github.com/mozilla/addons-code-manager/blob/fc93cfa4b1592ba8f2cddff3c5a1f0532bfdc479/stories/FileTree.stories.tsx#L67). We should always avoid executing code during startup because this pollutes the startup process. In any case, the optimization probably won't matter and since everyone was confused about it, I reverted my change.

kumar303:
Thanks for the reviews

PR #461, opened 2019-03-25
Add API host when not using the insecure proxy

### LINKED ISSUE #460: Use `REACT_APP_API_HOST` env var in API calls

1. Go to: https://code.addons-dev.allizom.org/
2. Open the devtools > network panel

### What happened?

API calls use the current domain (code.addons-dev.allizom.org)

API calls use the dev API hosted at `addons-dev.allizom.org`

We do not use `REACT_APP_API_HOST` in `src/api/index.tsx`. In local dev, we have an insecure proxy that forwards requests from the current domain to the API but we do not want to do that when the application is deployed.

We should probably use `REACT_APP_API_HOST` in the API code.

I believe the CSP issue is also related to this problem.

### LINKED ISSUE #462: `LoginButton` should initiate the login dance using REACT_APP_API_HOST
Similar to #460, the API host is not used in the `LoginButton` component.

For instance, in -dev, the button should point to:

```
https://addons-dev.allizom.org/api/v4dev/accounts/login/start/?config=amo&to=/
```

Currently, it points to:

```
https://code.addons-dev.allizom.org/api/v4dev/accounts/login/start/?config=amo&to=/
```

willdurand:
Fixes #460
Fixes #462 

---

How I tested it:

1. I tried to run the project like in prod:

```
rm -f .env.local && yarn build && yarn start
```

This should start the project at http://localhost:3000/ and the API call to fetch the user profile should have `addons.mozilla.org` as domain in the devtools > network panel (result should be a 401).

2. I ran `yarn dev` to make sure it behaves as before.

**Note:** I believe this won't fix the login issue entirely because we need some FxA changes (see: #463).

### DIFF
--- a/src/api/index.spec.tsx
+++ b/src/api/index.spec.tsx
@@ -8,6 +8,7 @@ import {
 import {
   HttpMethod,
   callApi,
+  getApiHost,
   getCurrentUser,
   getDiff,
   getVersion,
@@ -38,7 +39,7 @@ describe(__filename, () => {
       await callApiWithDefaultApiState();
 
       expect(fetch).toHaveBeenCalledWith(
-        `/api/${defaultVersion}/?lang=${defaultLang}`,
+        `${getApiHost()}/api/${defaultVersion}/?lang=${defaultLang}`,
         {
           headers: {},
           method: 'GET',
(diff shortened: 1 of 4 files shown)


willdurand:
Thanks!

> We need a shared function for creating a URL but that can come later in #466

:+1:
 
> We need documentation for how to run the production app. I filed #467

Yes, I'll do that.

Issue #438, opened 2019-03-21
Apply LinterProvider to DRY up logic for linter messages


kumar303:
A `LinterProvider` component was created in https://github.com/mozilla/addons-code-manager/issues/414

Let's use it in ~~`DiffView`~~ (https://github.com/mozilla/addons-code-manager/issues/441), `CodeView`, and ~~`FileTreeNode`~~ (https://github.com/mozilla/addons-code-manager/issues/451).

PR #464, opened 2019-03-26
Apply LinterProvider to CodeView for code re-use

kumar303:
This is the final patch to fix https://github.com/mozilla/addons-code-manager/issues/438

It moves `CodeView` to a `LinterProvider` implementation and removes unnecessary linter logic from `pages/Browse`. After this, all components are sharing the same linter logic without any duplicated code.

### DIFF
--- a/src/components/CodeView/index.spec.tsx
+++ b/src/components/CodeView/index.spec.tsx
@@ -1,8 +1,8 @@
 import * as React from 'react';
+import PropTypes from 'prop-types';
 import { Link } from 'react-router-dom';
 import { Location } from 'history';
 import { mount } from 'enzyme';
-import { Provider } from 'react-redux';
 
 import configureStore from '../../configureStore';
 import refractor from '../../refractor';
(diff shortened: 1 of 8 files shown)


src/components/CodeView/index.spec.tsx
```diff
@@ -48,36 +63,48 @@ describe(__filename, () => {
...
+
   const renderWithMount = ({
     location = createFakeLocation(),
     ...otherProps
   }: RenderParams = {}) => {
     return mount(
-      <CodeView {...getProps(otherProps)} />,
+      <Provider store={configureStore()}>
+        <CodeView {...getProps(otherProps)} />
+      </Provider>,
```
willdurand: Can we pass `store` in the `context` instead of using `Provider`?

src/components/CodeView/index.spec.tsx
```diff
@@ -26,11 +30,22 @@ describe(__filename, () => {
     location?: Location<{}>;
   };
 
+  const globalExternalMessage = (props = {}) => {
```
willdurand: Is `createGlobalExternalMessage` too long?

src/components/CodeView/index.spec.tsx
kumar303: whoops, I forgot this code was already passing options to `mount()`. I made that change but I had to add `childContextTypes` which feels wrong. I guess the `<Provider>` approach was nice because it abstracted all of that away. I dunno, I'll just use thte `childContextTypes` approach since it's already in use for the router. 🤷‍♀️

willdurand: Yeah, I think I mentioned this in another issue. We cannot simply remove `PropTypes` because we're using TS. The old context API uses `PropTypes` to pass metadata, which is weird but it is how it is..

kumar303: The invariant exception is a runtime error so it's more than TS. I guess it's React that throws the error but that is confusing. IIRC this is a development mode feature.

PR #468, opened 2019-03-26
Create a KeyboardShortcuts component

### LINKED ISSUE #459: Create a KeyboardShortcuts component
This supports #199.

As a starting point, create a component that renders a list of keyboard shortcuts and dispatches events when those keys are pressed. Follow-ups will be filed to respond to those events.

bobsilverberg:
Fixes #459 

This is ready for a review, but there are a number of outstanding issues, some of which could probably be addressed via follow-up issues:

- The styling (and possibly positioning) of the legend needs to be improved/tightened up.
- While the new selected file is automatically highlighted in the tree when `j` or `k` is pressed, folders do not automatically open, so a user cannot see the newly highlighted file if they navigate to a file in a closed folder.
- I have not added or implemented the ability to hide and unhide the tree itself. I will open a follow-up about that.
- There was discussion about either not looping around from the last file to the first (and vice versa), or at least displaying a message when we do loop around. Neither of those have been addressed so a follow-up should be opened about that.
- I'm using the original shortcuts for expand and collapse, of `e` and `c`, but as the wording says "open" and "close", maybe `o` and `c` would be better?
- There is a TS issue with the test for the `j` and `k` buttons which I wasn't able to figure out, so I added a `@ts-ignore` comment. I believe the problem is that TS is getting confused by the arguments being passed in via `it.each`. It would be nice to fix this if it's possible.

Note also that I am displaying the shortcuts as a child of the `FileTree` component, which means that they will not be displayed/activated unless a tree actually exists, which I believe is what makes sense.

### DIFF
--- a/src/components/KeyboardShortcuts/index.spec.tsx
+++ b/src/components/KeyboardShortcuts/index.spec.tsx
@@ -5,7 +5,8 @@ import configureStore from '../../configureStore';
 import { RelativePathPosition } from '../../reducers/fileTree';
 import { actions as versionActions } from '../../reducers/versions';
 import {
-  createFakeKeyboardEvent,
+  CreateKeydownEventParams,
+  createKeydownEvent,
   createFakeThunk,
   shallowUntilTarget,
   spyOn,
@@ -47,6 +48,19 @@ describe(__filename, () => {
     );
   };
 
+  const renderAndTriggerKeyEvent = (
+    eventProps: CreateKeydownEventParams = { key: '' },
+    renderProps: Partial<RenderParams>,
+  ) => {
+    const root = render(renderProps);
+
+    const { keydownListener } = root.instance() as KeyboardShortcutsBase;
+
+    const event = createKeydownEvent(eventProps);
+
+    keydownListener(event);
+  };
+
   it('renders a list of keyboard shortcuts', () => {
     const root = render();
 
@@ -54,20 +68,25 @@ describe(__filename, () => {
   });
 
   it('binds and unbinds keydown to the listener', () => {
-    document.addEventListener = jest.fn();
-    document.removeEventListener = jest.fn();
+    const createFakeDocument = (): Partial<Document> => {
+      return {
+        addEventListener: jest.fn(),
+        removeEventListener: jest.fn(),
+      };
+    };
+    const _document = createFakeDocument() as Document;
 
-    const root = render();
+    const root = render({ _document });
     const { keydownListener } = root.instance() as KeyboardShortcutsBase;
 
-    expect(document.addEventListener).toHaveBeenCalledWith(
+    expect(_document.addEventListener).toHaveBeenCalledWith(
       'keydown',
       keydownListener,
     );
 
     root.unmount();
 
-    expect(document.removeEventListener).toHaveBeenCalledWith(
+    expect(_document.removeEventListener).toHaveBeenCalledWith(
       'keydown',
       keydownListener,
     );
@@ -88,21 +107,16 @@ describe(__filename, () => {
       const fakeThunk = createFakeThunk();
       const _goToRelativeFile = fakeThunk.createThunk;
 
-      const root = render({
-        _goToRelativeFile,
-        currentPath,
-        pathList,
-        store,
-        versionId,
-      });
-
-      const { keydownListener } = root.instance() as KeyboardShortcutsBase;
-
-      // TS cannot seem to figure out that `key` is a string.
-      // @ts-ignore
-      const event = createFakeKeyboardEvent({ key }) as KeyboardEvent;
-
-      keydownListener(event);
+      renderAndTriggerKeyEvent(
+        { key: key as string },
+        {
+          _goToRelativeFile,
+          currentPath,
+          pathList,
+          store,
+          versionId,
+        },
+      );
 
       expect(dispatch).toHaveBeenCalledWith(fakeThunk.thunk);
       expect(_goToRelativeFile).toHaveBeenCalledWith({
@@ -114,22 +128,19 @@ describe(__filename, () => {
     },
   );
 
-  it('dispatches expandTree when "e" is pressed', () => {
+  it.each(['e', 'o'])('dispatches expandTree when "%s" is pressed', (key) => {
     const versionId = 123;
 
     const store = configureStore();
     const dispatch = spyOn(store, 'dispatch');
 
-    const root = render({
-      store,
-      versionId,
-    });
-
-    const { keydownListener } = root.instance() as KeyboardShortcutsBase;
-
-    const event = createFakeKeyboardEvent({ key: 'e' }) as KeyboardEvent;
-
-    keydownListener(event);
+    renderAndTriggerKeyEvent(
+      { key: key as string },
+      {
+        store,
+        versionId,
+      },
+    );
 
     expect(dispatch).toHaveBeenCalledWith(
       versionActions.expandTree({ versionId }),
@@ -142,16 +153,13 @@ describe(__filename, () => {
     const store = configureStore();
     const dispatch = spyOn(store, 'dispatch');
 
-    const root = render({
-      store,
-      versionId,
-    });
-
-    const { keydownListener } = root.instance() as KeyboardShortcutsBase;
-
-    const event = createFakeKeyboardEvent({ key: 'c' }) as KeyboardEvent;
-
-    keydownListener(event);
+    renderAndTriggerKeyEvent(
+      { key: 'c' },
+      {
+        store,
+        versionId,
+      },
+    );
 
     expect(dispatch).toHaveBeenCalledWith(
       versionActions.collapseTree({ versionId }),
@@ -164,16 +172,12 @@ describe(__filename, () => {
       const store = configureStore();
       const dispatch = spyOn(store, 'dispatch');
 
-      const root = render({ store });
-
-      const { keydownListener } = root.instance() as KeyboardShortcutsBase;
-
-      const event = createFakeKeyboardEvent({
-        key: 'k',
-        [modififierKey]: true,
-      }) as KeyboardEvent;
-
-      keydownListener(event);
+      renderAndTriggerKeyEvent(
+        { key: 'k', [modififierKey]: true },
+        {
+          store,
+        },
+      );
 
       expect(dispatch).not.toHaveBeenCalled();
     },
--- a/src/components/KeyboardShortcuts/index.spec.tsx
+++ b/src/components/KeyboardShortcuts/index.spec.tsx
@@ -1,0 +1,181 @@
+import * as React from 'react';
+import { Store } from 'redux';
+
+import configureStore from '../../configureStore';
+import { RelativePathPosition } from '../../reducers/fileTree';
+import { actions as versionActions } from '../../reducers/versions';
+import {
+  createFakeKeyboardEvent,
+  createFakeThunk,
+  shallowUntilTarget,
+  spyOn,
+} from '../../test-helpers';
+import styles from './styles.module.scss';
+
+import KeyboardShortcuts, {
+  DefaultProps,
+  PublicProps,
+  KeyboardShortcutsBase,
+} from '.';
+
+describe(__filename, () => {
+  type RenderParams = {
+    store?: Store;
+  } & Partial<PublicProps & DefaultProps>;
+
+  const render = ({
+    currentPath = 'file1.js',
+    pathList = ['file1.js'],
+    store = configureStore(),
+    versionId = 123,
+    ...moreProps
+  }: RenderParams = {}) => {
+    const props = {
+      _goToRelativeFile: jest.fn(),
+      currentPath,
+      pathList,
+      versionId,
+      ...moreProps,
+    };
+
+    return shallowUntilTarget(
+      <KeyboardShortcuts {...props} />,
+      KeyboardShortcutsBase,
+      {
+        shallowOptions: { context: { store } },
+      },
+    );
+  };
+
+  it('renders a list of keyboard shortcuts', () => {
+    const root = render();
+
+    expect(root.find(`.${styles.KeyboardShortcuts}`)).toHaveLength(1);
+  });
+
+  it('binds and unbinds keydown to the listener', () => {
+    document.addEventListener = jest.fn();
+    document.removeEventListener = jest.fn();
+
+    const root = render();
+    const { keydownListener } = root.instance() as KeyboardShortcutsBase;
+
+    expect(document.addEventListener).toHaveBeenCalledWith(
+      'keydown',
+      keydownListener,
+    );
+
+    root.unmount();
+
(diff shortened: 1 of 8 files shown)


src/components/KeyboardShortcuts/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+import styles from './styles.module.scss';
+import { gettext } from '../../utils';
+
+const keys = ['k', 'j', 'e', 'h'];
+
+type Props = {};
+
+export class KeyboardShortcutsBase extends React.Component<Props> {
+  keydownListener = (event: KeyboardEvent) => {
```
kumar303: I believe React gives us a synthetic keyboard event so the type should reflect that: https://reactjs.org/docs/events.html#keyboard-events

src/components/KeyboardShortcuts/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+    if (keys.includes(event.key)) {
```
kumar303: You'll probably also need to check that a meta key (ctrl, command) isn't present

src/components/KeyboardShortcuts/index.tsx
bobsilverberg: What I have gathered from the documentation and my observation, is that we are not receiving React synthetic keyboard events, but rather native keyboard events. I believe that if we use an `onKeyDown` attribute on a React element then React does generate synthetic keyboard events, but in this case we are adding an event listener to the `document` object and therefore we seem to be getting native keyboard events.

If I try to use `React.KeyboardEvent` or `React.KeyboardEvent<HTMLDivElement>` instead of `KeyboardEvent` then TS gets mad about me attempting to use `this.keydownListener` in my call to `document.addEventListener`:

```
Argument of type '(event: KeyboardEvent<HTMLDivElement>) => void' is not assignable to parameter of type 'EventListenerOrEventListenerObject'.
  Type '(event: KeyboardEvent<HTMLDivElement>) => void' is not assignable to type 'EventListener'.
    Types of parameters 'event' and 'evt' are incompatible.
      Type 'Event' is missing the following properties from type 'KeyboardEvent<HTMLDivElement>': altKey, charCode, ctrlKey, getModifierState, and 12 more.ts(2345)
```

I could be wrong about this, but from what I can tell `KeyboardEvent` is correct and `React.KeyboardEvent` is not.

willdurand: >  but in this case we are adding an event listener to the `document` object and therefore we seem to be getting native keyboard events.

exactly

kumar303:
>     * The styling (and possibly positioning) of the legend needs to be improved/tightened up.

I agree. I suggest filing a new issue. At the very least, the legend should be condensed so that it doesn't take up so much space.
 
>     * While the new selected file is automatically highlighted in the tree when `j` or `k` is pressed, folders do not automatically open, so a user cannot see the newly highlighted file if they navigate to a file in a closed folder.

Please file a follow-up. The tree should open up to the folder of the file that is currently selected.
 
>     * I'm using the original shortcuts for expand and collapse, of `e` and `c`, but as the wording says "open" and "close", maybe `o` and `c` would be better?

We should retain the old shortcuts since reviewers already know them. Maybe we can keep them just as internal aliases? We could show the official shortcut as `o` but trigger the same behavior internally from `e`.

src/components/KeyboardShortcuts/styles.module.scss
```diff
@@ -0,0 +1,8 @@
+@import '../../scss/variables';
+
+.KeyboardShortcuts {
+  border-radius: $border-radius;
+  border: 1px solid $light-gray;
+  margin-bottom: $default-padding;
```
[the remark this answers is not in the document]
kumar303: I suggested filing a separate issue for the styles but I like CSS so I couldn't resist 😄 

This looks better:

```diff
diff --git a/src/components/KeyboardShortcuts/index.tsx b/src/components/KeyboardShortcuts/index.tsx
index 306c27b..a2e8750 100644
--- a/src/components/KeyboardShortcuts/index.tsx
+++ b/src/components/KeyboardShortcuts/index.tsx
@@ -99,7 +99,7 @@ export class KeyboardShortcutsBase extends React.Component<Props> {
       <div className={styles.KeyboardShortcuts}>
         <h4>{gettext('Keyboard Shortcuts')}</h4>
 
-        <dl>
+        <dl className={styles.definitions}>
           <dt>k</dt>
           <dd>{gettext('Up file')}</dd>
           <dt>j</dt>
diff --git a/src/components/KeyboardShortcuts/styles.module.scss b/src/components/KeyboardShortcuts/styles.module.scss
index 9232e88..923147a 100644
--- a/src/components/KeyboardShortcuts/styles.module.scss
+++ b/src/components/KeyboardShortcuts/styles.module.scss
@@ -3,6 +3,16 @@
 .KeyboardShortcuts {
   border-radius: $border-radius;
   border: 1px solid $light-gray;
-  margin-bottom: $default-padding;
+  margin-top: $default-padding;
   padding: $default-padding;
 }
+
+.definitions {
+  display: grid;
+  grid-template-columns: auto auto;
+  margin: 0;
+
+  dd {
+    margin-bottom: 0;
+  }
+}
```

Screenshot 2019-04-10 15 03 11

src/components/KeyboardShortcuts/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+          <dd>{gettext('Open all folders')}</dd>
+          <dt>c</dt>
+          <dd>{gettext('Close all folders')}</dd>
+        </dl>
+      </div>
+    );
+  }
+}
+
+export default connect()(KeyboardShortcutsBase);
```
kumar303: I learned recently that the return type of `connect()` does not honor the component's prop types. This is probably a bug. Any time we use this pattern, we have to redefine the types:

```diff
diff --git a/src/components/KeyboardShortcuts/index.tsx b/src/components/KeyboardShortcuts/index.tsx
index 306c27b..9fd8d27 100644
--- a/src/components/KeyboardShortcuts/index.tsx
+++ b/src/components/KeyboardShortcuts/index.tsx
@@ -114,4 +114,6 @@ export class KeyboardShortcutsBase extends React.Component<Props> {
   }
 }
 
-export default connect()(KeyboardShortcutsBase);
+export default connect()(KeyboardShortcutsBase) as React.ComponentType<
+  PublicProps & Partial<DefaultProps>
+>;
```

This is very important for things like storybook since we rely on type safety to detect broken stories.

src/components/KeyboardShortcuts/index.spec.tsx
```diff
@@ -0,0 +1,181 @@
...
+  };
+
+  it('renders a list of keyboard shortcuts', () => {
+    const root = render();
+
+    expect(root.find(`.${styles.KeyboardShortcuts}`)).toHaveLength(1);
+  });
+
+  it('binds and unbinds keydown to the listener', () => {
+    document.addEventListener = jest.fn();
```
kumar303: This looks like it could pollute the context of other tests if they also access `document`. Are you relying on some magic jest behavior that gives us `document` in a sandbox or something like that?

I'd rather see this pass in a fake document like `render({ _document })`

src/components/KeyboardShortcuts/index.spec.tsx
```diff
@@ -0,0 +1,181 @@
...
+        pathList,
+        store,
+        versionId,
+      });
+
+      const { keydownListener } = root.instance() as KeyboardShortcutsBase;
+
+      // TS cannot seem to figure out that `key` is a string.
+      // @ts-ignore
+      const event = createFakeKeyboardEvent({ key }) as KeyboardEvent;
```
kumar303: I'm not sure what's going on either; it's possible a bug in the `it.each()` definition. Instead of ignoring all errors, what about this instead?

```diff
diff --git a/src/components/KeyboardShortcuts/index.spec.tsx b/src/components/KeyboardShortcuts/index.spec.tsx
index 3d4e098..4bc4e56 100644
--- a/src/components/KeyboardShortcuts/index.spec.tsx
+++ b/src/components/KeyboardShortcuts/index.spec.tsx
@@ -98,9 +98,7 @@ describe(__filename, () => {
 
       const { keydownListener } = root.instance() as KeyboardShortcutsBase;
 
-      // TS cannot seem to figure out that `key` is a string.
-      // @ts-ignore
-      const event = createFakeKeyboardEvent({ key }) as KeyboardEvent;
+      const event = createFakeKeyboardEvent({ key: key as string }) as KeyboardEvent;
 
       keydownListener(event);
 
```

src/components/KeyboardShortcuts/index.spec.tsx
```diff
@@ -0,0 +1,181 @@
...
+    const root = render({
+      store,
+      versionId,
+    });
+
+    const { keydownListener } = root.instance() as KeyboardShortcutsBase;
+
+    const event = createFakeKeyboardEvent({ key: 'c' }) as KeyboardEvent;
+
+    keydownListener(event);
```
kumar303: How about `renderAndTriggerKeyEvent(props, event)`? I think that would cut down on duplication across a couple tests.

src/test-helpers.tsx
```diff
@@ -541,6 +541,30 @@ export const createFakeEvent = (extraProps = {}) => {
...
 
+type CreateFakeKeyboardEventParams = {
+  altKey?: boolean;
+  ctrlKey?: boolean;
+  key: string;
+  metaKey?: boolean;
+  shiftKey?: boolean;
+};
+
+export const createFakeKeyboardEvent = ({
```
kumar303: I was playing around with this to understand why you needed so many type assertions and it looks like you don't need to create a fake event at all. This is working for me:

```diff
diff --git a/src/components/KeyboardShortcuts/index.spec.tsx b/src/components/KeyboardShortcuts/index.spec.tsx
index 3d4e098..60ae453 100644
--- a/src/components/KeyboardShortcuts/index.spec.tsx
+++ b/src/components/KeyboardShortcuts/index.spec.tsx
@@ -5,7 +5,7 @@ import configureStore from '../../configureStore';
 import { RelativePathPosition } from '../../reducers/fileTree';
 import { actions as versionActions } from '../../reducers/versions';
 import {
-  createFakeKeyboardEvent,
+  createKeydownEvent,
   createFakeThunk,
   shallowUntilTarget,
   spyOn,
@@ -100,7 +100,7 @@ describe(__filename, () => {
 
       // TS cannot seem to figure out that `key` is a string.
       // @ts-ignore
-      const event = createFakeKeyboardEvent({ key }) as KeyboardEvent;
+      const event = createKeydownEvent({ key }) as KeyboardEvent;
 
       keydownListener(event);
 
@@ -127,7 +127,7 @@ describe(__filename, () => {
 
     const { keydownListener } = root.instance() as KeyboardShortcutsBase;
 
-    const event = createFakeKeyboardEvent({ key: 'e' }) as KeyboardEvent;
+    const event = createKeydownEvent({ key: 'e' });
 
     keydownListener(event);
 
@@ -149,7 +149,7 @@ describe(__filename, () => {
 
     const { keydownListener } = root.instance() as KeyboardShortcutsBase;
 
-    const event = createFakeKeyboardEvent({ key: 'c' }) as KeyboardEvent;
+    const event = createKeydownEvent({ key: 'c' });
 
     keydownListener(event);
 
@@ -168,10 +168,10 @@ describe(__filename, () => {
 
       const { keydownListener } = root.instance() as KeyboardShortcutsBase;
 
-      const event = createFakeKeyboardEvent({
+      const event = createKeydownEvent({
         key: 'k',
         [modififierKey]: true,
-      }) as KeyboardEvent;
+      });
 
       keydownListener(event);
 
diff --git a/src/test-helpers.tsx b/src/test-helpers.tsx
index 4a67a9d..d4f6817 100644
--- a/src/test-helpers.tsx
+++ b/src/test-helpers.tsx
@@ -549,20 +549,20 @@ type CreateFakeKeyboardEventParams = {
   shiftKey?: boolean;
 };
 
-export const createFakeKeyboardEvent = ({
+export const createKeydownEvent = ({
   altKey = false,
   ctrlKey = false,
   key,
   metaKey = false,
   shiftKey = false,
 }: CreateFakeKeyboardEventParams) => {
-  return {
+  return new KeyboardEvent('keydown', {
     altKey,
     ctrlKey,
     key,
     metaKey,
     shiftKey,
-  };
+  });
 };
 
 /*
```

src/components/KeyboardShortcuts/styles.module.scss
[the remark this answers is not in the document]
bobsilverberg: The looks better, thanks @kumar303.

src/components/KeyboardShortcuts/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+    document.removeEventListener('keydown', this.keydownListener);
+  }
+
+  render() {
+    return (
+      <div className={styles.KeyboardShortcuts}>
+        <h4>{gettext('Keyboard Shortcuts')}</h4>
+
+        <dl>
+          <dt>k</dt>
```
willdurand: It would be great to use [`kbd` tags](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd) for each key.

src/components/KeyboardShortcuts/index.spec.tsx
bobsilverberg: Good idea, and no, I wasn't relying on any jest magic.

src/components/KeyboardShortcuts/index.spec.tsx
bobsilverberg: I thought I had tried that, but my syntax was off. Thanks, that works!

src/components/KeyboardShortcuts/index.spec.tsx
bobsilverberg: Good idea!

src/test-helpers.tsx
bobsilverberg: The refactor you mentioned above to create `renderAndTriggerKeyEvent` actually means that only one type assertion remained, but this still seems like a decent improvement, so I've made this change as well.

src/components/KeyboardShortcuts/index.tsx
bobsilverberg: Great idea!

bobsilverberg:
Thanks for the reviews @kumar303 and @willdurand. This is ready for another look.

bobsilverberg:
> Please file a follow-up. The tree should open up to the folder of the file that is currently selected.

I opened #562

src/components/KeyboardShortcuts/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+          dispatch(
+            _goToRelativeFile({
+              currentPath,
+              pathList,
+              position: RelativePathPosition.next,
+              versionId,
+            }),
+          );
+          break;
+        case 'e':
```
kumar303: Maybe add a comment here explaining that it's from the old site? Otherwise it's not obvious why we are supporting it.

PR #474, opened 2019-03-26
Add a toggleExpandedPath action

### LINKED ISSUE #473: Add a toggleExpandedPath action
This supports #185.

As discussed in IRC, we're going to store an array of expanded paths that the FileTree component can use to determine which folders are currently expanded. To support that we need a `toggleExpandedPath` action which can be dispatched whenever a user clicks on a folder in the tree.

bobsilverberg:
Fixes #473

### DIFF
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -113,19 +113,14 @@ describe(__filename, () => {
       const version = fakeVersion;
       let state = reducer(undefined, actions.loadVersionInfo({ version }));
 
-      expect(state).toHaveProperty(
-        `versionInfo.${version.id}.expandedPaths`,
-        [],
-      );
-
-      const selectedPath = 'new/selected/path';
+      const path = 'new/selected/path';
       state = reducer(
         state,
-        actions.toggleExpandedPath({ selectedPath, versionId: version.id }),
+        actions.toggleExpandedPath({ path, versionId: version.id }),
       );
 
       expect(state).toHaveProperty(`versionInfo.${version.id}.expandedPaths`, [
-        selectedPath,
+        path,
       ]);
     });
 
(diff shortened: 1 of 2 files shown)


src/reducers/versions.tsx
```diff
@@ -212,6 +213,10 @@ export const actions = {
     return (payload: { selectedPath: string; versionId: number }) =>
       resolve(payload);
   }),
+  toggleExpandedPath: createAction('TOGGLE_EXPANDED_PATH', (resolve) => {
+    return (payload: { selectedPath: string; versionId: number }) =>
```
kumar303: I suggest a name other than `selectedPath` to avoid confusion with `version.selectedPath`. Maybe just `path`?

src/reducers/versions.spec.tsx
```diff
@@ -109,6 +109,47 @@ describe(__filename, () => {
...
+        state,
+        actions.toggleExpandedPath({ selectedPath, versionId: version.id }),
+      );
+
+      expect(state).toHaveProperty(`versionInfo.${version.id}.expandedPaths`, [
+        selectedPath,
+      ]);
+    });
+
+    it('removes a path from expandedPaths', () => {
```
kumar303: I suggest one more test that adds two paths, removes one path, and makes sure the first is still in `expandedPaths`.

bobsilverberg:
Thanks for the review @kumar303.

> I have a feeling that the [`updateSelectedPaths`](https://github.com/mozilla/addons-code-manager/blob/e73de1ef154ee0b0fba54d27252330310c137bfb/src/reducers/versions.tsx#L603) reducer will also need to add its path to `expandedPaths` but you can figure that out later when you integrate with the UI.

I don't think we're going to need this, because `updateSelectedPaths` only ever gets called for a file, and not for a folder, and `toggleExpandedPath` only ever gets called for a folder, so I don't think we'll ever need to store a path selected via `updateSelectedPaths`. But yes, if it becomes an issue I'll address it when integrating with the UI.

PR #479, opened 2019-03-27
Dispatch addError() instead of logging an error message

### LINKED ISSUE #472: Handle API errors everywhere
Once ##182 lands, we can add proper error handling pretty much everywhere and replace the `TODO`: https://github.com/mozilla/addons-code-manager/search?q=%22TODO%3A+handle%22&unscoped_q=%22TODO%3A+handle%22.

willdurand:
Fixes #472 

---

Removed a bunch of TODOs by dispatching `addError()` instead.

Note: I did not change the behavior of the `linter` reducer here: https://github.com/mozilla/addons-code-manager/blob/c77356d60a358b1775dacb627f9f6630974c4773/src/reducers/linter.tsx#L198-L207

### DIFF
--- a/src/reducers/linter.spec.tsx
+++ b/src/reducers/linter.spec.tsx
@@ -18,6 +18,7 @@ import linterReducer, {
   initialState,
   selectMessageMap,
 } from './linter';
+import { actions as errorsActions } from './errors';
 
 describe(__filename, () => {
   const createExternalLinterResult = (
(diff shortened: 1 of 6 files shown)


kumar303:
> I did not change the behavior of the `linter` reducer

Yeah, we can defer this. The todo comment implied that the component would need to check for errors in addition to checking `messageMap`. That would not be possible with our current error architecture because there is no way to link an error back to the thunk that produced it.

PR #482, opened 2019-03-27
Create a function to construct an API URL and use it

### LINKED ISSUE #466: Use a function to construct an API URL
The `callApi()` function constructs a URL here:

https://github.com/mozilla/addons-code-manager/blob/99a91a44651ceb851fb009eeb2edfb7e9f8df022/src/api/index.tsx#L86

The `LoginButton` constructs a similar URL here:

https://github.com/mozilla/addons-code-manager/blob/99a91a44651ceb851fb009eeb2edfb7e9f8df022/src/components/LoginButton/index.tsx#L26

Let's introduce a shared function for both of them to use. This will be more important when they both need to reference an env var for the API host. See https://github.com/mozilla/addons-code-manager/pull/461

### LINKED ISSUE #480: Linter custom API code should use `REACT_APP_API_HOST`
We have some custom API code for fetching the linter messages, mainly because we are not relying on the addons-server API _per se_. In #460, we did not update this code though, so let's do it!

https://github.com/mozilla/addons-code-manager/blob/c6cad11653b45538292c5d2fd5b65033adb65b11/src/reducers/linter.tsx#L161-L163

willdurand:
Fixes #466
Fixes #480 

---

This patch ntroduces a `makeApiURL` function to construct an API URL with common values. Happy to change its name.

### DIFF
--- a/src/api/index.spec.tsx
+++ b/src/api/index.spec.tsx
@@ -15,6 +15,7 @@ import {
   getVersionsList,
   isErrorResponse,
   logOutFromServer,
+  makeApiURL,
 } from '.';
 
 describe(__filename, () => {
@@ -39,7 +40,7 @@ describe(__filename, () => {
       await callApiWithDefaultApiState();
 
       expect(fetch).toHaveBeenCalledWith(
-        `${getApiHost()}/api/${defaultVersion}/?lang=${defaultLang}`,
+        makeApiURL({ path: `/?lang=${defaultLang}` }),
         {
           headers: {},
           method: 'GET',
@@ -91,11 +92,12 @@ describe(__filename, () => {
 
     it('accepts an API version', async () => {
       const version = 'v123';
+      const endpoint = '/';
 
-      await callApiWithDefaultApiState({ endpoint: '/', version });
+      await callApiWithDefaultApiState({ endpoint, version });
 
       expect(fetch).toHaveBeenCalledWith(
-        expect.stringMatching(`/api/${version}/`),
+        expect.stringMatching(makeApiURL({ path: endpoint, version })),
         expect.any(Object),
       );
     });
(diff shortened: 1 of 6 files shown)


src/api/index.spec.tsx
```diff
@@ -91,11 +92,12 @@ describe(__filename, () => {
...
     it('accepts an API version', async () => {
       const version = 'v123';
+      const endpoint = '/';
 
-      await callApiWithDefaultApiState({ endpoint: '/', version });
+      await callApiWithDefaultApiState({ endpoint, version });
 
       expect(fetch).toHaveBeenCalledWith(
-        expect.stringMatching(`/api/${version}/`),
+        expect.stringMatching(makeApiURL({ path: endpoint, version })),
```
kumar303: This probably no longer needs `expect.stringMatching`, right?

src/api/index.spec.tsx
```diff
@@ -359,4 +349,80 @@ describe(__filename, () => {
...
+      const path = '/foo/';
+
+      expect(makeApiURL({ path, version: null })).toEqual(
+        expect.stringMatching(`/api${path}`),
+      );
+    });
+
+    it('can omit the prefix', () => {
+      const path = '/foo/';
+      // This is needed because `stringMatching` won't throw if `prefix` is there.
```
kumar303: I guess you are trying to prevent someone from using `stringMatching`? I don't really see how this comment is helpful.

src/api/index.spec.tsx
```diff
@@ -359,4 +349,80 @@ describe(__filename, () => {
...
+      const _getApiHost = jest.fn().mockReturnValue(apiHost);
+
+      expect(makeApiURL({ _getApiHost, path, prefix: null })).toEqual(
+        `${apiHost}/${defaultVersion}${path}`,
+      );
+    });
+
+    it('can omit both the prefix and version', () => {
+      const path = '/foo/';
+      // This is needed because `stringMatching` won't throw if `prefix` is there.
```
kumar303: To me, it's obvious that `stringMatching` couldn't be used for this test so, personally, I don't think this comment is necessary.

src/api/index.spec.tsx
```diff
@@ -359,4 +349,80 @@ describe(__filename, () => {
...
+
+    it('makes sure the path starts with a slash', () => {
+      const path = 'foo';
+
+      expect(makeApiURL({ path })).toEqual(
+        expect.stringMatching(`/api/${defaultVersion}/${path}`),
+      );
+    });
+
+    it('calls _getApiHost', () => {
```
kumar303: This test is already covered but it wouldn't hurt to add `expect(_getApiHost).toHaveBeenCalled()` to the end of it:

https://github.com/mozilla/addons-code-manager/blob/59578a7e340a9a4ab59d62a6c3872e69c28d3c73/src/api/index.spec.tsx#L354-L362

src/api/index.spec.tsx
willdurand: I kept `stringMatching` because we append `?lang=` to the URL.

Issue #327, opened 2019-03-06
Deploy dev


kumar303:
Deploy the app to dev as a step towards full deployment: https://github.com/mozilla/addons-code-manager/issues/25

The domain will probably be `code.addons-dev.allizom.org`

Notes:
- We have a build step that creates a static HTML site
- There is also a Node.js server that needs to handle all incoming requests

See [`yarn build`](https://github.com/mozilla/addons-code-manager#yarn-build) and `yarn start` (not documented for some reason).

willdurand:
The sooner, the better! Do we need ops?

willdurand:
I started the docker integration: https://github.com/mozilla/addons-code-manager/pull/343

>  `yarn start` (not documented for some reason).

yeah, I think we should remove this command as it is not used (or we alias it to another more useful command)

willdurand:
@bqbn and @autrilla: we'd like to create a hosted -dev env (the sooner, the better). A few things:

- this project already contains a `Dockerfile` but there is no `version.json` yet
- we have at least two environment variables to give to the running docker containers:
  - `REACT_APP_API_HOST` with `https://addons-dev.allizom.org` as value
  - `REACT_APP_SENTRY_DSN` with a public Sentry DSN (project has to be created in Sentry)

bqbn:
@willdurand, what do you mean by "a hosted -dev env"? 

Do you need the full stack? i.e. the frontend, the backends, the database, the elasticsearch cluster, redis and etc., or do you just need a separate frontend stack? And how would you like push to the said environment? and etc.

We need more info and background in order to know what ops need to do.

Because this sounds like a lot for ops to set up, and is also under such a short notice, we cannot commit to a timeline when we can set this up, at this point.

willdurand:
> @willdurand, what do you mean by "a hosted -dev env"?

Sorry, we want this project to be deployed and available at: `code.addons-dev.allizom.org`
 
> Do you need the full stack? i.e. the frontend, the backends, the database, the elasticsearch cluster, redis and etc., or do you just need a separate frontend stack? And how would you like push to the said environment? and etc.

It is a new frontend application that talks to the addons-server API, so we only need a separate "frontend stack" I believe.

We would like to have each commit to `master` deployed (like on addons-frontend for the -dev env).

> We need more info and background in order to know what ops need to do.
> 
> Because this sounds like a lot for ops to set up, and is also under such a short notice, we cannot commit to a timeline when we can set this up, at this point.

[...]

PR #485, opened 2019-03-27
Configure a CDN

### LINKED ISSUE #421: Add CDN endpoints to CSP
When CDN endpoints are known the CSP should be updated to configure `script-src`, `img-src` and `font-src` and any other relevant directives with the CDN endpoints. 

If the favicon is served from the main origin then `'self'` will need to remain under `img-src` otherwise it can be removed. For all others `'self'` should be removed when the CDN hosts are added.

When setting the directives, it's preferable to use a path rather than just an origin if at all possible. E.g. `https://cdn.blah.net/static/` is better than just `https://cdn.blah.net`

willdurand:
Fixes #421 

---

This _should_ work. When running `yarn start` locally, I see absolute URLs in the HTML and these URLs look good.

Static files are already available on the -dev CDN: http://addons-code-manager-dev-cdn.allizom.org/, so deploying this patch should not be a problem, unless there is a CSP issue. We might loose the `favicon.ico` though 😢

### DIFF
--- a/.env
+++ b/.env
@@ -15,3 +15,6 @@ REACT_APP_FXA_CONFIG=code-manager
 # which violates the CSP.
 INLINE_RUNTIME_CHUNK=false
 
+# CRA forces us to use a variable named `PUBLIC_URL` to set the CDN URL...
+# See: https://facebook.github.io/create-react-app/docs/advanced-configuration
+PUBLIC_URL=https://addons-code-manager.cdn.mozilla.net
--- a/.env.dev
+++ b/.env.dev
@@ -1,1 +1,3 @@
 REACT_APP_API_HOST=https://addons-dev.allizom.org
+# CDN URL
+PUBLIC_URL=https://addons-code-manager-dev-cdn.allizom.org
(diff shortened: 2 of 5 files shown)


.env
```diff
@@ -15,3 +15,6 @@ REACT_APP_FXA_CONFIG=code-manager
 # which violates the CSP.
 INLINE_RUNTIME_CHUNK=false
 
+# CRA forces us to use a variable named `PUBLIC_URL` to set the CDN URL...
+# See: https://facebook.github.io/create-react-app/docs/advanced-configuration
+PUBLIC_URL=https://addons-code-manager.cdn.mozilla.net
```
kumar303: Did you confirm all of these CDN URLs? I'm not sure what they willl be.

willdurand: Yes, it is from: https://github.com/mozilla/addons-code-manager/issues/327#issuecomment-473513809

willdurand:
Thanks, I guess we'll see :)

muffinresearch:
@willdurand is it possible to set a origin + path for the CDN hosted statics? That would be better than just an origin.

willdurand:
> @willdurand is it possible to set a origin + path for the CDN hosted statics? That would be better than just an origin.

Ah, I suppose yes.

PR #487, opened 2019-03-27
 Connect the FileTree component 

bobsilverberg:
~~Depends on #473.~~

It all works, but I'm not too sure about the tests. They feel a bit weird to me. Also, the Storybook story for `FileTree` is now broken, in that you cannot interact with it and expand the tree. I took a look but I'm not sure how to get it working.

### DIFF
--- a/src/components/FileTree/index.tsx
+++ b/src/components/FileTree/index.tsx
@@ -7,7 +7,7 @@ import FileTreeNode, {
   PublicProps as FileTreeNodeProps,
 } from '../FileTreeNode';
 import { ConnectedReduxProps } from '../../configureStore';
-import { Version, actions as versionActions } from '../../reducers/versions';
+import { Version, actions as versionsActions } from '../../reducers/versions';
 import { getLocalizedString } from '../../utils';
 
 type FileNode = {
@@ -150,7 +150,7 @@ export class FileTreeBase extends React.Component<Props> {
     const { dispatch, version } = this.props;
 
     dispatch(
-      versionActions.toggleExpandedPath({
+      versionsActions.toggleExpandedPath({
         path: node.id,
         versionId: version.id,
       }),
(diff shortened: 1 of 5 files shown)


bobsilverberg:
Note that this has been rebased to reduce it down to the pertinent commit.

bobsilverberg:
I haven't addressed the other issues yet, but I changed the code so the file tree is not rebuilt on every render. I think my approach will work, because the `FileTree` will always receive a `version` in its props.

@kumar303 Can you see if this helps the performance issue?

src/components/FileTree/index.tsx
```diff
@@ -135,29 +137,57 @@ export type PublicProps = {
...
-type Props = PublicProps;
+type Props = PublicProps & ConnectedReduxProps;
 
 export class FileTreeBase extends React.Component<Props> {
+  tree: DirectoryNode;
+
+  constructor(props: Props) {
+    super(props);
+    const { version } = this.props;
+    this.tree = buildFileTree(
```
kumar303: This optimization makes sense (good idea) but I'm not seeing much of a speedup after this change. Also, the tree should be updated when the `version` prop changes. I don't think it's worth finishing out this change for this patch. We should address the performance issues in a separate issue.

src/pages/Browse/index.spec.tsx
```diff
@@ -179,7 +179,7 @@ describe(__filename, () => {
       versionId: String(version.id),
     });
 
-    const fileTree = root.find(FileTree);
+    const fileTree = root.find(FileTree as FileTreeBase);
```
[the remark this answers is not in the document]
bobsilverberg: Yes, that seems better.

src/components/FileTree/index.tsx
bobsilverberg: Ok, I will revert the commit that introduced that change.

PR #495, opened 2019-03-29
Pass credentials when fetching the linter results

### LINKED ISSUE #490: Send credentials when fetching the linter messages
In order to successfully retrieve the linter messages, we should send the credentials when fetching these messages. This requires two changes:

1. Add `{ credentials: 'include' }` in https://github.com/mozilla/addons-code-manager/blob/63ba1b4379cd2a0366bd4c6f5c8733be3db2eebe/src/reducers/linter.tsx#L162-L164
2. Configure CORS headers in addons-server (https://github.com/mozilla/addons-server/issues/11048) and set the `mode` to `cors` in our code

In other words, we want to apply this patch in this project:

```diff
diff --git a/src/reducers/linter.tsx b/src/reducers/linter.tsx
index 01f159b..c83a12d 100644
--- a/src/reducers/linter.tsx
+++ b/src/reducers/linter.tsx
@@ -161,6 +161,7 @@ export const fetchLinterMessages = ({
     // This is a special URL and returns a non-standard JSON response.
     const response = await fetch(
       makeApiURL({ path: url, version: null, prefix: null }),
+      { credentials: 'include', mode: 'cors' },
     );
 
     try {
```

willdurand:
Fixes #490

---

This is safe to land since https://github.com/mozilla/addons-server/issues/11048 has been merged.

It is possible to try this patch in -dev by using Chrome and its "Local Overrides" feature:

Screen Shot 2019-03-29 at 12 23 00

### DIFF
--- a/src/reducers/linter.spec.tsx
+++ b/src/reducers/linter.spec.tsx
@@ -432,6 +432,7 @@ describe(__filename, () => {
 
       expect(fetchMock).toHaveBeenCalledWith(
         makeApiURL({ path: url, version: null, prefix: null }),
+        { credentials: 'include' },
       );
     });
 
--- a/src/reducers/linter.tsx
+++ b/src/reducers/linter.tsx
@@ -161,6 +161,7 @@ export const fetchLinterMessages = ({
     // This is a special URL and returns a non-standard JSON response.
     const response = await fetch(
       makeApiURL({ path: url, version: null, prefix: null }),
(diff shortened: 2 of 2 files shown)


willdurand:
> Huh. This seems like the right thing to do but I wonder why we didn't realize it sooner (I wonder why it was working without it).

Because of the insecure proxy, which allows our code to use the same domain for both the code and the api.

PR #497, opened 2019-03-29
Better CSP for static files

### LINKED ISSUE #493: Set a origin + path for the CDN hosted statics
From: https://github.com/mozilla/addons-code-manager/pull/485#issuecomment-477693502

willdurand:
Fixes #493

---

The `favicon.ico` file isn't hosted on the CDN yet, but I'll request an
ops change for that.  The rest should work as before, but it is better
to have the origin + path rather than the origin only.

Note: `static` (the name of the static folder) is hard-coded in CRA.


src/server/index.tsx
```diff
@@ -64,7 +65,12 @@ export const createServer = ({
     reportUri: '/__cspreport__',
   };
 
-  const cdnURL = env.PUBLIC_URL || "'none'";
+  const staticSrc = env.PUBLIC_URL
+    ? `${env.PUBLIC_URL}${STATIC_PATH}`
+    : "'none'";
+  const imgSrc = env.PUBLIC_URL
+    ? [staticSrc, `${env.PUBLIC_URL}/favicon.ico`]
```
kumar303: Can't we just put `favicon.ico` inside the `static` directory? That would keep things simple.

willdurand: Yep, that makes sense.

PR #498, opened 2019-03-29
Allow the FileTree to expand and collapse

### LINKED ISSUE #469: Enable the FileTree to fully expand and collapse
Depends on #185.

One of the keyboard shortcuts required for #199 is meant to expand the entire FileTree. I think it makes sense to allow the user to collapse the whole tree as well (possibly via the same keyboard shortcut, which would serve as a toggle). As far as I can tell, there's no simple way to just ask `react-treefold` to expand and/or collapse, the entire tree, so some code will need to be written to allow this to happen. Once that code is in place, we can have an action which can be dispatched to cause the tree to expand, and another to cause the tree to collapse.

I think it probably makes sense to complete #185 first, which I have raised for discussion in that issue.

bobsilverberg:
~~Depends on #503~~ 
Fixes #469

### DIFF
--- a/src/components/FileTree/index.tsx
+++ b/src/components/FileTree/index.tsx
@@ -1,9 +1,10 @@
 import log from 'loglevel';
 import * as React from 'react';
-import { ListGroup } from 'react-bootstrap';
+import { Button, ListGroup } from 'react-bootstrap';
 import { connect } from 'react-redux';
 import Treefold, { TreefoldRenderProps } from 'react-treefold';
 
+import styles from './styles.module.scss';
 import FileTreeNode, {
   PublicProps as FileTreeNodeProps,
 } from '../FileTreeNode';
@@ -191,21 +192,25 @@ export class FileTreeBase extends React.Component<Props> {
   onExpandTree = () => {
     const { dispatch, version } = this.props;
 
-    dispatch(
-      versionsActions.expandTree({
-        versionId: version.id,
-      }),
-    );
+    if (version) {
+      dispatch(
+        versionsActions.expandTree({
+          versionId: version.id,
+        }),
+      );
+    }
   };
 
   onCollapseTree = () => {
     const { dispatch, version } = this.props;
 
-    dispatch(
-      versionsActions.collapseTree({
-        versionId: version.id,
-      }),
-    );
+    if (version) {
+      dispatch(
+        versionsActions.collapseTree({
+          versionId: version.id,
+        }),
+      );
+    }
   };
 
   isNodeExpanded = (node: TreeNode) => {
@@ -229,18 +234,34 @@ export class FileTreeBase extends React.Component<Props> {
 
     return (
       <ListGroup>
+        <div className={styles.controlButtons}>
+          <Button
+            className={styles.button}
+            id="openAll"
+            onClick={this.onExpandTree}
+            size="sm"
+            type="button"
+            variant="light"
+          >
+            {gettext('Open all folders')}
+          </Button>
+          <Button
+            className={styles.button}
+            id="closeAll"
+            onClick={this.onCollapseTree}
+            size="sm"
+            type="button"
+            variant="light"
+          >
+            {gettext('Close all folders')}
+          </Button>
+        </div>
         <Treefold
           nodes={[tree]}
           render={this.renderNode}
           isNodeExpanded={this.isNodeExpanded}
           onToggleExpand={this.onToggleExpand}
         />
-        <button id="openAll" onClick={this.onExpandTree} type="button">
-          {gettext('Open all folders')}
-        </button>
-        <button id="closeAll" onClick={this.onCollapseTree} type="button">
-          {gettext('Close all folders')}
-        </button>
       </ListGroup>
     );
   }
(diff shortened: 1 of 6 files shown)


src/components/FileTree/index.tsx
```diff
@@ -179,6 +202,12 @@ export class FileTreeBase extends React.Component<Props> {
           isNodeExpanded={this.isNodeExpanded}
           onToggleExpand={this.onToggleExpand}
         />
+        <button onClick={this.onExpandTree} type="button">
+          {gettext('Expand the tree')}
```
kumar303: Something like *Open all folders* would probably make more sense in the UI

src/reducers/versions.spec.tsx
```diff
@@ -178,6 +180,43 @@ describe(__filename, () => {
...
+    });
+
+    it('removes all paths from expandedPaths when collapseTree is dispatched', () => {
+      const version = fakeVersion;
+      let state = reducer(undefined, actions.loadVersionInfo({ version }));
+
+      const path = 'new/selected/path';
+      state = reducer(
+        state,
+        actions.toggleExpandedPath({ path, versionId: version.id }),
```
kumar303: I think this test should expand more than one path

src/reducers/versions.tsx
```diff
@@ -624,6 +631,41 @@ const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
         },
       };
     }
+    case getType(actions.expandTree): {
+      const { versionId } = action.payload;
+
+      const version = state.versionInfo[versionId];
+      const { entries } = version;
+      const expandedPaths = entries
+        .filter((entry) => entry.type === 'directory')
```
kumar303: I didn't see any tests that mix directories and files. That would be a good test.

src/components/FileTree/index.tsx
```diff
@@ -179,9 +210,28 @@ export class FileTreeBase extends React.Component<Props> {
           isNodeExpanded={this.isNodeExpanded}
           onToggleExpand={this.onToggleExpand}
         />
+        <button id="openAll" onClick={this.onExpandTree} type="button">
```
kumar303: Always use bootstrap components to keep the look and feel consistent.

I have a patch to suggest (below):
* Move the buttons up top where they are more effective. Without this, the *close all* button is inaccessible after opening everything.
* Render small, light buttons side by side to conserve space
* Truncate the button text on small screens. This just looks nicer but, also, *Open all ...* still makes enough sense.

```diff
diff --git a/src/components/FileTree/index.tsx b/src/components/FileTree/index.tsx
index 947dece..de12414 100644
--- a/src/components/FileTree/index.tsx
+++ b/src/components/FileTree/index.tsx
@@ -1,8 +1,9 @@
 import * as React from 'react';
-import { ListGroup } from 'react-bootstrap';
+import { Button, ListGroup } from 'react-bootstrap';
 import { connect } from 'react-redux';
 import Treefold, { TreefoldRenderProps } from 'react-treefold';
 
+import styles from './styles.module.scss';
 import FileTreeNode, {
   PublicProps as FileTreeNodeProps,
 } from '../FileTreeNode';
@@ -204,18 +205,34 @@ export class FileTreeBase extends React.Component<Props> {
 
     return (
       <ListGroup>
+        <div className={styles.controlButtons}>
+          <Button
+            className={styles.button}
+            id="openAll"
+            onClick={this.onExpandTree}
+            size="sm"
+            type="button"
+            variant="light"
+          >
+            {gettext('Open all folders')}
+          </Button>
+          <Button
+            className={styles.button}
+            id="closeAll"
+            onClick={this.onCollapseTree}
+            size="sm"
+            type="button"
+            variant="light"
+          >
+            {gettext('Close all folders')}
+          </Button>
+        </div>
         <Treefold
           nodes={[tree]}
           render={this.renderNode}
           isNodeExpanded={this.isNodeExpanded}
           onToggleExpand={this.onToggleExpand}
         />
-        <button id="openAll" onClick={this.onExpandTree} type="button">
-          {gettext('Open all folders')}
-        </button>
-        <button id="closeAll" onClick={this.onCollapseTree} type="button">
-          {gettext('Close all folders')}
-        </button>
       </ListGroup>
     );
   }
diff --git a/src/components/FileTree/styles.module.scss b/src/components/FileTree/styles.module.scss
new file mode 100644
index 0000000..01fb104
--- /dev/null
+++ b/src/components/FileTree/styles.module.scss
@@ -0,0 +1,14 @@
+.controlButtons {
+  display: grid;
+  grid-column-gap: 10px;
+  grid-template-columns: 1fr 1fr;
+  margin-bottom: 10px;
+  width: 100%;
+}
+
+.button {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  width: 100%;
+}
```

bobsilverberg: Thanks, that looks awesome!

bobsilverberg:
Thanks @kumar303. I have addressed all of the comments. Please note that this is still built on top of #503.

src/reducers/versions.spec.tsx
```diff
@@ -178,6 +180,67 @@ describe(__filename, () => {
...
+
+      expect(state).toHaveProperty(`versionInfo.${version.id}.expandedPaths`, [
+        path1,
+        path2,
+        getRootPath(createInternalVersion(version)),
+      ]);
+    });
+
+    it('does not add paths for files to expandedPaths when expandTree is dispatched', () => {
+      const path1 = 'path/1';
```
kumar303: This test would be easier to read if you used realistic examples, maybe like `scripts/` and `scripts/background.js`. Even though I understood the spec, I was having a hard time deciphering `path1` and `path2`.

Issue #500, opened 2019-03-29
Optimize components that use LinterProvider


kumar303:
Part of the slowdown in https://github.com/mozilla/addons-code-manager/issues/494 is related to how `<LinterProvider>` components re-render each time `version` changes. Additionally, many components use an anonymous function to constantly force a re-render. This all became more apparent while expanding folders after https://github.com/mozilla/addons-code-manager/pull/487

Issue #494, opened 2019-03-28
Navigating the file tree is slow


kumar303:
### Describe the problem and steps to reproduce it:

* Browse a version like http://localhost:3000/en-US/browse/494431/versions/1532144/
* Click the root to expand it
* Click various directories to expand and collapse them

### What happened?

There is a bit of lagginess introduced by https://github.com/mozilla/addons-code-manager/pull/487

### What did you expect to happen?

File navigation should not be laggy

### Anything else we should know?

I [profiled](https://calibreapp.com/blog/react-performance-profiling-optimization/) to see what's happening here. The `CodeView` component seems to cause most of the slowdown. The patch in https://github.com/mozilla/addons-code-manager/pull/487 dispatches a Redux action every time a directory is expanded. This causes `CodeView` to re-render each time when previously it would not. There also seems to be some slowdown in `FileTreeNode` which I haven't dug into yet.

willdurand:
> The `CodeView` component seems to cause most of the slowdown. The patch in #487 dispatches a Redux action every time a directory is expanded. This causes `CodeView` to re-render each time when previously it would not.

This component does not have any guards and re-render all the time (because it is not using `connect()` for instance, which implements `shouldComponentUpdate()`). The `React.memo` helper might be a good solution here.

willdurand:
I tried the STR above in FF Nightly + React profiler (commit: aa565fbf0e2088c1988780890dd00eb777b3e3f7):

![Screen Shot 2019-04-01 at 11 14 33](https://user-images.githubusercontent.com/217628/55316756-612db800-546f-11e9-9413-532d80b7c8b9.png)

willdurand:
Similar capture with the patch from https://github.com/mozilla/addons-code-manager/pull/501:

![Screen Shot 2019-04-01 at 11 18 24](https://user-images.githubusercontent.com/217628/55317007-e87b2b80-546f-11e9-9b95-2438bc58684c.png)

I did not open/closed the exact same folders in the exact same order. I did open/closed lots of folders though. We should probably have a [protocol](https://en.wikipedia.org/wiki/Protocol_(science)) for such analyses though.

kumar303:
I forgot to mention that the slowness is dependent on the speed of rendering the currently selected file (which can be very slow depending on line length). I updated the PR with details: https://github.com/mozilla/addons-code-manager/pull/501#issuecomment-478617440

When I first noticed the problem, it was because I had a big-ish file selected.

[...]

PR #501, opened 2019-03-29
Stop <LinterProvider> from re-rendering so often

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/500

This makes a dent in https://github.com/mozilla/addons-code-manager/issues/494

When throttling my cpu to profile it, I saw a bit of a speedup after this patch. Here are some samples I grabbed that roughly show the average render times of expanding / collapsing a node. Even though these two samples I grabbed don't seem dramatic, it feels a bit faster. You can see in the second one that `<CodeView>` no longer re-renders.

**Before**

Screenshot 2019-03-29 11 02 44

**After**

Screenshot 2019-03-29 11 05 04


kumar303:
I forgot to mention that the slowness of navigation is dependent on the time it takes to render the currently selected file. Files with many lines of code will trigger a dramatic slowdown in navigation. Here is an example.

The steps I took:
* Open Chrome
* Disable all extensions (especially React which adds overhead)
* Load http://localhost:3000/en-US/browse/494431/versions/1532144/ 
* Click to expand the tree
* Load `lib/adblockplus.js`, a large file
* Begin profiling without CPU throttling
* Expand and collapse `qunit`
* Expand and collapse `icons`
* Expand and collapse `ext`

Here is a baseline using https://github.com/mozilla/addons-code-manager/commit/0739412. I am eyeballing the average time it takes to expand/collapse a folder. Here's one (2.25 **seconds**):

Screenshot 2019-04-01 09 54 11

Here is the same STR with the patch right here (https://github.com/mozilla/addons-code-manager/pull/501), 41.30 **milleseconds**:

Screenshot 2019-04-01 09 56 57

willdurand:
Thanks, this last comment above is very helpful. I don't remember where but I was mentioning the need for some sort of protocol to perform these analyses. I can confirm that [the current `master` branch](https://github.com/mozilla/addons-code-manager/commit/0739412) is super slow.

willdurand:
Do you really need to change the `LinterProvider` API? I get similar results with this patch:

```diff
diff --git a/src/components/CodeView/index.tsx b/src/components/CodeView/index.tsx
index 5e22100..31500ce 100644
--- a/src/components/CodeView/index.tsx
+++ b/src/components/CodeView/index.tsx
@@ -52,16 +52,22 @@ type RowProps = {
   ref?: typeof scrollToSelectedLine;
 };
 
-export const CodeViewBase = ({
-  _scrollToSelectedLine = scrollToSelectedLine,
-  content,
-  location,
-  mimeType,
-  version,
-}: Props) => {
-  const language = getLanguageFromMimeType(mimeType);
-
-  const renderWithLinterInfo = ({ selectedMessageMap }: LinterProviderInfo) => {
+export class CodeViewBase extends React.Component<Props> {
+  shouldComponentUpdate(nextProps: Props) {
+    return this.props.content !== nextProps.content;
+  }
+
+  renderWithLinterInfo = ({ selectedMessageMap }: LinterProviderInfo) => {
+    const {
+      _scrollToSelectedLine = scrollToSelectedLine,
+      content,
+      location,
+      mimeType,
+      version,
+    } = this.props;
+
+    const language = getLanguageFromMimeType(mimeType);
+
     return (
       <React.Fragment>
         {selectedMessageMap &&
@@ -130,9 +136,15 @@ export const CodeViewBase = ({
     );
   };
 
-  return (
-    <LinterProvider version={version}>{renderWithLinterInfo}</LinterProvider>
-  );
-};
+  render() {
+    const { version } = this.props;
+
+    return (
+      <LinterProvider version={version}>
+        {this.renderWithLinterInfo}
+      </LinterProvider>
+    );
+  }
+}
 
 export default withRouter(CodeViewBase);
```

The idea is to avoid a re-render for nothing, and in this case, it checks the `content`, which is pretty accurate (but we could have other more conditions).

kumar303:
> Do you really need to change the LinterProvider API? 

Yes, I think it would be a good idea; my patch will prevent *all* `LinterProvider` based components from constantly re-rendering, not just `CodeView`.

For example, the [old version](https://github.com/mozilla/addons-code-manager/pull/501/files#diff-8c92925c4ce87a2409eef2d34cb6c61bL213) of `FileTreeNode` was constantly re-rendering, say, if a component dispatched an error (or anything).

The `LinterProvider` should not have been receiving a complex `version` object because Redux only does a shallow comparison (iirc) and that makes the `connect()` wrapper update frequently. Additionally, attributes like `version.expandedPaths` should not automatically cause `LinterProvider` to re-render since they are not relevant to what `LinterProvider` does. If we need the underlying component to re-render based on such an attribute then we can set a key, like [I did](https://github.com/mozilla/addons-code-manager/pull/501/files#diff-8c92925c4ce87a2409eef2d34cb6c61bR203) for `FileTreeNode`.

Issue #442, opened 2019-03-21
Fix return type of getVersionInfo


kumar303:
The `getVersionInfo` selector currently looks like this:

https://github.com/mozilla/addons-code-manager/blob/5efef278ed8e2e3b7b2d7dfa1ceae7451cf3d801/src/reducers/versions.tsx#L309-L311

The return type should be `Version | void` because the `versionId` might be missing from `versions.versionInfo`. However, TypeScript infers the return type only as `Version`.

Let's declare the correct return type and fix the fallout (there is a lot). It should look more like:

```tsx
export const getVersionInfo = (
  versions: VersionsState,
  versionId: number,
): Version | void => {
  return versions.versionInfo[versionId];
};
```

This change will help ensure that all components are working with mapped Redux props correctly.

willdurand:
Sentry issue: [ADDONS-CODE-MANAGER-DEV-1](https://sentry.prod.mozaws.net/operations/addons-code-manager-dev/issues/5469345/)

PR #503, opened 2019-03-29
Load a version into the store for the FileTree story

### LINKED ISSUE #502: FileTree storybook no longer expands / collapses
This is a follow-up to https://github.com/mozilla/addons-code-manager/pull/487

Open the storybook for `FileTree` and try to click to expand. It doesn't expand.

bobsilverberg:
Fixes #502 

This doesn't actually fix #502, but it's a start. There was a problem before that we were not loading a version into the store, so when `toggleExpandedPath` was dispatched it generated an error. With this patch I can see that `toggleExpandedPath` is succeeding, and that `expandedPaths` in the store is being updated when we click on a folder, but the component isn't responding to that change. So it looks like the `FileTree` component in the story isn't actually connected to the store, even though we're using `renderWithStoreAndRouter`.

Do either of you have any ideas about why that might be?

As an aside, while doing some of this work on `FileTree` I keep running into situations in which it seems to me that it would make more sense to have `FileTree` get its `version` via `mapStateToProps` rather than being passed a `version` via `props` from its parent. I wonder if that would also fix this case as well, and if that's really what we should be doing here?

### DIFF
--- a/src/components/FileTree/index.spec.tsx
+++ b/src/components/FileTree/index.spec.tsx
@@ -8,10 +8,12 @@ import configureStore from '../../configureStore';
 import {
   Version,
   actions as versionActions,
+  createInternalVersion,
   createInternalVersionEntry,
   getVersionInfo,
 } from '../../reducers/versions';
 import {
+  createFakeLogger,
   fakeVersion,
   fakeVersionEntry,
   shallowUntilTarget,
@@ -366,12 +368,13 @@ describe(__filename, () => {
     };
 
     const render = ({
+      _log = createFakeLogger(),
       onSelect = jest.fn(),
       store = configureStore(),
       version = getVersion({ store }),
     } = {}) => {
       return shallowUntilTarget(
-        <FileTree versionId={version.id} onSelect={onSelect} />,
+        <FileTree _log={_log} versionId={version.id} onSelect={onSelect} />,
         FileTreeBase,
         {
           shallowOptions: { context: { store } },
@@ -495,6 +498,17 @@ describe(__filename, () => {
 
       const isNodeExpanded = treeFold.prop('isNodeExpanded');
       expect(isNodeExpanded(node)).toBeFalsy();
+    });
+
+    it('logs a warning message when no version is loaded', () => {
+      const _log = createFakeLogger();
+
+      render({
+        _log,
+        version: createInternalVersion({ ...fakeVersion, id: 0 }),
+      });
+
+      expect(_log.warn).toHaveBeenCalled();
     });
   });
 });
--- a/src/components/FileTree/index.spec.tsx
+++ b/src/components/FileTree/index.spec.tsx
@@ -8,7 +8,6 @@ import configureStore from '../../configureStore';
 import {
   Version,
   actions as versionActions,
-  createInternalVersion,
   createInternalVersionEntry,
   getVersionInfo,
 } from '../../reducers/versions';
@@ -372,10 +371,10 @@ describe(__filename, () => {
       _log = createFakeLogger(),
       onSelect = jest.fn(),
       store = configureStore(),
-      version = getVersion({ store }),
+      versionId = 1234,
     } = {}) => {
       return shallowUntilTarget(
-        <FileTree _log={_log} versionId={version.id} onSelect={onSelect} />,
+        <FileTree _log={_log} versionId={versionId} onSelect={onSelect} />,
         FileTreeBase,
         {
           shallowOptions: { context: { store } },
@@ -387,7 +386,7 @@ describe(__filename, () => {
       const store = configureStore();
       const version = getVersion({ store });
 
-      const root = render({ store, version });
+      const root = render({ store, versionId: version.id });
 
       expect(root.find(ListGroup)).toHaveLength(1);
       expect(root.find(Treefold)).toHaveLength(1);
@@ -401,7 +400,7 @@ describe(__filename, () => {
       const version = getVersion({ store });
       const onSelect = jest.fn();
 
-      const root = render({ onSelect, store, version });
+      const root = render({ onSelect, store, versionId: version.id });
 
       const node = (root.instance() as FileTreeBase).renderNode(
         getTreefoldRenderProps(),
@@ -417,7 +416,7 @@ describe(__filename, () => {
       const store = configureStore();
       const version = getVersion({ store });
 
-      const root = render({ store, version });
+      const root = render({ store, versionId: version.id });
 
       const node = (root.instance() as FileTreeBase).renderNode(
         getTreefoldRenderProps(),
@@ -440,7 +439,7 @@ describe(__filename, () => {
 
       const dispatch = spyOn(store, 'dispatch');
 
-      const root = render({ store, version });
+      const root = render({ store, versionId: version.id });
 
       const treeFold = root.find(Treefold);
       expect(treeFold).toHaveProp('onToggleExpand');
@@ -474,7 +473,7 @@ describe(__filename, () => {
 
       version = getVersionInfo(store.getState().versions, version.id);
 
-      const root = render({ store, version });
+      const root = render({ store, versionId: version.id });
 
       const treeFold = root.find(Treefold);
       expect(treeFold).toHaveProp('isNodeExpanded');
@@ -492,7 +491,7 @@ describe(__filename, () => {
       };
       const version = getVersion({ store });
 
-      const root = render({ store, version });
+      const root = render({ store, versionId: version.id });
 
       const treeFold = root.find(Treefold);
       expect(treeFold).toHaveProp('isNodeExpanded');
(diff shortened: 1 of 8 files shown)


bobsilverberg:
Thanks @kumar303. I agree that the first solution you mention is probably the best. I'm going to make that change now before addressing any of the other work I'm doing on the FileTree as it might impact other things as well.

bobsilverberg:
I have updated `FileTree` to retrieve a version via `mapStateToProps` and that does indeed fix the Storybook issue.

src/components/FileTree/index.tsx
```diff
@@ -184,4 +192,17 @@ export class FileTreeBase extends React.Component<Props> {
...
 
-export default connect()(FileTreeBase) as React.ComponentType<PublicProps>;
+const mapStateToProps = (
+  state: ApplicationState,
+  ownProps: PublicProps,
+): PropsFromState => {
+  const { versionId } = ownProps;
+
+  return {
+    version: getVersionInfo(state.versions, versionId),
```
kumar303: The return type of `getVersionInfo` [is wrong](https://github.com/mozilla/addons-code-manager/issues/442) so your component needs to account for the fact that it can be `void` (when not loaded).

Introducing the logic of fetching `version` from state means that `FileTree` *should* be responsible for loading that version into state when needed. This would be the only way to achieve true encapsulation and avoid confusion in future refactoring. In your current patch, `FileTree` is tightly coupled to the components that know how to load versions, which are [`Browse`](https://github.com/mozilla/addons-code-manager/blob/ee94f3282cd30df66a634f49bda895121fedb8a4/src/pages/Browse/index.tsx#L58) or [`Compare`](https://github.com/mozilla/addons-code-manager/blob/ee94f3282cd30df66a634f49bda895121fedb8a4/src/pages/Compare/index.tsx#L87). Before your patch, when `version` was passed as a direct prop, the coupling was obvious. Now it is no longer obvious and that can lead to future bugs and confusion.

So, what to do? The simplest thing is to keep passing in `version` because it makes the coupling obvious. I suggested not doing this because it made the story code more complicated.

The next options are:

1. Introduce loading code in `componentDidMount` for when `version` doesn't exist in state. This loading code would probably never get executed in real life but at least it would be there for when we need it.

2. We probably won't ever need the loading code and loading code adds complexity so I think it would be fine to log a warning to the console when `version` does not exist in state. This log message would help in debugging a story, test, or a future refactoring.

This was a long way of saying "just add a log message" but I wanted to explain why it is a problem and why we should strive to avoid this kind of pattern whenever possible.

Also, when we switch to hooks, we could probably make a shared hook based on [`useEffect`](https://reactjs.org/docs/hooks-reference.html#useeffect) like `useVersionLoader` and fix it The Right Way ™.

stories/FileTree.stories.tsx
```diff
@@ -60,21 +64,26 @@ const version = createInternalVersion({
...
+
+  const internalVersion = getVersionInfo(
+    store.getState().versions,
+    version.id,
+  ) as Version;
+
   const props: FileTreeProps = {
     onSelect: onSelectFile,
-    version,
+    versionId: internalVersion.id,
```
kumar303: Why do you need the full `internalVersion` object? Can't it just be `versionId: version.id`?

bobsilverberg: Ah, you're right, we don't need the full `Version` object in the story anymore. Thanks!

src/components/FileTree/index.tsx
bobsilverberg: This is all good information, thanks @kumar303. It does sound like the `FileTree` should be responsible for fetching its own `Version` if needed, but, as you report, we know that the `Browse` and `Compare` components will have already done that, so with the current set up we'll never need to do it.

I'll go ahead and add the warning when a version does not exist.

bobsilverberg: Ok, I've added the warning. Do you think we need a test for the warning as well, or are we happy to just have the code there, as we never expect it to be triggered anyway?

Also, should I change the type of the `version` prop to `Version | void`? I can do that, but then it also requires some extra code in the component to deal with the case where `version` is `null`. For example, we'd need to deal with it in `renderNode`.

src/components/FileTree/index.tsx
```diff
@@ -134,10 +139,14 @@ export const buildFileTree = (
 
 export type PublicProps = {
   onSelect: FileTreeNodeProps['onSelect'];
+  versionId: number;
+};
+
+type PropsFromState = {
   version: Version;
```
kumar303: In regards to making this `Version | void`, I don't think it's important for this patch. It will make more work for https://github.com/mozilla/addons-code-manager/issues/442 but that's fine.

src/components/FileTree/index.tsx
```diff
@@ -184,4 +193,26 @@ export class FileTreeBase extends React.Component<Props> {
...
+): PropsFromState => {
+  const { versionId } = ownProps;
+  const version = getVersionInfo(state.versions, versionId);
+
+  if (!version) {
+    // This should never happen as we are fetching the version in all of the
+    // parents on this component and only rendering the FileTree if we have a
+    // version, but let's log a warning in case we encounter a case where this
+    // does happen.
+    log.warn(`No version was loaded for version: `, versionId);
```
kumar303: You asked if we should add a test for this. Since we use TypeScript, I'm not worried about it breaking so a test is not very helpful. However, it decreases our code coverage score which is a bit annoying. I guess you could add a test just to keep code coverage up? Meh.

bobsilverberg:
@kumar303 I made the changes we just discussed in IRC. Let me know what you think.

bobsilverberg:
@kumar303 Just a reminder to take a final look at this, as I'm not terribly happy with what I had to do to address the `null version` state. It may be the best we can do right now, but I'd like your opinion before landing it.

src/components/FileTree/index.tsx
```diff
@@ -134,38 +139,65 @@ export const buildFileTree = (
...
+
   renderNode = (props: TreefoldRenderPropsForFileTree) => {
     const { onSelect, version } = this.props;
 
-    return <FileTreeNode {...props} onSelect={onSelect} version={version} />;
+    if (version) {
+      return <FileTreeNode {...props} onSelect={onSelect} version={version} />;
+    }
+    // This should never happen. See the comment in mapStateToProps.
+    return null;
```
kumar303: You can return `<Loading>` here and I don't think you need a comment. We would need a state like this anyway if the component knew how to load its own data.

src/components/FileTree/index.tsx
```diff
@@ -134,38 +139,65 @@ export const buildFileTree = (
...
   onToggleExpand = (node: TreeNode) => {
     const { dispatch, version } = this.props;
 
-    dispatch(
-      versionsActions.toggleExpandedPath({
-        path: node.id,
-        versionId: version.id,
-      }),
-    );
+    if (version) {
```
kumar303: I'd rather see something like:

```js
if (!version) {
  throw new Error('Cannot toggle expanded path without a version');
}
```

That should be easy to test

src/components/FileTree/index.tsx
```diff
@@ -134,38 +139,65 @@ export const buildFileTree = (
...
+        }),
+      );
+    }
   };
 
   isNodeExpanded = (node: TreeNode) => {
     const { version } = this.props;
 
-    return version.expandedPaths.includes(node.id);
+    if (version) {
```
kumar303: Same thing here. Something like:

```tsx
if (!version) {
  throw new Error('Cannot check if node is expanded without a version');
}
```

src/components/FileTree/index.tsx
```diff
@@ -134,38 +139,65 @@ export const buildFileTree = (
...
+    // This should never happen. See the comment in mapStateToProps.
+    return false;
   };
 
   render() {
     const { version } = this.props;
 
+    if (!version) {
+      // This should never happen. See the comment in mapStateToProps.
+      return null;
```
kumar303: This should render `<Loading>`

bobsilverberg:
Thanks @kumar303, I made all those changes and added tests for all of them. Did you want to take a final look before I land it?

src/components/FileTree/index.tsx
```diff
@@ -134,21 +140,40 @@ export const buildFileTree = (
...
+  };
+
   renderNode = (props: TreefoldRenderPropsForFileTree) => {
     const { onSelect, version } = this.props;
 
-    return <FileTreeNode {...props} onSelect={onSelect} version={version} />;
+    if (version) {
+      return <FileTreeNode {...props} onSelect={onSelect} version={version} />;
+    }
+    return <Loading message={gettext('Loading version...')} />;
```
kumar303: This doesn't appear to be tested. Can you add one for it?

src/components/FileTree/index.spec.tsx
```diff
@@ -496,5 +500,58 @@ describe(__filename, () => {
       const isNodeExpanded = treeFold.prop('isNodeExpanded');
       expect(isNodeExpanded(node)).toBeFalsy();
     });
+
+    it('logs a warning message when no version is loaded', () => {
+      const _log = createFakeLogger();
+
+      render({
+        _log,
+        version: createInternalVersion({ ...fakeVersion, id: 0 }),
```
kumar303: This could probably just link to a non-existent version like `versionId: 1234`

src/components/FileTree/index.spec.tsx
```diff
@@ -496,5 +500,58 @@ describe(__filename, () => {
...
+      });
+
+      expect(_log.warn).toHaveBeenCalled();
+    });
+
+    it('renders a Loading component when no version is loaded', () => {
+      const root = render({
+        version: createInternalVersion({ ...fakeVersion, id: 0 }),
```
kumar303: The `id: 0` bit is a little confusing. Could you use `versionId: 1234` in these other tests, too?

bobsilverberg: Yes, I cleaned up the `render` function to accept just a `versionId` and it's much better now.

PR #505, opened 2019-03-29
Pass absolute url to api login (fxa)

### LINKED ISSUE #486: `LoginButton` should pass an absolute URL in the FxA URL
We pass `location.pathname` to the API when initiating the login dance in the [`LoginButton` component](https://github.com/mozilla/addons-code-manager/blob/c6cad11653b45538292c5d2fd5b65033adb65b11/src/components/LoginButton/index.tsx) but this won't work.

First, `location.pathname` is a relative URL. Second, the API is hosted on a domain that is not the same as this app (`addons-dev.allizom.org` vs `code.addons-dev.allizom.org`). Passing a relative URL in the `to` query parameter works when both the API and the app are on the same domain because the API returns a redirect with the relative URL, but in the context of code-manager, we should pass an absolute URL.

willdurand:
Fixes #486

---

This is needed because the API (which sends the redirect after FxA login page) is not on the same domain, hence a relative URL cannot work. ~~This is not working yet because of an API issue (cf. https://github.com/mozilla/addons-server/issues/11018)~~.

When we use the `local` FxA config, we still want relative URLs because we are using the -dev API for local dev and the API will not allow `localhost:3000` as a valid domain (valid domains are: `code.addons-dev.allizom.org` or `addons-dev.allizom.org`). If we're using the API locally too, then it would recognize `localhost:3000` or `olympia.test` as valid domains, but that would work with relative URLs too so it is much simpler to always pass a relative URL when FxA config is `local`. This also means: the behavior before and after this patch remains the same locally (but it fixes the login issue on https://code.addons-dev.allizom.org).

### DIFF
--- a/src/components/LoginButton/index.tsx
+++ b/src/components/LoginButton/index.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react';
 import { Button } from 'react-bootstrap';
-import { withRouter, RouteComponentProps } from 'react-router-dom';
 
 import { makeApiURL } from '../../api';
 import { gettext } from '../../utils';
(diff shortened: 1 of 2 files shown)


src/components/LoginButton/index.tsx
```diff
@@ -38,6 +43,6 @@ export class LoginButtonBase extends React.Component<Props> {
   }
 }
 
-export default withRouter(LoginButtonBase) as React.ComponentType<
+export default LoginButtonBase as React.ComponentType<
```
kumar303: You can remove the type assertion now 🎉

PR #510, opened 2019-04-01
Add ops endpoints

### LINKED ISSUE #428: standard endpoints for ops to onboard the app

In order for ops to onboard an application, we need the app to provide a few standard endpoints, namely,

* `/__version__`
* `/__heartbeat__`
* `/__lbheartbeat__`

Please see https://mana.mozilla.org/wiki/display/SVCOPS/New+Service+Guide for the specifics of these endpoints.

### What happened?

n/a

The app provides the above mentioned 3 endpoints.

Please let us know if there are questions.

willdurand:
Fixes #428 

---

For the heartbeat endpoint, we do not depend on any service (except the API that has its own heartbeat endpoint already) so that's why both heartbeat and lbheartbeat are the same.

### DIFF
--- a/src/server/index.spec.tsx
+++ b/src/server/index.spec.tsx
@@ -1,4 +1,5 @@
 /* eslint @typescript-eslint/no-object-literal-type-assertion: 0 */
+import fs from 'fs';
 import http from 'http';
 import path from 'path';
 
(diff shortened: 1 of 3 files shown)


src/server/index.spec.tsx
```diff
@@ -341,6 +342,46 @@ describe(__filename, () => {
...
+        const response = await server.get('/__version__');
+
+        expect(response.status).toEqual(200);
+        expect(response.header).toHaveProperty(
+          'content-type',
+          'application/json; charset=UTF-8',
+        );
+
+        // This file is in `src/server/fixtures/`.
+        const content = fs.readFileSync(path.join(rootPath, 'version.json'), {
```
kumar303: if you renamed `rootPath` to `fixturesPath` then I think this would be more clear and you wouldn't need a comment about it.

src/server/index.tsx
```diff
@@ -158,6 +158,16 @@ export const createServer = ({
   // Return 200 for csp reports - this will need to be overridden when deployed.
   app.post('/__cspreport__', (req, res) => res.status(200).end('ok'));
 
+  app.get('/__version__', (req, res) => {
+    const versionFile = path.join(rootPath, 'version.json');
+
+    return res.sendFile(versionFile);
+  });
+
+  app.get(['/__heartbeat__', '/__lbheartbeat__'], (req, res) => {
```
kumar303: The spec [says](https://mana.mozilla.org/wiki/display/SVCOPS/New+Service+Guide#NewServiceGuide-/__heartbeat__) that `__heartbeat__` should check dependent services. Is there an API endpoint we can check? We definitely depend on the API. This could be done in a follow-up patch if that's easier.

src/server/index.spec.tsx
willdurand: Yes, I thought about it too tbh. I left it because it is the name of the argument in the `createServer()` function but that's confusing.

src/server/index.tsx
willdurand: I made a wrong assumption in my PR description. I cannot find the API endpoint. I naively thought it would be http://addons.mozilla.org/__heartbeat__, but that does not work, heh.

Otherwise I would have suggested the following: we could maybe proxy the request to `https://<api>/__heartbeat__`?

willdurand: So, AMO does not have these endpoints (yet?). I am not sure which API endpoint we could use to check whether the API is healthy, so I'd like to defer than in another issue: https://github.com/mozilla/addons-code-manager/issues/524

bqbn: I think it's OK for now, at least from ops perspective. We'll be using the __lbheartbeat__ to start with.

PR #521, opened 2019-04-02
Use existing CSP report URLs

### LINKED ISSUE #423: Ensure CSP report endpoint is setup to consume reports
There's a no-op CSP report endpoint configured on the app, this should be superseded by a real CSP report endpoint setup by ops that feeds the actual security logs.

Endpoint is `/__cspreport__`

willdurand:
Fixes #423

---

Use real CSP report URLs instead of the no-op one. In local dev, it should still be the no-op endpoint (`/__cspreport__`) but one can build and serve the app (make sure `.env.local` does not contain `REACT_APP_USE_INSECURE_PROXY=true`, then run `yarn build && yarn serve`) to see absolute CSP report URLs.

### DIFF
--- a/src/server/index.spec.tsx
+++ b/src/server/index.spec.tsx
@@ -417,6 +417,13 @@ describe(__filename, () => {
           const policy = cspParser(response.header['content-security-policy']);
           expect(policy['connect-src']).toEqual(["'self'"]);
         });
+
+        it('uses a relative for the CSP report URI', async () => {
+          const response = await server.get('/');
+
+          const policy = cspParser(response.header['content-security-policy']);
+          expect(policy['report-uri']).toEqual(['/__cspreport__']);
+        });
       });
     });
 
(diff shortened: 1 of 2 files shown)


src/server/index.spec.tsx
```diff
@@ -191,8 +195,13 @@ describe(__filename, () => {
         expect(staticPolicy['script-src']).toEqual(undefined);
         expect(staticPolicy['style-src']).toEqual(undefined);
         expect(staticPolicy['worker-src']).toEqual(undefined);
+      });
+
+      it('exposes a default no-op CSP report endpoint', async () => {
+        const server = request(
+          createServer({ env: prodEnv as ServerEnvVars, rootPath }),
```
kumar303: The value of `REACT_APP_USE_INSECURE_PROXY` seems crucial to this test so I suggest setting it explicitly rather than relying on defaults.

willdurand: I double-checked and it appears that the server responds to this endpoint in any case, but you're right, there seems to be a missing test case for when `REACT_APP_USE_INSECURE_PROXY` is set to `true`.

PR #525, opened 2019-04-02
Add credentials: 'include' to logout call

### LINKED ISSUE #515: Logout does not seem to work in -dev

1. Go to https://code.addons-dev.allizom.org/
2. Make sure you're logged in
3. Click the "Logout" button
4. Refresh the page

### What happened?

You appear logged in again. Logout did not work.

You should be successfully logged out.

The API call succeeds. It returns the `set-cookie` headers to clear the cookies for the user on FF, Chrome does not display them though, so that might be an hint. Not sure why it is not working though. I believe it has something to do with having 2 different domains or CORS or something in between.

It works in local dev.

willdurand:
Fixes #515 

---

I don't know if a new parameter to `callApi` is what we want because only one endpoint will accept credentials and it will fail in all other cases, but it is simpler than re-implementing `callApi` in `logOutFromServer`.

This patch requires https://github.com/mozilla/addons-server/issues/11112 and I want to live-test this change in Chrome before landing this patch, but a quick review is always welcomed :)

### DIFF
--- a/src/api/index.spec.tsx
+++ b/src/api/index.spec.tsx
@@ -189,6 +189,19 @@ describe(__filename, () => {
         expect.any(Object),
       );
     });
+
+    it('accepts credentials', async () => {
+      const credentials = 'include';
+
+      await callApiWithDefaultApiState({ endpoint: '/url', credentials });
+
+      expect(fetch).toHaveBeenCalledWith(
+        expect.any(String),
+        expect.objectContaining({
+          credentials,
+        }),
+      );
+    });
   });
 
   describe('isErrorResponse', () => {
@@ -267,6 +280,7 @@ describe(__filename, () => {
       expect(fetch).toHaveBeenCalledWith(
         expect.stringMatching(`/api/${defaultVersion}/accounts/session/`),
         {
+          credentials: 'include',
           headers: {},
           method: HttpMethod.DELETE,
         },
--- a/src/api/index.spec.tsx
+++ b/src/api/index.spec.tsx
@@ -203,6 +203,20 @@ describe(__filename, () => {
         }),
       );
     });
+
+    it('can omit credentials', async () => {
+      await callApiWithDefaultApiState({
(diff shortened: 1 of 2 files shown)


src/api/index.tsx
```diff
@@ -67,6 +67,7 @@ export enum HttpMethod {
 
 type CallApiParams = {
   apiState: ApiState;
+  credentials?: 'omit' | 'same-origin' | 'include';
```
kumar303: I would rather see this as `includeCredentials: boolean` which would only pass `credentials: 'include'` to `fetch()` when `true`. I just don't see us ever needing the other values and I'd rather not introduce them as footguns. We would need to carefully consider the use of `omit` or `same-origin`.

willdurand: Yep, that works for me.

src/api/index.spec.tsx
```diff
@@ -189,6 +189,20 @@ describe(__filename, () => {
         expect.any(Object),
       );
     });
+
+    it('can include credentials', async () => {
+      await callApiWithDefaultApiState({
+        endpoint: '/url',
+        includeCredentials: true,
```
kumar303: It would be nice to get an additional test to make sure `credentials` is undefined without this.

src/api/index.tsx
```diff
@@ -191,6 +194,7 @@ export const getVersionsList = async ({
 export const logOutFromServer = async (apiState: ApiState) => {
   return callApi<{}>({
     apiState,
+    includeCredentials: true,
```
kumar303: Maybe it's just me but a comment would help explain why this is needed. It wasn't obvious to me before I researched this that you would need to *send* cookies to the server in order for the response to delete those cookies. I still don't know if this is the real problem, heh. We'll see.

willdurand:
I just tried a local override in chrome and logout works when sending the credentials. This patch should work then.

Issue #492, opened 2019-03-28
The Browse component does not react to Redux state correctly


kumar303:
### Describe the problem and steps to reproduce it:

* Open Redux Devtools
* Sign in
* Load http://localhost:3000/en-US/browse/494431/versions/1532144/
* Click `AndroidADBPlus` to expand the file list
* Click the `ext` folder to expand it
* Click `background.js` to load the file
* In the Redux Devtool, jump back to the `LOAD_VERSION_FILE` action. This loads the default file, `manifest.json`.
* Jump forward to the `UPDATE_SELECTED_PATH` action which had `{ selectedPath: "ext/background.js" }` in its payload

### What happened?

You see an infinite loading spinner

### What did you expect to happen?

The contents of `ext/background.js` should load

### Anything else we should know?

This implementation **will block https://github.com/mozilla/addons-code-manager/issues/199** because it's currently not possible to change the selected file and trigger a re-fetch in `Browse`. 

~~Additionally, the current implementation will result in us not being able to use the Redux Devtool for development on this page and it will also prevent us from using a stack of actions to replay user interaction.~~ (that's actually still possible)

The root cause is that `Browse` [only reacts to user events](https://github.com/mozilla/addons-code-manager/blob/1c9c9369d43fe3d0f4352567d9d2116db2f8669b/src/pages/Browse/index.tsx#L53-L76). It should react to Redux state changes instead. See `Compare` for [how it should work](https://github.com/mozilla/addons-code-manager/blob/1c9c9369d43fe3d0f4352567d9d2116db2f8669b/src/pages/Compare/index.tsx#L69-L95).

PR #530, opened 2019-04-03
Prepare Browse to handle keyboard navigation

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/492

This adds some complexity but it's necessary to support keyboard navigation (and other things). Instead of reacting to user events, the `Browse` component now reacts to state changes to know when to load files.


src/pages/Browse/index.spec.tsx
```diff
@@ -84,6 +85,57 @@ describe(__filename, () => {
...
+    const fakeThunk = createFakeThunk();
+    const _fetchVersionFile = fakeThunk.createThunk;
+
+    return {
+      _fetchVersionFile,
+      addonId,
+      fakeThunk,
+      store,
+      updateProps: (props = {}) => {
+        const dispatchSpy = spyOn(store, 'dispatch');
```
willdurand: We tend to use `dispatch` without the `Spy` suffix.

src/pages/Browse/index.spec.tsx
```diff
@@ -84,6 +85,57 @@ describe(__filename, () => {
...
+
+    const fakeThunk = createFakeThunk();
+    const _fetchVersionFile = fakeThunk.createThunk;
+
+    return {
+      _fetchVersionFile,
+      addonId,
+      fakeThunk,
+      store,
+      updateProps: (props = {}) => {
```
willdurand: It would be maybe a bit more explicit to name this function `renderAndUpdateWithProps` or something like that.

src/reducers/versions.spec.tsx
```diff
@@ -428,7 +431,8 @@ describe(__filename, () => {
     it('returns a version file', () => {
       const downloadURL = 'http://example.org/download/file';
       const mimeType = 'mime/type';
-      const path = 'test.js';
+      const filename = 'test.js';
+      const path = 'some/dir/test.js';
```
willdurand: We could write:

```js
const path = `some/dir/${filename}`;
```

src/pages/Browse/index.spec.tsx
kumar303: In this case, I think it would be confusing to call this variable `dispatch` because one might think you could dispatch actions with it. In other cases it's probably OK but when using this helper, it's not obvious that this is a spy object (the `spyOn()` call is hidden).

PR #532, opened 2019-04-04
Handle absolute URLs coming from the API

### LINKED ISSUE #531: Validation URL is now absolute + validation/download URLs should be proxied when insecure proxy is used
All URLs to something external to the API, including Django views that are currently served under the same domain as the API will be returned as absolute URLs by addons-server. Therefore, we need to make a change to our code once https://github.com/mozilla/addons-server/issues/11034 lands.

We have to change the way we construct the API URL for retrieving the linter messages given that we currently assume it is a relative URL. This change will fix the issue in -dev (and stage, prod).

Error in -dev:

> Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://addons-dev.allizom.org/https://addons-dev.allizom.org/en-US/developers/addon/androidadbplus2/file/252861/validation.json. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing).

In addition to that, we should remove the base URL of the URLs returned by the API so that requests are proxied _via_ the insecure proxy in local dev. This will allow us to fetch validation results (like before) but also to download files without any authentication problem (currently broken in local dev).

Error in local dev:

> Got status 404 for URL https://addons-dev.allizom.org/en-US/developers/addon/androidadbplus2/file/252861/validation.json

willdurand:
Fixes mozilla/addons-server#531

---

Without this patch in local dev, it is not possible to download a file (authentication issue). Linter messages won't be retrieved anymore either because of https://github.com/mozilla/addons/issues/6512. In -dev, linter messages cannot be retrieved anymore either.

This patch fixes all these issues.

### DIFF
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -26,33 +26,53 @@ export const getApiHost = ({
 };
 
 type MakApiUrlParams = {
-  _getApiHost?: typeof getApiHost;
-  version?: string | null;
-  path: string;
+  apiHost?: string | null;
+  path?: string;
   prefix?: string | null;
+  url?: string;
+  useInsecureProxy?: boolean;
+  version?: string | null;
 };
 
 export const makeApiURL = ({
-  _getApiHost = getApiHost,
+  apiHost = process.env.REACT_APP_API_HOST,
   path,
   prefix = 'api',
+  url,
+  useInsecureProxy = process.env.REACT_APP_USE_INSECURE_PROXY === 'true',
   version = process.env.REACT_APP_DEFAULT_API_VERSION,
 }: MakApiUrlParams) => {
-  const parts = [_getApiHost()];
-
-  if (prefix) {
-    parts.push(`/${prefix}`);
+  if (path && url) {
+    throw new Error('Cannot receive both `path` and `url` parameters.');
   }
 
-  if (version) {
-    parts.push(`/${version}`);
-  }
+  const parts = [];
 
-  let adjustedPath = path;
-  if (!path.startsWith('/')) {
-    adjustedPath = `/${adjustedPath}`;
+  if (path) {
+    parts.push(getApiHost({ apiHost, useInsecureProxy }));
+
+    if (prefix) {
+      parts.push(`/${prefix}`);
+    }
+
+    if (version) {
+      parts.push(`/${version}`);
+    }
+
+    let adjustedPath = path;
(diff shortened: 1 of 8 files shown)


src/api/index.tsx
```diff
@@ -26,33 +26,53 @@ export const getApiHost = ({
...
+    if (prefix) {
+      parts.push(`/${prefix}`);
+    }
+
+    if (version) {
+      parts.push(`/${version}`);
+    }
+
+    let adjustedPath = path;
+    if (!path.startsWith('/')) {
```
kumar303: It might be safer to make this `!adjustedPath.startsWith('/')` in case some future refactoring moves this code around some more

src/components/FileMetadata/index.tsx
```diff
@@ -28,7 +29,9 @@ const FileMetadataBase = ({ file }: PublicProps) => {
           <React.Fragment>
             <dt>{gettext('Download link')}</dt>
             <dd className={styles.downloadURL}>
-              <a href={file.downloadURL}>{file.filename}</a>
+              <a href={makeApiURL({ url: file.downloadURL })}>
```
kumar303: When I download `popup.js` from http://localhost:3000/en-US/browse/502955/versions/1541786/ I get a blank file. The network panel says it's a 200 but the response is empty there, too. Is this a known issue?

kumar303 (approved):
r+wc, thanks for fixing this so quickly. I noticed it because local development was broken for me.

I see an issue with the download link but I'm not sure if that's related to your patch or not. Is it working for you? Maybe the proxy is messing with it but I can't think of how.

src/components/FileMetadata/index.tsx
willdurand: Yeah, it looks like you're not the only one: https://github.com/mozilla/addons-server/issues/11118#issuecomment-479914017

src/api/index.tsx
willdurand: ah right, thanks!

src/components/FileMetadata/index.tsx
willdurand: I don't know if the issue comes from code-manager or the API, so I am currently updating my addons-server local dev env to see if the API replies correctly.

willdurand: This looks like a separate bug because I can reproduce without code-manager (i.e. using httpie only)

Issue #534, opened 2019-04-04
An error occurs when trying to change the "new" version in the Compare page


willdurand:
### Describe the problem and steps to reproduce it:

1. Go to https://code.addons-dev.allizom.org/en-US/compare/502955/versions/1541794...1541798/
2. Select a "new" version, e.g., `1.4.0`

### What happened?

It crashes.

### What did you expect to happen?

It does not crash.

### Anything else we should know?

Error in local dev is:

```
TypeError: version is undefined
render
src/components/DiffView/index.tsx:205

  202 | 
  203 | return (
  204 |   <LinterProvider
> 205 |     versionId={version.id}
      | ^  206 |     validationURL={version.validationURL}
  207 |     selectedPath={version.selectedPath}
  208 |   >
```

Selecting another "old" version seems to work.

willdurand:
This is likely related to https://github.com/mozilla/addons-code-manager/issues/442

AlexandraMoga:
I haven't run into error with the compare file component. I've also tried with the following example:
https://code.addons-dev.allizom.org/en-US/compare/616548/versions/1655414...1655415/

PR #535, opened 2019-04-04
Filter old/new versions depending on the selected values

### LINKED ISSUE #415: "New" version list should list versions newer than the currently selected "old" version

1. Open a compare view: http://localhost:3000/en-US/compare/502955/versions/1541794...1541798/
2. Open the "Choose a new version" dropdown

### What happened?

The list contains all the versions including `1.3.0`, which is lower than the selected "old" version (`1.4.0`).

The list only contains "newer" versions.

I don't know whether we want to list the version that is equal to the selected old version or only strictly newer versions.

willdurand:
Fixes #415

---

Comparisons are strict because I am not sure we want to compare the same old/new version but it is easy to change if needed.

### DIFF
--- a/src/components/VersionChooser/index.tsx
+++ b/src/components/VersionChooser/index.tsx
@@ -15,12 +15,22 @@ import {
 import { gettext } from '../../utils';
 import styles from './styles.module.scss';
 
+export const higherVersionsThan = (versionId: string) => {
+  return (version: VersionsListItem) => version.id > parseInt(versionId, 10);
+};
+
+export const lowerVersionsThan = (versionId: string) => {
+  return (version: VersionsListItem) => version.id < parseInt(versionId, 10);
+};
+
 export type PublicProps = {
   addonId: number;
 };
 
 export type DefaultProps = {
   _fetchVersionsList: typeof fetchVersionsList;
+  _higherVersionsThan: typeof higherVersionsThan;
+  _lowerVersionsThan: typeof lowerVersionsThan;
 };
 
 type PropsFromState = {
@@ -41,17 +51,11 @@ type Props = PublicProps &
   PropsFromState &
   RouterProps;
 
-export const higherVersionsThan = (versionId: string) => {
-  return (version: VersionsListItem) => version.id > parseInt(versionId, 10);
-};
-
-export const lowerVersionsThan = (versionId: string) => {
-  return (version: VersionsListItem) => version.id < parseInt(versionId, 10);
-};
-
 export class VersionChooserBase extends React.Component<Props> {
   static defaultProps: DefaultProps = {
     _fetchVersionsList: fetchVersionsList,
+    _higherVersionsThan: higherVersionsThan,
+    _lowerVersionsThan: lowerVersionsThan,
   };
 
   componentDidMount() {
@@ -88,7 +92,12 @@ export class VersionChooserBase extends React.Component<Props> {
   };
 
   render() {
-    const { match, versionsMap } = this.props;
+    const {
+      _higherVersionsThan,
+      _lowerVersionsThan,
+      match,
+      versionsMap,
+    } = this.props;
     const { baseVersionId, headVersionId } = match.params;
 
     return (
@@ -104,27 +113,21 @@ export class VersionChooserBase extends React.Component<Props> {
             <Form.Row>
               <VersionSelect
                 className={styles.baseVersionSelect}
+                isSelectable={_lowerVersionsThan(headVersionId)}
                 label={gettext('Choose an old version')}
-                listedVersions={versionsMap.listed.filter(
-                  lowerVersionsThan(headVersionId),
-                )}
+                listedVersions={versionsMap.listed}
                 onChange={this.onOldVersionChange}
-                unlistedVersions={versionsMap.unlisted.filter(
-                  lowerVersionsThan(headVersionId),
-                )}
+                unlistedVersions={versionsMap.unlisted}
                 value={baseVersionId}
               />
 
               <VersionSelect
                 className={styles.headVersionSelect}
+                isSelectable={_higherVersionsThan(baseVersionId)}
                 label={gettext('Choose a new version')}
-                listedVersions={versionsMap.listed.filter(
-                  higherVersionsThan(baseVersionId),
-                )}
+                listedVersions={versionsMap.listed}
                 onChange={this.onNewVersionChange}
-                unlistedVersions={versionsMap.unlisted.filter(
-                  higherVersionsThan(baseVersionId),
-                )}
+                unlistedVersions={versionsMap.unlisted}
                 value={headVersionId}
                 withLeftArrow
               />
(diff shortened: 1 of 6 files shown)


src/components/VersionChooser/index.tsx
```diff
@@ -36,6 +40,22 @@ type Props = PublicProps &
   PropsFromState &
   RouterProps;
 
+export const higherVersionsThan = (versionId: string) => {
+  if (!versionId) {
+    return Boolean;
```
willdurand: This is only needed for Storybook (otherwise we won't have any value). I don't know if we want that, but on the other hand, it won't hurt.

src/components/VersionChooser/index.tsx
```diff
@@ -36,6 +40,22 @@ type Props = PublicProps &
...
+  }
+
+  return (version: VersionsListItem) => version.id > parseInt(versionId, 10);
+};
+
+export const lowerVersionsThan = (versionId: string) => {
+  if (!versionId) {
+    return Boolean;
```
willdurand: This is only needed for Storybook (otherwise we won't have any value). I don't know if we want that, but on the other hand, it won't hurt.

src/components/VersionChooser/index.tsx
```diff
@@ -36,6 +40,22 @@ type Props = PublicProps &
   PropsFromState &
   RouterProps;
 
+export const higherVersionsThan = (versionId: string) => {
+  if (!versionId) {
```
kumar303: How about `if (versionId === '')` ? I think it could also benefit from a comment saying that the source is a query parameter.

willdurand: I will try. I am not sure it will fix the storybook issue though, which is the main point of having this check. Otherwise, we should never have an empty string, everything else would break before :D

willdurand: Na, it does not work as intended. I see why you asked for `''` now. The test case is another issue: TS does not want to let me call this function with `undefined` or `null`, only an empty string works here.

As said in my 3 previous comments above, this is mainly for Storybook, which does not pass any routing params, and I don't really know how to simulate that (other than using the `Route` component). Passing `match` to the component does not work.

kumar303: I get why you need this but it indicates a flawed design. It's fine as is but the problem can be fixed in two ways. These are not change requests, just patterns to consider for future scenarios.

1. We could give storybook a utility that creates fake router locations to simulate data normally extracted from the URL. We probably have some of this already in the test helper.
2. We could make parameters such as `baseVersionId` required (which they are) and wrap the component in a helper like this:

```tsx
const VersionChooserFromRouter = (props) => {
  const { match, ...otherProps } = props;
  const { baseVersionId, headVersionId } = match.params;
  return (
    <VersionChooser
      baseVersionId={baseVersionId}
      headVersionId={headVersionId}
      {...otherProps}
    />
  );
};

export default withRouter(VersionChooserFromRouter);
```
The storybook code would then have an easier time rendering `<VersionChooser />`

willdurand:
> It doesn't happen on master but maybe this just needs a rebase? I know we fixed something related.

it's indeed fixed in `master` and how I first discovered #534

src/components/VersionChooser/index.tsx
willdurand: > 1. We could give storybook a utility that creates fake router locations to simulate data normally extracted from the URL. We probably have some of this already in the test helper.

That's unfortunately hard. I tried and there is nothing we can do to make everyone (React Router, TS, etc) happy it seems. It works in the test suite because we're not using a `Router` component. Wrapping the component into a `Route` would do the trick.

> 2. We could make parameters such as baseVersionId required (which they are) and wrap the component in a helper like this

Even more code to write \o/ That's probably the best solutions (of the 2), but writing such a component _just_ because Storybook isn't React Router friendly is... not fun.

I think we should make Storybook more React Router friendly, and that would mean adopting something like https://www.npmjs.com/package/storybook-react-router.

willdurand:
Rebased ✨

src/components/VersionChooser/index.tsx
willdurand: I guess it also means it's time to fix this issue between Storybook and React Router, but maybe in a follow-up?

kumar303: > Even more code to write \o/ 

Ha, well, it was just an example. It could probably be made generic for all components like this:

```tsx
<RouterProvider>
  {(match) => (
      <VersionChooser
        baseVersionId={match.params.baseVersionId}
        headVersionId={match.params.headVersionId}
        {...otherProps}
      />
  )}
</RouterProvider>
```

> ...but writing such a component just because Storybook isn't React Router friendly is... not fun.

It highlights that the component is coupled to router in an inflexible way but, yeah, wrapping it in another component adds more complexity.

In any case, I'm not sure if we should file a follow-up or not. It's something to consider for the next component that needs a storybook *and* URL parameters.

willdurand: This makes me think that we could maybe export the component without the router HOC before the `export default`.

willdurand: I took this approach by creating a `ConnectedVersionChooser`, which accepts a `match` prop.

willdurand:
> The old site just disabled the unselectable versions which makes it more clear IMO. Is there a reason you didn't go that route?

Good question, initially I did not really want to show unselectable options to the user but thinking about it again, it's probably better.

src/components/VersionChooser/index.tsx
```diff
@@ -132,6 +155,10 @@ const mapStateToProps = (
   };
 };
 
+export const ConnectedVersionChooser = connect(mapStateToProps)(
```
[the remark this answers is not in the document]
willdurand: awww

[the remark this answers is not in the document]
willdurand: I pushed a change that fixes the issue and guarantees only expected props are passed in Storybook. The rest of the code won't be affected.

src/components/VersionChooser/index.tsx
```diff
@@ -132,6 +155,19 @@ const mapStateToProps = (
   };
 };
 
+const ConnectedVersionChooser = connect(mapStateToProps)(VersionChooserBase);
+
+export const VersionChooserWithoutRouter = ConnectedVersionChooser as React.ComponentType<
+  PublicProps &
+    Partial<
+      DefaultProps & {
+        match: {
```
kumar303: It would be confusing to understand why this type is declared because it's not necessary for normal functionality. A comment would help.

src/components/VersionSelect/index.spec.tsx
```diff
@@ -129,4 +139,29 @@ describe(__filename, () => {
...
+      {
+        ...fakeVersionsListItem,
+        id: 456,
+        channel: 'unlisted',
+        version: 'v2',
+      },
+    ];
+
+    const root = render({
+      isSelectable: () => false,
```
kumar303: It would be helpful to add another test that makes sure the right `version` object gets passed to `isSelectable()`. It should also test multiple versions where one gets disabled and the other does not.

PR #536, opened 2019-04-04
Change the signature of `getVersionInfo()`

### LINKED ISSUE #442: Fix return type of getVersionInfo

### LINKED ISSUE #534: An error occurs when trying to change the "new" version in the Compare page

willdurand:
Fixes #534 
Fixes #442

---

### DIFF
--- a/src/pages/Browse/index.tsx
+++ b/src/pages/Browse/index.tsx
@@ -40,7 +40,7 @@ type PropsFromState = {
   apiState: ApiState;
   file: VersionFile | null | void;
   fileIsLoading: boolean;
-  version: Version;
+  version: Version | void;
 };
 
 export type Props = RouteComponentProps<PropsFromRouter> &
--- a/src/pages/Compare/index.spec.tsx
+++ b/src/pages/Compare/index.spec.tsx
@@ -129,17 +129,11 @@ describe(__filename, () => {
   });
 
   it('renders a loading message when no diff has been loaded', () => {
-    const addonId = 123;
-    const baseVersionId = 1;
-    const version = { ...fakeVersionWithDiff, id: baseVersionId + 1 };
+    const version = fakeVersionWithDiff;
     const store = configureStore();
     store.dispatch(versionsActions.loadVersionInfo({ version }));
 
-    const root = render({
-      addonId: String(addonId),
-      headVersionId: String(version.id),
-      store,
-    });
+    const root = render({ headVersionId: String(version.id), store });
 
     expect(root.find(Loading)).toHaveLength(1);
     expect(root.find(Loading)).toHaveProp('message', 'Loading diff...');
--- a/src/pages/Compare/index.spec.tsx
+++ b/src/pages/Compare/index.spec.tsx
@@ -113,7 +113,7 @@ describe(__filename, () => {
     );
   };
 
-  it('renders loading messages when no diff has been loaded', () => {
+  it('renders loading messages when no version has been loaded', () => {
     const addonId = 123;
     const root = render({ addonId: String(addonId) });
 
@@ -126,6 +126,23 @@ describe(__filename, () => {
     // This component is always displayed.
     expect(root.find(VersionChooser)).toHaveLength(1);
     expect(root.find(VersionChooser)).toHaveProp('addonId', addonId);
+  });
+
+  it('renders a loading message when no diff has been loaded', () => {
+    const addonId = 123;
+    const baseVersionId = 1;
+    const version = { ...fakeVersionWithDiff, id: baseVersionId + 1 };
+    const store = configureStore();
+    store.dispatch(versionsActions.loadVersionInfo({ version }));
+
+    const root = render({
+      addonId: String(addonId),
+      headVersionId: String(version.id),
+      store,
+    });
+
+    expect(root.find(Loading)).toHaveLength(1);
+    expect(root.find(Loading)).toHaveProp('message', 'Loading diff...');
   });
 
   it('renders a FileTree component when a diff has been loaded', () => {
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -352,6 +352,39 @@ describe(__filename, () => {
       expect(_log.debug).toHaveBeenCalled();
       expect(versionsState).toEqual(previousState);
     });
+
+    it('logs a debug message when headVersion is missing on loadDiff()', () => {
+      const addonId = 1;
+      const baseVersionId = 2;
+      const path = 'some/other/file.js';
+      const _log = createFakeLogger();
+
+      const version = {
+        ...fakeVersionWithDiff,
+        file: {
+          ...fakeVersionWithDiff.file,
+          // eslint-disable-next-line @typescript-eslint/camelcase
+          selected_file: path,
+        },
+      };
+      const headVersionId = version.id;
+
+      const versionsState = reducer(
+        undefined,
+        actions.loadDiff({
+          addonId,
+          baseVersionId,
+          headVersionId,
+          version,
+        }),
+        // TS looks confused about the third optional argument.
+        // @ts-ignore
+        { _log },
+      );
+
+      expect(_log.debug).toHaveBeenCalled();
+      expect(versionsState).toEqual(initialState);
+    });
   });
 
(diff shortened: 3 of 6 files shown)


src/reducers/versions.tsx
```diff
@@ -756,7 +759,13 @@ const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
     case getType(actions.loadDiff): {
       const { baseVersionId, headVersionId, version } = action.payload;
 
-      const { entries, selectedPath } = getVersionInfo(state, headVersionId);
+      const headVersion = getVersionInfo(state, headVersionId);
+      if (!headVersion) {
+        _log.debug(`Version missing for headVersionId: ${headVersionId}`);
```
bobsilverberg: Should this be `log.warn` because this seems like a pretty critical situation? Just returning `state` makes sense, but it will appear as if nothing happened. I suppose `debug` will also put a message in the console, but `warn` seems more appropriate.

src/reducers/versions.spec.tsx
```diff
@@ -352,6 +352,39 @@ describe(__filename, () => {
...
+      const baseVersionId = 2;
+      const path = 'some/other/file.js';
+      const _log = createFakeLogger();
+
+      const version = {
+        ...fakeVersionWithDiff,
+        file: {
+          ...fakeVersionWithDiff.file,
+          // eslint-disable-next-line @typescript-eslint/camelcase
+          selected_file: path,
```
bobsilverberg: Why do you need to override `selected_file` with anything? It doesn't seem to have any impact on what you are testing here. It only cares about whether the version being requested is present or not.

src/reducers/versions.spec.tsx
```diff
@@ -352,6 +352,39 @@ describe(__filename, () => {
...
+
+      const versionsState = reducer(
+        undefined,
+        actions.loadDiff({
+          addonId,
+          baseVersionId,
+          headVersionId,
+          version,
+        }),
+        // TS looks confused about the third optional argument.
```
bobsilverberg: This is odd. Can we do something better with the definition to make TS less confused?

src/pages/Compare/index.spec.tsx
```diff
@@ -128,6 +128,23 @@ describe(__filename, () => {
...
 
+  it('renders a loading message when no diff has been loaded', () => {
+    const addonId = 123;
+    const baseVersionId = 1;
+    const version = { ...fakeVersionWithDiff, id: baseVersionId + 1 };
+    const store = configureStore();
+    store.dispatch(versionsActions.loadVersionInfo({ version }));
+
+    const root = render({
+      addonId: String(addonId),
```
bobsilverberg: I'm not sure why you're overriding `addonId` - it doesn't seem to have any impact on the test. Note that this is reflected in by diff above.

src/pages/Compare/index.spec.tsx
```diff
@@ -128,6 +128,23 @@ describe(__filename, () => {
     expect(root.find(VersionChooser)).toHaveProp('addonId', addonId);
   });
 
+  it('renders a loading message when no diff has been loaded', () => {
+    const addonId = 123;
+    const baseVersionId = 1;
```
bobsilverberg: I'm not sure why you're specifying a `baseVersionId` here. The only place you use it is to set the `id` when creating `version`, and then later you just use `version.id`, so it seems like you might as well just remove `baseVersionId`, and not bother overriding `id` for the version.

I.e., this looks like the same test:

```diff
diff --git a/src/pages/Compare/index.spec.tsx b/src/pages/Compare/index.spec.tsx
index b3111d9..b7cfddd 100644
--- a/src/pages/Compare/index.spec.tsx
+++ b/src/pages/Compare/index.spec.tsx
@@ -129,14 +129,11 @@ describe(__filename, () => {
   });
 
   it('renders a loading message when no diff has been loaded', () => {
-    const addonId = 123;
-    const baseVersionId = 1;
-    const version = { ...fakeVersionWithDiff, id: baseVersionId + 1 };
+    const version = fakeVersionWithDiff;
     const store = configureStore();
     store.dispatch(versionsActions.loadVersionInfo({ version }));
 
     const root = render({
-      addonId: String(addonId),
       headVersionId: String(version.id),
       store,
     });
```

src/pages/Browse/index.tsx
```diff
@@ -40,7 +40,7 @@ type PropsFromState = {
   apiState: ApiState;
   file: VersionFile | null | void;
   fileIsLoading: boolean;
-  version: Version;
+  version: Version | void;
```
bobsilverberg: I'm surprised that this can just be changed without any fallout in this component, but I guess it can!

src/reducers/versions.tsx
willdurand: I don't really when this is supposed to occur to be honest.

src/reducers/versions.spec.tsx
willdurand: It does not matter indeed.

src/reducers/versions.spec.tsx
willdurand: No, we can't. This is existing code.

willdurand: This is because TS does not expect a third argument to `reducer` but there is no reason: an optional argument can be added to a function without any problem in JS.

src/reducers/versions.spec.tsx
```diff
@@ -352,6 +352,30 @@ describe(__filename, () => {
...
+        // @ts-ignore
+        { _log },
```
kumar303: It makes sense that TS won't allow this because Redux has a specific reducer signature and this doesn't match it. We can make a pretty good guess that the reducer will still work in the wild with a `_log` option *but* I think it would be fine to honor the type signature and just change the design. We could do it like this:

```tsx
const createReducer = ({ _log = log } = {}) => {
  const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
    state = initialState,
    action,
  ): VersionsState => {
    switch (action.type) {
      // ...
    }
  };

  return reducer;
};

const reducer = createReducer();
export default reducer;
```

This might be super annoying -- just throwing it out as an idea (not a change request).

src/reducers/versions.tsx
```diff
@@ -756,7 +759,13 @@ const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
     case getType(actions.loadDiff): {
       const { baseVersionId, headVersionId, version } = action.payload;
 
-      const { entries, selectedPath } = getVersionInfo(state, headVersionId);
+      const headVersion = getVersionInfo(state, headVersionId);
+      if (!headVersion) {
+        _log.error(`Version missing for headVersionId: ${headVersionId}`);
```
kumar303: I'll be honest. I don't think it's important to use dependency injection to test a log message 😄 If you add a test that runs this line of code to make sure no errors are thrown, that's enough. TypeScript will catch all other mistakes that could go wrong with this log message. It was way more important to fully test function calls like this before static analysis.

willdurand: Yep, this is more to reach 100% code coverage than anything else now.

PR #537, opened 2019-04-04
Move buildFileTree into a fileTree reducer

### LINKED ISSUE #533: Move buildFileTree into a fileTree reducer
Supports #522 

In preparation for also storing the ordered fileTree paths and linter messages, move the `buildFileTree` function from `FileTree` into a `fileTree` reducer and store the tree in the redux store.

bobsilverberg:
Fixes #533 

This patch moves all of the functionality that creates the `tree` structure out of the `FileTree` component and into a new `fileTree` reducer. This includes:

- The types for `FileNode`, `DirectoryNode` and `TreeNode`.
- The functions `getVersionName`, `getRootPath`, `recursiveSortInPlace` and `buildFileTree`, all of which are used for generating the `tree` and were all moved exactly as is.
- The `FileTree` component now gets its `tree` prop via `mapStateToProps` from the `fileTree` reducer.
- The `FileTree` component now has a `loadData` function which is called from both `componentDidMount` and `componentDidUpdate` which checks whether has a value in the `tree` prop and if not dispatches `fileTreeActions.buildTree`.
- The new `fileTree` reducer exposes one action: `buildTree` which accepts a version and stores a `tree` structure in its state, and one selector: `getTree` which returns either the `tree` structure, or `undefined` if it hasn't been built yet for the requested `version`.

### DIFF
--- a/src/components/FileTree/index.spec.tsx
+++ b/src/components/FileTree/index.spec.tsx
@@ -1,372 +1,33 @@
 /* eslint @typescript-eslint/camelcase: 0 */
-import * as React from 'react';
 import { shallow } from 'enzyme';
-import Treefold from 'react-treefold';
+import * as React from 'react';
 import { ListGroup } from 'react-bootstrap';
+import Treefold from 'react-treefold';
+import { Store } from 'redux';
 
 import configureStore from '../../configureStore';
 import {
+  actions as fileTreeActions,
+  buildFileTree,
+} from '../../reducers/fileTree';
+import {
   Version,
   actions as versionActions,
-  createInternalVersion,
-  createInternalVersionEntry,
   getVersionInfo,
 } from '../../reducers/versions';
 import {
   createFakeLogger,
-  createVersionWithEntries,
   fakeVersion,
-  fakeVersionEntry,
   shallowUntilTarget,
   spyOn,
 } from '../../test-helpers';
-import { getLocalizedString } from '../../utils';
 import { getTreefoldRenderProps } from '../FileTreeNode/index.spec';
 import FileTreeNode from '../FileTreeNode';
 import Loading from '../Loading';
 
-import FileTree, {
-  DirectoryNode,
-  FileTreeBase,
-  buildFileTree,
-  getRootPath,
-} from '.';
+import FileTree, { DefaultProps, FileTreeBase, PublicProps } from '.';
 
 describe(__filename, () => {
-  describe('getRootPath', () => {
-    const version = createInternalVersion(fakeVersion);
-    const addonName = getLocalizedString(version.addon.name);
-
-    expect(getRootPath(version)).toEqual(`root-${addonName}`);
-  });
-
-  describe('buildFileTree', () => {
-    it('creates a root node', () => {
-      const version = createVersionWithEntries([]);
-      const addonName = getLocalizedString(version.addon.name);
-
-      const tree = buildFileTree(version);
-
-      expect(tree).toEqual({
-        id: `root-${addonName}`,
-        name: addonName,
-        children: [],
-      });
-    });
-
-    it('converts a non-directory entry to a file node', () => {
-      const filename = 'some-filename';
-      const entries = [
-        createInternalVersionEntry({
-          ...fakeVersionEntry,
-          mime_category: 'text',
-          filename,
-        }),
-      ];
-      const version = createVersionWithEntries(entries);
-
-      const tree = buildFileTree(version);
-
-      expect(tree.children).toEqual([
-        {
-          id: entries[0].path,
-          name: filename,
-        },
-      ]);
-    });
-
-    it('converts a directory entry to a directory node', () => {
-      const filename = 'some-directory';
-      const entries = [
-        createInternalVersionEntry({
-          ...fakeVersionEntry,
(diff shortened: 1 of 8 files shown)


bobsilverberg:
My apologies for the large diff, but this is pretty much the minimum amount of changes required to move this function. The large number of lines changed has to do with the fact that quite a bit of code was moved from `FileTree` (component) to `fileTree` (reducer), and the same was done for the tests (a wholesale move of a lot of code from tests for the component to tests for the reducer). The diff for the test is super ugly though. :-(

src/components/FileTree/index.spec.tsx
```diff
@@ -1,399 +1,120 @@
...
-        },
-      );
-    };
+      expect(dispatch).not.toHaveBeenCalled();
+    });
 
     it('renders a ListGroup component with a Treefold', () => {
       const store = configureStore();
       const version = getVersion({ store });
+      _buildTree(store, version);
```
willdurand: Could this be moved in `render()`? It looks like it is used everywhere.

bobsilverberg: There are a couple of issues with that, which is why I didn't do it.

1. We don't always want to build the file tree, so I would have needed an argument like `buildFileTree: boolean`.
2. The `render` function is designed to accept a `versionId`, not a `version`. It's up to the test to decide whether to load a `version` that matches the `versionId` into the store before calling `render`. `buildTree` requires an actual `version`, so it would require changes to the `render` function, which is currently working pretty well for the tests we need.

I did experiment with this, but decided that keeping `render` the way it is would be best. Suggestions for improvements are welcome, though.

src/reducers/fileTree.tsx
```diff
@@ -0,0 +1,179 @@
...
+        for (let i = 0; i < parts.length; i++) {
+          const foundNode = currentNode.children.find(
+            (child: TreeNode) => child.name === parts[i],
+          ) as DirectoryNode;
+
+          if (foundNode) {
+            currentNode = foundNode;
+          }
+
+          // TODO: this should not happen but what if we don't find a node?
```
kumar303: I know this was in the original patch but I think it would be nice if this threw an error or logged a warning. I'm imagining that hitting this line would result in a broken tree so it would be helpful to at least get some kind of signal of what went wrong. You could probably inject a stub `root` to test it -- whatever works.

src/reducers/fileTree.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+export const actions = {
+  buildTree: createAction('BUILD_TREE', (resolve) => {
+    return (payload: { version: Version }) => resolve(payload);
+  }),
+};
+
+export const getTree = (
+  treeState: FileTreeState,
+  version: Version | void,
```
kumar303: can this just be `versionId`?

src/reducers/fileTree.tsx
```diff
@@ -0,0 +1,179 @@
...
+};
+
+export const initialState: FileTreeState = {
+  forVersionId: undefined,
+  tree: undefined,
+};
+
+export const actions = {
+  buildTree: createAction('BUILD_TREE', (resolve) => {
+    return (payload: { version: Version }) => resolve(payload);
```
kumar303: I think it's better to keep actions light whenever possible since they live on the stack of all user activity. What about accepting `versionId` and `addonName` (or `rootName`)?

bobsilverberg: I'm not sure I follow. `buildTree` needs a version to work with to build the tree. Are you suggesting we get that from the `versions` reducer, rather than passing it in? Is it okay to call a selector from `versions` when inside the `fileTree` reducer?

bobsilverberg:
Thanks for the initial reviews @kumar303 and @willdurand. I addressed the first two of @kumar303's comments, and added some replies to the other comments.

src/reducers/fileTree.tsx
kumar303: Where does it need a `version`? I only saw two places:

- to build the root name: https://github.com/mozilla/addons-code-manager/blob/69cdeeebdc789e8cbc922c6a8c1ae9ba0b93feec/src/reducers/fileTree.tsx#L45
- to use `version.id` as a key https://github.com/mozilla/addons-code-manager/blob/69cdeeebdc789e8cbc922c6a8c1ae9ba0b93feec/src/reducers/fileTree.tsx#L170

Thus, I think the action payload can just be `dispatch(buildTree({ versionId: 123, rootName: 'some-name' }))`.

Am I missing some other use of `version`?

willdurand: Aren't we supposed to use the `entries` too? https://github.com/mozilla/addons-code-manager/blob/69cdeeebdc789e8cbc922c6a8c1ae9ba0b93feec/src/reducers/fileTree.tsx#L53

kumar303:
> Aren't we supposed to use the entries too? 

Yes! That's what I was missing. OK, I guess you might as well pass `version` into the action. Carry on.

src/reducers/fileTree.tsx
bobsilverberg: Yes, the entries are kind of the point of the whole thing. 😉

PR #539, opened 2019-04-05
Store a list of ordered file paths in the fileTree reducer 

### LINKED ISSUE #538: Store a list of ordered file paths in the fileTree reducer
Supports #522 
Depends on #533 

Store a list of file paths, in the same order as they appear in the `FileTree`, in the `fileTree` reducer.

bobsilverberg:
Fixes #538 

This builds a list of file paths that can then be used for keyboard navigation. Note that the `FileTree` component does not currently use the `treePathList` prop, but it does need it to know whether to request that the path list be built by the reducer.

### DIFF
--- a/src/components/FileTree/index.tsx
+++ b/src/components/FileTree/index.tsx
@@ -15,6 +15,7 @@ import {
   TreeNode,
   actions as fileTreeActions,
   getTree,
+  getTreePathList,
 } from '../../reducers/fileTree';
 import {
   Version,
@@ -39,6 +40,7 @@ export type DefaultProps = {
 
 type PropsFromState = {
   tree: DirectoryNode | void;
+  treePathList: string[] | void;
   version: Version | void;
 };
 
@@ -66,10 +68,15 @@ export class FileTreeBase extends React.Component<Props> {
   }
 
   _loadData = () => {
-    const { dispatch, tree, version } = this.props;
+    const { dispatch, tree, treePathList, version } = this.props;
 
-    if (version && !tree) {
-      dispatch(fileTreeActions.buildTree({ version }));
+    if (version) {
+      if (!tree) {
+        dispatch(fileTreeActions.buildTree({ version }));
+      } else if (!treePathList) {
+        console.log(tree);
+        dispatch(fileTreeActions.buildTreePathList({ versionId: version.id }));
+      }
     }
   };
 
@@ -190,6 +197,9 @@ const mapStateToProps = (
 
   return {
     tree: version ? getTree(state.fileTree, version.id) : undefined,
+    treePathList: version
+      ? getTreePathList(state.fileTree, version.id)
+      : undefined,
     version,
   };
 };
--- a/src/components/FileTree/index.tsx
+++ b/src/components/FileTree/index.tsx
@@ -15,7 +15,6 @@ import {
   TreeNode,
   actions as fileTreeActions,
   getTree,
-  getTreePathList,
 } from '../../reducers/fileTree';
 import {
   Version,
@@ -39,8 +38,7 @@ export type DefaultProps = {
 };
 
 type PropsFromState = {
-  tree: DirectoryNode | void;
-  treePathList: string[] | void;
+  treeNodes: DirectoryNode | void;
   version: Version | void;
 };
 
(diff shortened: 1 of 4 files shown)


src/reducers/fileTree.tsx
```diff
@@ -172,6 +207,20 @@ const reducer: Reducer<FileTreeState, ActionType<typeof actions>> = (
         tree,
       };
     }
+    case getType(actions.buildTreePathList): {
+      const { versionId } = action.payload;
+      const { forVersionId, tree } = state;
+
+      if (!versionId || forVersionId !== versionId || !tree) {
+        _log.warn('No version loaded for versionId: ', versionId);
+        return state;
```
kumar303: The coverage report says this is missing a test but I didn't verify. Is it?

src/components/FileTree/index.tsx
```diff
@@ -66,10 +68,15 @@ export class FileTreeBase extends React.Component<Props> {
...
-    const { dispatch, tree, version } = this.props;
+    const { dispatch, tree, treePathList, version } = this.props;
 
-    if (version && !tree) {
-      dispatch(fileTreeActions.buildTree({ version }));
+    if (version) {
+      if (!tree) {
+        dispatch(fileTreeActions.buildTree({ version }));
+      } else if (!treePathList) {
+        console.log(tree);
```
kumar303: eslint says no 💣

src/reducers/fileTree.tsx
```diff
@@ -172,6 +207,20 @@ const reducer: Reducer<FileTreeState, ActionType<typeof actions>> = (
         tree,
       };
     }
+    case getType(actions.buildTreePathList): {
+      const { versionId } = action.payload;
+      const { forVersionId, tree } = state;
+
+      if (!versionId || forVersionId !== versionId || !tree) {
```
kumar303: It's not a good idea to depend on `state.tree` because it puts the burden of control on the component. Currently, it's not in full control otherwise you wouldn't need to log a warning and bail. 

I suggest performing this reducer logic at the same time you [build the tree](https://github.com/mozilla/addons-code-manager/blob/5d836b97265d8b49368a0c12eba8ac0646260f3f/src/reducers/fileTree.tsx#L200). The only downside might be performance but I don't think that will be an issue. If it becomes an issue, we could split the tree building reducer into smaller pieces.

If you really think it's a better idea to depend on `state.tree` then you should pass it through the action payload instead. With that approach, the component would be fully in control.

src/reducers/fileTree.tsx
```diff
@@ -131,20 +132,43 @@ export const buildFileTree = (version: Version): DirectoryNode => {
...
+
+  extractPaths(tree);
+
+  return treePathList;
+};
+
 export type FileTreeState = {
   forVersionId: void | number;
   tree: DirectoryNode | void;
+  treePathList: string[] | void;
```
kumar303: Since I suggested doing all the reducer work in response to `buildTreePathList`, it would probably be more convenient to expand the tree object instead of adding a new state attribute. Something like this:

```tsx
tree: {
  nodes: DirectoryNode | void;
  pathList: string[] | void;
}
```

What do you think?

src/reducers/fileTree.tsx
bobsilverberg: My original intention was to build this at the same time as the tree, but I thought you requested I not do that and make it distinct actions. I now realize that that was probably about storing the linter messages, and not simply the paths. I'll make that change which will change a lot of this patch.

src/reducers/fileTree.tsx
bobsilverberg: Yes, sounds good.

bobsilverberg:
Ok, I changed this to build the file paths in the same action as the tree nodes, and that cleaned up pretty much everything from the original comments.

src/reducers/fileTree.tsx
kumar303: > My original intention was to build this at the same time as the tree, but I thought you requested I not do that and make it distinct actions

Ah, sorry for the confusion. I was just requesting to start with a smaller patch (the first one for just files, not linter messages) and improve it incrementally. That makes each patch easier to review.

src/components/FileTree/index.tsx
```diff
@@ -38,7 +38,7 @@ export type DefaultProps = {
 };
 
 type PropsFromState = {
-  tree: DirectoryNode | void;
+  treeNodes: DirectoryNode | void;
```
kumar303: It would be helpful to make this `treeNodes: FileTree['nodes']` so they remain in sync.

src/reducers/fileTree.spec.tsx
```diff
@@ -20,11 +21,15 @@ describe(__filename, () => {
     it('builds and loads a tree', () => {
       const version = createInternalVersion(fakeVersion);
       const state = reducer(undefined, actions.buildTree({ version }));
+      const nodes = buildFileTree(version);
 
       expect(state).toEqual({
         ...initialState,
         forVersionId: version.id,
-        tree: buildFileTree(version),
+        tree: {
```
kumar303: I'm seeing this pattern of calling two functions repeatedly. Why not make it one?

```tsx
      expect(state).toEqual({
        ...initialState,
        forVersionId: version.id,
        tree: buildFileTree(version),
      });
```

This would mean renaming the current `buildFileTree()` to `buildFileTreeNode()` and adding a new function, `buildFileTree()`, that combines everything and returns the final tree.

Issue #523, opened 2019-04-02
Visualize and navigate inline linter messages


kumar303:
User story:

> As a reviewer, I want to quickly see inline linter messages so I can understand their impact when reviewing code.

The current tool offers a vertical strip to the left of the file pane that attempts to let a reviewer visualize what slice of the file they are viewing and what lines have linter messages. It is buggy and unintuitive and does not show you actual linter messages.

It attempts to implement a common UX pattern of showing a condensed overview of a long file, found in IDEs. Here is an example of one, along the right side of this screen:

<details>
<summary>Screenshot</summary>

<img width="1277" alt="Screenshot 2019-04-02 at 19 08 51" src="https://user-images.githubusercontent.com/55398/55422685-d49ffa00-5541-11e9-832b-5358337db515.png">

</details>

Maybe we can implement something like this so that it addresses the user story above and possibly future user stories about navigating reviewer comments in files and other annotations from analysis. However, starting with a quicker implementation would be fine, too.

See https://github.com/mozilla/addons-code-manager/issues/520 for addressing this same user story purely with keyboard shortcuts

vcarciu:
@kumar303 : If the message is at the beginning of the file, then you can't visualize the linter message directly from the right side.  For example : https://code.addons-dev.allizom.org/en-US/browse/494431/versions/1532144/?path=manifest.json .

Please let me know if this is expected.

vcarciu:
I will log a new issue for comment from above.

PR #542, opened 2019-04-08
Visualize and navigate inline linter messages

kumar303:
This UI provides an overview of the entire file so one can visualize linter messages (and later things like comments). It's a proof of concept for https://github.com/mozilla/addons-code-manager/issues/523

<details>
<summary>Screencast</summary>

</details>

@willdurand @bobsilverberg do you have any feedback on the look and feel of the UX?

Notes:
- I'll break this up into smaller parts before requesting any code reviews
- This is just a foundation. We could add more things later like a magnifier to indicate which part of the overview is in view.
- I only added the overview to the `<Browse>` but it wouldn't be too hard to support `<Compare>` as well.


willdurand:
I suppose we can close this PR because the feature has landed.

PR #543, opened 2019-04-08
Configure lockfile maintenance PRs for Renovate

willdurand:
This tells Renovate to send a lockfile maintenance PR once a week (on
Monday). The second added option tells Renovate to wait until branch
tests have passed or failed before opening a PR.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,8 @@
 {
   "extends": [
-    "config:base"
+    "config:base",
+    ":maintainLockFilesWeekly",
+    ":prNotPending"
   ],
   "packageRules": [
     {


renovate.json
```diff
@@ -1,6 +1,8 @@
 {
   "extends": [
-    "config:base"
+    "config:base",
+    ":maintainLockFilesWeekly",
```
bobsilverberg: What is the default behaviour without this option?

willdurand: No lockfile maintenance PR.

bobsilverberg: Fair enough.

PR #547, opened 2019-04-08
Create actions for selecting the next and previous file in the tree

### LINKED ISSUE #471: Create actions for selecting the next and previous file in the tree
~~Depends on #185.~~

Two of the keyboard shortcuts required for #199 are meant to select the next file in the FileTree, and the previous file in the FileTree. We therefore require actions that will do each of these things (i.e., change the currently selected file to the next/previous file in the tree). This should be reflected in the UI by the new file being loaded on the right-hand side, and also being indicated as the selected file in the FileTree.

Once we have these actions, we can wire them up to the corresponding keyboard shortcuts.

I think it probably makes sense to complete #185 first.

bobsilverberg:
Fixes #471 

This is just a proof of concept for executing this action which needs data from both the `fileTree` state and the `versions` state, and needs to modify the `versions` state, which I am doing by dispatching `versionActions.updateSelectedPath`. 

I thought I could get away with not having to access `getState` in the thunk, by accepting `currentPath` as a parameter, but we still need it to access the internal state of the `fileTree` reducer, because we also need to dispatch.

I may have gotten a number of things wrong here, because it still seems a bit convoluted to me. Submitting this for feedback on the approach.

@willdurand this is something I discussed with @kumar303 via Vidyo today, but please feel free to comment if you have any suggestions.

### DIFF
--- a/src/reducers/fileTree.tsx
+++ b/src/reducers/fileTree.tsx
@@ -160,7 +160,7 @@ export const buildFileTree = (version: Version): FileTree => {
 };
 
 type GetRelativePathParams = {
-  _log: typeof log;
+  _log?: typeof log;
   currentPath: string;
   pathList: string[];
   position: 'previous' | 'next';
@@ -173,7 +173,7 @@ export const getRelativePath = ({
   position,
 }: GetRelativePathParams): string | void => {
   const currentIndex = pathList.indexOf(currentPath);
-  if (!currentIndex) {
+  if (currentIndex < 0) {
     _log.debug(`Cannot find ${currentPath} in pathList: ${pathList}`);
     return undefined;
   }
--- a/src/reducers/fileTree.tsx
+++ b/src/reducers/fileTree.tsx
@@ -1,7 +1,9 @@
+import log from 'loglevel';
 import { Reducer } from 'redux';
 import { ActionType, createAction, getType } from 'typesafe-actions';
 
-import { Version } from './versions';
+import { Version, actions as versionActions } from './versions';
+import { ThunkActionCreator } from '../configureStore';
 import { getLocalizedString } from '../utils';
 
 type FileNode = {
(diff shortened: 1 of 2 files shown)


src/reducers/fileTree.tsx
```diff
@@ -157,6 +159,74 @@ export const buildFileTree = (version: Version): FileTree => {
...
+      _log,
+      currentPath,
+      pathList,
+      position: 'next',
+    });
+
+    if (!nextPath) {
+      // There is no next file, so do nothing.
+      // TODO: I don't think we need to log in this case.
+      return;
```
kumar303: If I'm pressing a key to see the next message and nothing happens, I'd be pretty confused. You could dispatch an action here that shows a message like 'There are no more linter messages'. Because of this, I think using a thunk for this feature is the right way to go.

src/reducers/fileTree.tsx
```diff
@@ -157,6 +159,74 @@ export const buildFileTree = (version: Version): FileTree => {
...
+    }
+
+    dispatch(
```
kumar303: Were you implying on IRC that the existence of a sole dispatch suggests you don't need a thunk and that you could move all of this to the versions reducer? I don't think it would buy you much because you'd have to duplicate the tests for `updateSelectedPath` and that would be annoying. 

Also, I can see a need to dispatch more actions (as commented above) so I think you probably need a thunk. Moving the thunk to the versions reducer won't buy you much either.

src/reducers/fileTree.tsx
```diff
@@ -157,6 +159,74 @@ export const buildFileTree = (version: Version): FileTree => {
...
+  _log = log,
+  currentPath,
+}: GoToNextFileParams): ThunkActionCreator => {
+  return async (dispatch, getState) => {
+    const { fileTree: fileTreeState } = getState();
+    const { forVersionId, tree } = fileTreeState;
+    const { pathList } = tree;
+
+    if (!forVersionId || !pathList) {
+      _log.debug('Cannot navigate to next file with no version loaded');
```
kumar303: Since this thunk will be triggered from a key press, it seems likely that the user will encounter this case. You could dispatch an action here to show a message that the data is still loading, similar to my other comment.

I wonder if it's better to read these variables from action parameters rather than from state but I'm not sure. If you read them from parameters, the controlling component would know when to load data (when they are empty) but maybe that's not necessary. You could keep it reading from state (as is) for now.

src/reducers/fileTree.tsx
```diff
@@ -157,6 +159,74 @@ export const buildFileTree = (version: Version): FileTree => {
...
+  position,
+}: GetRelativePathParams): string | void => {
+  const currentIndex = pathList.indexOf(currentPath);
+  if (!currentIndex) {
+    _log.debug(`Cannot find ${currentPath} in pathList: ${pathList}`);
+    return undefined;
+  }
+  const newIndex =
+    position === 'previous' ? currentIndex - 1 : currentIndex + 1;
+  const validIndex = newIndex >= 0 && pathList.length >= newIndex;
```
kumar303: I think this probably needs to be `pathList.length > newIndex`

src/reducers/fileTree.tsx
```diff
@@ -157,6 +159,74 @@ export const buildFileTree = (version: Version): FileTree => {
...
+  if (!validIndex) {
+    _log.debug(`There is not a ${position} file wrt ${currentPath}`);
```
kumar303: Could you do this without an abbreviation? I know it was just a draft PR but there ya go 😄

src/reducers/fileTree.tsx
willdurand: Couldn't we loop over the files/messages?

bobsilverberg: > Couldn't we loop over the files/messages?

This is a good question. Do we think it would make sense to go to the first file if a user chooses "next file" when we're at the last file? The same question for "previous file" - should that go to the last file if the user is on the first file? In some ways this makes sense, but in other ways it does not. I think I'm leaning towards "yes" though.

@wagnerand What do you think?

bobsilverberg: FYI, for now I will implement @willdurand's suggestion as it saves us from having to inform the user if there are no more files in the direction they wish to navigate.

src/reducers/fileTree.tsx
bobsilverberg: @kumar303 Is your suggestion here that the component dispatching `gotToNextFile` could pass the state variables into the action? I can see how that could be useful for your first comment, as we wouldn't even bother dispatching unless a version was loaded. In fact, the keyboard shortcuts don't make any sense without a tree, so we could not even render the shortcut list until the tree is present. This makes a lot of sense, but I'm not sure passing that amount of data is a good idea. The action needs the entire `pathList` from the tree, which could be quite large. Is that considered okay?

src/reducers/fileTree.tsx
kumar303: > If I'm pressing a key to see the next message and nothing happens, I'd be pretty confused. 
>> Couldn't we loop over the files/messages?

I considered this. I find this type of behavior (which is common for a find/replace UX) very confusing because with a long list of results, it's hard to know when you're *finished*. In this case, navigating through messages is a task that has a finite end state. We could probably address this problem in another way, though. For example, maybe the first item in the list has some kind of animation to give you a cue.

src/reducers/fileTree.tsx
```diff
@@ -157,6 +159,88 @@ export const buildFileTree = (version: Version): FileTree => {
...
+      currentPath,
+      pathList,
+      position,
+    });
+
+    if (!nextPath) {
+      // This will only happen if currentPath is not found in pathList, which
+      // should never happen, so I think we can just ignore this and log a debug
+      // message.
+      _log.debug(`Cannot find ${currentPath} in pathList: ${pathList}`);
```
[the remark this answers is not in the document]
bobsilverberg: Nice, killing two birds with one stone. 👍

bobsilverberg:
Thanks @kumar303, I've addressed your last two comments.

PR #550, opened 2019-04-09
Enable HMR

### LINKED ISSUE #275: Get hot reloading to work for development

willdurand:
Fixes #275

---

I started from scratch, trying to understand whether CRA officially
supports HMR (hint: it is not as clear as it should be). In the end, I
added some code to enable HMR for React [1] and Redux [2]. It's similar
to what I've done for addons-frontend in the past, albeit much simpler.

Creating `src/reducers/index.tsx` is required so that we can track
changes on all reducers at once.

---

[1]: I tested by live editing a React component, which is updated in FF
without reloading the page. It's hot patched instead
[2]: I tested by adding a new prop to a reducer and it works too (it can
be monitored in the redux devtools)

### DIFF
--- a/src/components/App/index.tsx
+++ b/src/components/App/index.tsx
@@ -11,7 +11,8 @@ import makeClassName from 'classnames';
 import log from 'loglevel';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
-import { ApplicationState, ConnectedReduxProps } from '../../configureStore';
+import { ApplicationState } from '../../reducers';
+import { ConnectedReduxProps } from '../../configureStore';
 import styles from './styles.module.scss';
 import { ApiState, actions as apiActions } from '../../reducers/api';
 import {
--- a/src/components/FileTree/index.tsx
+++ b/src/components/FileTree/index.tsx
@@ -9,7 +9,8 @@ import FileTreeNode, {
   PublicProps as FileTreeNodeProps,
 } from '../FileTreeNode';
 import Loading from '../Loading';
-import { ApplicationState, ConnectedReduxProps } from '../../configureStore';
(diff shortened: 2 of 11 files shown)


willdurand:
> Could you look to see what's going on here?

I'll try to fix it but I bet this is: https://github.com/reduxjs/redux/issues/1940

kumar303:
Hmm, that issue suggest that it is unfixable, no?

willdurand:
> Hmm, that issue suggest that it is unfixable, no?

It seems it should just work, but it does not. We could maybe backup the state before hot reloading.

kumar303:
> We could maybe backup the state before hot reloading.

I suppose that will work. Creating a root reducer with a predefined state is well supported. I guess it's all still a workaround for https://github.com/facebook/create-react-app/issues/6503 but that's OK.

willdurand:
Actually, I think this is an issue with our app and not with the HMR setup.

willdurand:
Also, we can't really backup the state before applying the HMR patch. Redux does not allow that.

kumar303:
> I think this is an issue with our app and not with the HMR setup.

Can you file it with STR so we don't forget to fix it?

willdurand:
> Can you file it with STR so we don't forget to fix it?

I filed https://github.com/mozilla/addons-code-manager/issues/551

PR #553, opened 2019-04-09
The fileTree reducer should not store an empty FileTree object 

### LINKED ISSUE #552: The fileTree reducer should not store an empty FileTree object
Depends on #547 

(This doesn't strictly depend on #547, but to avoid conflicts I plan to build it on top of that)

The `fileTree` reducer is [currently set up to store an empty `FileTree` object in its `tree` state](https://github.com/mozilla/addons-code-manager/blob/d75a64f80630ab6f544ce40ed199357603229cdc/src/reducers/fileTree.tsx#L160-L173) before the tree is built:

```js
export type FileTree = {
  nodes: DirectoryNode | void;
  pathList: PathList | void;
};

export type FileTreeState = {
  forVersionId: void | number;
  tree: FileTree;
};

export const initialState: FileTreeState = {
  forVersionId: undefined,
  tree: { nodes: undefined, pathList: undefined },
};
```

This seemed like a good idea at the time, but I realize now it is making it more difficult to work with. We should change this so that `tree` can be `void` before being populated, which will simplify some of the code I'm working on for the `FileTree` component.

bobsilverberg:
Fixes #552 
~~Depends on #547~~

Only the final commit in this PR is relevant to this patch.

Note that this change is being made to simplify the changes required for #459, which will require that the `FileTree` component also have access to `pathList` from the `fileTree` state.

### DIFF
--- a/src/components/FileTree/index.tsx
+++ b/src/components/FileTree/index.tsx
@@ -39,7 +39,7 @@ export type DefaultProps = {
 };
 
 type PropsFromState = {
-  treeNodes: FileTree['nodes'] | void;
+  tree: FileTree | void;
   version: Version | void;
 };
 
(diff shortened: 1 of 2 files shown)


bobsilverberg:
Note: It will probably be much easier to review this once #547 lands, and it's fine to wait for that.

kumar303:
This looks like it's ready for a rebase

bobsilverberg:
Thanks @kumar303. It's been rebased!

bobsilverberg:
Thanks @kumar303!

PR #558, opened 2019-04-10
Do not dispatch setAuthToken when token exists in state

### LINKED ISSUE #557: Do not dispatch `setAuthToken` on mount
See also #551 

---

STR:

1. Navigate to http://localhost:3000/
2. Open the redux devtools
3. Edit `Index.tsx`
4. Observe the dispatched actions

It should not dispatch `setAuthToken`.

willdurand:
Fixes #557

### DIFF
--- a/src/components/App/index.spec.tsx
+++ b/src/components/App/index.spec.tsx
@@ -88,7 +88,7 @@ describe(__filename, () => {
     expect(dispatch).not.toHaveBeenCalled();
   });
 
-  it('dispatches setAuthToken on mount when authToken is valid', () => {
+  it('dispatches setAuthToken on mount when authToken is valid and not already set', () => {
     const authToken = 'my-token';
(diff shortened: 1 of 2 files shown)


src/components/App/index.spec.tsx
```diff
@@ -77,6 +77,17 @@ describe(__filename, () => {
...
+    const store = configureStore();
+    store.dispatch(apiActions.setAuthToken({ authToken }));
+    const dispatch = spyOn(store, 'dispatch');
+
+    render({ authToken, store });
+
+    expect(dispatch).not.toHaveBeenCalled();
+  });
+
   it('dispatches setAuthToken on mount when authToken is valid', () => {
```
bobsilverberg: You might want to rename this test to indicate `when authToken is not present in state`, or something similar.

Issue #546, opened 2019-04-08
Generate an anchor link to a line of code


kumar303:
For the code overview panel (https://github.com/mozilla/addons-code-manager/issues/523), each line needs to link to a line of code. We already generate these links (see https://github.com/mozilla/addons-code-manager/issues/317) so now we need a shared function that more than one component can use.

Note that the compare view is a little different because we rely on the upstream library to generate the anchor links.

kumar303:
Ultimately, we'll want the code overview to link to lines of code in the Compare view as well as the Browse view.

The Compare view will be tricky because the anchor tags we'll deal with will either look like `#N123` (for the normal part of the diff) or `#I123` (for insertions).

These anchors are generated here: https://github.com/otakustay/react-diff-view/blob/4bca86cec1168b2ca05c16dd6d202053a727931a/src/utils/diff/getChageKey.js

I'm not going to worry about it for this first patch.

willdurand:
@ioanarusiczki added QA instructions

AlexandraMoga:
I've verified this on -dev for the browse view and each line of code, for any of the inspected files has an anchor link:

![2019-04-16_16-03-48](https://user-images.githubusercontent.com/31961530/56214738-19aa4d00-6067-11e9-8b6f-f7d2ce4e1155.gif)

PR #560, opened 2019-04-10
Use common functions to get code line anchors

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/546

This seems like a small thing to add a function for but it will help to synchronize the line linking functionality between `Browse` and the forthcoming code overview component (see https://github.com/mozilla/addons-code-manager/pull/542/files#diff-4a3375c826127680658868411c3f42d3R236).

### DIFF
--- a/src/components/CodeView/utils.tsx
+++ b/src/components/CodeView/utils.tsx
@@ -37,16 +37,10 @@ export const getLines = (content: string) => {
   return content.replace('\r\n', '\n').split('\n');
 };
 
-/*
- * This gets an ID (for a DOM element) representing a line of code.
- */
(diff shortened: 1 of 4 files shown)


src/components/CodeView/utils.tsx
```diff
@@ -36,3 +36,17 @@ export const mapWithDepth = (depth: number, _mapChild = mapChild) => {
 export const getLines = (content: string) => {
   return content.replace('\r\n', '\n').split('\n');
 };
+
```
willdurand: I don't think the two comment blocks below are useful. They both paraphrase the function names.

Issue #544, opened 2019-04-08
Add array chunking utility


kumar303:
In order to fit code of varying lengths into a defined grid (see https://github.com/mozilla/addons-code-manager/issues/523), we need a utility that splits arrays into chunks.

kumar303:
Actually, we can just use [lodash.chunk](https://lodash.com/docs/4.17.11#chunk)

PR #561, opened 2019-04-10
Add array chunking utility

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/544

This will be used to fit long code files into a predefined grid: https://github.com/mozilla/addons-code-manager/pull/542/files#diff-4a3375c826127680658868411c3f42d3R221

### DIFF
--- a/src/utils.tsx
+++ b/src/utils.tsx
@@ -49,3 +49,25 @@ export const nl2br = (text: string | null) => {
 export const formatFilesize = (size: number): string => {
   return filesize(size, { standard: 'iec' });
 };
+
+export const splitArrayIntoChunks = <ItemType extends {}>(
+  array: ItemType[],
+  chunkSize: number,
+): ItemType[][] => {
+  if (chunkSize <= 0) {
+    // This would create an infinite loop.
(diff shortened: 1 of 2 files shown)


src/utils.tsx
```diff
@@ -49,3 +49,25 @@ export const nl2br = (text: string | null) => {
 export const formatFilesize = (size: number): string => {
   return filesize(size, { standard: 'iec' });
 };
+
+export const splitArrayIntoChunks = <ItemType extends {}>(
```
willdurand: Can we have a comment to indicate that `extends {}` is needed because TS does not understand that it's a generics otherwise?

kumar303:
> That looks good but, before r+ing it, any reason not to use [lodash.chunk](https://lodash.com/docs/#chunk)?

No reason! I can use `lodash.chunk`. Thanks for the tip. I didn't bother looking for libraries because it's such a small function and I figured there would be a hundred on NPM that I'd have to comb through and audit. The lodash code looks [straight forward](https://github.com/lodash/lodash/blob/750067f42d3aa5f927604ece2c6df0ff2b2e9d72/chunk.js) so we might as well use it.

Issue #551, opened 2019-04-09
Page components should not always dispatch an action on mount


willdurand:
From: https://github.com/mozilla/addons-code-manager/pull/550

---

### Describe the problem and steps to reproduce it:

1. Load http://localhost:3000/en-US/browse/494431/versions/1532144
2. Click on the root folder to expand it
3. Change line 98 so that it says LOL instead of a line number (or any edit like this).

## What happened?

The component updates but the tree collapses. It looks like it's replaying all Redux actions starting with initialState.

### What did you expect to happen?

The `CodeView` component updates but the rest of the UI does not change, i.e., the tree is still expanded.

### Anything else we should know?

This now occurs because of HMR https://github.com/mozilla/addons-code-manager/pull/550. We should prevent our main components to always dispatch actions because HMR re-mounts them and we did not handle this edge case.

In addons-frontend, we handled this edge case because of (or thanks to) SSR.

willdurand:
I gave `useEffect` a try in the `Browse` component but Enzyme (shallow) does not support it, so here is the sketch of patch:

```diff
diff --git a/src/pages/Browse/index.tsx b/src/pages/Browse/index.tsx
index 96890a4..7805af5 100644
--- a/src/pages/Browse/index.tsx
+++ b/src/pages/Browse/index.tsx
@@ -50,102 +50,83 @@ export type Props = RouteComponentProps<PropsFromRouter> &
   DefaultProps &
   ConnectedReduxProps;
 
-export class BrowseBase extends React.Component<Props> {
-  static defaultProps = {
-    _fetchVersion: fetchVersion,
-    _fetchVersionFile: fetchVersionFile,
-    _log: log,
-  };
-
-  componentDidMount() {
-    const { _fetchVersion, dispatch, match } = this.props;
-    const { addonId, versionId } = match.params;
-
-    // This also fetches / loads the default version file.
-    // Example: manifest.json
-    dispatch(
-      _fetchVersion({
-        addonId: parseInt(addonId, 10),
-        versionId: parseInt(versionId, 10),
-      }),
-    );
-  }
-
-  componentDidUpdate() {
-    const {
-      _fetchVersionFile,
-      dispatch,
-      file,
-      fileIsLoading,
-      match,
-      version,
-    } = this.props;
+export const BrowseBase = (props: Props) => {
+  const {
+    _fetchVersion,
+    _fetchVersionFile,
+    dispatch,
+    file,
+    fileIsLoading,
+    match,
[...]

Issue #246, opened 2019-02-21
Experiment with the useEffect hook


kumar303:
A good place to start on hooks is to consider porting some components to [useEffect](https://reactjs.org/docs/hooks-reference.html#useeffect). A lot of bugs I've seen on `addons-frontend` were related to all the if statement logic we needed in `constructor()` and `componentDidUpdate()` to know when to load data. I think `useEffect` will simplify that kind of logic and we already have some components in `addons-code-manager` like `App` which are relying on this logic.

kumar303:
See https://github.com/mozilla/addons-code-manager/issues/363

PR #564, opened 2019-04-11
Add `?path` to query string when selecting a file

### LINKED ISSUE #286: Link to a URL for viewing a single file (browse page)

willdurand:
Fixes #286
~~⚠️ depends on #48~~

---

This patch adds a new `?path` parameter to the query string when a file is selected. It works in all pages, with keyboard navigation and mouse selection. This patch also allows to load a specific file **in the browse page only**.

Note 1: this patch also has a fix for #551 (only browse component)
Note 2: I'll file an issue to add this feature to the Compare page too

### DIFF
--- a/src/pages/Browse/index.spec.tsx
+++ b/src/pages/Browse/index.spec.tsx
@@ -48,7 +48,7 @@ describe(__filename, () => {
     _fetchVersion?: BrowseProps['_fetchVersion'];
     _fetchVersionFile?: BrowseProps['_fetchVersionFile'];
     _log?: BrowseProps['_log'];
-    _updateSelectedPath?: BrowseProps['_updateSelectedPath'];
+    _viewVersionFile?: BrowseProps['_viewVersionFile'];
     addonId?: string;
     history?: History;
     store?: Store;
@@ -59,7 +59,7 @@ describe(__filename, () => {
     _fetchVersion,
     _fetchVersionFile,
     _log,
-    _updateSelectedPath,
+    _viewVersionFile,
     addonId = '999',
     history = createFakeHistory(),
     store = configureStore(),
@@ -73,7 +73,7 @@ describe(__filename, () => {
       _fetchVersion,
       _fetchVersionFile,
       _log,
-      _updateSelectedPath,
+      _viewVersionFile,
     };
 
     return shallowUntilTarget(<Browse {...props} />, BrowseBase, {
@@ -240,7 +240,7 @@ describe(__filename, () => {
     expect(root.find(Loading)).toHaveProp('message', 'Loading content...');
   });
 
-  it('dispatches updateSelectedPath() when a file is selected', () => {
+  it('dispatches viewVersionFile when a file is selected', () => {
     const addonId = 123456;
     const version = { ...fakeVersion, id: 98765 };
     const path = 'some-path';
@@ -250,10 +250,10 @@ describe(__filename, () => {
     const dispatch = spyOn(store, 'dispatch');
 
     const fakeThunk = createFakeThunk();
-    const _updateSelectedPath = fakeThunk.createThunk;
+    const _viewVersionFile = fakeThunk.createThunk;
 
     const root = render({
-      _updateSelectedPath,
+      _viewVersionFile,
       store,
       addonId: String(addonId),
       versionId: String(version.id),
@@ -266,7 +266,7 @@ describe(__filename, () => {
     onSelectFile(path);
 
     expect(dispatch).toHaveBeenCalledWith(fakeThunk.thunk);
-    expect(_updateSelectedPath).toHaveBeenCalledWith({
+    expect(_viewVersionFile).toHaveBeenCalledWith({
       selectedPath: path,
       versionId: version.id,
     });
@@ -418,7 +418,7 @@ describe(__filename, () => {
     expect(dispatch).not.toHaveBeenCalled();
   });
 
-  it('dispatches updateSelectedPath on mount if the query string contains a `path` that is different than `version.selectedPath`', () => {
+  it('dispatches viewVersionFile on mount if the query string contains a `path` that is different than `version.selectedPath`', () => {
     const path = 'a/different/file.js';
     const version = fakeVersion;
     const history = createFakeHistory({
@@ -431,10 +431,10 @@ describe(__filename, () => {
     const dispatch = spyOn(store, 'dispatch');
 
     const fakeThunk = createFakeThunk();
-    const _updateSelectedPath = fakeThunk.createThunk;
+    const _viewVersionFile = fakeThunk.createThunk;
 
     render({
-      _updateSelectedPath,
+      _viewVersionFile,
       history,
       store,
       versionId: String(version.id),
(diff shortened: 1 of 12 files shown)


willdurand:
@kumar303 @bobsilverberg Let me know what you think of this approach. It fixes a bunch of issues at once and one single method will ease the transition to `useEffect` (https://github.com/mozilla/addons-code-manager/issues/246) once Enzyme supports this hook, but maybe I overlooked something. Loading the `version` first avoids some problems.

src/pages/Browse/index.spec.tsx
```diff
@@ -252,24 +266,31 @@ describe(__filename, () => {
...
 
     expect(dispatch).toHaveBeenCalledWith(
-      versionActions.updateSelectedPath({
+      versionsActions.updateSelectedPath({
         versionId: version.id,
         selectedPath: path,
       }),
     );
+
+    expect(history.push).toHaveBeenCalledWith(
```
bobsilverberg: Because you've added this assertion to the test, you should probably either update the description of the test, or have a separate test with just this assertion. As it stands, this test is just called "dispatches updateSelectedPath() when a file is selected".

src/pages/Browse/index.spec.tsx
```diff
@@ -358,4 +379,119 @@ describe(__filename, () => {
...
+    expect(dispatch).toHaveBeenCalledWith(
+      versionsActions.updateSelectedPath({
+        versionId: version.id,
+        selectedPath: path,
+      }),
+    );
+    expect(dispatch).toHaveBeenCalledTimes(1);
+  });
+
+  it('does not dispatch updateSelectedPath on mount when the query string does not contain a `path`', () => {
```
bobsilverberg: Should there also be tests that assert that it does not dispatch `updateSelectedPath` if the path in the query string is the same as `version.selectedPath`, both for `mount` and `update`?

kumar303:
I started looking at this and I'm iffy on how the URL doesn't change consistently. I'd like the URL to change to `?path=file` *every* time `version.selectedPath` changes, aside from when we load the default file. This includes keyboard navigation. A reviewer should be able to always share the URL they are on and have it load the same for someone else (with regard to files).

The only way I can think to do this is to make all code call `history.push()` whenever it wants to select a file. That is, nothing should ever dispatch `updateSelectedPath()`.

Really, what I want is [`connected-router`](https://github.com/supasate/connected-react-router), I think, because updating the selected file should just naturally change the URL.

I'm fine with taking baby steps to that approach; these are just some quick thoughts before I go on lunch. I can add a more concrete review later.

willdurand:
ah well, this was written before any keyboard navigation feature... Of course, this new feature changes a few things. I also feel like the connected-router might help and we should definitely synchronize the URL when any path is selected, no matter how.

willdurand:
> One thing I noticed, which has nothing to do with this patch, but I noticed when I was testing it and looking at the console, is the fact that `LOAD_VERSION_FILE` is called twice any time a new file is loaded. I'm not sure if that's necessary, or if this is a bug we should investigate.

I think it does that only when loading the URL for the first time. It's because we need the `version` first, which also loads the default file (often `manifest.json`).

willdurand:
It looks like rebasing this branch on top of `master`, which contains the keyboard navigation, revealed a bug. The keyboard navigation does not work anymore 💣

willdurand:
So I don't think there is a better solution than relying on `connected-router` so I am putting this PR on hold and will work on https://github.com/mozilla/addons-code-manager/issues/48 first.

kumar303:
> So I don't think there is a better solution than relying on `connected-router` so I am putting this PR on hold and will work on #48 first.

Good idea. I'm still not sure that `connected-router` will allow us to synchronize the URL to state. I don't see anything in their docs or examples to suggest that it's possible. 

I guess one way would be to always access `selectedPath` from `state.router.location.search.path` instead of accessing it from `version.selectedPath`. Just a thought.

willdurand:
> Good idea. I'm still not sure that `connected-router` will allow us to synchronize the URL to state. I don't see anything in their docs or examples to suggest that it's possible.

Yep, I did not find that either and I have the same concern, but I believe we still need the `connected-router` anyway...

willdurand:
I refactored the code to use `connected-react-router`. Because we need the `versionId` to update the selected path, we cannot listen to `LOCATION_CHANGE` and react accordingly.

Therefore, I went for an action dispatched in the `Browse` component if and only if there is a valid `path` query parameter that is not the current selected path. We'll have to use the same principle in the `Compare` component (which is not taken into account in this patch).

src/reducers/versions.tsx
```diff
@@ -610,6 +611,23 @@ export const fetchDiff = ({
...
+
+export const updateSelectedPath = ({
+  versionId,
+  selectedPath,
+}: UpdateSelectedPathParams): ThunkActionCreator => {
+  return async (dispatch, getState) => {
+    const { router } = getState();
+
+    dispatch(actions.updateSelectedPath({ versionId, selectedPath }));
+    dispatch(push(`${router.location.pathname}?path=${selectedPath}`));
```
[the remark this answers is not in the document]
willdurand: > This should only add the `path` query string param, not replace all others. Could you file a follow-up to address that if it's too involved?

ah good call, I totally missed that :x yeah, we'll have to rebuild our own query param I suppose, but with `query-string` + `location.search`, this should be doable.

src/reducers/versions.tsx
```diff
@@ -610,6 +611,23 @@ export const fetchDiff = ({
   };
 };
 
+type UpdateSelectedPathParams = {
+  versionId: number;
+  selectedPath: string;
+};
+
+export const updateSelectedPath = ({
```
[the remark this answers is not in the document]
willdurand: I like the idea!

src/pages/Browse/index.tsx
```diff
@@ -55,38 +56,49 @@ export class BrowseBase extends React.Component<Props> {
...
+          versionId: parseInt(versionId, 10),
+        }),
+      );
+      return;
+    }
+
+    const path = getPathFromQueryString(history);
+
+    if (path && path !== version.selectedPath) {
+      this.onSelectFile(path);
```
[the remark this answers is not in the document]
willdurand: yep, makes sense ❤️

willdurand:
I rebased this branch and updated the patch according to @kumar303's comments :)

src/reducers/fileTree.tsx
```diff
@@ -199,14 +199,18 @@ export const getRelativePath = ({
...
   _getRelativePath?: typeof getRelativePath;
+  _viewVersionFile?: typeof viewVersionFile;
   currentPath: string;
   pathList: string[];
   position: RelativePathPosition;
   versionId: number;
 };
 
 export const goToRelativeFile = ({
+  /* istanbul ignore next */
```
bobsilverberg: Why do we need these comments? We are exercising these via the tests. Is there some issue with the way the `thunkTester` works that makes it so that they seem not covered even though they are?

src/reducers/versions.tsx
```diff
@@ -610,6 +612,29 @@ export const fetchDiff = ({
   };
 };
 
+type UpdateSelectedPathParams = {
```
bobsilverberg: This should probably be renamed to `ViewVersionFileParams`.

src/reducers/fileTree.tsx
willdurand: We never test the real implementation hence some missing code coverage. This comment should tell the coverage tool to ignore the default assignation.

src/reducers/versions.tsx
willdurand: Ha! :facepalm:

src/reducers/fileTree.tsx
bobsilverberg: Maybe we should have a test that uses the real implementation? Although I don't know what that test would do. 🤷‍♂️

PR #566, opened 2019-04-11
 Open the parent folder(s) when navigating to a new file 

### LINKED ISSUE #562: Open the parent folder(s) when navigating to a new file
Depends on #468 

This is a follow-up to PR #468. When navigating to a new file via a keyboard shortcut, the new file is highlighted in the tree, but if the parent folder was closed, it remains closed, so the user cannot see that the new file is now highlighted.

We should fix this so that any parent folders of the current file are expanded automatically when a new file is selected. It seems like doing this as part of the response to the `updateSelectedPath` action would make the most sense.

bobsilverberg:
Fixes #562 
~~Depends on #468~~

Note that only the final commit in this PR is relevant. Hat tip to @willdurand for the super-simple solution. 🎩

### DIFF
--- a/src/reducers/versions.tsx
+++ b/src/reducers/versions.tsx
@@ -690,20 +690,13 @@ const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
       const { expandedPaths } = version;
 
       // Make sure the root folder is expanded.
-      const parents = [];
-      const rootPath = getRootPath(version);
-      if (!expandedPaths.includes(rootPath)) {
-        parents.push(rootPath);
-      }
+      const parents = [getRootPath(version)];
 
       const folders = selectedPath.split('/');
 
       while (folders.length > 1) {
         folders.pop();
-        const newPath = folders.join('/');
-        if (!expandedPaths.includes(newPath)) {
-          parents.push(newPath);
-        }
+        parents.push(folders.join('/'));
       }
 
       return {
@@ -714,8 +707,8 @@ const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
             ...state.versionInfo[versionId],
             selectedPath,
             expandedPaths: [
-              ...state.versionInfo[versionId].expandedPaths,
-              ...parents,
+              ...expandedPaths,
+              ...parents.filter((newPath) => !expandedPaths.includes(newPath)),
             ],
           },
         },
--- a/src/reducers/versions.tsx
+++ b/src/reducers/versions.tsx
@@ -687,14 +687,23 @@ const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
       const { selectedPath, versionId } = action.payload;
 
       const version = state.versionInfo[versionId];
+      const { expandedPaths } = version;
 
       // Make sure the root folder is expanded.
-      const parents = [getRootPath(version)];
+      const parents = [];
+      const rootPath = getRootPath(version);
+      if (!expandedPaths.includes(rootPath)) {
+        parents.push(rootPath);
+      }
+
       const folders = selectedPath.split('/');
 
       while (folders.length > 1) {
         folders.pop();
-        parents.push(folders.join('/'));
+        const newPath = folders.join('/');
+        if (!expandedPaths.includes(newPath)) {
+          parents.push(newPath);
+        }
       }
 
       return {
--- a/src/reducers/versions.tsx
+++ b/src/reducers/versions.tsx
@@ -686,6 +686,17 @@ const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
     case getType(actions.updateSelectedPath): {
       const { selectedPath, versionId } = action.payload;
 
+      const version = state.versionInfo[versionId];
+
+      // Make sure the root folder is expanded.
+      const parents = [getRootPath(version)];
+      const folders = selectedPath.split('/');
+
+      while (folders.length > 1) {
+        folders.pop();
+        parents.push(folders.join('/'));
+      }
+
       return {
         ...state,
         versionInfo: {
@@ -693,6 +704,10 @@ const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
           [versionId]: {
             ...state.versionInfo[versionId],
             selectedPath,
+            expandedPaths: [
+              ...state.versionInfo[versionId].expandedPaths,
+              ...parents,
+            ],
           },
         },
       };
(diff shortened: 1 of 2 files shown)


willdurand:
> Hmm. Why do we need to dispatch more data? We already have all the data in state to figure this out.

Do we? Instead of using `version.selectedPath.startsWith(node.id)`, it looks like it only adds the needed sub-paths to the state so that it simulates what a user would do.

kumar303:
> Do we dispatch more data?

Dispatch wasn't the right word. I meant to say why do we need to [add more data](https://github.com/mozilla/addons-code-manager/pull/566/files#diff-849d3d7e822fc2a7e48991bf6abc02a2R680) to the state for this? I think we have it all. The components just need to react accordingly. It's just a thought for a simpler approach. It might not work.




> this patch only adds the needed sub-paths to the state, which simulates what a user would do (opening each directory one by one).

We needed to add `expandedPaths` to state for when the user opens a folder that is *not* related to the selected path. For the case of opening all folders leading up to the selected path, we might already have what we need.

bobsilverberg:
I haven't had a chance to test it yet, but I suspect that @kumar303's solution will work, and it's definitely simpler. I'll try it out tomorrow.

willdurand:
Given a path `foo/bar/file.js`, a user who wants to see `file.js` will expand `foo` and then `bar`. These two actions should result in the following list of `expandedPaths`: `['foo', 'bar']` in the state if I understand the existing code correctly. The current patch _seems to guarantee_ that for any path like `foo/bar/file.js` where we need to expand the tree, we'll end up with the exact same list of `expandedPaths`.

I like that the redux state is the same, no matter who (user clicks vs keyboard shortcut vs the URL _via_ https://github.com/mozilla/addons-code-manager/issues/286) initiated the action (the action being "[user] wants to see `file.js`").

bobsilverberg:
@willdurand yes, I was thinking it was good that the value of `expandedPaths` in state will always reflect the folders that are open. But I'm not sure we need that. Anywhere we have access to `expandedPaths`, we also have access to `selectedPath`, so the logic to determine whether a folder is/should be open can make use of that, which is what @kumar303's suggestion entails.

Although I get a "good feeling" about the fact that `expandedPaths` is kept in sync, I'm not sure that there are any other tangible benefits.

What do you think?

kumar303:
I just don't want us to store more data than we need to in the state. In any case, I'm finishing up a review of https://github.com/mozilla/addons-code-manager/pull/468 which will make this patch easier to review.

willdurand:
> I just don't want us to store more data than we need to in the state.

I am sorry but I don't understand. We store the paths when a user toggles directories, so how is that different? The `selectedPath` is used for a different purpose.

Having the exact same state is not just about having a "good feeling" but I feel like everything else is based on that and we assume that paths are stored in the expanded paths, so that we can toggle/collapse any directory. The sketch of patch above will likely break the parent directories of the selected path. A user won't be able to collapse any of these parent directories and I think that'd be nice to keep this feature because for versions with lots of file, it can be useful to quickly close a directory to access other files.

kumar303:
> A user won't be able to collapse any of these parent directories and I think that'd be nice to keep this feature because for versions with lots of file, it can be useful to quickly close a directory to access other files.

I'm not proposing that we break any existing features 😄 

I only offered the diff up above so that it could be considered as a solution since it's simpler and won't introduce additional state logic. It's not a finished solution, just an idea. 

The concept of *the user wants to close the file tree* is currently implemented entirely with `expandedPaths` so I understand that this needs thought if we were to rely on `version.selectedPath` to open a folder. Maybe we could add another state property like `treeIsOpen: boolean` or maybe we just continue to use `expandedPaths` for everything.

bobsilverberg:
I think @willdurand's point about users not being able to close parent folders is a valid one, so I do think that we need to do something along the lines of what I've done in this patch.

Also, this has now been rebased and is ready for a review.

src/reducers/versions.tsx
```diff
@@ -669,13 +669,28 @@ const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
     case getType(actions.updateSelectedPath): {
```
kumar303: It feels wrong that parent directories are only added to `expandedPaths` on `updateSelectedPath` and not when `version.selectedPath` is [first created](https://github.com/mozilla/addons-code-manager/blob/e91ed7269bc47cf15ecae73cb6cce44a1fbb803e/src/reducers/versions.tsx#L639). Adding `selectedPath` to `expandedPaths` at creation time would expand the tree when the page first loads which seems like the right thing to do. I don't understand why anyone would want to start the page with a collapsed tree.

This is not a change request, just me thinking out loud.

bobsilverberg: I agree that we should do this as well, but I take from your final sentence that you feel this could be a follow-up?

bobsilverberg:
Thanks for the review @kumar303. This is read for another look.

src/reducers/versions.tsx
kumar303: Yeah, no change request. I'm not really sure what to do about it yet. I wonder if storing the initial file at `version.defaultFile` would just make everything more clear.

src/reducers/versions.tsx
```diff
@@ -686,13 +686,37 @@ const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
...
       return {
         ...state,
         versionInfo: {
           ...state.versionInfo,
           [versionId]: {
             ...state.versionInfo[versionId],
             selectedPath,
+            expandedPaths: [
+              ...state.versionInfo[versionId].expandedPaths,
+              ...parents,
```
kumar303: How about this?

```js
...parents.filter((newPath) => !expandedPaths.includes(newPath))
```

With that, I think you won't have to duplicate the checks up above.

src/reducers/versions.tsx
```diff
@@ -686,13 +686,37 @@ const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
...
+
       return {
         ...state,
         versionInfo: {
           ...state.versionInfo,
           [versionId]: {
             ...state.versionInfo[versionId],
             selectedPath,
+            expandedPaths: [
+              ...state.versionInfo[versionId].expandedPaths,
```
kumar303: This could use the local `expandedPaths` variable instead

src/reducers/versions.tsx
bobsilverberg: I thought your original thought made a lot of sense. I.e., when we first store `selectedPath` for a version, we also add all of its parents to `expandedPaths`. That seems pretty simple and straightforward.

src/reducers/versions.tsx
bobsilverberg: Nice! Yes, that's better. 👍

Issue #545, opened 2019-04-08
Add CodeShape, a micro representation of code


kumar303:
To condense many lines of code into a small panel (https://github.com/mozilla/addons-code-manager/issues/523), we need a CodeShape component to represent the general shape of how code looks.

PR #567, opened 2019-04-11
Add a CodeLineShapes component

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/545

This component will more or less replace the call to `renderRow()` in [this proof](https://github.com/mozilla/addons-code-manager/pull/542/files#diff-4a3375c826127680658868411c3f42d3R245). The array of line shapes needs to be pre-generated because the `CodeOverview` component may choose to render linter messages *instead* of line shapes. This makes `CodeShape` different from typical components.

Storybook example:

Screenshot 2019-04-11 12 26 48


stories/CodeShape.stories.tsx
```diff
@@ -0,0 +1,79 @@
+import * as React from 'react';
+import { storiesOf } from '@storybook/react';
+
+import CodeShape, { generateLineShapes } from '../src/components/CodeShape';
+
+const exampleCode = `\
+browser.runtime.onMessage.addListener(
```
willdurand: This looks like the longest line but it is only 39 chars. Could we have a story with a line that has more chars than the `maxLineLength`?

stories/setup/styles.scss
```diff
@@ -1,3 +1,41 @@
...
+
+  &:first-child {
+    margin-top: 0;
+  }
+}
+
+.CodeShapeStory-actualCode {
+  overflow: hidden;
+
+  pre {
```
willdurand: We tend to avoid referencing html tags in sass. Maybe we could introduce a className for that too? It could be `.CodeShapeStory-actualCode-line`.

src/components/CodeShape/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+  return allLineShapes;
+};
+
+export type PublicProps = {
+  lineShapes: LineShapes;
+};
+
+type Props = PublicProps;
+
+const CodeShape = ({ lineShapes }: Props) => {
```
willdurand: I thought this component was used to render a whole file but it only renders a shape for a single line of code, so maybe `Line` could be part of the name of the component?

src/components/CodeShape/index.spec.tsx
```diff
@@ -0,0 +1,232 @@
...
+            }),
+            // Check 'BB', the final 25%
+            expect.objectContaining({ token: Token.code, percentOfWidth: 25 }),
+          ],
+        },
+      ]);
+    });
+
+    it('respects maxLineLength when counting tokens', () => {
+      const line = 'AA  BBB ; the rest will be ignored';
```
willdurand: Oh, so the code is not scaled down according to `maxLineLength`? It is truncated instead, is that correct?

src/components/CodeShape/index.spec.tsx
```diff
@@ -0,0 +1,232 @@
...
+          line: 1,
+          tokens: [
+            // Expect 'AA' to be 2
+            expect.objectContaining({ token: Token.code, count: 2 }),
+            // Expect '  ' to be 2
+            expect.objectContaining({ token: Token.whitespace, count: 2 }),
+            // Expect 'BBB' to be 3
+            expect.objectContaining({ token: Token.code, count: 3 }),
+            // Expect ' ' to be 1
+            expect.objectContaining({ token: Token.whitespace, count: 1 }),
```
willdurand: I must say that each comment above each `expect` in all the test cases are ❤️

src/components/CodeShape/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+// This is a collection of shapes for one line of code.
+export type LineShapes = {
+  line: number;
+  tokens: TokenShape[];
+};
+
+// These are shape collections for all lines of code.
+export type AllLineShapes = LineShapes[];
+
+export const generateLineShapes = (
```
willdurand: Could we move this function to `CodeShape/utils.tsx`? I don't see it used in the component and that's (in part) why I thought the component would render shapes for all the lines of a file.

src/components/CodeShape/index.spec.tsx
kumar303: Yes, it's truncated. I think this is a good idea for minified files that have really long lines.

src/components/CodeShape/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+        return (
+          <div
+            key={[
+              String(lineShapes.line),
+              String(shape.percentOfWidth),
+              className,
+            ].join(':')}
+            className={className}
+            style={{ width: `${shape.percentOfWidth}%` }}
```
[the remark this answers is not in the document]
kumar303: I think it will work. We do something [similar](https://github.com/mozilla/addons-frontend/blob/5c2db379c80a5ab1d0830c6fab5b49dd11ec3e6a/src/ui/components/LoadingText/index.js#L42) in `LoadingText` on the frontend without CSP conflicts.

src/components/CodeShape/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+export type PublicProps = {
+  lineShapes: LineShapes;
+};
+
+type Props = PublicProps;
+
+const CodeShape = ({ lineShapes }: Props) => {
+  return (
+    <React.Fragment>
```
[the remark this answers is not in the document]
kumar303: > ...maybe it would make sense to have a `div` or `span` to enclose the logic of this component 🤷‍♂️

The container needs to be a flex box so that everything expands and the `CodeOverview` component will need to control this so the linter messages look consistent. If I add a default container that may not work.

src/components/CodeShape/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+export enum Token {
+  code = 'code',
+  whitespace = 'whitespace',
+}
+
+// This represents the shape of one or more tokens.
+export type TokenShape = {
+  count: number;
+  percentOfWidth: number;
+  token: Token.code | Token.whitespace;
```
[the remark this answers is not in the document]
kumar303: Huh, that works. I was getting confusing errors before saying that `Token` is incompatible. Oh, well, they're gone now.

stories/CodeShape.stories.tsx
kumar303: If you keep scrolling in *all variants* there is one already:

Screenshot 2019-04-12 10 32 57

stories/setup/styles.scss
kumar303: I think that was just for performance reasons, right? Since this is just storybook css, it shouldn't be a problem. 

I wasn't totally sure where this storybook css should go.

src/components/CodeShape/index.tsx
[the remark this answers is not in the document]
kumar303: > Do you even need this? I thought components could return arrays nowadays

Enzyme doesn't support arrays, it seems.

src/components/CodeShape/index.tsx
kumar303: Sure. `CodeLineShapes` is all I could think of. I didn't want it to conflict with the `LineShapes` type since that's a pain with `LinterMessage`.

kumar303:
@willdurand thanks for the review. I think I addressed all questions but let me know.

src/components/CodeShape/index.tsx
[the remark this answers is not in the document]
willdurand: d'oh 😞

stories/CodeShape.stories.tsx
willdurand: ah oops, I saw it actually. Sorry.

src/components/CodeShape/index.spec.tsx
willdurand: Sure, I think it's simpler than scaling down the code.

src/components/CodeLineShapes/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+        return (
+          <div
+            key={[
+              String(lineShapes.line),
+              String(shape.percentOfWidth),
+              className,
+            ].join(':')}
+            className={className}
+            style={{ width: `${shape.percentOfWidth}%` }}
```
willdurand: I am still a bit worried that this will be a problem with CSP but maybe not.. We'll see soon.

PR #571, opened 2019-04-12
Fix log call in FileTree

### LINKED ISSUE #563: In `FileTree`, call to `_log` in `mapStateToProps()` will result in an error
The `_log` prop is not available in `ownProps` because this is set as a default prop inside the (base) component. In `mapStateToProps()`, `ownProps` are the props received by the component, not those set inside the base component (because `connect()` wraps the base component).

Instead of logging a warning, an error will be thrown because `_log` is `undefined`.

willdurand:
Fixes #563 

---

I removed the test case because I could not find an easy way to fix it... 🙈

### DIFF
--- a/src/components/FileTree/index.spec.tsx
+++ b/src/components/FileTree/index.spec.tsx
@@ -252,14 +252,6 @@ describe(__filename, () => {
       expect(isNodeExpanded(node)).toBeFalsy();
     });
 
-    it('logs a warning message when no version is loaded', () => {
-      const _log = createFakeLogger();
-
-      render({ _log });
-
-      expect(_log.warn).toHaveBeenCalled();
-    });
-
     it('renders a Loading component when no version is loaded', () => {
       const root = render();
 
--- a/src/components/FileTree/index.tsx
+++ b/src/components/FileTree/index.tsx
@@ -178,7 +178,7 @@ const mapStateToProps = (
   state: ApplicationState,
   ownProps: PublicProps & DefaultProps,
 ): PropsFromState => {
-  const { _log, versionId } = ownProps;
+  const { versionId } = ownProps;
   const version = getVersionInfo(state.versions, versionId);
 
   if (!version) {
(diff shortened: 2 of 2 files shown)


src/components/FileTree/index.tsx
```diff
@@ -178,15 +178,15 @@ const mapStateToProps = (
   state: ApplicationState,
   ownProps: PublicProps & DefaultProps,
```
kumar303: This seems more realistic and it would have produced a TS error:

```diff
diff --git a/src/components/FileTree/index.tsx b/src/components/FileTree/index.tsx
index d394eaa..9ceb8ad 100644
--- a/src/components/FileTree/index.tsx
+++ b/src/components/FileTree/index.tsx
@@ -176,7 +176,7 @@ export class FileTreeBase extends React.Component<Props> {
 
 const mapStateToProps = (
   state: ApplicationState,
-  ownProps: PublicProps & DefaultProps,
+  ownProps: PublicProps & Partial<DefaultProps>,
 ): PropsFromState => {
   const { versionId } = ownProps;
   const version = getVersionInfo(state.versions, versionId);
```

src/components/FileTree/index.spec.tsx
```diff
@@ -252,14 +252,6 @@ describe(__filename, () => {
       expect(isNodeExpanded(node)).toBeFalsy();
     });
 
-    it('logs a warning message when no version is loaded', () => {
```
kumar303: Removing this is fine by me. The execution of the if statement is already covered by five other tests.

src/components/FileTree/index.tsx
kumar303: I checked all other components and I don't see any incorrect types for `ownProps`. Typically we set it to `PublicProps` to be safe.

willdurand: ah right, but here it should be `ownProps: PublicProps` only as well.

Issue #559, opened 2019-04-10
Clicking on a line of code should go to that line


kumar303:
### Describe the problem and steps to reproduce it:

- Sign and load http://localhost:3000/en-US/browse/494431/versions/1532144/
- Click Line 13 of the code

### What happened?

Nothing happens

### What did you expect to happen?

Line 13 should be selected

### Anything else we should know?

If you reload the page after clicking, line 13 will appear selected. This is how it should behave.

Line anchors were added in https://github.com/mozilla/addons-code-manager/issues/317

willdurand:
Marking this issue as regression because it used to work.

willdurand:
I think the problem comes from `LinterProvider`, which blocks the update of the rest of the `CodeView` component. Clicking an anchor does update the routing data (i.e. `locationhash`) but because nothing else changes, the `CodeView` does not update the selected line...

kumar303:
ok, I see what the problem is. I need it for https://github.com/mozilla/addons-code-manager/issues/523 so I'll fix this.

AlexandraMoga:
Verified on -dev with FF66, WIn10x64 for both the Browse and Compare views.

Lines of code are immediately selected when clicking on them:

![2019-04-16_16-03-48](https://user-images.githubusercontent.com/31961530/56212102-06e14980-6062-11e9-8a5e-92be4c09952a.gif)

PR #584, opened 2019-04-15
Always re-render LinterProvider on location change

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/559

### DIFF
--- a/src/components/LinterProvider/index.spec.tsx
+++ b/src/components/LinterProvider/index.spec.tsx
@@ -10,6 +10,7 @@ import {
   getMessageMap,
 } from '../../reducers/linter';
 import {
+  createContextWithFakeRouter,
   createFakeExternalLinterResult,
   createFakeThunk,
   fakeVersion,
@@ -47,11 +48,19 @@ describe(__filename, () => {
       ...moreProps,
     };
 
+    const shallowOptions = createContextWithFakeRouter();
+
     return shallowUntilTarget(
       <LinterProvider {...props}>{children}</LinterProvider>,
       LinterProviderBase,
       {
-        shallowOptions: { context: { store } },
+        shallowOptions: {
+          ...shallowOptions,
+          context: {
+            ...shallowOptions.context,
+            store,
+          },
+        },
       },
     );
   };
--- a/src/components/LinterProvider/index.tsx
+++ b/src/components/LinterProvider/index.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { connect } from 'react-redux';
+import { withRouter, RouteComponentProps } from 'react-router-dom';
 
 import { ApplicationState } from '../../reducers';
 import { ConnectedReduxProps } from '../../configureStore';
@@ -37,7 +38,8 @@ type PropsFromState = LinterProviderInfo;
 export type Props = PublicProps &
   DefaultProps &
   PropsFromState &
-  ConnectedReduxProps;
+  ConnectedReduxProps &
+  RouteComponentProps;
 
 export class LinterProviderBase extends React.Component<Props> {
   loadData: LoadData;
@@ -93,7 +95,7 @@ export class LinterProviderBase extends React.Component<Props> {
 
 const mapStateToProps = (
   state: ApplicationState,
-  ownProps: PublicProps,
+  ownProps: PublicProps & RouteComponentProps,
 ): PropsFromState => {
   const { selectedPath, versionId } = ownProps;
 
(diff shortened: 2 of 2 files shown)


src/components/LinterProvider/index.tsx
```diff
@@ -113,6 +115,8 @@ const mapStateToProps = (
   };
 };
 
-export default connect(mapStateToProps)(
-  LinterProviderBase,
-) as React.ComponentType<PublicProps & Partial<DefaultProps>>;
```
kumar303: I removed this type hint and tested it by rendering with unknown props. I got the expected type error. I'm not sure why but... 🎉

bobsilverberg: Interesting. I just added something like this to a recent patch at your request. Is it somehow specific to the component?

src/components/LinterProvider/index.spec.tsx
```diff
@@ -47,11 +48,19 @@ describe(__filename, () => {
...
     };
 
+    const shallowOptions = createContextWithFakeRouter();
+
     return shallowUntilTarget(
       <LinterProvider {...props}>{children}</LinterProvider>,
       LinterProviderBase,
       {
-        shallowOptions: { context: { store } },
+        shallowOptions: {
```
bobsilverberg: This fixes the test to work with the router, but do we also want a test that proves that the bug this patch fixes is in fact fixed? Is it possible to write a test for that?

src/components/LinterProvider/index.tsx
kumar303: Well, previously, the loss of type safety only affected `connect()`. For `withRouter`, it worked unless you were wrapping `connect()`, of course.

willdurand:
> This looks good, although it feels a bit funny to me. It could be that I don't understand the issue this is fixing. Is it that the child of the `LinterProvider` isn't being notified of an update because `mapStateToProps` is being memoized at the level of the `LinterProvider`?

`connect()` implements a `shouldComponentUpdate()` function for each component it decorates. In the case of `LinterProvider`, updating the routing data would not impact the component because the component does not subscribe to it (via `mapStateToProps()`). By decorating the connected component with `withRouter`, we inject the routing data and change the behavior of the redux's `shouldComponentUpdate()` function.

kumar303:
> This looks good, although it feels a bit funny to me. It could be that I don't understand the issue this is fixing. Is it that the child of the `LinterProvider` isn't being notified of an update because `mapStateToProps` is being memoized at the level of the `LinterProvider`?

Yes. The `react-redux` docs recently got an overhaul and they explain it much better now: https://react-redux.js.org/using-react-redux/connect-mapstate#return-values-determine-if-your-component-re-renders

The behavior my patch relies on is that `ownProps` contains `location` and new values will automatically trigger a re-render.

src/components/LinterProvider/index.spec.tsx
kumar303: I think it would be tricky to write a test for this given our current tools and testing approach. We would need a higher level functional test to cover it and we don't have a framework for that yet.

I agree that a test would be good but the Redux API is very stable so I think it's OK to rely on this fix via configuration. 

The risks for regression here are if someone removes the configuration without knowing it. I had originally put a comment above `withRouter()` to prevent someone from removing. You made me think that we need another comment above the `ownProps` definition, too, since removing that would also break the configuration. I just added a comment there.

kumar303:
@bobsilverberg thanks, I addressed your comments.

src/components/LinterProvider/index.spec.tsx
willdurand: I tend to agree that writing an integration test for that would be tricky.

In other news, I filed https://github.com/mozilla/addons-code-manager/issues/585 for removing more type hints!

bobsilverberg:
> Yes. The `react-redux` docs recently got an overhaul and they explain it much better now: https://react-redux.js.org/using-react-redux/connect-mapstate#return-values-determine-if-your-component-re-renders
>
> The behavior my patch relies on is that `ownProps` contains `location` and new values will automatically trigger a re-render.

I guess what I'm confused about is that it didn't seem to me that we needed `LinterProvider` to re-render, but rather than we needed its child to re-render. Does the fact that `LinterProvider` was not re-rendering prevent the child from re-rendering on its own?

kumar303:
> Does the fact that `LinterProvider` was not re-rendering prevent the child from re-rendering on its own?

Yes, exactly. Internally, `LinterProvider` was not returning anything new so React didn't even know that one of its children changed.

PR #589, opened 2019-04-15
Open parent folders for the default file

### LINKED ISSUE #586: Open parent folders for the default file
This is a follow-up to #566 

When we first load a version, and set the `selectedPath` to whatever the API returned, we should also add the parent folders for that file to `expandedPaths` so that the tree appears expanded to that point. We know for now that it will always just be a file in the root folder, so we could just make sure that the root folder is added to `expandedPaths`, but I think it's worth implementing a proper solution which takes into account the value of `selectedPath`.

bobsilverberg:
Fixes #586

### DIFF
--- a/src/reducers/fileTree.tsx
+++ b/src/reducers/fileTree.tsx
@@ -4,7 +4,7 @@ import { ActionType, createAction, getType } from 'typesafe-actions';
 
 import { Version, viewVersionFile } from './versions';
 import { ThunkActionCreator } from '../configureStore';
-import { getLocalizedString } from '../utils';
+import { getLocalizedString, LocalizedStringMap } from '../utils';
 
 type FileNode = {
   id: string;
@@ -43,19 +43,19 @@ const recursiveSortInPlace = (node: DirectoryNode): void => {
   });
 };
 
-const getVersionName = (version: Version) => {
-  return getLocalizedString(version.addon.name);
+const getVersionName = (versionAddonName: LocalizedStringMap) => {
+  return getLocalizedString(versionAddonName);
 };
 
-export const getRootPath = (version: Version) => {
-  return `root-${getVersionName(version)}`;
+export const getRootPath = (versionAddonName: LocalizedStringMap) => {
+  return `root-${getVersionName(versionAddonName)}`;
 };
 
 export const buildFileTreeNodes = (version: Version): DirectoryNode => {
   const { entries } = version;
   const root: DirectoryNode = {
-    id: getRootPath(version),
-    name: getVersionName(version),
+    id: getRootPath(version.addon.name),
+    name: getVersionName(version.addon.name),
     children: [],
   };
 
--- a/src/reducers/fileTree.tsx
+++ b/src/reducers/fileTree.tsx
@@ -4,7 +4,9 @@ import { ActionType, createAction, getType } from 'typesafe-actions';
 
 import { Version, viewVersionFile } from './versions';
 import { ThunkActionCreator } from '../configureStore';
-import { getLocalizedString, LocalizedStringMap } from '../utils';
+import { getLocalizedString } from '../utils';
+
+export const ROOT_PATH = '~root~';
 
 type FileNode = {
   id: string;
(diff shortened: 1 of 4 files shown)


src/reducers/fileTree.tsx
```diff
@@ -43,19 +43,19 @@ const recursiveSortInPlace = (node: DirectoryNode): void => {
...
-const getVersionName = (version: Version) => {
-  return getLocalizedString(version.addon.name);
+const getVersionName = (versionAddonName: LocalizedStringMap) => {
+  return getLocalizedString(versionAddonName);
 };
 
-export const getRootPath = (version: Version) => {
-  return `root-${getVersionName(version)}`;
+export const getRootPath = (versionAddonName: LocalizedStringMap) => {
+  return `root-${getVersionName(versionAddonName)}`;
```
willdurand: Why is `getVersionName()` kept here?

src/reducers/fileTree.tsx
```diff
@@ -43,19 +43,19 @@ const recursiveSortInPlace = (node: DirectoryNode): void => {
   });
 };
 
-const getVersionName = (version: Version) => {
-  return getLocalizedString(version.addon.name);
+const getVersionName = (versionAddonName: LocalizedStringMap) => {
+  return getLocalizedString(versionAddonName);
 };
```
willdurand: Why did you change this function? I like that it gives the name of version based on a version object. Now, all it does is returning a localized string for `version.addon.name`, which could be done either directly (if only used in `getRootPath`) or done in a `getAddonName` function instead.

willdurand: But I think I like this `getVersionName` function because it encapsulates the logic for retrieving a good version name.

src/reducers/fileTree.tsx
bobsilverberg: I want to be able to call `getRootPath` without having to pass in a complete `Version` object, because we're now calling it from `createInternalVersion`.

src/reducers/fileTree.tsx
bobsilverberg: I'm a bit confused. Have you answered all of your own questions above? I changed the function because of the changes to `getRootPath`.

bobsilverberg:
@willdurand I'm not sure if I've answered your questions, but I'm also not sure if there are any changes being requested to the patch.

src/reducers/fileTree.tsx
willdurand: Sorry, I like the original version of `getVersionName`. If you can't pass a version to it anymore, I don't think it makes a lot of sense to keep such a function.

src/reducers/fileTree.tsx
willdurand: I see. Could you inline the content of `getVersionName` here? And get rid of `getVersionName` entirely?

bobsilverberg: I do also call it in `buildFileTreeNodes`, so I'd have to inline the logic twice. That's why it exists.

willdurand: The `getVersionName` function should be renamed to `getAddonName` if you want to keep it, but it's confusing because this function will receive the addon's name, so calling `getLocalizedString` directly seems to make more sense to me.

bobsilverberg: I think I'd rather rename it to `getAddonName` than have to call `getLocalizedString` in two different places. I'll do that.

bobsilverberg:
Ok @willdurand, I renamed the function. Are there any other changes required?

src/reducers/fileTree.tsx
```diff
@@ -43,19 +43,19 @@ const recursiveSortInPlace = (node: DirectoryNode): void => {
...
-const getVersionName = (version: Version) => {
-  return getLocalizedString(version.addon.name);
+const getAddonName = (versionAddonName: LocalizedStringMap) => {
+  return getLocalizedString(versionAddonName);
 };
 
-export const getRootPath = (version: Version) => {
-  return `root-${getVersionName(version)}`;
+export const getRootPath = (versionAddonName: LocalizedStringMap) => {
+  return `root-${getAddonName(versionAddonName)}`;
```
willdurand: So, that's probably a silly question but why do we need the name of the addon here? Couldn't we return `root` only? I am thinking about that because I am uncertain about the changes related to this function here.

bobsilverberg: That's a valid question, which I hadn't even considered. The answer appears to be "no", which definitely simplifies things. Thanks for asking that, @willdurand!

bobsilverberg:
Ok, I think this is looking much better now. I also rebased it on top of the latest changes.

PR #594, opened 2019-04-16
Preserve query parameters and hash in CodeView/Browse

### LINKED ISSUE #593: The file path is not preserved in the URL when a line of code is selected inside the file viewer
Follow up for #286 

**STR:**
1. Open the browse file view for an add-on - see https://code.addons-dev.allizom.org/en-US/browse/616653/versions/1655542/?path=options.js
2. Make sure that the `?path` to file is present in the URL
3. Select a line of code inside the opened file

**Actual result:**
The  `?path` to file disappears from the URL

**Expected result:**
The  `?path` to file is preserved, in case reviewers will need to share a specific line of code for a selected file

**Notes:**
- reproduced on -dev with FF66, Win10x64

willdurand:
Fixes #593

---

This patch makes sure the query string is kept in the `CodeView` line anchors. It also ensures we can load any file with a specific line selected. Last but not least, when selecting a new file, it removes the `hash` from the URL.

### DIFF
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -1521,7 +1521,11 @@ describe(__filename, () => {
         }),
       );
       expect(dispatch).toHaveBeenCalledWith(
-        push(`${pathname}?path=${selectedPath}`),
+        push({
+          ...location,
+          search: `?path=${selectedPath}`,
+          hash: undefined,
+        }),
       );
     });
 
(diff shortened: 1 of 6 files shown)


src/reducers/versions.spec.tsx
```diff
@@ -1521,7 +1521,11 @@ describe(__filename, () => {
         }),
       );
       expect(dispatch).toHaveBeenCalledWith(
-        push(`${pathname}?path=${selectedPath}`),
+        push({
+          ...location,
+          search: `?path=${selectedPath}`,
+          hash: undefined,
```
bobsilverberg: Do we need `hash: undefined` here, or are you just doing it to make the test more explicit? The code does a `delete newLocation.hash` so I wouldn't expect that our assert would need to include `hash: undefined`.

src/reducers/versions.spec.tsx
```diff
@@ -1538,7 +1542,54 @@ describe(__filename, () => {
       await thunk();
 
       expect(dispatch).toHaveBeenCalledWith(
-        push(`${pathname}${search}&path=${selectedPath}`),
+        push({
+          ...location,
+          search: `${search}&path=${selectedPath}`,
+          hash: undefined,
```
bobsilverberg: As above, do we need `hash: undefined` here, or are you just doing it to make the test more explicit? The code does a `delete newLocation.hash` so I wouldn't expect that our assert would need to include `hash: undefined`.

src/reducers/versions.spec.tsx
willdurand: I did that because I spread `location` above.

src/reducers/versions.spec.tsx
willdurand: Same here, I spread `location`.

willdurand:
> This is great and it works well @willdurand. Just a couple of questions on the tests, and I also have a general note about src/pages/Browse/index.spec.tsx: You have updated some existing tests to assert that preserveHash is passed with the appropriate value in the cases where it matters, but there is no explicit test (explicitly named) for this new preserveHash behaviour. I think that's fine, as I believe the changes you made give us all the coverage we need, but I just thought I'd point it out as it's a bit unusual to add a feature and not end up with any tests that mention the feature.

I have updated one test case description to indicate when we don't want to preserve the URL hash (which is when user clicks on a file in the tree).

PR #596, opened 2019-04-16
Compare load file

### LINKED ISSUE #551: Page components should not always dispatch an action on mount

### LINKED ISSUE #590: Link to a URL for viewing a single file (compare page)
This is similar to #286 but for the compare page, e.g,  https://code.addons-dev.allizom.org/en-US/compare/502955/versions/1541794...1541798/.

willdurand:
Fixes #590
Fixes #551
~~⚠️ depends on #593~~

---

I found new bugs after having fixed other unrelated bugs and rebased 😞 Now, both `Browse` and `Compare` components have a similar behavior and the keyboard navigation works for both as intended and without any side effect (hopefully).

It seemed easier to fix #551 as part of this patch, but it resulted in a bigger diff than expected. Sorry about that..

### DIFF
--- a/src/pages/Browse/index.spec.tsx
+++ b/src/pages/Browse/index.spec.tsx
@@ -480,9 +480,7 @@ describe(__filename, () => {
     expect(dispatch).not.toHaveBeenCalled();
   });
 
-  // This could happen when a keyboard navigation updates the selected path. We
-  // do not want to update the selected path again, so we apply this logic to
-  // the first render (mount) only.
+  // This could happen when a keyboard navigation updates the selected path.
   it('does not dispatch viewVersionFile on update if the query string contains a `path` that is different than `version.selectedPath`', () => {
     const { renderAndUpdate } = setUpVersionFileUpdate({
       loadVersionAndFile: true,
--- a/src/pages/Browse/index.spec.tsx
+++ b/src/pages/Browse/index.spec.tsx
@@ -480,8 +480,11 @@ describe(__filename, () => {
     expect(dispatch).not.toHaveBeenCalled();
   });
 
-  it('dispatches viewVersionFile on update if the query string contains a `path` that is different than `version.selectedPath`', () => {
-    const { renderAndUpdate, version } = setUpVersionFileUpdate({
+  // This could happen when a keyboard navigation updates the selected path. We
+  // do not want to update the selected path again, so we apply this logic to
+  // the first render (mount) only.
+  it('does not dispatch viewVersionFile on update if the query string contains a `path` that is different than `version.selectedPath`', () => {
+    const { renderAndUpdate } = setUpVersionFileUpdate({
       loadVersionAndFile: true,
     });
     const path = 'a/different/file.js';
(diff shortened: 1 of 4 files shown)


kumar303:
> It seemed easier to fix #551 as part of this patch, but it resulted in a bigger diff than expected. Sorry about that..

It doesn't look too big. +/- 400 changes is roughly the limit, I'd say. This patch is under that.

willdurand:
> It doesn't look too big. +/- 400 changes is roughly the limit, I'd say. This patch is under that.

Sure but the diff mixes changes for two different issues, which does not make it easy to follow. Especially the test file has changed quite a lot.

src/pages/Browse/index.spec.tsx
```diff
@@ -480,8 +480,11 @@ describe(__filename, () => {
     expect(dispatch).not.toHaveBeenCalled();
   });
 
-  it('dispatches viewVersionFile on update if the query string contains a `path` that is different than `version.selectedPath`', () => {
-    const { renderAndUpdate, version } = setUpVersionFileUpdate({
+  // This could happen when a keyboard navigation updates the selected path. We
+  // do not want to update the selected path again, so we apply this logic to
+  // the first render (mount) only.
```
kumar303: Oh, you do have a comment 😄 Can you move this to the `loadData()` method? That's a better place for it.

src/pages/Compare/index.tsx
```diff
@@ -66,45 +73,55 @@ export class CompareBase extends React.Component<Props> {
...
-      (prevProps.path && path !== prevProps.path) ||
+    if (!prevProps) {
+      if (path && path !== version.selectedPath) {
+        // We preserve the hash in the URL (if any) when we load the file from
+        // an URL that has likely be shared.
+        this.viewVersionFile(path, { preserveHash: true });
+      }
+    } else if (
+      (prevProps.version &&
+        version.selectedPath !== prevProps.version.selectedPath) ||
```
kumar303: I guess nothing here is broken but we should update this someday to not use `prevProps` values for comparison. It is fragile. The best way to do it is to compare Redux mapped properties.

src/pages/Compare/index.spec.tsx
```diff
@@ -435,63 +445,148 @@ describe(__filename, () => {
...
+      addonId,
+      baseVersionId,
+      headVersionId,
+      path,
+    });
+  });
+
+  // This could happen when a keyboard navigation updates the selected path. We
+  // do not want to update the selected path again, so we apply this logic to
+  // the first render (mount) only.
```
kumar303: This comment is helpful but I think it should move to `loadData()` where the logic happens.

willdurand:
> It's a shame there is so much duplication between this and the `Browse` component.

Yes 😞

PR #600, opened 2019-04-17
Make sure we run fetchLinterMessages only once

### LINKED ISSUE #591: Too many actions are dispatched when loading a file by path

1. Open the Redux devtools
2. Load http://localhost:3000/en-US/browse/494431/versions/1532144/?path=notification.js

### What happened?

There are lots of `BEGIN_LINTER_RESULT` and `LOAD_LINTER_RESULT` actions dispatched.

Only one `BEGIN_LINTER_RESULT` and one `LOAD_LINTER_RESULT` are dispatched.

The default file is not yet affected because of #586.

It looks like we cannot prevent all but the first action to be dispatched even though we have a loading flag in the state to avoid this issue... Thunks are asynchronous and that's probably the reason.

This diff "fixes" the issue:

```diff
diff --git a/src/reducers/linter.tsx b/src/reducers/linter.tsx
index 4b41911..d7d0239 100644
--- a/src/reducers/linter.tsx
+++ b/src/reducers/linter.tsx
@@ -177,7 +177,12 @@ export const fetchLinterMessages = ({
   url: string;
   versionId: number;
 }): ThunkActionCreator => {
-  return async (dispatch) => {
+  return async (dispatch, getState) => {
+    if (getState().linter.isLoading) {
+      log.info('early return');
+      return;
+    }
+
     dispatch(actions.beginFetchLinterResult({ versionId }));
 
     // This is a special URL and returns a non-standard JSON response.
```

willdurand:
Fixes #591

---

This patch fixes an issue with the `LinterProvider` dispatching too many asynchronous actions.

### DIFF
--- a/src/reducers/linter.spec.tsx
+++ b/src/reducers/linter.spec.tsx
@@ -22,6 +22,7 @@ import linterReducer, {
 } from './linter';
 import { actions as errorsActions } from './errors';
 import { makeApiURL } from '../api';
+import configureStore from '../configureStore';
 
 describe(__filename, () => {
   const createExternalLinterResult = (
@@ -500,6 +501,22 @@ describe(__filename, () => {
       expect(dispatch).toHaveBeenCalledWith(
         actions.abortFetchLinterResult({ versionId }),
       );
+    });
+
+    it('early returns and does not do anything when linter messages are already being fetched', async () => {
+      const url = '/some/url';
+      const versionId = 123;
+      const store = configureStore();
+      store.dispatch(actions.beginFetchLinterResult({ versionId }));
+
+      const { dispatch, thunk } = thunkTester({
+        createThunk: () => fetchLinterMessages({ url, versionId }),
+        store,
+      });
+
+      await thunk();
+
+      expect(dispatch).not.toHaveBeenCalled();
     });
   });
 
--- a/src/reducers/linter.tsx
+++ b/src/reducers/linter.tsx
@@ -177,7 +177,12 @@ export const fetchLinterMessages = ({
   url: string;
   versionId: number;
 }): ThunkActionCreator => {
-  return async (dispatch) => {
+  return async (dispatch, getState) => {
+    // See: https://github.com/mozilla/addons-code-manager/issues/591
+    if (getState().linter.isLoading) {
+      return;
+    }
+
     dispatch(actions.beginFetchLinterResult({ versionId }));
 
     // This is a special URL and returns a non-standard JSON response.


willdurand:
I filed https://github.com/mozilla/addons-code-manager/issues/601 for the other thunks.

willdurand:
> Was there a reason you didn't like this approach when you first suggested it at [#591 (comment)](https://github.com/mozilla/addons-code-manager/issues/591#issue-433475862)?

I did not want to overlook anything, so I spent some time reading and seeing what others have done to fix this issue. Finding the recommended ways of doing things helped a lot in being sure that this was the right approach.

src/reducers/linter.tsx
```diff
@@ -177,7 +177,12 @@ export const fetchLinterMessages = ({
   url: string;
   versionId: number;
 }): ThunkActionCreator => {
-  return async (dispatch) => {
+  return async (dispatch, getState) => {
+    // See: https://github.com/mozilla/addons-code-manager/issues/591
+    if (getState().linter.isLoading) {
```
kumar303: Please remove the `isLoading` logic from [LinterProvider](https://github.com/mozilla/addons-code-manager/blob/50be0327f1ae670d743a570855d23bef3eff14c3/src/components/LinterProvider/index.tsx#L75) (and tests) because we no longer need it.

kumar303: This should only abort loading if `getState().forVersionId === versionId` is also true. Consider the compare page. Imagine that the user is viewing a version but the API for linter messages is slow to respond. Before it responds, the user switches to a new version. This should trigger a re-fetch of the linter messages. This seems likely because one could switch versions in the select box quickly.

src/reducers/linter.spec.tsx
```diff
@@ -501,6 +502,22 @@ describe(__filename, () => {
...
     });
+
+    it('early returns and does not do anything when linter messages are already being fetched', async () => {
+      const url = '/some/url';
+      const versionId = 123;
+      const store = configureStore();
+      store.dispatch(actions.beginFetchLinterResult({ versionId }));
+
+      const { dispatch, thunk } = thunkTester({
+        createThunk: () => fetchLinterMessages({ url, versionId }),
```
kumar303: What about renaming `fetchLinterMessages` to `fetchLinterMessagesIfNeeded`? I think it would help. When looking at [this code](https://github.com/mozilla/addons-code-manager/blob/50be0327f1ae670d743a570855d23bef3eff14c3/src/components/LinterProvider/index.tsx#L77) it's not intuitive that the fetching action might *do nothing*. The name can provide a hint about what's going on.

src/reducers/linter.tsx
kumar303: Could you also log a warning here? It could be surprising that nothing is happening when the action is dispatched.

willdurand: I thought about doing this but the existing logic avoids a `dispatch` call on update, so that still seems worthy, no?

willdurand: yep, although I'll rather log a `debug` message.

willdurand: Answering myself: controlling the `isLoading` logic in one location is actually better.

willdurand: Thanks, I missed this condition. I wonder why we did not have it in the `LinterProvider` code though, any ideas?

kumar303: > Thanks, I missed this condition. I wonder why we did not have it in the `LinterProvider` code though, any ideas?

Because the mapper already factored in version mismatches when creating `messageMap` (it would be undefined for mismatches): 

https://github.com/mozilla/addons-code-manager/blob/50be0327f1ae670d743a570855d23bef3eff14c3/src/components/LinterProvider/index.tsx#L105

Issue #565, opened 2019-04-11
Infinite loop when fetching a file has errored


willdurand:
In the `versions` reducer, `fetchVersionFile` dispatches `abortFetchVersionFile` when an API error is returned. The reducer reacts to this action by setting the `loading` flag to `false`.

This is problematic because a component (like `Browse`) won't know that an error has occured and will retry to fetch the file given that `file` is not defined and `fileIsLoading` is `false`. This behavior will be triggered again and again, hence the infinite loop.

willdurand:
Maybe we should always use this convention: https://github.com/mozilla/addons-code-manager/blob/d7e02107869b4b2a5f63e5abdfb32af3d07f9f6e/src/reducers/versions.tsx#L257-L259 so that we could detect API errors and avoid re-fetching forever.

kumar303:
In `addons-frontend`, it was always possible to see if a saga had produced an error or not. All loaders would check for an error before loading. Hmm.

willdurand:
@ioanarusiczki added QA instructions

willdurand:
>  As a sidenote, I don't think we should have a 500 error in this case, which is why I've filed [mozilla/addons-server#11196](https://github.com/mozilla/addons-server/issues/11196) a little while ago.

:+1:

PR #603, opened 2019-04-17
Dispatch begin/abort actions when fetching a version and avoid infinite fetch/error loop

### LINKED ISSUE #556: `fetchFile` should dispatch `beginFetchVersionFile` before loading version info/file
While working on [a different patch for the `Browse` component](https://github.com/mozilla/addons-code-manager/issues/551#issuecomment-481714926), I discovered an inconsistency between `fetchVersion` and `fetchVersionFile`: `fetchVersion` does not set the `loading` boolean to `true` before loading the file.

This might be a problem given that the `Browse` component tries to fetch a file (by calling `fetchVersionFile`) when the loading state is `false`. We should probably dispatch `beginFetchVersionFile` before loading the info/file.

### LINKED ISSUE #602: Infinite loop for 404 errors generated by an invalid addon/version id in the browse view
**STR:**

Open the browse view for an add-on and make sure that either the addon or the version Id is incorrect:
see https://code.addons-dev.allizom.org/en-US/browse/616653/versions/1655549/

**Actual result:**
An infinite loop for the 404 error 

**Expected result:**
The 404 error and its corresponding message should be caught only once

willdurand:
Fixes #602
Fixes #556 

---

This patch is similar to the one merged to solve https://github.com/mozilla/addons-code-manager/issues/565. I added a bunch of early returns in the `reducer` (with `log.error` calls and test cases).

### DIFF
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -127,7 +127,7 @@ describe(__filename, () => {
           undefined,
           actions.updateSelectedPath({ selectedPath: 'pa/th', versionId: 123 }),
         );
-      }).toThrow();
+      }).toThrow(/Version missing/);
     });
 
     it('expands all parent folders when updateSelectedPath is dispatched', () => {
@@ -184,7 +184,7 @@ describe(__filename, () => {
           undefined,
           actions.toggleExpandedPath({ path: 'pa/th', versionId: 123 }),
         );
-      }).toThrow();
+      }).toThrow(/Version missing/);
     });
 
     it('does not duplicate paths in expandedPaths when updateSelectedPath is dispatched', () => {
@@ -317,7 +317,7 @@ describe(__filename, () => {
     it('throws an error when expandTree is called for an unknown version', () => {
       expect(() => {
         reducer(undefined, actions.expandTree({ versionId: 123 }));
-      }).toThrow();
+      }).toThrow(/Version missing/);
     });
 
     it('does not add paths for files to expandedPaths when expandTree is dispatched', () => {
@@ -364,7 +364,7 @@ describe(__filename, () => {
     it('throws an error when collapseTree is called for an unknown version', () => {
       expect(() => {
         reducer(initialState, actions.collapseTree({ versionId: 123 }));
-      }).toThrow();
+      }).toThrow(/Version missing/);
     });
 
     it('stores lists of versions by add-on ID', () => {
@@ -469,7 +469,7 @@ describe(__filename, () => {
             version,
           }),
         );
-      }).toThrow();
+      }).toThrow(/Entry missing for headVersionId/);
     });
 
     it('throws an error message when headVersion is missing on loadDiff()', () => {
@@ -488,7 +488,7 @@ describe(__filename, () => {
             version,
           }),
         );
-      }).toThrow();
+      }).toThrow(/Version missing for headVersionId/);
     });
   });
 
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -121,6 +121,20 @@ describe(__filename, () => {
       );
     });
 
+    it('logs an error when updateSelectedPath is called for an unknown version', () => {
+      const _log = createFakeLogger();
+      const state = reducer(
+        initialState,
+        actions.updateSelectedPath({ selectedPath: 'pa/th', versionId: 123 }),
+        // TS looks confused about the third optional argument.
+        // @ts-ignore
+        { _log },
+      );
+
+      expect(_log.error).toHaveBeenCalled();
+      expect(state).toEqual(initialState);
+    });
+
     it('expands all parent folders when updateSelectedPath is dispatched', () => {
       const selectedPath = 'folder1/folder2/file1.js';
 
@@ -167,6 +181,20 @@ describe(__filename, () => {
         initialPath,
         newFolder,
       ]);
+    });
+
+    it('logs an error when toggleExpandedPath is called for an unknown version', () => {
+      const _log = createFakeLogger();
+      const state = reducer(
+        initialState,
+        actions.toggleExpandedPath({ path: 'pa/th', versionId: 123 }),
+        // TS looks confused about the third optional argument.
+        // @ts-ignore
+        { _log },
+      );
+
+      expect(_log.error).toHaveBeenCalled();
+      expect(state).toEqual(initialState);
     });
 
     it('does not duplicate paths in expandedPaths when updateSelectedPath is dispatched', () => {
@@ -296,6 +324,20 @@ describe(__filename, () => {
       ]);
     });
 
+    it('logs an error when expandTree is called for an unknown version', () => {
+      const _log = createFakeLogger();
+      const state = reducer(
+        initialState,
+        actions.expandTree({ versionId: 123 }),
+        // TS looks confused about the third optional argument.
+        // @ts-ignore
+        { _log },
+      );
+
+      expect(_log.error).toHaveBeenCalled();
+      expect(state).toEqual(initialState);
+    });
+
     it('does not add paths for files to expandedPaths when expandTree is dispatched', () => {
       const path1 = 'scripts/';
       const path2 = 'scripts/background.js';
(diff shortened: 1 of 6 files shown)


src/reducers/versions.tsx
```diff
@@ -735,6 +764,12 @@ const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
       const { selectedPath, versionId } = action.payload;
 
       const version = state.versionInfo[versionId];
+
+      if (!version) {
+        _log.error(`Version missing for versionId: ${versionId}`);
```
bobsilverberg: I wonder if this should throw an actual error, rather than just logging. It shouldn't really be possible to have called `updateSelectedPath` without having a corresponding version loaded.

src/reducers/versions.tsx
```diff
@@ -757,12 +792,19 @@ const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
     case getType(actions.toggleExpandedPath): {
       const { path, versionId } = action.payload;
 
-      const { expandedPaths } = state.versionInfo[versionId];
+      const version = state.versionInfo[versionId];
+
+      if (!version) {
+        _log.error(`Version missing for versionId: ${versionId}`);
```
bobsilverberg: I wonder if this should throw an actual error, rather than just logging. It shouldn't really be possible to have called `toggleExpandedPath` without having a corresponding version loaded.

src/reducers/versions.tsx
```diff
@@ -776,8 +818,13 @@ const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
       const { versionId } = action.payload;
 
       const version = state.versionInfo[versionId];
-      const { entries } = version;
-      const expandedPaths = entries
+
+      if (!version) {
+        _log.error(`Version missing for versionId: ${versionId}`);
```
bobsilverberg: I wonder if this should throw an actual error, rather than just logging. It shouldn't really be possible to have called `expandedTree` without having a corresponding version loaded.

src/reducers/versions.tsx
```diff
@@ -796,12 +843,19 @@ const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
     case getType(actions.collapseTree): {
       const { versionId } = action.payload;
 
+      const version = state.versionInfo[versionId];
+
+      if (!version) {
+        _log.error(`Version missing for versionId: ${versionId}`);
```
bobsilverberg: I wonder if this should throw an actual error, rather than just logging. It shouldn't really be possible to have called `collapseTree` without having a corresponding version loaded.

src/reducers/versions.spec.tsx
```diff
@@ -121,6 +121,20 @@ describe(__filename, () => {
       );
     });
 
+    it('logs an error when updateSelectedPath is called for an unknown version', () => {
```
bobsilverberg: The four test cases added that check that an error was logged for different actions are all essentially identical, other than the action being called, so I wonder if either a helper, or an `it.each()` would be appropriate for these to reduce the duplication in this file?

src/reducers/versions.spec.tsx
```diff
@@ -1618,4 +1719,94 @@ describe(__filename, () => {
...
+      const version = fakeVersion;
+      let state = reducer(undefined, actions.loadVersionInfo({ version }));
+
+      const abortFetchVersionId = version.id + 123;
+      state = reducer(
+        state,
+        actions.abortFetchVersion({ versionId: abortFetchVersionId }),
+      );
+
+      expect(state.versionInfo[version.id]).toEqual(expect.any(Object));
```
bobsilverberg: Why are you using `any(Object)` here rather than `createInternalVersion(version)`? You are checking that the other version wasn't set to `null` but not necessarily that it was untouched in any way.

willdurand:
> This looks good, thanks @willdurand. I have some questions about the tests, and also about whether we should be throwing errors instead of logging.

@bobsilverberg for a similar case, we used `log.error` instead of throwing: https://github.com/mozilla/addons-code-manager/blob/aadb04c6983341e10d56ab754403dd18093fd803/src/reducers/versions.tsx#L838, so 🤷‍♂️

src/reducers/versions.spec.tsx
```diff
@@ -1618,4 +1719,94 @@ describe(__filename, () => {
...
+      const version = fakeVersion;
+      let state = reducer(undefined, actions.loadVersionInfo({ version }));
+
+      const beginFetchVersionId = version.id + 123;
+      state = reducer(
+        state,
+        actions.beginFetchVersion({ versionId: beginFetchVersionId }),
+      );
+
+      expect(state.versionInfo[version.id]).toEqual(expect.any(Object));
```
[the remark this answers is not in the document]
willdurand: The purpose of the test case isn't to assert this value, only making sure it's not `null` or `undefined`. That's why `expect.any(Object)` is used here. The exact shape of this object is covered by another test case.

bobsilverberg:
> @bobsilverberg for a similar case, we used `log.error` instead of throwing: https://github.com/mozilla/addons-code-manager/blob/aadb04c6983341e10d56ab754403dd18093fd803/src/reducers/versions.tsx#L838
> 
> so.. 🤷‍♂️ 

I would argue that maybe that other function should be changed too. It seems reasonable to throw an error when we encounter a case that we don't believe should ever have happened.

willdurand:
> I would argue that maybe that other function should be changed too. It seems reasonable to throw an error when we encounter a case that we don't believe should ever have happened.

I don't know. `log.error` will send the error to sentry, without breaking the site for the end user and, for developers, logged errors are in red in the console.

I don't have a strong opinion on that but that would break the contract of a reducer: https://redux.js.org/glossary#reducer (pureness and side-effect free).

src/reducers/versions.tsx
kumar303: > I wonder if this should throw an actual error

I guess either way works but it would be a bit better to throw an error so that Sentry gets a stack trace. It might render an error boundary and make the app unusable but at least that would be less confusing than the user seeing no change at all from their action. Thrown errors are easier to test, too 😅

src/reducers/versions.tsx
```diff
@@ -356,10 +365,7 @@ export const getVersionFiles = (versions: VersionsState, versionId: number) => {
   return versions.versionFiles[versionId];
 };
 
-export const getVersionInfo = (
-  versions: VersionsState,
-  versionId: number,
-): Version | void => {
```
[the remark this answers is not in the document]
kumar303: Well, I should ask, is TypeScript still inferring the wrong type?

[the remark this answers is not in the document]
willdurand: If I remove `as Version` in a test file (on a `getVersionInfo` call), I get this TS error:

```
Type 'Version | null | undefined' is not assignable to type 'Version'. Type 'undefined' is not assignable to type 'Version'.
```

So I guess it has been fixed either _via_ a TS upgrade or _via_ the type change in this patch.

willdurand:
@bobsilverberg I updated this patch to throw errors all the time as both @kumar303 and you think it's better (I am fine with that, I removed all the log calls to be consistent).

willdurand:
> I think this infinite loop is new behavior because I've definitely triggered 404s in the past.

Yep, I agree. There is a link to an error page in the `Index` component. We would have noticed the infinite loop before.

src/reducers/versions.tsx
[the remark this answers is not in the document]
kumar303: ok, good

src/reducers/versions.spec.tsx
```diff
@@ -121,6 +121,15 @@ describe(__filename, () => {
       );
     });
 
+    it('throws an error when updateSelectedPath is called for an unknown version', () => {
+      expect(() => {
+        reducer(
+          undefined,
+          actions.updateSelectedPath({ selectedPath: 'pa/th', versionId: 123 }),
+        );
+      }).toThrow();
```
bobsilverberg: In a lot of our tests we assert on the actual error message. Maybe that's overkill (and introduces a maintenance burden if the messages change), but I wanted to raise it here as an inconsistency. This same comment applies to all the `expect(...).toThrow()` tests.

kumar303: Yeah, since so many errors *could* be thrown, I think it's always helpful to assert some part of the message.

PR #606, opened 2019-04-17
Introduce a Skeleton component

### LINKED ISSUE #379: Render loading states with skeleton text, not spinners

### LINKED ISSUE #607: Create a `Skeleton` component
Supports #379 

---

We need a `Skeleton` component to render skeletons.

### LINKED ISSUE #608: Use `Skeleton` components in `VersionChooser`
Instead of rendering a `Retrieving all the versions [...]` message when navigating to [a new "compare" page](https://code.addons-dev.allizom.org/en-US/compare/502955/versions/1541794...1541798/), we should render the component with most of its content (all the static content can be rendered) and `Skeleton` (a.k.a. fake animations) where we have dynamic content.

The dynamic content for this component are the two `select`/dropdown lists.

willdurand:
Fixes #607 
Fixes #608 

Supports #379

---

This is a very simple `Skeleton` component that should fill the place where it is added. We can iterate if we need to add constraints (like `range` and `minWidth` props) but it's only the beginning of the long journey to fix #379.

The second commit refactors the `VersionChooser` component to see the `Skeleton` in action:

### DIFF
--- a/src/components/Skeleton/index.tsx
+++ b/src/components/Skeleton/index.tsx
@@ -1,0 +1,9 @@
+import * as React from 'react';
+
+import styles from './styles.module.scss';
+
+const SkeletonBase = () => {
+  return <div className={styles.skeleton} />;
+};
+
+export default SkeletonBase;
--- a/src/components/Skeleton/styles.module.scss
+++ b/src/components/Skeleton/styles.module.scss
@@ -1,0 +1,21 @@
+@import '../../scss/variables';
+
+.skeleton {
+  animation: pulse 1s infinite ease-in-out;
+  border-radius: $border-radius;
+  height: 100%;
+  min-height: 1rem;
+  width: 100%;
+}
+
+@keyframes pulse {
+  0% {
+    background-color: rgba($light-gray, 0.5);
+  }
+  50% {
+    background-color: rgba($light-gray, 1);
+  }
+  100% {
+    background-color: rgba($light-gray, 0.5);
+  }
+}
--- a/src/components/VersionSelect/index.tsx
+++ b/src/components/VersionSelect/index.tsx
@@ -23,7 +23,6 @@ export type PublicProps = {
 
 class VersionSelectBase extends React.Component<PublicProps> {
   static defaultProps = {
-    isLoading: false,
     withLeftArrow: false,
   };
 
--- a/src/components/VersionSelect/index.tsx
+++ b/src/components/VersionSelect/index.tsx
@@ -4,12 +4,14 @@ import { Col, Form } from 'react-bootstrap';
 import { FormControlProps } from 'react-bootstrap/lib/FormControl';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
+import Skeleton from '../Skeleton';
 import { VersionsList, VersionsListItem } from '../../reducers/versions';
 import { gettext } from '../../utils';
 import styles from './styles.module.scss';
 
 export type PublicProps = {
   className?: string;
+  isLoading: boolean;
   isSelectable: (version: VersionsListItem) => boolean;
   label: string;
   listedVersions: VersionsList;
@@ -21,6 +23,7 @@ export type PublicProps = {
 
 class VersionSelectBase extends React.Component<PublicProps> {
   static defaultProps = {
+    isLoading: false,
     withLeftArrow: false,
   };
 
@@ -47,6 +50,7 @@ class VersionSelectBase extends React.Component<PublicProps> {
   render() {
     const {
       className,
+      isLoading,
       label,
       listedVersions,
       unlistedVersions,
@@ -64,24 +68,31 @@ class VersionSelectBase extends React.Component<PublicProps> {
 
         <Form.Group as={Col} className={className}>
           <Form.Label>{label}</Form.Label>
-          <Form.Control as="select" value={value} onChange={this.onChange}>
-            {listedVersions.length && (
-              <optgroup
-                className={styles.listedGroup}
-                label={gettext('Listed')}
-              >
-                {listedVersions.map(this.renderOption)}
-              </optgroup>
-            )}
-            {unlistedVersions.length && (
-              <optgroup
-                className={styles.unlistedGroup}
-                label={gettext('Unlisted')}
-              >
-                {unlistedVersions.map(this.renderOption)}
-              </optgroup>
-            )}
-          </Form.Control>
+
+          {isLoading ? (
+            <div className="form-control">
+              <Skeleton />
+            </div>
+          ) : (
+            <Form.Control as="select" value={value} onChange={this.onChange}>
+              {listedVersions.length && (
+                <optgroup
+                  className={styles.listedGroup}
+                  label={gettext('Listed')}
+                >
+                  {listedVersions.map(this.renderOption)}
+                </optgroup>
+              )}
+              {unlistedVersions.length && (
+                <optgroup
+                  className={styles.unlistedGroup}
+                  label={gettext('Unlisted')}
+                >
+                  {unlistedVersions.map(this.renderOption)}
+                </optgroup>
+              )}
+            </Form.Control>
+          )}
         </Form.Group>
       </React.Fragment>
     );
--- a/src/components/Skeleton/index.spec.tsx
+++ b/src/components/Skeleton/index.spec.tsx
@@ -1,0 +1,18 @@
+import * as React from 'react';
+import { shallow } from 'enzyme';
+
+import styles from './styles.module.scss';
+
+import Skeleton from '.';
+
+describe(__filename, () => {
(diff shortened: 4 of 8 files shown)


src/components/VersionSelect/index.tsx
```diff
@@ -21,6 +23,7 @@ export type PublicProps = {
 
 class VersionSelectBase extends React.Component<PublicProps> {
   static defaultProps = {
+    isLoading: false,
```
bobsilverberg: Nit: I wonder if the default should be `true` as it likely will usually be true when this component is initially rendered. I realize that the default doesn't really matter, but if we're going to have one, `true` seems more realistic.

Actually, on second thought, why do we need a default value for `isLoading` at all? It seems reasonable to expect that it will always be provided as a public prop.

willdurand: > Actually, on second thought, why do we need a default value for `isLoading` at all? It seems reasonable to expect that it will always be provided as a public prop.

Oh, that's right. Let's always require this prop.

src/components/Skeleton/styles.module.scss
```diff
@@ -0,0 +1,21 @@
+@import '../../scss/variables';
+
+.skeleton {
+  animation: pulse 1s infinite ease-in-out;
+  border-radius: $border-radius;
+  height: 100%;
+  min-height: 1rem;
```
kumar303: I think this should be `em` because that references the font-size of the current element (`rem` references that of the root). The skeleton container might inherit font-size adjustments and that seems important.

src/components/Skeleton/styles.module.scss
```diff
@@ -0,0 +1,21 @@
+@import '../../scss/variables';
+
+.skeleton {
+  animation: pulse 1s infinite ease-in-out;
+  border-radius: $border-radius;
+  height: 100%;
```
kumar303: The skeleton should always be the same height as the text it's replacing so I don't think `100%` is a good value for text.

How about aiming to simulate the select box more closely, like this?

```diff
diff --git a/src/components/Skeleton/styles.module.scss b/src/components/Skeleton/styles.module.scss
index 197d0a7..c87798a 100644
--- a/src/components/Skeleton/styles.module.scss
+++ b/src/components/Skeleton/styles.module.scss
@@ -3,7 +3,6 @@
 .skeleton {
   animation: pulse 1s infinite ease-in-out;
   border-radius: $border-radius;
-  height: 100%;
   min-height: 1rem;
   width: 100%;
 }
diff --git a/src/components/VersionSelect/index.tsx b/src/components/VersionSelect/index.tsx
index 953540e..bd3edcf 100644
--- a/src/components/VersionSelect/index.tsx
+++ b/src/components/VersionSelect/index.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import makeClassName from 'classnames';
 import { Col, Form } from 'react-bootstrap';
 // eslint-disable-next-line import/no-unresolved
 import { FormControlProps } from 'react-bootstrap/lib/FormControl';
@@ -69,7 +70,12 @@ class VersionSelectBase extends React.Component<PublicProps> {
           <Form.Label>{label}</Form.Label>
 
           {isLoading ? (
-            <div className="form-control">
+            <div
+              className={makeClassName(
+                'form-control',
+                styles.simulatedFormControl,
+              )}
+            >
               <Skeleton />
             </div>
           ) : (
diff --git a/src/components/VersionSelect/styles.module.scss b/src/components/VersionSelect/styles.module.scss
index eae3a48..0ce2103 100644
--- a/src/components/VersionSelect/styles.module.scss
+++ b/src/components/VersionSelect/styles.module.scss
@@ -2,3 +2,8 @@
   margin: auto;
   padding-top: 14px;
 }
+
+.simulatedFormControl {
+  align-items: center;
+  display: flex;
+}
```

The skeleton still needs a width adjustment (see the other comment) but I think this layout transition is smoother because the text appears in the same place as the skeleton:

Screenshot 2019-04-17 11 53 13
Screenshot 2019-04-17 11 53 28

src/components/Skeleton/index.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+
+import styles from './styles.module.scss';
+
+const SkeletonBase = () => {
+  return <div className={styles.skeleton} />;
```
kumar303: This should accept a `className` prop so that the style can be adjusted more easily.

src/components/VersionSelect/index.tsx
```diff
@@ -64,24 +67,31 @@ class VersionSelectBase extends React.Component<PublicProps> {
...
-                label={gettext('Unlisted')}
-              >
-                {unlistedVersions.map(this.renderOption)}
-              </optgroup>
-            )}
-          </Form.Control>
+
+          {isLoading ? (
+            <div className="form-control">
+              <Skeleton />
```
kumar303: It would be nice to give this a specific width to mimic a version number (like `1.5.1`). We should strive to make skeleton screens as close to the final layout as possible. In this case, it would be easy to predict the average width of a version number. They will most likely be in the range of `0.1` to `99.0.0`.

src/components/VersionSelect/index.tsx
```diff
@@ -64,24 +67,31 @@ class VersionSelectBase extends React.Component<PublicProps> {
...
-                className={styles.unlistedGroup}
-                label={gettext('Unlisted')}
-              >
-                {unlistedVersions.map(this.renderOption)}
-              </optgroup>
-            )}
-          </Form.Control>
+
+          {isLoading ? (
+            <div className="form-control">
```
kumar303: Could you make it clear somehow that `form-control` is a hardcoded value that must match that of `<Form.Control>`? If possible, maybe you could control the class name? I think there is at least an option to specify a class name prefix.

kumar303:
Sorry for the late review. I think some of my comments are worth following up on.

src/components/VersionSelect/index.tsx
willdurand: I did not set a width because the select also has an arrow on the right, so rendering the skeleton full width covers both.

src/components/Skeleton/styles.module.scss
willdurand: I was thinking about embedding the skeleton component inside the container of the thing it is supposed to mimic, hence 100% but I did not try all the possibilities.

What you're saying is still relevant of course.

src/components/Skeleton/index.tsx
willdurand: Yep, I did not add that prop yet because I did not need to inject a className. We might also want to have other props.

src/components/Skeleton/styles.module.scss
willdurand: Correct, thanks

willdurand:
> I would say just hold off on skeletonizing all the existing components because that will take a lot of time and I'd like to make some headway on the comments work first. However, we should try to skeletonize new components after you land this 🎉

I thought I left such a comment somewhere.. I totally agree and that was my plan too.

src/components/VersionSelect/index.tsx
willdurand: unfortunately, we have to use `.form-control`. It comes from the Bootstrap CSS package. We could use `@extend .form-control` in our own CSS class definition, but that would be very similar.

src/components/Skeleton/styles.module.scss
willdurand: So, if I remove `height: 100%`, then the `Skeleton` does not fit the container's dimensions, which I think is a problem.

willdurand: Ok, I found a way to combine both idea:

Screen Shot 2019-04-18 at 10 41 07

PR #610, opened 2019-04-17
Create a function to return the Uid of the next/previous message

### LINKED ISSUE #609: Create a function to return the Uid of the next/previous message
Supports #520 

This is work towards #520, and is about creating a reducer function which can be called and will return the next/previous linter message id.

bobsilverberg:
Fixes #609 

I have written a lot of tests for this, testing the various situations that might exist when navigating between messages. I'm not sure if I have too many, not enough, or just the right number of tests, so that's something to check.

There are some paths as yet untested, for example some errors (I think), but I wanted to submit this for an initial review before going any further with it.

I created an internal function called `findRelativeMessageUid`, but I didn't export it and didn't write any tests for it. I think it makes more sense to just write the tests for `getRelativeMessageUid` to test every scenario, rather than a suite of tests for `findRelativeMessageUid` and then also use `findRelativeMessageUid` in the tests for `getRelativeMessageUid`. And yes, I know those names are kind of confusing and can probably be improved upon.

### DIFF
--- a/src/reducers/fileTree.spec.tsx
+++ b/src/reducers/fileTree.spec.tsx
@@ -662,6 +662,31 @@ describe(__filename, () => {
       }).toThrow(`No messages found for current path: ${currentPath}`);
     });
 
+    it('throws an error with an empty pathList', () => {
+      const externalMessages1 = [
+        { line: null, uid: global1 },
+        { line: line1, uid: line1FirstUid },
+      ];
+      const externalMessages2 = [{ line: null, uid: global2 }];
+
+      const messageMap = createMessageMap([
+        { path: file1, messages: externalMessages1 },
+        { path: file3, messages: externalMessages2 },
+      ]);
+
+      expect(() => {
+        _getRelativeMessageUid({
+          currentMessageUid: global2,
+          currentPath: file3,
+          messageMap,
+          pathList: [],
+          position: RelativePathPosition.next,
+        });
+      }).toThrow(
+        `findRelativeMessageUid was unable to find a message using currentPath: ${file3}`,
+      );
+    });
+
     it('returns the first global message if no currentMessageUid exists', () => {
       const externalMessages = [
         { line: null, uid: global1 },
@@ -676,6 +701,27 @@ describe(__filename, () => {
         _getRelativeMessageUid({
           currentMessageUid: '',
           messageMap,
+        }),
+      ).toEqual(global1);
+    });
+
+    it('returns the first global message from a path with messages if no currentMessageUid exists', () => {
+      const externalMessages = [
+        { line: null, uid: global1 },
+        { line: line1, uid: line1FirstUid },
+      ];
+      const pathList = [file1, file2];
+
+      const messageMap = createMessageMap([
+        { path: file2, messages: externalMessages },
+      ]);
+
+      expect(
+        _getRelativeMessageUid({
+          currentMessageUid: '',
+          currentPath: file2,
+          messageMap,
+          pathList,
         }),
       ).toEqual(global1);
     });
--- a/src/reducers/fileTree.spec.tsx
+++ b/src/reducers/fileTree.spec.tsx
@@ -2,18 +2,22 @@
 import reducer, {
   ROOT_PATH,
   DirectoryNode,
+  GetRelativeMessageUidParams,
   RelativePathPosition,
   actions,
   buildFileTree,
   buildFileTreeNodes,
   buildTreePathList,
+  getRelativeMessageUid,
   getRelativePath,
   getTree,
   goToRelativeFile,
   initialState,
 } from './fileTree';
 import { createInternalVersion, createInternalVersionEntry } from './versions';
+import { ExternalLinterMessage, LinterMessageMap } from './linter';
 import {
+  createFakeLinterMessagesByPath,
   createFakeThunk,
   createVersionWithEntries,
   fakeVersion,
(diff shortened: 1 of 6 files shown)


src/reducers/fileTree.tsx
```diff
@@ -191,6 +190,100 @@ export const getRelativePath = ({
...
+        return nextMessagesForPath[0].uid;
+      }
+    }
+    path = nextPath;
+  }
+  // If we got here then we never found another message, which would be odd,
+  // because we should have at the very least gotten back to our original
+  // message.
+  // We should probably log this.
+  return null;
```
bobsilverberg: It probably makes more sense to throw an error here, as this should never happen. I'm not sure how I would be able to test this though, again because it shouldn't be possible.

src/reducers/fileTree.tsx
```diff
@@ -191,6 +190,100 @@ export const getRelativePath = ({
...
+      if (messages) {
+        const messagesForPath = getMessagesForPath(messages);
+        if (messagesForPath.length) {
+          return messagesForPath[0].uid;
+        }
+      }
+    }
+    // No messages found in the map (which shouldn't be possible because of the
+    // length check, so throw an error.
+    throw new Error(
```
bobsilverberg: I don't know how to test this, because it is an impossibility. Maybe that means we don't need to throw here at all? Maybe we should just return `null`?

src/reducers/fileTree.tsx
```diff
@@ -191,6 +190,100 @@ export const getRelativePath = ({
...
+  for (let i = 0; i < pathList.length; i++) {
+    const nextPath = getRelativePath({
+      currentPath: path,
+      pathList,
+      position,
+    });
+    const nextMessages = messageMap[nextPath];
+    if (nextMessages) {
+      const nextMessagesForPath = getMessagesForPath(nextMessages);
+      if (position === RelativePathPosition.previous) {
```
kumar303: I found this part hard to read because of the nested if branching and the verbose variable names. Maybe after improving test coverage you can shorten it? Something like this:

```diff
diff --git a/src/reducers/fileTree.tsx b/src/reducers/fileTree.tsx
index c956ffd..d20dc6a 100644
--- a/src/reducers/fileTree.tsx
+++ b/src/reducers/fileTree.tsx
@@ -205,14 +205,11 @@ const findRelativeMessageUid = (
     });
     const nextMessages = messageMap[nextPath];
     if (nextMessages) {
-      const nextMessagesForPath = getMessagesForPath(nextMessages);
-      if (position === RelativePathPosition.previous) {
-        const lastMessageIndex = nextMessagesForPath.length - 1;
-        if (lastMessageIndex >= 0) {
-          return nextMessagesForPath[lastMessageIndex].uid;
-        }
-      } else if (nextMessagesForPath.length) {
-        return nextMessagesForPath[0].uid;
+      const nextForPath = getMessagesForPath(nextMessages);
+      const nextIndex =
+        position === RelativePathPosition.previous ? nextForPath.length - 1 : 0;
+      if (nextForPath[nextIndex]) {
+        return nextForPath[nextIndex].uid;
       }
     }
     path = nextPath;
```

src/reducers/fileTree.tsx
```diff
@@ -191,6 +190,100 @@ export const getRelativePath = ({
...
+        const lastMessageIndex = nextMessagesForPath.length - 1;
+        if (lastMessageIndex >= 0) {
```
kumar303: This needs a test because I don't see any failures after this change:

```diff
diff --git a/src/reducers/fileTree.tsx b/src/reducers/fileTree.tsx
index c956ffd..e092303 100644
--- a/src/reducers/fileTree.tsx
+++ b/src/reducers/fileTree.tsx
@@ -208,9 +208,7 @@ const findRelativeMessageUid = (
       const nextMessagesForPath = getMessagesForPath(nextMessages);
       if (position === RelativePathPosition.previous) {
         const lastMessageIndex = nextMessagesForPath.length - 1;
-        if (lastMessageIndex >= 0) {
-          return nextMessagesForPath[lastMessageIndex].uid;
-        }
+        return nextMessagesForPath[lastMessageIndex].uid;
       } else if (nextMessagesForPath.length) {
         return nextMessagesForPath[0].uid;
       }
```

src/reducers/fileTree.tsx
```diff
@@ -191,6 +190,100 @@ export const getRelativePath = ({
...
+  if (!currentMessageUid) {
+    // There is no current message, so get the first one.
+    if (!Object.keys(messageMap).length) {
+      return null;
+    }
+    for (const path of pathList) {
+      const messages = messageMap[path];
+      if (messages) {
+        const messagesForPath = getMessagesForPath(messages);
+        if (messagesForPath.length) {
```
kumar303: This needs a test for empty messages because I don't see a failure with this patch:

```diff
diff --git a/src/reducers/fileTree.tsx b/src/reducers/fileTree.tsx
index c956ffd..c51fb95 100644
--- a/src/reducers/fileTree.tsx
+++ b/src/reducers/fileTree.tsx
@@ -249,9 +249,7 @@ export const getRelativeMessageUid = ({
       const messages = messageMap[path];
       if (messages) {
         const messagesForPath = getMessagesForPath(messages);
-        if (messagesForPath.length) {
-          return messagesForPath[0].uid;
-        }
+        return messagesForPath[0].uid;
       }
     }
     // No messages found in the map (which shouldn't be possible because of the

```

src/reducers/fileTree.tsx
```diff
@@ -191,6 +190,100 @@ export const getRelativePath = ({
...
+  position,
+}: GetRelativeMessageUidParams): string | null => {
+  if (!currentMessageUid) {
+    // There is no current message, so get the first one.
+    if (!Object.keys(messageMap).length) {
+      return null;
+    }
+    for (const path of pathList) {
+      const messages = messageMap[path];
+      if (messages) {
```
kumar303: Is it possible to test a falsy `messageMap[path]`? I don't see any failures with this patch:

```diff
diff --git a/src/reducers/fileTree.tsx b/src/reducers/fileTree.tsx
index c956ffd..6139699 100644
--- a/src/reducers/fileTree.tsx
+++ b/src/reducers/fileTree.tsx
@@ -247,11 +247,9 @@ export const getRelativeMessageUid = ({
     }
     for (const path of pathList) {
       const messages = messageMap[path];
-      if (messages) {
-        const messagesForPath = getMessagesForPath(messages);
-        if (messagesForPath.length) {
-          return messagesForPath[0].uid;
-        }
+      const messagesForPath = getMessagesForPath(messages);
+      if (messagesForPath.length) {
+        return messagesForPath[0].uid;
       }
     }
     // No messages found in the map (which shouldn't be possible because of the
```

src/reducers/fileTree.tsx
```diff
@@ -191,6 +190,100 @@ export const getRelativePath = ({
...
+};
+
+export const getRelativeMessageUid = ({
+  currentMessageUid,
+  currentPath,
+  messageMap,
+  pathList,
+  position,
+}: GetRelativeMessageUidParams): string | null => {
+  if (!currentMessageUid) {
```
kumar303: The edge case of not looking for a relative message is simpler but it's not much different than looking for a relative message. You could combine the logic so that there aren't so many code paths -- the fewer diverging code paths the better.

After improving test coverage, you could try combining them like this:

```diff
diff --git a/src/reducers/fileTree.tsx b/src/reducers/fileTree.tsx
index c956ffd..83093f4 100644
--- a/src/reducers/fileTree.tsx
+++ b/src/reducers/fileTree.tsx
@@ -240,25 +240,9 @@ export const getRelativeMessageUid = ({
   pathList,
   position,
 }: GetRelativeMessageUidParams): string | null => {
-  if (!currentMessageUid) {
-    // There is no current message, so get the first one.
-    if (!Object.keys(messageMap).length) {
-      return null;
-    }
-    for (const path of pathList) {
-      const messages = messageMap[path];
-      if (messages) {
-        const messagesForPath = getMessagesForPath(messages);
-        if (messagesForPath.length) {
-          return messagesForPath[0].uid;
-        }
-      }
-    }
-    // No messages found in the map (which shouldn't be possible because of the
-    // length check, so throw an error.
-    throw new Error(
-      'The messageMap is not empty, but we were unable to find any messages',
-    );
+  if (!Object.keys(messageMap).length) {
+    // There are no messages at all.
+    return null;
   }
 
   const messages = messageMap[currentPath];
@@ -269,14 +253,22 @@ export const getRelativeMessageUid = ({
   }
 
   const messagesForPath = getMessagesForPath(messages);
-  const currentMessageIndex = messagesForPath.findIndex(
-    (message) => message.uid === currentMessageUid,
-  );
 
-  const newIndex =
-    position === RelativePathPosition.previous
-      ? currentMessageIndex - 1
-      : currentMessageIndex + 1;
+  let newIndex;
+  if (!currentMessageUid) {
+    // Since we aren't looking a message relative to an existing one,
+    // just get the first message.
+    newIndex = 0;
+  } else {
+    const currentMessageIndex = messagesForPath.findIndex(
+      (message) => message.uid === currentMessageUid,
+    );
+
+    newIndex =
+      position === RelativePathPosition.previous
+        ? currentMessageIndex - 1
+        : currentMessageIndex + 1;
+  }
 
   if (newIndex >= 0 && newIndex < messagesForPath.length) {
     return messagesForPath[newIndex].uid;
```

src/reducers/fileTree.tsx
kumar303: > It probably makes more sense to throw an error here, as this should never happen. I'm not sure how I would be able to test this though, again because it shouldn't be possible.

You should throw an error, if possible.

A test using an empty `pathList` will make it throw:

```diff
diff --git a/src/reducers/fileTree.spec.tsx b/src/reducers/fileTree.spec.tsx
index 925c39e..6061b2b 100644
--- a/src/reducers/fileTree.spec.tsx
+++ b/src/reducers/fileTree.spec.tsx
@@ -932,6 +932,27 @@ describe(__filename, () => {
         ).toEqual(global11);
       });
 
+      it('throws...', () => {
+        const externalMessages1 = [
+          { line: null, uid: global11 },
+          { line: null, uid: global21 },
+        ];
+        const externalMessages2 = [{ line: null, uid: global12 }];
+
+        const messageMap = createMessageMap([
+          { path: file1, messages: externalMessages1 },
+          { path: file3, messages: externalMessages2 },
+        ]);
+
+        _getRelativeMessageUid({
+          currentMessageUid: global12,
+          currentPath: file3,
+          messageMap,
+          pathList: [],
+          position: RelativePathPosition.next,
+        });
+      });
+
       it('wraps around from the first message in the first file to the last message in the last file', () => {
         const externalMessages1 = [
           { line: null, uid: global11 },
```

src/reducers/fileTree.tsx
bobsilverberg: Looking at this again, I don't think we need this `if` statement at all. I think this was a result of some verbose code I wrote and then refactored. This `if` statement is checking to make sure that what we calculate as the index of the last message for the file (via `nextMessagesForPath.length - 1`) is in fact a valid index, but as long as `nextMessagesForPath.length` is not `0`, it will always be valid, and if `nextMessagesForPath.length` *is* `0` we don't even need to do this calculation. I will make some changes to the code so this `if` statement is no longer needed.

src/reducers/fileTree.tsx
bobsilverberg: This is another interesting one. The `if` statement is protecting us against an empty array of messages, but I don't believe it's possible to have an empty array of messages. *If* `messages` is not `undefined`, then some messages exist in the `messageMap` and therefore the call to `getMessagesForPath` should return an array that is not empty (unless there's a problem with the data, or a bug in `getMessagesForPath`). I actually tried to write a test for this originally, but I was getting an error in my test code when I tried to do:

```
      const messageMap = createMessageMap([
        { path: file1, messages: [] },
      ]);
```

because the `createFakeLinterMessagesByPath` helper (which is used by `createMessageMap` in the test) is [designed to throw an error](https://github.com/mozilla/addons-code-manager/blob/4af1b350023ef6fa54da5ad0f0de4e4f8a23c29b/src/test-helpers.tsx#L328) if you try to use it to create some `LinterMessagesByPath` with an empty array of messages. I took this at the time to mean that this would be an invalid test, but maybe the helper needs to be changed to allow for an empty array of messages, or maybe that particular test needs to be changed to not make use of `createFakeLinterMessagesByPath`?

But I think the question of whether this check is even needed is also valid, as the app is designed to not allow this situation to exist. I think we do need it, but I just wanted to raise this as a discussion point.

src/reducers/fileTree.tsx
bobsilverberg: I've added a test case to support this.

src/reducers/fileTree.tsx
bobsilverberg: @kumar303 I'm interested to hear what you think about the above.

src/reducers/fileTree.tsx
bobsilverberg: Yes, that's much better. Thanks!

src/reducers/linter.tsx
```diff
@@ -139,6 +139,18 @@ export const findMostSevereType = (
   );
 };
 
+export const getMessagesForPath = (
+  messages: LinterMessagesByPath,
+): LinterMessage[] => {
+  const allMessages = [...messages.global];
```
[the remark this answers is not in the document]
bobsilverberg: > In fact, you should be able to replace this line:
> 
> [addons-code-manager/src/components/FileTreeNode/index.tsx](https://github.com/mozilla/addons-code-manager/blob/50be0327f1ae670d743a570855d23bef3eff14c3/src/components/FileTreeNode/index.tsx#L74-L76)
> 
> Lines 74 to 76 in [50be032](/mozilla/addons-code-manager/commit/50be0327f1ae670d743a570855d23bef3eff14c3)
> 
>  if (m.global.length > 1 || Object.keys(m.byLine).length > 0) { 
>    return false; 
>  } 
> with:
> 
> ```ts
> if (getMessagesForPath(linterMessageMap[path]).length) {
>   return false;
> }
> ```
>

That does not seem to be accurate. If I do that I get a couple of test failures. AFAIU the current `if` statement is checking for more than one `global` message, or more than 0 `byLine` messages, but `getMessagesForPath(linterMessageMap[path]).length` will return `true` with only 1 `global` message.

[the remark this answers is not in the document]
bobsilverberg: I suppose that `if (getMessagesForPath(linterMessageMap[path]).length > 1)` would work though.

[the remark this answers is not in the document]
bobsilverberg: Actually, that doesn't work either, because it doesn't shortcut when there are no global messages.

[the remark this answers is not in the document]
bobsilverberg: Hmm, that's not exactly the issue either. The problem is, with the code above, a file that has just one `byLine` message, and no global messages, will still execute the final `return m.global[0].code.includes(LINTER_KNOWN_LIBRARY_CODE);`, which will error due to there being no `global` messages.

src/reducers/fileTree.tsx
kumar303: You shouldn't need the helper to create a map of empty arrays. Just take a look at [the type](https://github.com/mozilla/addons-code-manager/blob/4af1b350023ef6fa54da5ad0f0de4e4f8a23c29b/src/reducers/linter.tsx#L72) and try creating it by hand.

src/reducers/linter.tsx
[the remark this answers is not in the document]
bobsilverberg: I don't think there's a way to reuse `getMessagesForPath` effectively here, but I did extract a function that checks for unexpected keys which I will use in both places.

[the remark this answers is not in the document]
bobsilverberg: Ok, I found a way to use `getMessagesForPath` in `isKnownLibrary`. It's a bit questionable (as it's not really needed), but I added a comment to explain why we're using it.

src/reducers/fileTree.tsx
bobsilverberg: This code is outdated now so I think this comment is now irrelevant.

src/reducers/fileTree.tsx
bobsilverberg: I believe the code has changed enough to render this comment inapplicable as well, so I'm marking it as resolved.

bobsilverberg:
Thanks for the detailed review and all of the great suggestions, @kumar303. I believe I have addressed all of the issues. This is ready for another look.

src/reducers/fileTree.spec.tsx
```diff
@@ -585,4 +589,417 @@ describe(__filename, () => {
...
+    const line2 = 2;
+    const line3 = 3;
+    const line1FirstUid = 'line1-uid-1';
+    const line1SecondUid = 'line1-uid-2';
+    const line2Uid = 'line2-uid';
+    const line3Uid = 'line3-uid';
+    const global1 = 'global-uid-1';
+    const global2 = 'global-uid-2';
+
+    const createMessageMap = (
```
kumar303: I think you have made more work for yourself than necessary with this helper. I'm pretty sure you can just use `getMessageMap()`, right? I tried replacing a few tests and it seemed to work:

```diff
diff --git a/src/reducers/fileTree.spec.tsx b/src/reducers/fileTree.spec.tsx
index 5e0f2b7..a002f1a 100644
--- a/src/reducers/fileTree.spec.tsx
+++ b/src/reducers/fileTree.spec.tsx
@@ -15,8 +15,13 @@ import reducer, {
   initialState,
 } from './fileTree';
 import { createInternalVersion, createInternalVersionEntry } from './versions';
-import { ExternalLinterMessage, LinterMessageMap } from './linter';
 import {
+  ExternalLinterMessage,
+  LinterMessageMap,
+  getMessageMap,
+} from './linter';
+import {
+  createFakeExternalLinterResult,
   createFakeLinterMessagesByPath,
   createFakeThunk,
   createVersionWithEntries,
@@ -640,7 +645,11 @@ describe(__filename, () => {
       expect(
         _getRelativeMessageUid({
           currentMessageUid: '',
-          messageMap: createMessageMap([]),
+          messageMap: getMessageMap(
+            createFakeExternalLinterResult({
+              messages: [],
+            }),
+          ),
         }),
       ).toEqual(null);
     });
@@ -649,9 +658,11 @@ describe(__filename, () => {
       const currentPath = 'currentPath.js';
       const messagePath = 'notTheCurrentPath.js';
 
-      const messageMap = createMessageMap([
-        { path: messagePath, messages: [{ line: line1, uid: 'line1' }] },
-      ]);
+      const messageMap = getMessageMap(
+        createFakeExternalLinterResult({
+          messages: [{ file: messagePath, line: line1, uid: 'line1' }],
+        }),
+      );
 
       expect(() => {
         _getRelativeMessageUid({
@@ -663,16 +674,15 @@ describe(__filename, () => {
     });
 
     it('throws an error with an empty pathList', () => {
-      const externalMessages1 = [
-        { line: null, uid: global1 },
-        { line: line1, uid: line1FirstUid },
-      ];
-      const externalMessages2 = [{ line: null, uid: global2 }];
-
-      const messageMap = createMessageMap([
-        { path: file1, messages: externalMessages1 },
-        { path: file3, messages: externalMessages2 },
-      ]);
+      const messageMap = getMessageMap(
+        createFakeExternalLinterResult({
+          messages: [
+            { file: file1, line: null, uid: global1 },
+            { file: file1, line: line1, uid: line1FirstUid },
+            { file: file3, line: null, uid: global2 },
+          ],
+        }),
+      );
 
       expect(() => {
         _getRelativeMessageUid({
```

src/reducers/fileTree.tsx
```diff
@@ -191,6 +190,90 @@ export const getRelativePath = ({
...
+): LinterMessage['uid'] | null => {
+  let path = currentPath;
+  const maxAttempts = pathList.length;
+  for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+    const nextPath = getRelativePath({
+      currentPath: path,
+      pathList,
+      position,
+    });
+    const nextMessages = messageMap[nextPath];
```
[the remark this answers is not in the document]
bobsilverberg: I shortened them a bit, but I'm not sure what you're looking for. Did you want things as short as `const m = getMessagesForPath(nextMessages)`? I try to avoid things like that. Take a look at the new version (once I push it) and see if it's short enough for you.

src/reducers/linter.tsx
```diff
@@ -139,6 +139,33 @@ export const findMostSevereType = (
...
+    }
+  });
+};
+
+export const getMessagesForPath = (
+  messages: LinterMessagesByPath,
+  _checkMessageKeys: typeof checkMessageKeys = checkMessageKeys,
+): LinterMessage[] => {
+  // Make sure the messages do not have any unexpected keys.
+  _checkMessageKeys(messages);
```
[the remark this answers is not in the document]
bobsilverberg: It doesn't. That was the result of me extracting it to use in two places, and then figuring out how to use it in one, and not moving it back. I'll fix that.

src/reducers/fileTree.spec.tsx
bobsilverberg: Yes, this is better. I think my way meant not having to specify a file name multiple times for multiple messages, but I agree that this is better.

bobsilverberg:
Thanks for the review @kumar303. This is ready for another look.

src/reducers/linter.spec.tsx
```diff
@@ -599,4 +601,78 @@ describe(__filename, () => {
...
+      expect(messagesForPath.length).toEqual(5);
+      expect(messagesForPath[0]).toHaveProperty('uid', uid1);
+      expect(messagesForPath[1]).toHaveProperty('uid', uid2);
+      expect(messagesForPath[2]).toHaveProperty('uid', uid3);
+      expect(messagesForPath[3]).toHaveProperty('uid', uid4);
+      expect(messagesForPath[4]).toHaveProperty('uid', uid5);
+    });
+
+    it('returns with only global messages', () => {
+      const externalMessages = [message1, message2];
```
kumar303: The same goes for this. Otherwise, it's hard to tell where the `uid` values are coming from.

src/reducers/linter.spec.tsx
```diff
@@ -599,4 +601,78 @@ describe(__filename, () => {
...
+      });
+
+      const messagesForPath = getMessagesForPath(messages);
+      expect(messagesForPath.length).toEqual(2);
+      expect(messagesForPath[0]).toHaveProperty('uid', uid1);
+      expect(messagesForPath[1]).toHaveProperty('uid', uid2);
+    });
+
+    it('returns with only byLine messages', () => {
+      const externalMessages = [message3, message4, message5];
```
kumar303: Same here. I think having variables for the UIDs is fine. As a reader, I'm trying to connect the UIDs so that's the hard part when the configuration is abstracted into variables.

PR #612, opened 2019-04-18
CSS fixes for Skeleton/VersionSelect

willdurand:
See: https://github.com/mozilla/addons-code-manager/pull/606#pullrequestreview-227866846

---

Before:

After:

### DIFF
--- a/src/components/Skeleton/styles.module.scss
+++ b/src/components/Skeleton/styles.module.scss
@@ -4,7 +4,7 @@
   animation: pulse 1s infinite ease-in-out;
   border-radius: $border-radius;
   height: 100%;
-  min-height: 1rem;
+  min-height: 1em;
   width: 100%;
 }
 
--- a/src/components/Skeleton/styles.module.scss
+++ b/src/components/Skeleton/styles.module.scss
@@ -3,7 +3,6 @@
 .skeleton {
   animation: pulse 1s infinite ease-in-out;
   border-radius: $border-radius;
-  height: 100%;
   min-height: 1em;
   width: 100%;
 }
--- a/src/components/VersionSelect/styles.module.scss
+++ b/src/components/VersionSelect/styles.module.scss
@@ -16,7 +16,6 @@
   display: flex;
 
   .simulatedFormControlSkeleton {
-    height: auto;
     width: 60px;
   }
 }
--- a/src/components/VersionSelect/styles.module.scss
+++ b/src/components/VersionSelect/styles.module.scss
@@ -15,7 +15,7 @@
   align-items: center;
   display: flex;
 
-  div {
+  .simulatedFormControlSkeleton {
     height: auto;
     width: 60px;
   }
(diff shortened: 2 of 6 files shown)


src/components/VersionSelect/styles.module.scss
```diff
@@ -10,3 +10,13 @@
     font-weight: 300;
   }
 }
+
+.simulatedFormControl {
+  align-items: center;
+  display: flex;
+
+  div {
```
kumar303: This is fragile because the internal structure of `Skeleton` could change. I'd like to see the code use a custom class like:

```tsx
<div className={makeClassName(styles.simulatedFormControl, ...)}>
  <Skeleton className={styles.simulatedFormControlSkeleton} />
</div>
```

With that, you can reference it directly:

```css
.simulatedFormControlSkeleton {
  height: auto;
  width: 60px;
}
```

If someone were working on the internal structure of `Skeleton`, they would see the `className` as a public prop which would make them think about all the components that might be depending on that class name.

src/components/Skeleton/styles.module.scss
```diff
@@ -4,7 +4,7 @@
   animation: pulse 1s infinite ease-in-out;
   border-radius: $border-radius;
   height: 100%;
```
kumar303: I'd like to see this removed because I don't think it's a good default. It's better to default to the size of text within the container.

Also, when you remove this, `VersionSelect` will no longer need to override it to `height: auto`.

willdurand: If the container is a `p`, then the default size of text is used according to the `p` story.

Removing this rule would break the `div` story though, which I think is valid as one of the constraints in the original issue was to match the dimensions of the containers.

For what it's worth, we need to change the `height` in the `VersionSelect` because what we want to mimic is an HTML `option`.

src/components/VersionSelect/styles.module.scss
willdurand: Oh nooes /o\

I changed laptop in the meantime and definitely did not want to push that change (and missed it when I self-reviewed this patch). This needs a CSS class of course.

willdurand:
@kumar303 sorry I did not re-request a review and only left a few comments here. I worked on most of the requested changes except for the `width` that I think would not work well with what I had in mind, see: https://github.com/mozilla/addons-code-manager/pull/612#discussion_r276741678.

src/components/Skeleton/styles.module.scss
kumar303: What is the real use case of the div story? I can't think of why we would want that behavior. If we did want it, we could introduce a custom class and set the height explicitly. I don't think it should *default* to this behavior.

kumar303:
> sorry I did not re-request a review and only left a few comments here. 

Whoops, sometimes I have a hard time keeping up with comment replies. The best thing is to mention me in a comment; I'll always get an instant notification 😁

src/components/Skeleton/styles.module.scss
willdurand: > What is the real use case of the div story?

That's a good question... I don't know what I was thinking 😞 I updated this code to get rid of the `height: 100%`.

willdurand:
rebased

Issue #621, opened 2019-04-19
Add a CodeOverview component


kumar303:
Add a `CodeOverview` component to support https://github.com/mozilla/addons-code-manager/issues/523

PR #623, opened 2019-04-19
Add a CodeOverview component

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/621

I tried to keep the patch small and mostly failed. Here are to-do items for future patches:
* ~~Support linter messages.~~ (in https://github.com/mozilla/addons-code-manager/pull/631)
* ~~Add a CodeOverview story~~ (in https://github.com/mozilla/addons-code-manager/pull/631)


src/components/CodeOverview/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+      allLineShapes,
+    );
+    const overview = [];
+
+    for (let rowIndex = 0; rowIndex < numberOfRows; rowIndex++) {
+      const shapes = chunkedLineShapes[rowIndex] || undefined;
+      // Use the first line in the group.
+      const shapeIndex = 0;
+      const line =
+        shapes && shapes.length ? shapes[shapeIndex].line : undefined;
```
bobsilverberg: How can we end up with an empty `shapes` array?

src/components/CodeOverview/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+    );
+  };
+
+  render() {
+    const { overviewHeight } = this.state;
+    const { version } = this.props;
+
+    return (
+      <LinterProvider
+        key={overviewHeight ? String(overviewHeight) : ''}
```
bobsilverberg: Just wondering, why did you decide to use `overviewHeight` as the `key`?

src/components/CodeLineShapes/index.tsx
```diff
@@ -24,6 +24,7 @@ const CodeLineShapes = ({ lineShapes }: Props) => {
           <div
             key={[
               String(lineShapes.line),
+              String(shapeIndex),
```
bobsilverberg: Why did you need to add this? Is it now possible to have more than one shape with the same `line`? Or is it because a `shape` doesn't necessarily have a `line` (which I don't fully understand, tbh)?

src/components/CodeOverview/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+    rowHeight: 10,
+  };
+
+  public state = { overviewHeight: null };
+
+  private overviewRef = React.createRef<HTMLDivElement>();
+
+  componentDidMount() {
+    const { _window } = this.props;
+    this.setOverviewHeight();
```
bobsilverberg: This line doesn't seem to be covered by a test. If I comment it out all the tests still pass.

src/components/CodeOverview/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+    _debounce: debounce,
+    _window: window,
+    // This is the padding of the overview container.
+    overviewPadding: 10,
+    rowTopPadding: 2,
+    // This is the height of the row, including rowTopPadding.
+    rowHeight: 10,
+  };
+
+  public state = { overviewHeight: null };
```
bobsilverberg: `overviewHeight` defaulting to `null` does not seem to be explicitly covered via a test. If I change the value to something other than `null` then the test for `fitLineShapesIntoOverview -> requires overviewHeight` fails, but that's really testing something different.

Do you think we should have a test that asserts that `overviewHeight` defaults to `null`, or does that seem unnecessary?

src/components/CodeOverview/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+    this.setState({ overviewHeight: null });
+  };
+
+  setOverviewHeight = (ref = this.overviewRef) => {
+    if (ref && ref.current) {
+      this.setState({ overviewHeight: ref.current.clientHeight });
+    }
+  };
+
+  waitAndSetNewOverviewHeight = this.props._debounce(
```
bobsilverberg: This doesn't seem to be covered by a test. Is there something we can do to test this?

src/components/CodeOverview/index.spec.tsx
```diff
@@ -0,0 +1,452 @@
...
+    expect(link1).toHaveStyle('paddingTop', undefined);
+
+    const link2 = innerRoot.find(Link).at(1);
+
+    expect(link2).toHaveStyle('height', `${rowHeight}px`);
+    expect(link2).toHaveStyle('paddingTop', `${rowTopPadding}px`);
+
+    const link3 = innerRoot.find(Link).at(2);
+
+    expect(link3).toHaveStyle('height', `${rowHeight}px`);
```
bobsilverberg: I'm not sure that you need to keep repeating the assertion for `height` as it's a property that doesn't depend on the particular `Link`, but it's fine by me.

src/components/CodeOverview/index.spec.tsx
```diff
@@ -0,0 +1,452 @@
...
+
+    const innerRoot = renderWithLinterProvider({ root });
+
+    const lineShapes = innerRoot.find(CodeLineShapes);
+
+    expect(lineShapes.at(0)).toHaveProp('lineShapes', allLineShapes[0]);
+    expect(lineShapes.at(1)).toHaveProp('lineShapes', allLineShapes[1]);
+    expect(lineShapes.at(2)).toHaveProp('lineShapes', allLineShapes[2]);
+
+    expect(lineShapes).toHaveLength(3);
```
bobsilverberg: Nit: I would expect this assertion to appear above the ones for the individual `lineShapes`, because if there are less than 3 `lineShapes` then the test should fail and give a more descriptive message than it would if it encountered a `toHaveProp` for a nonexistent line.

src/components/CodeOverview/index.spec.tsx
```diff
@@ -0,0 +1,452 @@
...
+  });
+
+  it('renders CodeLineShapes for groups of lines that fit', () => {
```
bobsilverberg: I'm not sure what the `lines that fit` part means? Why not just `renders CodeLineShapes for groups of lines`?

src/components/CodeOverview/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    return {
+      numberOfRows,
+      chunkedLineShapes: chunk(allLineShapes, chunkSize),
+    };
+  }
+
+  renderRow(
+    selectedMessageMap: LinterProviderInfo['selectedMessageMap'],
```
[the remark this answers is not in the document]
kumar303: Yes, that's correct. They are utilized in https://github.com/mozilla/addons-code-manager/pull/631

src/components/CodeOverview/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+    if (ref && ref.current) {
+      this.setState({ overviewHeight: ref.current.clientHeight });
+    }
+  };
+
+  waitAndSetNewOverviewHeight = this.props._debounce(
+    () => this.setOverviewHeight(),
+    200,
+    {
+      leading: false,
```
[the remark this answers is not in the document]
kumar303: Sure, I guess we could just rely on the defaults. It's very important to make this a trailing debounce.

src/components/CodeOverview/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    const allLineShapes = generateLineShapes(getLines(content));
+    const { numberOfRows, chunkedLineShapes } = this.fitLineShapesIntoOverview(
+      allLineShapes,
+    );
+    const overview = [];
+
+    for (let rowIndex = 0; rowIndex < numberOfRows; rowIndex++) {
+      const shapes = chunkedLineShapes[rowIndex] || undefined;
```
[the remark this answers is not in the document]
kumar303: > Does a falsey `chunkedLineShapes[rowIndex]` indicate an empty row that was created to fill up the overview when there is no more content for it? 

Yes, exactly

> If so, maybe a comment somewhere around here to that effect will make it easier for the next person who encounters this code.

Sure.

src/components/CodeOverview/index.tsx
kumar303: I added a comment in the code to explain it.

src/components/CodeOverview/index.tsx
kumar303: The key is needed because `<LinterProvider>` should re-render any time `overviewHeight` changes. Key values in React can be used to force re-renders -- this is a bit odd to see when you're unfamiliar with that magic.

src/components/CodeLineShapes/index.tsx
kumar303: There can be multiple shapes per line, yes. A shape is either whitespace or a code silhouette. That's all.

src/components/CodeOverview/index.tsx
kumar303: Good catch, thanks

src/components/CodeOverview/index.tsx
kumar303: Ah, there was this test https://github.com/mozilla/addons-code-manager/blob/2bf0c9f8c50cfe8a441ef1152c631fb43bed8751/src/components/CodeOverview/index.spec.tsx#L187

...but it was using `setState()` which isn't so helpful. I fixed it.

src/components/CodeOverview/index.tsx
kumar303: Ah, yes there is. I couldn't figure it out at first but I got it to work.

src/components/CodeOverview/index.spec.tsx
kumar303: It's just making sure that the value gets passed from props to the style attribute on every element. The `paddingTop` assertion is more important.

src/components/CodeOverview/index.spec.tsx
kumar303: The failure message for `toHaveLength()` is not a good one so I try to put it at the end when possible. Additionally, in this case, it's only there to catch an unexpected amount of rows so it's secondary to the main assertions.

src/components/CodeOverview/index.spec.tsx
kumar303: Fitting into the grid is an important part of the integration test but I made the spec more explicit.

kumar303:
@bobsilverberg thanks for catching the test coverage issues. I think I resolved everything. Can you take another look?

PR #627, opened 2019-04-22
Retain expandedPaths when loading version info

### LINKED ISSUE #592: The parent folder closes when a new file is selected in the compare view
**STR:**

1. Open the compare view for an addon - see https://code.addons-dev.allizom.org/en-US/compare/616548/versions/1655558...1655559/
2. Click on the Parent folder and select another file from the list

**Actual result:**
The parent folder closes

**Expected result:**
The parent folder stays open

**Notes:**
- this might be fixed when a solution for #586 is found, but I wasn't sure about this case
- I don't remember seeing this behavior before, so it might be a very recent regression

bobsilverberg:
Fixes #592

### DIFF
--- a/src/reducers/versions.tsx
+++ b/src/reducers/versions.tsx
@@ -671,7 +671,7 @@ export const fetchDiff = ({
   path,
 }: FetchDiffParams): ThunkActionCreator => {
   return async (dispatch, getState) => {
-    const { api: apiState } = getState();
+    const { api: apiState, versions: versionsState } = getState();
 
     dispatch(actions.beginFetchDiff());
 
@@ -687,7 +687,9 @@ export const fetchDiff = ({
       dispatch(errorsActions.addError({ error: response.error }));
       dispatch(actions.abortFetchDiff());
     } else {
-      dispatch(actions.loadVersionInfo({ version: response }));
+      if (!getVersionInfo(versionsState, response.id)) {
+        dispatch(actions.loadVersionInfo({ version: response }));
+      }
       dispatch(
         actions.loadDiff({
           addonId,
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -1370,11 +1370,13 @@ describe(__filename, () => {
       addonId = 1,
       baseVersionId = 2,
       headVersionId = 3,
+      store = configureStore(),
       version = fakeVersionWithDiff,
       _getDiff = jest.fn().mockReturnValue(Promise.resolve(version)),
       path = undefined as string | undefined,
     } = {}) => {
       return thunkTester({
+        store,
         createThunk: () =>
           fetchDiff({
             _getDiff,
@@ -1417,13 +1419,26 @@ describe(__filename, () => {
       });
     });
 
-    it('dispatches loadVersionInfo() when API response is successful', async () => {
+    it('dispatches loadVersionInfo() when API response is successful and version is not already loaded', async () => {
       const version = fakeVersionWithDiff;
 
       const { dispatch, thunk } = _fetchDiff({ version });
       await thunk();
 
       expect(dispatch).toHaveBeenCalledWith(
+        actions.loadVersionInfo({ version }),
+      );
+    });
+
+    it('does not dispatch loadVersionInfo() when version is already loaded', async () => {
+      const store = configureStore();
+      const version = fakeVersionWithDiff;
+      store.dispatch(actions.loadVersionInfo({ version }));
+
+      const { dispatch, thunk } = _fetchDiff({ version, store });
+      await thunk();
+
+      expect(dispatch).not.toHaveBeenCalledWith(
         actions.loadVersionInfo({ version }),
       );
     });


bobsilverberg:
> This approach works but it doesn't seem like the right thing to do. A version can have any tree of files so it seems wrong to assume that the last version has the same tree of files. It might but it might not.

I'm not sure I follow what the problem is. Versions are loaded and stored by `id` and AFAIU, each version has a unique `id`, so it seems impossible that an "earlier" version with and `id` of, say `123`, could possibly have a different file tree than a "later" version with an `id` of `123`. They should always be identical, shouldn't they?

If that's the case, then doing what I'm doing seems like it couldn't introduce a problem, although having just explained that out loud it now seems to me that maybe, when we're loading a version, if we already have a version with that `id` loaded we could just make it a null op, which seems like it would also solve the problem.

@kumar303 what do you think?

kumar303:
> I'm not sure I follow what the problem is. Versions are loaded and stored by `id` and AFAIU, each version has a unique `id`, so it seems impossible that an "earlier" version with and `id` of, say `123`, could possibly have a different file tree than a "later" version with an `id` of `123`. They should always be identical, shouldn't they?

It's very possible for an earlier version to have a different file tree than a newer version. I was just looking at [LastPassword](https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/lastpass-password-manager) the other day. Version 3.1.1 has totally different files from 4.18.2.

> 
> if we already have a version with that `id` loaded we could just make it a null op, which seems like it would also solve the problem.
> 

Yeah, I think so. If we already have the version loaded, why do we need to reload it? I think `fetchDiff` is doing the wrong thing here and that's what's causing the bug. 

The `fetchDiff` thunk is trying to serve two cases: 1) the first time you load a diff, when the page loads; and 2) updating diffs. The updating part doesn't need to reload the version.

bobsilverberg:
> It's very possible for an earlier version to have a different file tree than a newer version. I was just looking at [LastPassword](https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/lastpass-password-manager) the other day. Version 3.1.1 has totally different files from 4.18.2.

But that doesn't address my point. Yes an earlier version of an add-on can have a different tree, but if we load a `Version` with an `id` of `123`, and later load a `Version` again, which also has an `id` of `123`, they must necessarily be the same version, and therefore have the same tree. In your example, Version 3.1.1 of LastPassword would *not* have the same internal `id` as Version 4.18.2 of LastPassword.

> Yeah, I think so. If we already have the version loaded, why do we need to reload it? I think `fetchDiff` is doing the wrong thing here and that's what's causing the bug. 

> The `fetchDiff` thunk is trying to serve two cases: 1) the first time you load a diff, when the page loads; and 2) updating diffs. The updating part doesn't need to reload the version.

I agree that the `fetchDiff` thunk shouldn't be loading the `version` every time, but that seems like a separate issue. We should probably fix that, via a different issue/patch, but it also seems reasonable to me to fix the issue of overwriting versions in any scenario in which that might happen, by changing the loading logic to do nothing if a version with that `id` is already loaded. I propose we make that change via this issue, and open a separate one for the `fetchDiff` issue.

@kumar303 what do you think?

bobsilverberg:
@kumar303 I went ahead and changed this patch to prevent reloading a version, which seems like a decent thing to have in place. Please let me know what you think.

src/reducers/versions.tsx
```diff
@@ -666,6 +666,10 @@ const reducer: Reducer<VersionsState, ActionType<typeof actions>> = (
   switch (action.type) {
     case getType(actions.loadVersionInfo): {
       const { version } = action.payload;
+      if (state.versionInfo[version.id]) {
+        // We shouldn't load the same version again.
+        return state;
```
kumar303: I know that this solves the problem but this isn't the right place to do it. What if we need to re-fetch the same version from the API because it changed on the server? That would no longer be possible.

What about not dispatching a version in `fetchDiff()` when it already exists? https://github.com/mozilla/addons-code-manager/blob/fcac49f0babb33fa301522e932dd969343c0957c/src/reducers/versions.tsx#L621

I still think it's weird that `fetchDiff()` loads a version when switching paths. Maybe we need a new action, like `fetchDiffForVersion()`, that we can call when we know we are only fetching a diff for the same version.

kumar303: The reason this isn't happening in `Browse` is because we already have two different actions.

We have `fetchVersion()`:

https://github.com/mozilla/addons-code-manager/blob/fcac49f0babb33fa301522e932dd969343c0957c/src/pages/Browse/index.tsx#L92

And we have `fetchVersionFile()`:

https://github.com/mozilla/addons-code-manager/blob/fcac49f0babb33fa301522e932dd969343c0957c/src/pages/Browse/index.tsx#L114

bobsilverberg: It does seem like redesigning the way `Compare` works to be more like `Browse` might make sense, and I started looking at that, but it was becoming a bigger patch. I think that making the change you suggest above, to just not dispatch `loadVersionInfo` from inside `fetchDiff` if the version is already loaded is also a decent solution to the immediate problem and much simpler, so thats what I've done.

bobsilverberg:
@kumar303 This is ready for another look, based on my comment at https://github.com/mozilla/addons-code-manager/pull/627#discussion_r279395983

src/reducers/versions.tsx
```diff
@@ -687,7 +687,9 @@ export const fetchDiff = ({
       dispatch(errorsActions.addError({ error: response.error }));
       dispatch(actions.abortFetchDiff());
     } else {
-      dispatch(actions.loadVersionInfo({ version: response }));
+      if (!getVersionInfo(versionsState, response.id)) {
+        dispatch(actions.loadVersionInfo({ version: response }));
```
kumar303: This seems fine as a short term fix but it doesn't seem like the right thing to do. It's better than the last iteration but it still would never allow us to reload a version from the server when calling `fetchDiff()`. I suggest filing an issue to introduce another action, like `fetchDiffPath()`, which would only be for switching paths within the same version. That seems like a better way to fix it. 

Alternatively, we could introduce a new action parameter like `loadVersion: boolean` to indicate whether a version should be loaded.

bobsilverberg: Thanks @kumar303. I filed #671

Issue #622, opened 2019-04-19
Add a CodeOverview story


kumar303:
A `CodeOverview` component was added in https://github.com/mozilla/addons-code-manager/issues/621 to support https://github.com/mozilla/addons-code-manager/issues/523

This issue is to add a story for it so we can see all screens in isolation.

PR #631, opened 2019-04-23
Add linter message support to CodeOverview

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/523
Fixes https://github.com/mozilla/addons-code-manager/issues/622
Fixes https://github.com/mozilla/addons-code-manager/issues/665

~~⚠️ depends on https://github.com/mozilla/addons-code-manager/pull/623~~

Storybook sample:

Screenshot 2019-04-23 10 36 15

### DIFF
--- a/src/components/CodeOverview/index.spec.tsx
+++ b/src/components/CodeOverview/index.spec.tsx
@@ -275,6 +275,64 @@ describe(__filename, () => {
     expect(link).toHaveProp('title', `Jump to line ${line}`);
   });
 
+  it('links to the first line that has a linter message', () => {
+    const firstLineWithMsg = 5;
+
+    const { innerRoot } = renderWithMessages({
+      // Create a file with 6 lines.
+      contentLines: generateFileLines({ count: 6 }),
+      // Make a grid of height 2 so that each row will contain 3 lines.
+      overviewHeight: 2,
+      // Put a linter message on line 5.
+      messages: [{ line: firstLineWithMsg }],
+      // Normalize the grid parameters.
+      rowHeight: 1,
+      overviewPadding: 0,
+      rowTopPadding: 0,
+    });
+
+    const allLinks = innerRoot.find(Link);
+    expect(allLinks).toHaveLength(2);
+
+    const link = allLinks.at(1);
+
+    expect(link).toHaveProp(
+      'to',
+      expect.objectContaining({
+        hash: getCodeLineAnchor(firstLineWithMsg),
+      }),
+    );
+  });
+
+  it('links to the first of multiple lines containing linter messages', () => {
+    const firstLineWithMsg = 5;
+
+    const { innerRoot } = renderWithMessages({
+      // Create a file with 6 lines.
+      contentLines: generateFileLines({ count: 6 }),
+      // Make a grid of height 2 so that each row will contain 3 lines.
+      overviewHeight: 2,
+      // Put a linter message on line 5 and line 6.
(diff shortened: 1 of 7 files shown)


kumar303:
@bobsilverberg @willdurand rebased to resolve conflicts

src/components/CodeOverview/index.spec.tsx
```diff
@@ -377,6 +412,105 @@ describe(__filename, () => {
...
+    const links = innerRoot.find(Link);
+
+    expect(links.at(0).childAt(0)).not.toHaveClassName(styles.linterMessage);
+    expect(links.at(0).find(CodeLineShapes)).toHaveLength(1);
+
+    expect(links.at(1).childAt(0)).toHaveClassName(styles.linterMessage);
+    expect(links.at(1).find(CodeLineShapes)).toHaveLength(0);
+  });
+
+  it('uses the most severe linter type', () => {
```
bobsilverberg: Nit: "uses the most severe linter *message* type".

stories/CodeOverview.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+  fakeVersionEntry,
+} from '../src/test-helpers';
+import { newLinterMessageUID, renderWithStoreAndRouter } from './utils';
+
+const render = ({
+  content = JS_SAMPLE,
+  messages = [],
+  store = configureStore(),
+}: {
+  store?: Store;
```
bobsilverberg: Nit: These are not in the same order as the properties above.

stories/CodeOverview.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+  content?: string;
+  messages?: Partial<LinterMessage>[];
+} = {}) => {
+  let version = createInternalVersion(fakeVersion);
+
+  if (messages.length) {
+    const path = 'background.js';
+    version = createInternalVersion({
```
bobsilverberg: You could just create this version as the default version (i.e., at line 29, via a `const` instead of a `let`), instead of having it inside this `if` block. It would be fine for all stories, and would allow you to have only the special logic for the `with messages` story inside this `if`. I was a bit confused about why you had to create a different `version` in this block just for the `with messages` story and then realized that this version could work for all stories.

I think that only including the logic that is specific to the `with messages` story in this `if` block makes it a bit easier to understand.

stories/fixtures/long-file-sample.tsx
```diff
@@ -0,0 +1,193 @@
+export default `/* @flow */
```
bobsilverberg: It's a bit funny (as in "ha ha" funny, not "odd" funny) that our sample uses `flow`. Not a change request, just an observation.

kumar303:
> I wonder if linking to the actual line where the first message in the chunk resides (as opposed to the first line in the chunk) is something we might consider doing? I know that complicates the implementation, but maybe it would be an improvement?

Thanks, yeah, it's a bit confusing. I don't think it will be too complicated to support it. I filed https://github.com/mozilla/addons-code-manager/issues/665

kumar303:
> Thanks, yeah, it's a bit confusing. I don't think it will be too complicated to support it. I filed #665

Actually, I found time to squeeze this in: https://github.com/mozilla/addons-code-manager/pull/631/commits/2dbeb704f608bef527ee53570ea08a5e7d96ff2b

It was simpler than I thought 😄

Issue #634, opened 2019-04-23
Add a fullscreen grid component


kumar303:
We need a fullscreen grid component for https://github.com/mozilla/addons-code-manager/issues/582 so that the entire app can be rendered fullscreen (overflowing content should scroll).

I tried getting things to work with the bootstrap grid but I couldn't make it full screen.

Also, I made a lot of effort trying to get https://github.com/mozilla/addons-code-manager/issues/582 to work *without* a fullscreen layout (i.e. letting you scroll past the header) and I couldn't think of a way to do it.

PR #635, opened 2019-04-23
Add a FullscreenGrid component

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/634

This may need tweaks once I try to fit the layout into it but I'm pretty sure it will work! 🔥 

Storybook samples:

Screenshot 2019-04-23 16 42 47
Screenshot 2019-04-23 16 42 21

### DIFF
--- a/src/components/FullscreenGrid/styles.module.scss
+++ b/src/components/FullscreenGrid/styles.module.scss
@@ -1,8 +1,10 @@
 @import '../../scss/variables';
 
+$grid-gap: 10px;
+
 .FullscreenGrid {
   display: grid;
-  grid-gap: 10px;
+  grid-gap: $grid-gap;
   grid-template-areas:
     'FGHeader FGHeader FGHeader'
     'FGNavPanel FGContentPanel FGAltPanel'
@@ -18,12 +20,12 @@
 
 .NavPanel {
   grid-area: FGNavPanel;
-  margin-left: 10px;
+  margin-left: $grid-gap;
 }
 
 .AltPanel {
   grid-area: FGAltPanel;
-  margin-right: 10px;
+  margin-right: $grid-gap;
 }
 
 .ContentPanel {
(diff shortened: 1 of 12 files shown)


src/components/FullscreenGrid/styles.module.scss
```diff
@@ -0,0 +1,41 @@
...
+}
+
+.NavPanel {
+  grid-area: FGNavPanel;
+  margin-left: 10px;
+}
+
+.AltPanel {
+  grid-area: FGAltPanel;
+  margin-right: 10px;
```
willdurand: I suppose we could introduce a variable for this `10px` value.

stories/utils.tsx
```diff
@@ -27,3 +27,30 @@ export const newLinterMessageUID = () => {
...
 };
+
+export const rootAttributeParams = ({ fullscreenByDefault = false } = {}) => {
+  type Attribute = {
+    name: string;
+    value: string | null;
+  };
+
+  const fullscreen: Attribute = { name: 'Fullscreen', value: 'fullscreen' };
+  const normal: Attribute = { name: 'Normal', value: null };
```
willdurand: Could we get rid of this `normal` mode? It's confusing but this mode is exactly the same as without the add-on.

I'd also suggest putting a comment for:

> the main thing is that this story only makes sense when it's full screen, no padding

kumar303: The normal mode is required (otherwise all other stories will be fullscreen) but I found a better way to do it. I disabled the buttons for toggling between states and I think that makes it more clear.

src/components/FullscreenGrid/styles.module.scss
kumar303: Did you mean in the whole app or just this style module? I started with one just in this style module. That makes sense.

kumar303: I just added a module variable for now.

PR #640, opened 2019-04-24
Fix regression in Browse component

### LINKED ISSUE #626: Links to specific file paths are no longer working in the browse page
**STR:**

Open the browse files page for the following addon: https://code.addons-dev.allizom.org/en-US/browse/385621/versions/1655456/?path=html%2Ffeedback.html

**Actual result:**
You are redirected to the default file - manifest.json - even though a different file path is specified in the URL

**Expected result:**
You are viewing the file specified in the URL

**Notes:**
 - this is not happening in the compare page
 - could be a regression from #596  because this function used to work before

willdurand:
Fixes #626

---

A regression has been recently merged, breaking the `Browse` component when loading a file from the URL. The regression was supposed to fix an issue but introduced a new one. This patch should solve all the known issues:

- it brings the feature back
- it avoids the issue related to too many actions dispatched when updating the selected path
- it prevents the component to enter an infinite loop in case of an error

### DIFF
--- a/src/pages/Browse/index.tsx
+++ b/src/pages/Browse/index.tsx
@@ -118,6 +118,8 @@ export class BrowseBase extends React.Component<Props> {
       _viewVersionFile({
         versionId: parseInt(versionId, 10),
         selectedPath: path,
+        // When selecting a new file to view, we do not want to preserve the
+        // hash in the URL (this hash highlights a specific line of code).
         preserveHash: false,
       }),
     );
--- a/src/pages/Browse/index.tsx
+++ b/src/pages/Browse/index.tsx
@@ -63,11 +63,11 @@ export class BrowseBase extends React.Component<Props> {
     this.loadData();
   }
 
-  componentDidUpdate(prevProps: Props) {
-    this.loadData(prevProps);
+  componentDidUpdate() {
+    this.loadData();
   }
 
-  loadData(prevProps?: Props) {
+  loadData() {
     const {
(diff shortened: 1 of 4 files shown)


src/pages/Browse/index.tsx
```diff
@@ -119,17 +110,15 @@ export class BrowseBase extends React.Component<Props> {
     }
   }
 
-  // When selecting a new file to view, we do not want to preserve the hash in
```
kumar303: Maybe this comment should be added back underneath `preserveHash: false`? It might be helpful if someone thinks about adding hash support in the future.

src/reducers/versions.spec.tsx
```diff
@@ -779,6 +779,29 @@ describe(__filename, () => {
...
+        createThunk: () =>
+          fetchVersion({ _getVersion, addonId, versionId, path }),
+      });
+
+      await thunk();
+
+      expect(_getVersion).toHaveBeenCalledWith({
+        addonId,
+        apiState: store.getState().api,
+        path,
```
kumar303: For tests like these I think it's nice to limit the assertion to the specific thing you're testing. In this case it would be:

```tsx

      expect(_getVersion).toHaveBeenCalledWith(
        expect.objectContaining({
          path,
        }),
      );
```

With that, you can remove unnecessary setup and it will also allow the call signature of `fetchVersion()` to evolve in unrelated ways without breaking this test.

PR #643, opened 2019-04-24
Create a function to return the anchor for the next diff in a file

### LINKED ISSUE #641: Create a function to return the anchor for the next diff in a file
This supports #615.

Given a current line, return an anchor for the first line of the next block of changes in the current file.

This will be used for an initial implementation of the keyboard shortcuts to navigate between diffs, but will only work for moving around in the current file.

bobsilverberg:
Fixes #641 

Note that the tests use a fixture for the diff, which makes verifying them visually a bit difficult. I.e., it's hard to know that the expectation that `['D2', 'D16', 'D18', 'D30', 'D34']` is the correct array of anchors is accurate. I could put a diff into the test file itself, but we do ideally want a diff with multiple hunks and changes and non-changes interspersed, and that's going to look pretty ugly.

I suppose that the tests for `getRelativeDiffAnchor` are easy to verify because the test for `getDiffAnchors` shows what the expected order of anchors is. I thought about injecting a mock for `getDiffAnchors` into `getRelativeDiffAnchor`  for the purposes of the tests, but that seems unnecessary when we can just allow it to use `getDiffAnchors`.

### DIFF
--- a/src/components/DiffView/fixtures/diffWithDeletions.tsx
+++ b/src/components/DiffView/fixtures/diffWithDeletions.tsx
@@ -30,6 +30,10 @@ index b7ee4f8..08a29b0 100644
 +
 +  render() {
      const { diff, viewType } = this.props;
+// Here's an unchanged line in the middle of a hunk.
+-  // This comment was deleted.
++  // This comment was added.
+// Here's another unchanged line.
 @@ -29,3 +47,5 @@ class DiffView extends React.Component<Props> {
              viewType={viewType}
 -          />
--- a/src/components/DiffView/fixtures/diffWithDeletions.tsx
+++ b/src/components/DiffView/fixtures/diffWithDeletions.tsx
@@ -30,10 +30,6 @@ index b7ee4f8..08a29b0 100644
 +
 +  render() {
      const { diff, viewType } = this.props;
-// Here's an unchanged line in the middle of a hunk.
--  // This comment was deleted.
-+  // This comment was added.
-// Here's another unchanged line.
 @@ -29,3 +47,5 @@ class DiffView extends React.Component<Props> {
              viewType={viewType}
 -          />
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -1,5 +1,6 @@
-import { getType } from 'typesafe-actions';
 import { push } from 'connected-react-router';
+import { parseDiff } from 'react-diff-view';
+import { getType } from 'typesafe-actions';
 
 import { actions as errorsActions } from './errors';
 import reducer, {
@@ -18,7 +19,9 @@ import reducer, {
   fetchVersion,
   fetchVersionFile,
   fetchVersionsList,
+  getDiffAnchors,
   getParentFolders,
+  getRelativeDiffAnchor,
   getVersionFile,
   getVersionFiles,
   getVersionInfo,
@@ -26,8 +29,9 @@ import reducer, {
   isFileLoading,
   viewVersionFile,
 } from './versions';
-import { ROOT_PATH } from './fileTree';
+import { ROOT_PATH, RelativePathPosition } from './fileTree';
 import configureStore from '../configureStore';
+import diffWithDeletions from '../components/DiffView/fixtures/diffWithDeletions';
 import {
   createFakeEntry,
   createFakeHistory,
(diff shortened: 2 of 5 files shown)


src/reducers/versions.tsx
```diff
@@ -423,6 +423,68 @@ export const isFileLoading = (
...
 
+export const getDiffAnchors = (diff: DiffInfo): string[] => {
+  const anchors: string[] = [];
+
+  for (const hunk of diff.hunks) {
+    let seekingChange = true;
+    for (const change of hunk.changes) {
+      if (change.type === 'normal') {
+        seekingChange = true;
+      } else if (seekingChange) {
```
kumar303: I'm confused as to why this ignores all the changes that come before the first `normal` change. Why not just filter out all normal changes? I know there is a reason because I tried it and the tests failed, heh. Could you explain? I think if you were to rewrite the tests with my suggestion for a different approach then it would become clear.

src/reducers/versions.tsx
```diff
@@ -423,6 +423,68 @@ export const isFileLoading = (
...
+    // There are no changes in the diff.
+    // This can happen if a file is saved but not changed internally.
+    return null;
+  }
+
+  let newIndex;
+  if (!currentAnchor) {
+    // Since we aren't looking for an anchor relative to an existing one,
+    // just get the first anchor.
+    newIndex = 0;
```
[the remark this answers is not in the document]
bobsilverberg: If `currentAnchor` is falsey then a user has requested movement (likely via a keyboard shortcut) before any diff has been selected. This means they are *initiating* movement, and therefore I think it makes sense to start at the very first diff regardless of whether they chose `next` or `previous`. That's my logic anyway. As you mention, we can deal with this again later when the function is used for something.

src/reducers/versions.spec.tsx
```diff
@@ -1774,4 +1778,80 @@ describe(__filename, () => {
...
+        getRelativeDiffAnchor({
+          currentAnchor: 'D16',
+          diff,
+          position: RelativePathPosition.next,
+        }),
+      ).toEqual('D18');
+    });
+
+    it('returns the first previous anchor in the diff', () => {
+      const diff = parseDiff(diffWithDeletions)[0];
```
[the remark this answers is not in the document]
bobsilverberg: This is a great idea, and I took your suggestion and implemented it. The tests are much easier to reason about now.

src/reducers/versions.tsx
bobsilverberg: This does not ignore all changes that come before the first `normal` change, as the test bears out. I'm not sure why you think it does. Or maybe I'm misinterpreting what you're saying. Please take a look at the updated test and see if you still feel this problem exists.

As for why I cannot just filter out all `normal` changes, if I did that then I'd have no way to identify *blocks* of changes, which is really the point of this function.

bobsilverberg:
Thanks for the review @kumar303. This is ready for another look.

bobsilverberg:
I also added a new test for `returns null if there are no changes in the diff`, which provides some coverage that was missing.

src/components/DiffView/fixtures/diffWithDeletions.tsx
```diff
@@ -30,6 +30,10 @@ index b7ee4f8..08a29b0 100644
 +
 +  render() {
      const { diff, viewType } = this.props;
+// Here's an unchanged line in the middle of a hunk.
```
kumar303: I don't think you need this anymore, right?

src/reducers/versions.spec.tsx
```diff
@@ -1774,4 +1778,203 @@ describe(__filename, () => {
...
+  type TestHunkChange = {
+    lineNumber: number;
+    type: string;
+  };
+
+  type TestHunk = TestHunkChange[];
+
+  // This helper accepts an array of arrays, each of which is a list of changes
+  // which will be contained in a hunk.
+  const createDiff = (testHunks: TestHunk[]): DiffInfo => {
```
kumar303: `DiffInfo` can be imported from `react-diff-view`

src/reducers/versions.spec.tsx
```diff
@@ -1774,4 +1778,203 @@ describe(__filename, () => {
...
+        changes: hunk.map((change) => {
+          return {
+            content: 'the content of the change is irrelevant to the tests',
+            isDelete: change.type === 'delete',
+            isInsert: change.type === 'insert',
+            isNormal: change.type === 'normal',
+            lineNumber: change.lineNumber,
+            oldLineNumber:
+              change.type === 'normal' ? change.lineNumber : undefined,
+            type: change.type as ChangeType,
```
kumar303: You don't have to make this type assertion if you reference the real property in `TestHunkChange`. I think that would also be a good idea to keep the two types in sync. Example:

```diff
diff --git a/src/@types/react-diff-view/index.d.ts b/src/@types/react-diff-view/index.d.ts
index 0f07a6d..7a2e56a 100644
--- a/src/@types/react-diff-view/index.d.ts
+++ b/src/@types/react-diff-view/index.d.ts
@@ -3,7 +3,7 @@
 declare module 'react-diff-view' {
   type ChangeType = 'delete' | 'insert' | 'normal';
 
-  type ChangeInfo = {
+  export type ChangeInfo = {
     content: string;
     isDelete?: boolean;
     isInsert?: boolean;
diff --git a/src/reducers/versions.spec.tsx b/src/reducers/versions.spec.tsx
index 89d7176..097392d 100644
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -1,5 +1,5 @@
 import { push } from 'connected-react-router';
-import { ChangeType, parseDiff } from 'react-diff-view';
+import { ChangeInfo, ChangeType, DiffInfo, parseDiff } from 'react-diff-view';
 import { getType } from 'typesafe-actions';
 
 import { actions as errorsActions } from './errors';
@@ -1780,15 +1780,13 @@ describe(__filename, () => {
   });
 
   type TestHunkChange = {
-    lineNumber: number;
-    type: string;
+    lineNumber: ChangeInfo['lineNumber'];
+    type: ChangeInfo['type'];
   };
 
-  type TestHunk = TestHunkChange[];
-
   // This helper accepts an array of arrays, each of which is a list of changes
   // which will be contained in a hunk.
-  const createDiff = (testHunks: TestHunk[]): DiffInfo => {
+  const createDiff = (testHunks: TestHunkChange[][]): DiffInfo => {
     const diffSample = parseDiff(diffWithDeletions)[0];
 
     const hunks = testHunks.map((hunk) => {
@@ -1802,7 +1800,7 @@ describe(__filename, () => {
             lineNumber: change.lineNumber,
             oldLineNumber:
               change.type === 'normal' ? change.lineNumber : undefined,
-            type: change.type as ChangeType,
+            type: change.type,
           };
         }),
         content: 'the content of hunks is irrelevant to the tests',
```

src/reducers/versions.spec.tsx
kumar303: `createFakeDiffWithChanges` is more verbose but I think it would help to make it clear that it's a simulated diff and that it only accepts changes (nothing else).

src/reducers/versions.spec.tsx
```diff
@@ -1774,4 +1778,203 @@ describe(__filename, () => {
...
+    lineNumber: number;
+    type: string;
+  };
+
+  type TestHunk = TestHunkChange[];
+
+  // This helper accepts an array of arrays, each of which is a list of changes
+  // which will be contained in a hunk.
+  const createDiff = (testHunks: TestHunk[]): DiffInfo => {
+    const diffSample = parseDiff(diffWithDeletions)[0];
```
kumar303: It would help to put a comment above this saying how it's only used to initialize some fields and that the hunks / changes will be overwritten.

bobsilverberg:
Thanks for the review @kumar303. This is ready for another look.

PR #646, opened 2019-04-25
Add status information to VersionEntry

### LINKED ISSUE #645: Add status information to VersionEntry
Supports #413 
Supports #615 

The compare API now provides a `status` property for each `entry` and we need to use that to identify which files have changed. There are some complexities in the data, such as files that are marked with a `status` of `M` but contain no actual changes, but ultimately we need this data in `VersionEntry`.

This issue is about adding it to `VersionEntry` and assumes that what we're getting back from the API in `status` is what we want.

bobsilverberg:
Fixes #645

### DIFF
--- a/src/reducers/versions.tsx
+++ b/src/reducers/versions.tsx
@@ -26,7 +26,7 @@ type VersionLicense = {
   url: string;
 };
 
-export type VersionEntryStatus = '' | 'M' | 'A' | 'D' | 'R' | 'C' | 'U';
+export type VersionEntryStatus = '' | 'M' | 'A' | 'D' | 'R' | 'C';
 export type VersionEntryType = 'image' | 'directory' | 'text' | 'binary';
 
 export type ExternalVersionEntry = {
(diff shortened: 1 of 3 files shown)


src/reducers/versions.tsx
```diff
@@ -26,6 +26,7 @@ type VersionLicense = {
   url: string;
 };
 
+export type VersionEntryStatus = '' | 'M' | 'A' | 'D' | 'R' | 'C' | 'U';
```
willdurand: Can we really have `U`?

> U: file is unmerged (you must complete the merge before it can be committed)

willdurand: In `createInternalDiff`, we support `A, C, D, M, R` and we fallback to `M` otherwise. I don't think it is ultra relevant to know that here but now you know :p

EnTeQuAk: > Can we really have U?

theoretically, we don't have anything in place to avoid that status code to be issued but practically we don't. We don't do actual merges between versions to avoid complexity so `U` most certainly won't happen in real life.

bobsilverberg: Ok, based on all that I will remove `U` from the list.

PR #650, opened 2019-04-26
Create a function to return the next/previous path with a diff

### LINKED ISSUE #649: Create a function to return the next/previous path with a diff
Supports #615 

We currently have [`getRelativeDiffAnchor`](https://github.com/mozilla/addons-code-manager/blob/eba7dbd0d53e77c423a77c585028fabd4b4e35f5/src/reducers/versions.tsx#L454) which will give us the next/previous diff in the current file, but we also want to be able to get the next/previous file with a diff, if we are at the end/beginning of the current file. To do that we need a function which will return the next/previous path with a diff.

This issue is about creating such a function.

bobsilverberg:
Fixes #649

### DIFF
--- a/src/reducers/fileTree.spec.tsx
+++ b/src/reducers/fileTree.spec.tsx
@@ -15,7 +15,11 @@ import reducer, {
   goToRelativeFile,
   initialState,
 } from './fileTree';
-import { createInternalVersion, createInternalVersionEntry } from './versions';
+import {
+  ExternalVersionEntry,
+  createInternalVersion,
+  createInternalVersionEntry,
+} from './versions';
 import { getMessageMap } from './linter';
 import {
   createFakeExternalLinterResult,
@@ -1018,38 +1022,72 @@ describe(__filename, () => {
     const file1 = 'file1.js';
     const file2 = 'file2.js';
     const file3 = 'file3.js';
+    const file4 = 'file4.js';
 
-    const entries = [
-      createInternalVersionEntry({
-        ...fakeVersionEntry,
-        filename: file1,
-        mime_category: 'text',
-        path: file1,
-        status: 'M',
-      }),
-      // Note that file2 does not contain any changes.
-      createInternalVersionEntry({
-        ...fakeVersionEntry,
-        filename: file2,
-        mime_category: 'text',
-        path: file2,
-        status: '',
-      }),
-      createInternalVersionEntry({
-        ...fakeVersionEntry,
-        filename: file3,
-        mime_category: 'text',
-        path: file3,
-        status: 'M',
-      }),
-    ];
-    const version = createVersionWithEntries(entries);
+    // const entries = [
+    //   createInternalVersionEntry({
+    //     ...fakeVersionEntry,
+    //     filename: file1,
+    //     mime_category: 'text',
+    //     path: file1,
+    //     status: 'M',
+    //   }),
+    //   // Note that file2 does not contain any changes.
+    //   createInternalVersionEntry({
+    //     ...fakeVersionEntry,
+    //     filename: file2,
+    //     mime_category: 'text',
+    //     path: file2,
+    //     status: '',
+    //   }),
+    //   createInternalVersionEntry({
+    //     ...fakeVersionEntry,
+    //     filename: file3,
+    //     mime_category: 'text',
+    //     path: file3,
+    //     status: 'M',
+    //   }),
+    //   createInternalVersionEntry({
+    //     ...fakeVersionEntry,
+    //     filename: file4,
+    //     mime_category: 'text',
+    //     path: file4,
+    //     status: 'M',
+    //   }),
+    // ];
+
+    const getFakeVersionAndPathList = (
+      entries: ({ path: string } & Partial<ExternalVersionEntry>)[],
+    ) => {
+      const pathList = entries.map((e) => e.path);
+
+      const version = createVersionWithEntries(
+        entries.map((params) =>
+          createInternalVersionEntry({
+            ...fakeVersionEntry,
+            mime_category: 'text',
+            status: 'M',
+            ...params,
+            filename: params.path,
+          }),
+        ),
+      );
+
+      return { pathList, version };
+    };
 
     it('gets the next path which has a diff', () => {
+      const { pathList, version } = getFakeVersionAndPathList([
+        { path: file1, status: 'M' },
+        { path: file2, status: '' },
+        { path: file3, status: 'M' },
+        { path: file4, status: 'M' },
+      ]);
+
       expect(
         findRelativePathWithDiff({
           currentPath: file1,
-          pathList: [file1, file2, file3],
+          pathList,
           position: RelativePathPosition.next,
           version,
         }),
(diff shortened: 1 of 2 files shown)


src/reducers/fileTree.spec.tsx
```diff
@@ -1012,4 +1013,125 @@ describe(__filename, () => {
...
+        }),
+      ).toEqual(file3);
+    });
+
+    it('gets the previous path which has a diff', () => {
+      expect(
+        findRelativePathWithDiff({
+          currentPath: file3,
+          pathList: [file1, file2, file3],
+          position: RelativePathPosition.previous,
```
kumar303: This test passes if you pass `position: RelativePathPosition.next` which means it's not a great test. It needs more than two paths with diff changes. The same goes for your `previous` test.

src/reducers/fileTree.spec.tsx
```diff
@@ -1012,4 +1013,125 @@ describe(__filename, () => {
...
+        }),
+      ).toEqual(file1);
+    });
+
+    it('wraps around to the first path on next', () => {
+      expect(
+        findRelativePathWithDiff({
+          currentPath: file3,
+          pathList: [file1, file2, file3],
+          position: RelativePathPosition.next,
```
kumar303: This test has the same problem: you can change it to `.previous` and it still passes because there are only two file paths with diff changes. The `next` test also has the problem.

src/reducers/fileTree.spec.tsx
```diff
@@ -1012,4 +1013,125 @@ describe(__filename, () => {
...
+        status: 'M',
+      }),
+    ];
+    const version = createVersionWithEntries(entries);
+
+    it('gets the next path which has a diff', () => {
+      expect(
+        findRelativePathWithDiff({
+          currentPath: file1,
+          pathList: [file1, file2, file3],
```
kumar303: It's not easy to look at this test and know what the important inputs are. In this case, `status` is an important input. I'd like to see this use a helper that will let you declare only the important inputs to the test.

Here is a quick idea:

```diff
diff --git a/src/reducers/fileTree.spec.tsx b/src/reducers/fileTree.spec.tsx
index 59d6670..bd41867 100644
--- a/src/reducers/fileTree.spec.tsx
+++ b/src/reducers/fileTree.spec.tsx
@@ -17,6 +17,7 @@ import reducer, {
 } from './fileTree';
 import { createInternalVersion, createInternalVersionEntry } from './versions';
 import { getMessageMap } from './linter';
+import { ExternalVersionEntry } from './versions';
 import {
   createFakeExternalLinterResult,
   createFakeThunk,
@@ -1045,11 +1046,37 @@ describe(__filename, () => {
     ];
     const version = createVersionWithEntries(entries);
 
+    const getFakeVersionAndPathList = (
+      entries: ({ path: string } & Partial<ExternalVersionEntry>)[],
+    ) => {
+      const pathList = entries.map((e) => e.path);
+
+      const version = createVersionWithEntries(
+        entries.map((params) =>
+          createInternalVersionEntry({
+            ...fakeVersionEntry,
+            mime_category: 'text',
+            status: 'M',
+            ...params,
+            filename: params.path,
+          }),
+        ),
+      );
+
+      return { pathList, version };
+    };
+
     it('gets the next path which has a diff', () => {
+      const { pathList, version } = getFakeVersionAndPathList([
+        { path: file1, status: 'M' },
+        { path: file2, status: '' },
+        { path: file3, status: 'M' },
+      ]);
+
       expect(
         findRelativePathWithDiff({
           currentPath: file1,
-          pathList: [file1, file2, file3],
+          pathList,
           position: RelativePathPosition.next,
           version,
         }),
```

This makes it easier for the reader to see that `status: ''` gets skipped.

Also, a helper like this would give you more flexibility. You wouldn't have to include `status: ''` in every test. You could easily define multiple paths, etc.

bobsilverberg: Yes, this is a great suggestion, thanks @kumar303!

bobsilverberg:
Thanks for the suggestions @kumar303. This is ready for another look.

src/reducers/fileTree.spec.tsx
```diff
@@ -1012,4 +1017,183 @@ describe(__filename, () => {
...
     });
   });
+
+  describe('findRelativePathWithDiff', () => {
+    const file1 = 'file1.js';
+    const file2 = 'file2.js';
+    const file3 = 'file3.js';
+    const file4 = 'file4.js';
+
+    // const entries = [
```
kumar303: 👻

src/reducers/fileTree.spec.tsx
```diff
@@ -1012,4 +1017,183 @@ describe(__filename, () => {
...
+          position: RelativePathPosition.previous,
+          version,
+        }),
+      ).toEqual(file1);
+    });
+
+    it('wraps around to the first path on next', () => {
+      const { pathList, version } = getFakeVersionAndPathList([
+        { path: file1, status: 'M' },
+        { path: file2, status: '' },
```
kumar303: A path without changes doesn't seem necessary for this test

src/reducers/fileTree.spec.tsx
```diff
@@ -1012,4 +1017,183 @@ describe(__filename, () => {
...
+          position: RelativePathPosition.next,
+          version,
+        }),
+      ).toEqual(file1);
+    });
+
+    it('wraps around to the last path on previous', () => {
+      const { pathList, version } = getFakeVersionAndPathList([
+        { path: file1, status: 'M' },
+        { path: file2, status: '' },
```
kumar303: > A path without changes doesn't seem necessary for this test

Same here. I only pointed it out because it makes the setup a little confusing.

src/reducers/fileTree.spec.tsx
```diff
@@ -1012,4 +1017,183 @@ describe(__filename, () => {
...
+          pathList,
+          position: RelativePathPosition.next,
+          version,
+        }),
+      ).toEqual(file1);
+    });
+
+    it('throws an error if no entry is found for the next path', () => {
+      const { version } = getFakeVersionAndPathList([
+        { path: file1, status: 'M' },
```
kumar303: Wouldn't this be a problem for a version with only one file? An add-on only needs `manifest.json` so it's technically possible.

src/reducers/fileTree.spec.tsx
bobsilverberg: I guess we don't have an `eslint` rule to catch stuff like that. I usually spot it after pushing but I guess I didn't check the full diff this time.

src/reducers/fileTree.spec.tsx
bobsilverberg: No, but this test's name and setup were a bit misleading, so I fixed that. The error is only thrown when the `pathList` contains a file for which there is not an `entry`. Having just one file is no problem. I updated the test to make it more clear that what we are testing is a `pathList` without a matching `entry`.

src/reducers/fileTree.spec.tsx
kumar303: This is what code review is good for 😎

PR #651, opened 2019-04-26
View an image file

### LINKED ISSUE #109: View an image file

bobsilverberg:
Fixes #109 

A couple of notes:

1. I had to modify the CSP in order to display images from `data:` URLs. I'm not sure if what I did is okay and entirely safe.
2. I only implemented this for the `Browse` page. I don't think it makes sense for the `Compare` page, because if an image was changed I imagine we'd just have a diff of the contents of the file, not a diff showing us two different images, and therefore that wouldn't really be useful to display.

### DIFF
--- a/src/components/ImageView/index.spec.tsx
+++ b/src/components/ImageView/index.spec.tsx
@@ -36,36 +36,15 @@ describe(__filename, () => {
     );
   });
 
-  it('does not render an img tag when btoa fails', () => {
+  it('throws an error when btoa fails', () => {
+    const message = 'An error message';
     const _btoa = jest.fn().mockImplementation(() => {
-      throw new Error();
+      throw new Error(message);
     });
 
-    const root = render({ _btoa });
-
-    expect(root.find('img')).toHaveLength(0);
-  });
-
-  it('renders a message when btoa fails', () => {
-    const _btoa = jest.fn().mockImplementation(() => {
-      throw new Error();
-    });
-
-    const root = render({ _btoa });
-
-    expect(root.find('p')).toHaveText('Unrecognized image format');
-  });
-
-  it('logs an error message when btoa fails', () => {
-    const _btoa = jest.fn().mockImplementation(() => {
-      throw new Error();
-    });
-    const _log = createFakeLogger();
-
-    render({ _btoa, _log });
-
-    expect(_btoa).toHaveBeenCalled();
-    expect(_log.error).toHaveBeenCalled();
+    expect(() => {
+      render({ _btoa });
+    }).toThrow(message);
   });
 
   it('logs an error message if an image has an invalid mimeType', () => {
--- a/src/components/ImageView/index.spec.tsx
+++ b/src/components/ImageView/index.spec.tsx
@@ -46,6 +46,16 @@ describe(__filename, () => {
     expect(root.find('img')).toHaveLength(0);
   });
 
+  it('renders a message when btoa fails', () => {
+    const _btoa = jest.fn().mockImplementation(() => {
+      throw new Error();
+    });
+
+    const root = render({ _btoa });
+
+    expect(root.find('p')).toHaveText('Unrecognized image format');
+  });
+
   it('logs an error message when btoa fails', () => {
     const _btoa = jest.fn().mockImplementation(() => {
       throw new Error();
@@ -58,10 +68,18 @@ describe(__filename, () => {
     expect(_log.error).toHaveBeenCalled();
   });
 
-  it('throws an error if an image has an invalid mimeType', () => {
-    expect(() => {
-      render({ mimeType: 'invalid/type' });
-    }).toThrow('Unexpected mimeType: "invalid/type"');
+  it('logs an error message if an image has an invalid mimeType', () => {
+    const _log = createFakeLogger();
+
+    render({ _log, mimeType: 'invalid/type' });
+
+    expect(_log.error).toHaveBeenCalled();
+  });
+
+  it('renders a message if an image has an invalid mimeType', () => {
+    const root = render({ mimeType: 'invalid/type' });
+
+    expect(root.find('p')).toHaveText('Unrecognized image format');
   });
 
   it('calls sanitize on svg files', () => {
--- a/src/components/ImageView/index.spec.tsx
+++ b/src/components/ImageView/index.spec.tsx
@@ -9,6 +9,7 @@ describe(__filename, () => {
   const render = ({
     _btoa = btoa,
     _log = createFakeLogger(),
+    _sanitize = jest.fn(),
     content = 'some image content',
     mimeType = 'mime/type',
   } = {}) => {
@@ -16,6 +17,7 @@ describe(__filename, () => {
       <ImageView
         _btoa={_btoa}
         _log={_log}
+        _sanitize={_sanitize}
         mimeType={mimeType}
         content={content}
       />,
@@ -54,5 +56,23 @@ describe(__filename, () => {
 
     expect(_btoa).toHaveBeenCalled();
     expect(_log.debug).toHaveBeenCalled();
+  });
+
+  it('calls sanitize on svg files', () => {
+    const _sanitize = jest.fn();
+    const content = 'some content';
+    const mimeType = 'image/svg+xml';
+    render({ _sanitize, content, mimeType });
+
+    expect(_sanitize).toHaveBeenCalledWith(content);
+  });
+
+  it('does not call sanitize on non-svg files', () => {
+    const _sanitize = jest.fn();
+    const content = 'some content';
+    const mimeType = 'image/png';
+    render({ _sanitize, content, mimeType });
+
+    expect(_sanitize).not.toHaveBeenCalled();
   });
 });
--- a/src/components/ImageView/index.spec.tsx
+++ b/src/components/ImageView/index.spec.tsx
@@ -1,25 +1,25 @@
 import * as React from 'react';
-import { shallow } from 'enzyme';
+import { shallow, ShallowWrapper } from 'enzyme';
 
 import { createFakeLogger } from '../../test-helpers';
 
-import ImageView from '.';
+import ImageView, { PublicProps } from '.';
 
 describe(__filename, () => {
   const render = ({
     _btoa = btoa,
     _log = createFakeLogger(),
-    _sanitize = jest.fn(),
     content = 'some image content',
-    mimeType = 'mime/type',
-  } = {}) => {
+    mimeType = 'image/png',
+    ...props
+  }: Partial<PublicProps> = {}) => {
     return shallow(
       <ImageView
         _btoa={_btoa}
         _log={_log}
-        _sanitize={_sanitize}
         mimeType={mimeType}
         content={content}
+        {...props}
       />,
     );
   };
@@ -46,7 +46,7 @@ describe(__filename, () => {
     expect(root.find('img')).toHaveLength(0);
   });
 
-  it('logs a debug message when btoa fails', () => {
+  it('logs an error message when btoa fails', () => {
     const _btoa = jest.fn().mockImplementation(() => {
       throw new Error();
     });
@@ -55,13 +55,28 @@ describe(__filename, () => {
     render({ _btoa, _log });
 
     expect(_btoa).toHaveBeenCalled();
-    expect(_log.debug).toHaveBeenCalled();
+    expect(_log.error).toHaveBeenCalled();
+  });
+
+  it('throws an error if an image has an invalid mimeType', () => {
+    expect(() => {
+      render({ mimeType: 'invalid/type' });
+    }).toThrow('Unexpected mimeType: "invalid/type"');
   });
 
   it('calls sanitize on svg files', () => {
     const _sanitize = jest.fn();
     const content = 'some content';
     const mimeType = 'image/svg+xml';
+    render({ _sanitize, content, mimeType });
+
+    expect(_sanitize).toHaveBeenCalledWith(content);
+  });
+
+  it('recognizes upper cased svg mimeType', () => {
+    const _sanitize = jest.fn();
+    const content = 'some content';
+    const mimeType = 'IMAGE/SVG+XML';
     render({ _sanitize, content, mimeType });
 
     expect(_sanitize).toHaveBeenCalledWith(content);
@@ -74,5 +89,24 @@ describe(__filename, () => {
     render({ _sanitize, content, mimeType });
 
     expect(_sanitize).not.toHaveBeenCalled();
+  });
+
+  it('sanitizes an svg file', () => {
+    const content = `
+      <svg version="1.1">
+        <circle fill="red" />
+        <script>alert('XSS via SVG')</script>
+        </svg>
+      `;
+    const root = render({ content, mimeType: 'image/svg+xml' });
+
+    const img: ShallowWrapper = root.find('img');
+    const src: string = img.prop('src');
+    const parts = src.split(',');
+    const contentPart = parts[parts.length - 1];
+
+    const svg = String(atob(contentPart));
+    expect(svg).toContain('<circle');
+    expect(svg).not.toContain('<script>');
   });
 });
--- a/src/components/ImageView/index.spec.tsx
+++ b/src/components/ImageView/index.spec.tsx
@@ -36,25 +36,6 @@ describe(__filename, () => {
     );
   });
 
-  it('throws an error when btoa fails', () => {
-    const message = 'An error message';
-    const _btoa = jest.fn().mockImplementation(() => {
-      throw new Error(message);
-    });
-
-    expect(() => {
-      render({ _btoa });
-    }).toThrow(message);
-  });
-
-  it('logs an error message if an image has an invalid mimeType', () => {
-    const _log = createFakeLogger();
-
-    render({ _log, mimeType: 'invalid/type' });
-
-    expect(_log.error).toHaveBeenCalled();
-  });
-
   it('renders a message if an image has an invalid mimeType', () => {
     const root = render({ mimeType: 'invalid/type' });
 
@@ -93,8 +74,8 @@ describe(__filename, () => {
       <svg version="1.1">
         <circle fill="red" />
         <script>alert('XSS via SVG')</script>
-        </svg>
-      `;
+      </svg>
+    `;
     const root = render({ content, mimeType: 'image/svg+xml' });
 
     const img: ShallowWrapper = root.find('img');
--- a/src/components/ImageView/index.tsx
+++ b/src/components/ImageView/index.tsx
@@ -47,15 +47,10 @@ const convertImageData = ({
     return null;
   }
 
-  try {
-    return _btoa(
-      // Run sanitize on svg files.
-      mimeTypeCased === 'image/svg+xml' ? _sanitize(content) : content,
-    );
-  } catch (error) {
-    _log.error('Error converting image data to ascii:', error);
-    return null;
-  }
+  return _btoa(
+    // Run sanitize on svg files.
+    mimeTypeCased === 'image/svg+xml' ? _sanitize(content) : content,
+  );
 };
 
 const ImageViewBase = ({
--- a/src/components/ImageView/index.tsx
+++ b/src/components/ImageView/index.tsx
@@ -1,3 +1,4 @@
+import purify from 'dompurify';
 import log from 'loglevel';
 import * as React from 'react';
 
@@ -6,6 +7,7 @@ import styles from './styles.module.scss';
 type PublicProps = {
   _btoa?: typeof btoa;
   _log?: typeof log;
+  _sanitize?: typeof purify.sanitize;
   mimeType: string;
   content: string;
 };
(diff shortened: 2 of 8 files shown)


wagnerand:
Would it be possible to show the new version of an image on the `Compare` page, along with a note that the displayed image is the new one?

bobsilverberg:
> Would it be possible to show the new version of an image on the Compare page, along with a note that the displayed image is the new one?

I don't think that's possible with the current API, which just returns a diff to us between the old and new image, but maybe it's possible with an API change?

@EnTeQuAk What do you think?

bobsilverberg:
Following up on @muffinresearch's comment in IRC that:

> it's possible to create an SVG that will execute script as a data-uri if you right-click and open it in the browser. That requires an specific action, and that's  not the same as the app context though, so I'm not sure to what extent that would be useful. I'd suggest running this past security in-case of other ways this could potentially be.

@psiinon Can you comment on any issues you might forsee with allowing `data:` urls for `img-src`?

psiinon:
@bobsilverberg could we add a message telling users not to right click and open in the browser? Or even better detect script tags in svgs and then if present just display them as text rather than images?

muffinresearch:
> @bobsilverberg could we add a message telling users not to right click and open in the browser? Or even better detect script tags in svgs and then if present just display them as text rather than images?

One possible solution for that specific case would be to run any SVG through DOMPurify before base64 encoding it.

bobsilverberg:
> > @bobsilverberg could we add a message telling users not to right click and open in the browser? Or even better detect script tags in svgs and then if present just display them as text rather than images?
> 
> One possible solution for that specific case would be to run any SVG through DOMPurify before base64 encoding it.

I'll give that a try, thanks @muffinresearch and @psiinon!

bobsilverberg:
Ok, I'm now running the svg source through DOMPurify. Do we think that's enough to deal with the fact that we're now allowing `data:` URLs in `img-src`? @psiinon 

@kumar303 This is ready for a review now.

src/components/ImageView/index.spec.tsx
```diff
@@ -55,4 +57,22 @@ describe(__filename, () => {
     expect(_btoa).toHaveBeenCalled();
     expect(_log.debug).toHaveBeenCalled();
   });
+
+  it('calls sanitize on svg files', () => {
+    const _sanitize = jest.fn();
+    const content = 'some content';
+    const mimeType = 'image/svg+xml';
+    render({ _sanitize, content, mimeType });
```
kumar303: I think we should add an additional test to cover the integration of all parts. I'm mostly concerned about this because the test I'm commenting on doesn't cover the actual `src` value.

This is a bit paranoid and usually I don't like to test functionality of third party libraries but I think this could go a long way to ensure everything is hooked up correctly:

```diff
diff --git a/src/components/ImageView/index.spec.tsx b/src/components/ImageView/index.spec.tsx
index 0ff5fd2..fb48bb0 100644
--- a/src/components/ImageView/index.spec.tsx
+++ b/src/components/ImageView/index.spec.tsx
@@ -3,23 +3,23 @@ import { shallow } from 'enzyme';
 
 import { createFakeLogger } from '../../test-helpers';
 
-import ImageView from '.';
+import ImageView, { PublicProps } from '.';
 
 describe(__filename, () => {
   const render = ({
     _btoa = btoa,
     _log = createFakeLogger(),
-    _sanitize = jest.fn(),
     content = 'some image content',
     mimeType = 'mime/type',
-  } = {}) => {
+    ...props
+  }: Partial<PublicProps> = {}) => {
     return shallow(
       <ImageView
         _btoa={_btoa}
         _log={_log}
-        _sanitize={_sanitize}
         mimeType={mimeType}
         content={content}
+        {...props}
       />,
     );
   };
@@ -75,4 +75,24 @@ describe(__filename, () => {
 
     expect(_sanitize).not.toHaveBeenCalled();
   });
+
+  it('sanitizes an svg file', () => {
+    const content = `
+      <svg version="1.1">
+        <circle fill="red" />
+        <script>alert('XSS via SVG')</script>
+      </svg>
+    `;
+    const root = render({ content, mimeType: 'image/svg+xml' });
+
+    const parts = root
+      .find('img')
+      .prop('src')
+      .split(',');
+    const contentPart = parts[parts.length - 1];
+
+    const svg = String(atob(contentPart));
+    expect(svg).toContain('<circle');
+    expect(svg).not.toContain('<script>');
+  });
 });
diff --git a/src/components/ImageView/index.tsx b/src/components/ImageView/index.tsx
index 971d781..93c040f 100644
--- a/src/components/ImageView/index.tsx
+++ b/src/components/ImageView/index.tsx
@@ -4,7 +4,7 @@ import * as React from 'react';
 
 import styles from './styles.module.scss';
 
-type PublicProps = {
+export type PublicProps = {
   _btoa?: typeof btoa;
   _log?: typeof log;
   _sanitize?: typeof purify.sanitize;

```

src/components/ImageView/index.tsx
```diff
@@ -6,31 +7,51 @@ import styles from './styles.module.scss';
...
   _log = log,
+  _sanitize = purify.sanitize,
   content,
+  mimeType,
 }: ConvertImageDataParams): string | null => {
   try {
-    return _btoa(content);
+    return _btoa(
+      // Run sanitize on svg files.
+      mimeType === 'image/svg+xml' ? _sanitize(content) : content,
```
kumar303: The security of this sanitization relies on `mimeType`. I think that's probably OK but if someone could figure out a way to set the mime type to `text/html` then that would result in an XSS'able data URL (maybe). I know it sounds paranoid but I think it wouldn't hurt to just add a simple check right above this line, like:

```tsx
if (!mimeType.startsWith('image')) {
  throw new Error(`Unexpected mimeType: "${mimeType}"`);
}
```

kumar303: Another bit of paranoia 😛 : what about converting `mimeType` to lower case before making the comparison? I think an upper case `IMAGE/SVG+XML` would still be valid in the browser (if someone figured out how to do that).

src/pages/Browse/index.tsx
```diff
@@ -159,11 +160,15 @@ export class BrowseBase extends React.Component<Props> {
...
+              ) : (
+                <CodeView
+                  mimeType={file.mimeType}
+                  content={file.content}
+                  version={version}
+                />
+              )}
             </Col>
             <Col md="1">
               <CodeOverview content={file.content} version={version} />
```
kumar303: Funny enough, passing image content to `CodeOverview` totally works. We should probably not do that, though, since it could be confusing. I think you could pass it an empty string for the case of images to not show anything. If that doesn't work, file an issue.

src/server/index.tsx
```diff
@@ -78,6 +78,10 @@ export const createServer = ({
   const staticSrc = env.PUBLIC_URL
     ? `${env.PUBLIC_URL}${STATIC_PATH}`
     : "'none'";
+  const imgSrc = [staticSrc];
+  if (env.PUBLIC_URL) {
```
kumar303: Why is `data:` only added when `PUBLIC_URL` exists? How are they related?

src/server/index.tsx
```diff
@@ -248,7 +252,7 @@ export const createServer = ({
       helmet.contentSecurityPolicy({
         directives: {
           ...prodCSP,
-          imgSrc: ["'self'"],
+          imgSrc: ["'self'", 'data:'],
```
kumar303: Could this be: `[...prodCSP.imgSrc, "'self'"]` ?

src/components/ImageView/index.tsx
```diff
@@ -6,31 +7,51 @@ import styles from './styles.module.scss';
...
+    );
   } catch (error) {
     _log.debug('Error converting image data to ascii:', error);
```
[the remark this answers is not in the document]
bobsilverberg: I agree that we shouldn't be logging a `debug` message. I'm not sure why I did that. I did a quick test, and it looks like if we throw an error here, then the entire site is replaced with an error message. It looks like this is a development-only thing, but I'm not sure why the message isn't displayed inside the site. I think displaying an error in the content area would be the best thing to do, but I'm not sure what other changes are needed to support that.

Perhaps it's best to just log an `error` for now and open a follow-up about throwing an error that can be properly displayed in the content area.

src/pages/Browse/index.tsx
bobsilverberg: Passing an empty string works and the `CodeOverview` appears as empty.

src/server/index.tsx
bobsilverberg: Tbh, I'm not sure, but I was trying to preserve the logic from above which seems to say "if `env.PUBLIC_URL` is falsey, then set the value to `"'none'"`. But I could very well not understand what's going on here. Without this `if`, we end up with an `imgSrc` of `["'none'", 'data:']` when `env.PUBLIC_URL` is falsey. Is that what we want?

bobsilverberg: Ok, I made that change. You can comment on re-review if this is what we want.

src/components/ImageView/index.spec.tsx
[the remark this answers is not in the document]
bobsilverberg: I think the integration test you suggested below will cover this. Correct?

bobsilverberg: Yes, this seems like a good test. I'll add it.

src/pages/Browse/index.tsx
bobsilverberg: I think that ideally we'd also pass no `version`, and `CodeOverview` would be smart enough not to do all the message processing if no `version` were passed, but that could be a follow-up. Or maybe we render `null` instead of `CodeOverview` when `file.type === 'image'`?

bobsilverberg:
@kumar303 this is ready for another look, thanks.

src/components/ImageView/index.tsx
```diff
@@ -1,36 +1,62 @@
...
+
   try {
-    return _btoa(content);
+    return _btoa(
+      // Run sanitize on svg files.
+      mimeTypeCased === 'image/svg+xml' ? _sanitize(content) : content,
+    );
   } catch (error) {
-    _log.debug('Error converting image data to ascii:', error);
+    _log.error('Error converting image data to ascii:', error);
```
kumar303: > I did a quick test, and it looks like if we throw an error here, then the entire site is replaced with an error message. It looks like this is a development-only thing, but I'm not sure why the message isn't displayed inside the site. 

How would I reproduce this? I am trying out the patch locally and I can't get it to throw an error after editing the code and viewing http://localhost:3000/en-US/browse/494431/versions/1532144/

bobsilverberg: Here's a patch I used to force an error:

```diff
diff --git a/src/components/ImageView/index.tsx b/src/components/ImageView/index.tsx
index fd37d13..a858781 100644
--- a/src/components/ImageView/index.tsx
+++ b/src/components/ImageView/index.tsx
@@ -21,7 +21,9 @@ type ConvertImageDataParams = {
 };
 
 const convertImageData = ({
-  _btoa = btoa,
+  _btoa = () => {
+    throw new Error('an error from btoa');
+  },
   _log = log,
   _sanitize = purify.sanitize,
   content,
@@ -38,8 +40,7 @@ const convertImageData = ({
       mimeTypeCased === 'image/svg+xml' ? _sanitize(content) : content,
     );
   } catch (error) {
-    _log.error('Error converting image data to ascii:', error);
-    return null;
+    throw new Error(`error: ${error.message}`);
   }
 };
```

Wnen I visit this URL: http://localhost:3000/en-US/browse/494431/versions/1532144/?path=skin%2Fabb-logo.png I see it as a full page error.

muffinresearch:
@bobsilverberg One question, is there a safe-list of image content-types? Might be a nice idea to limit the output of images in this way to a specific list of content mime types.

bobsilverberg:
> @bobsilverberg One question, is there a safe-list of image content-types? Might be a nice idea to limit the output of images in this way to a specific list of content mime types.

@muffinresearch I did add a check, at @kumar303's suggestion, to throw an error if we encounter a `mimeType` that doesn't start with `image`. I assume we can find a list of valid image mimeTypes somewhere that we could use to secure this even further. Do you feel we should do that, or is just checking for `image` at the start enough?

muffinresearch:
> > @bobsilverberg One question, is there a safe-list of image content-types? Might be a nice idea to limit the output of images in this way to a specific list of content mime types.
> 
> @muffinresearch I did add a check, at @kumar303's suggestion, to throw an error if we encounter a `mimeType` that doesn't start with `image`. I assume we can find a list of valid image mimeTypes somewhere that we could use to secure this even further. Do you feel we should do that, or is just checking for `image` at the start enough?

Wearing my tinfoil hat, being more restrictive might be better, it would protect against a case like SVG being available via a future different mime-type to the one being used for sanitization. Of course on the flip-side the list might need to be updated for new image formats occasionally.

bobsilverberg:
> Wearing my tinfoil hat, being more restrictive might be better, it would protect against a case like SVG being available via a future different mime-type to the one being used for sanitization. Of course on the flip-side the list might need to be updated for new image formats occasionally.

There's a pretty good list of mime types at https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Complete_list_of_MIME_types, which lists a number of image formats that I'm not even sure Firefox supports displaying. Would it make sense to limit our list to only those types of images that Firefox can display, and just display nothing for any other mime types?

We probably don't even need to throw an error in that case, we can either transform and display the image if we recognize it, or display a message like "Unsupported mime type".

What do you think @muffinresearch / @kumar303?

muffinresearch:
> There's a pretty good list of mime types at https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Complete_list_of_MIME_types, which lists a number of image formats that I'm not even sure Firefox supports displaying. Would it make sense to limit our list to only those types of images that Firefox can display, and just display nothing for any other mime types?
> 
> We probably don't even need to throw an error in that case, we can either transform and display the image if we recognize it, or display a message like "Unsupported mime type".
> 
> What do you think @muffinresearch / @kumar303?

That also raises another question. Do web-exts only support certain image formats?

bobsilverberg:
> That also raises another question. Do web-exts only support certain image formats?

I (probably naively) assume that any format that Firefox can display would also be supported via a Web Extension, but maybe that's not accurate. In any case, as what *we* are doing is displaying the image to a user via an `` tag, then I think we only need to care about whether we can display it or not.

kumar303:
> Here's a list of image formats supported by Firefox (according to MDN): https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Firefox

I like the idea of restricting output to this list ^

This adds a good layer of security.

src/components/ImageView/index.tsx
kumar303: Thanks for the example patch. At first, I thought you were saying that there is an expected error that sometimes happens in real life.

Because we don't expect this to throw an error, I would like to see this try block removed so that the error gets thrown. This will show us the error screen you mentioned which has stack traces and code highlighting. This is what we want. If there is an error thrown by `btoa()` then that sounds like a bug we need to fix asap.

bobsilverberg:
Ok, I am now allowing `btoa` to throw an error, and I am also using a list of image mime types that are supported by Firefox. If a mime type is received that is not in that list, I display a message to the user.

@kumar303 This is ready for another look.

src/components/ImageView/index.tsx
```diff
@@ -1,41 +1,81 @@
...
   content,
+  mimeType,
 }: ConvertImageDataParams): string | null => {
-  try {
-    return _btoa(content);
-  } catch (error) {
-    _log.debug('Error converting image data to ascii:', error);
+  const mimeTypeCased = mimeType.toLowerCase();
+  if (!supportedImageMimeTypes.includes(mimeTypeCased)) {
+    _log.error('Unrecognized mimeType :', mimeType);
```
kumar303: Since this message will be mixed with all other Sentry messages, it would help to make it specifically about images

src/components/ImageView/index.tsx
```diff
@@ -1,41 +1,81 @@
...
-  return imageData ? (
+const ImageViewBase = ({
+  _btoa,
+  _log,
+  _sanitize,
+  content,
+  mimeType,
+}: PublicProps) => {
+  const imageData = convertImageData({
+    _btoa,
```
kumar303: TypeScript makes it safe enough to use spreads here:

```diff
diff --git a/src/components/ImageView/index.tsx b/src/components/ImageView/index.tsx
index 26934ef..8e6955a 100644
--- a/src/components/ImageView/index.tsx
+++ b/src/components/ImageView/index.tsx
@@ -53,20 +53,8 @@ const convertImageData = ({
   );
 };
 
-const ImageViewBase = ({
-  _btoa,
-  _log,
-  _sanitize,
-  content,
-  mimeType,
-}: PublicProps) => {
-  const imageData = convertImageData({
-    _btoa,
-    _log,
-    _sanitize,
-    content,
-    mimeType,
-  });
+const ImageViewBase = ({ mimeType, ...props }: PublicProps) => {
+  const imageData = convertImageData({ mimeType, ...props });
   return (
     <div className={styles.ImageView}>
       {imageData ? (
```

Without TypeScript it's risky because you could have a hard time keeping track of the spread properties.

src/components/ImageView/index.tsx
```diff
@@ -1,41 +1,81 @@
+import purify from 'dompurify';
 import log from 'loglevel';
 import * as React from 'react';
 
+import { gettext } from '../../utils';
 import styles from './styles.module.scss';
 
-type PublicProps = {
+export type PublicProps = {
```
kumar303: These types are the same:

```diff
diff --git a/src/components/ImageView/index.tsx b/src/components/ImageView/index.tsx
index 26934ef..946a3e0 100644
--- a/src/components/ImageView/index.tsx
+++ b/src/components/ImageView/index.tsx
@@ -5,14 +5,16 @@ import * as React from 'react';
 import { gettext } from '../../utils';
 import styles from './styles.module.scss';
 
-export type PublicProps = {
+type ConvertImageDataParams = {
   _btoa?: typeof btoa;
   _log?: typeof log;
   _sanitize?: typeof purify.sanitize;
-  mimeType: string;
   content: string;
+  mimeType: string;
 };
 
+export type PublicProps = ConvertImageDataParams;
+
 // From https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Firefox.
 const supportedImageMimeTypes = [
   'image/jpeg',
@@ -26,14 +28,6 @@ const supportedImageMimeTypes = [
   'image/webp',
 ];
 
-type ConvertImageDataParams = {
-  _btoa?: typeof btoa;
-  _log?: typeof log;
-  _sanitize?: typeof purify.sanitize;
-  content: string;
-  mimeType: string;
-};
-
 const convertImageData = ({
   _btoa = btoa,
   _log = log,
```

I don't think you even need an alias. I was just using it to illustrate the issue.

src/components/ImageView/index.spec.tsx
```diff
@@ -34,25 +36,74 @@ describe(__filename, () => {
...
-    const _btoa = jest.fn().mockImplementation(() => {
-      throw new Error();
-    });
+  it('logs an error message if an image has an invalid mimeType', () => {
     const _log = createFakeLogger();
 
-    render({ _btoa, _log });
+    render({ _log, mimeType: 'invalid/type' });
+
+    expect(_log.error).toHaveBeenCalled();
```
kumar303: I don't mind if you keep this test but I really think these type of tests are not helpful. It will just slow down the test suite and carry a maintenance burden. You already have a test to catch that `null` is returned so you will get the code coverage points 🏀

src/components/ImageView/index.spec.tsx
```diff
@@ -34,25 +36,74 @@ describe(__filename, () => {
...
+
+    expect(_sanitize).not.toHaveBeenCalled();
+  });
+
+  it('sanitizes an svg file', () => {
+    const content = `
+      <svg version="1.1">
+        <circle fill="red" />
+        <script>alert('XSS via SVG')</script>
+        </svg>
```
kumar303: My brain really wants this to be indented two spaces to the left but no biggie 😆

src/server/index.tsx
```diff
@@ -112,7 +114,7 @@ export const createServer = ({
     formAction: ["'none'"],
     frameAncestors: ["'none'"],
     frameSrc: ["'none'"],
-    imgSrc: [staticSrc],
```
kumar303: This looks like it can be inline, no?

```js
imgSrc: [staticSrc, 'data:']
```

src/pages/Browse/index.tsx
```diff
@@ -138,6 +139,22 @@ export class BrowseBase extends React.Component<Props> {
       );
     }
 
+    const getContent = () => {
```
kumar303: Since this doesn't *need* to be a closure, I think it would be slightly nicer as an instance method.

Issue #652, opened 2019-04-26
Use the FullscreenGrid


kumar303:
As a step towards https://github.com/mozilla/addons-code-manager/issues/582 , put the entire app in the `FullscreenGrid` component (added in https://github.com/mozilla/addons-code-manager/issues/634).

Issue #653, opened 2019-04-26
Make side panels accessible when the file tree is open


kumar303:
After https://github.com/mozilla/addons-code-manager/issues/652 we'll have a side panel that stays in one place and lets you scroll through the overflow. This is nice for long file trees but it obscures the metadata and keyboard shortcuts panels. 

Let's make it so you can still scroll through long file trees but also access the other panels more easily. I'm thinking of an [accordion menu](http://ui-patterns.com/patterns/AccordionMenu).

Ultimately, this aims to make the file tree easier to use. See https://github.com/mozilla/addons-code-manager/issues/582

AlexandraMoga:
@kumar303 I've used the new side panel menus to navigate the file tree and I didn't find any significant issues with this new structure.

I can mention a couple of 'hindrances' that one might consider:

1. having to keep switching between different views, especially between the 'Files' and 'Information' menus is probably not an ideal workflow. I know this will steal some space in the file tree, but maybe we should consider leaving the information panel open whenever a file is selected?
2. I'm not sure if I remember correctly, but wasn't the code manager supposed to be usable on smaller viewports? The new layout doesn't look very responsive to me:

![image](https://user-images.githubusercontent.com/31961530/57440489-ccc62a80-7250-11e9-9eea-54ea13ee95d3.png)

kumar303:
>     1. having to keep switching between different views, especially between the 'Files' and 'Information' menus is probably not an ideal workflow. I know this will steal some space in the file tree, but maybe we should consider leaving the information panel open whenever a file is selected?

I asked the reviewers about this and they don't think it will be a problem. They don't need to frequently access the information panel.

> 
>     2. I'm not sure if I remember correctly, but wasn't the code manager supposed to be usable on smaller viewports?

We don't intend to support smaller viewports at this time.

PR #655, opened 2019-04-26
Use FullscreenGrid for everything

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/652

**TODO**: the file tree area is an improvement over master but I'd like to improve it more in https://github.com/mozilla/addons-code-manager/issues/653

Notes:
- I made the Navbar a bit smaller since the header is always present
- I removed borders from some things since the grid panels have borders now
- Everything else looks more or less the same except all screens are in the grid
- I adapted `FullscreenGrid` to make it easier to work with for our specific app

<details>
<summary>Screencast</summary>

</details>

<details>
<summary>Screenshots</summary>

acm_panels_initial
acm_panels_logging_in
acm_panels_loading
acm_panels_browse
acm_panels_compare
acm_panels_errors

</details>

### DIFF
--- a/src/components/FullscreenGrid/index.tsx
+++ b/src/components/FullscreenGrid/index.tsx
@@ -3,38 +3,63 @@ import makeClassName from 'classnames';
 
 import styles from './styles.module.scss';
 
-type PanelProps = {
-  children: React.ReactNode | React.ReactNode[];
+type AnyReactNode = React.ReactNode | React.ReactNode[];
+
+export const Header = ({
+  children,
+  className,
+}: {
+  children?: AnyReactNode;
   className?: string;
-};
-
-export const Header = ({ children, className }: PanelProps) => {
+}) => {
   return (
-    <div className={makeClassName(styles.Header, className)}>{children}</div>
-  );
-};
-
-export const NavPanel = ({ children, className }: PanelProps) => {
-  return (
-    <div className={makeClassName(styles.NavPanel, className)}>{children}</div>
-  );
-};
-
-export const AltPanel = ({ children, className }: PanelProps) => {
-  return (
-    <div className={makeClassName(styles.AltPanel, className)}>{children}</div>
-  );
-};
-
-export const ContentPanel = ({ children, className }: PanelProps) => {
-  return (
-    <div className={makeClassName(styles.ContentPanel, className)}>
+    <header className={makeClassName(styles.Header, className)}>
       {children}
-    </div>
+    </header>
   );
 };
 
-const FullscreenGrid = ({ children, className }: PanelProps) => {
+type ContentShellProps = {
+  altSidePanel?: AnyReactNode;
+  altSidePanelClass?: string;
+  children?: AnyReactNode;
+  className?: string;
+  mainSidePanel?: AnyReactNode;
+  mainSidePanelClass?: string;
+};
+
+export const ContentShell = ({
+  altSidePanel,
+  altSidePanelClass,
+  children,
+  className,
+  mainSidePanel,
+  mainSidePanelClass,
+}: ContentShellProps) => {
+  return (
+    <React.Fragment>
+      <aside
+        className={makeClassName(styles.mainSidePanel, mainSidePanelClass)}
+      >
+        {mainSidePanel}
+      </aside>
+      <main className={makeClassName(styles.content, className)}>
+        {children}
+      </main>
+      <aside className={makeClassName(styles.altSidePanel, altSidePanelClass)}>
+        {altSidePanel}
+      </aside>
+    </React.Fragment>
+  );
+};
+
+const FullscreenGrid = ({
+  children,
+  className,
+}: {
+  children?: AnyReactNode;
+  className?: string;
+}) => {
   return (
     <div className={makeClassName(styles.FullscreenGrid, className)}>
       {children}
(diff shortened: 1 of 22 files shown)


kumar303:
@bobsilverberg I rebased this to get the linter message support in the overview. Clicking on a line always scrolls to the exact line like magic!

src/components/FullscreenGrid/index.tsx
```diff
@@ -3,38 +3,63 @@ import makeClassName from 'classnames';
...
+      <aside
+        className={makeClassName(styles.mainSidePanel, mainSidePanelClass)}
+      >
+        {mainSidePanel}
+      </aside>
+      <main className={makeClassName(styles.content, className)}>
+        {children}
+      </main>
+      <aside className={makeClassName(styles.altSidePanel, altSidePanelClass)}>
+        {altSidePanel}
```
bobsilverberg: So if `altSidePanel` is undefined then it would show an empty panel? There's no current way to hide the `aside` for this `altSidePanel`?

src/components/FullscreenGrid/styles.module.scss
```diff
@@ -1,15 +1,13 @@
...
-$grid-gap: 10px;
-
 .FullscreenGrid {
   display: grid;
-  grid-gap: $grid-gap;
+  grid-gap: $default-padding;
   grid-template-areas:
     'FGHeader FGHeader FGHeader'
-    'FGNavPanel FGContentPanel FGAltPanel'
+    'FGMainSidePanel FGContent FGAltSidePanel'
```
bobsilverberg: What's "FG" stand for?

src/pages/Compare/index.spec.tsx
```diff
@@ -260,7 +261,10 @@ describe(__filename, () => {
 
     const root = render({ store });
 
-    expect(root.find(`.${styles.error}`)).toHaveLength(2);
+    expect(
+      getContentShellPanel(root, 'mainSidePanel').find(`.${styles.error}`),
```
bobsilverberg: Shouldn't this error be displayed in the `Header`, rather than in the `mainSidePanel`?

src/test-helpers.tsx
```diff
@@ -590,3 +592,15 @@ export const simulateLinterProvider = (
     selectedMessageMap,
   });
 };
+
+/*
+ * Given a ShallowWrapper for a component that renders ContentShell,
+ * return a wrapper around one of its panel attributes.
+ */
+export const getContentShellPanel = (
```
bobsilverberg: Nice helper! 👍

stories/CodeView.stories.tsx
```diff
@@ -56,7 +56,12 @@ const render = ({
     version: createInternalVersion(fakeVersion),
     ...moreProps,
   };
-  return renderWithStoreAndRouter(<CodeView {...props} />, store);
+  return renderWithStoreAndRouter(
+    <div className="CodeView-panel">
```
bobsilverberg: Why is this needed?

src/components/FullscreenGrid/index.tsx
```diff
@@ -3,38 +3,63 @@ import makeClassName from 'classnames';
 
 import styles from './styles.module.scss';
 
-type PanelProps = {
-  children: React.ReactNode | React.ReactNode[];
-  className?: string;
-};
+type AnyReactNode = React.ReactNode | React.ReactNode[];
```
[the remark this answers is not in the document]
kumar303: 'Any react node' seems like a way of saying 'a single react node or an array of react nodes' to me. In JSX, there isn't much difference between a single node or an array of nodes.

kumar303:
> My main question is about the right-hand side panel. There seems to be no way to hide it, even if there is no content. This is particulary evident on the `Compare` page, which, AFAICT doesn't support having a `CodeOverview`. Do we plan to implement a `CodeOverview` to work with the `Compare` page? 

Definitely! I thought I'd filed but I guess not. Filed: https://github.com/mozilla/addons-code-manager/issues/672

> Do we want to be able to hide the side panel when it serves no purpose?

I can't really think of when the side panel won't be in use aside from loading screens. We could hide it on the loading screens but I think showing an empty outline is fine, too.

src/components/FullscreenGrid/styles.module.scss
kumar303: It stands for `FullscreenGrid`. As I understand it, `grid-template-area` names are global to all CSS. I wanted to protect this namespace a little bit but I didn't want to prefix it with `FullscreenGrid` because that's too long. The longer these names become, the harder it is to see the pattern of columns and rows.

src/pages/Compare/index.spec.tsx
kumar303: This component is far removed from where the `Header` is rendered so it would be tricky to do that. Also, I think it's still readable to see an error in this panel.

stories/CodeView.stories.tsx
kumar303: `CodeView` no longer renders its own border. This makes it hard to tell where the title of the story ends and the component content begins. I added a border around the story content which highlights the component's container.

kumar303:
@bobsilverberg thanks for looking! Let me know if you have further questions.

PR #664, opened 2019-04-29
getRelativeMessageUid should also return a path

### LINKED ISSUE #663: getRelativeMessageUid should also return a path
Supports #520 

We landed `getRelativeMessageUid` which provides the next message uid to enable navigating between messages, but if we're moving to another file, we need to know which file that is as well, so `getRelativeMessageUid` should also return the path for the file.

bobsilverberg:
Fixes #663


bobsilverberg:
> This change looks good. Thanks for the small PR.

It's nice for a change, isn't it? 😉

Issue #668, opened 2019-04-30
Add a reducer for AccordionMenu


kumar303:
An AccordionMenu component will be added in https://github.com/mozilla/addons-code-manager/issues/667

This will be connected to Redux state so it needs a reducer.

PR #669, opened 2019-04-30
Add a reducer for AccordionMenu

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/668

There's a screencast in https://github.com/mozilla/addons-code-manager/pull/670 that shows how the reducer will be used.

I removed the *registers all the reducers* test because I don't see how this is helpful to us. TypeScript will tell us if we forget to register a reducer in all the right places. This kind of test was more useful in the wild west days of dynamic code.


src/reducers/accordionMenu.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+      let state;
+      // Expand it:
+      state = reducer(state, actions.toggleItem({ itemId }));
+      // This should collapse it:
+      state = reducer(state, actions.toggleItem({ itemId }));
+
+      expect(state.expanded).toEqual(null);
+    });
+
+    it('expands an item if collapsed', () => {
```
bobsilverberg: Do we really need this case? The test above, for `collapses an item if expanded` proves that an item that was expanded will be collapsed when toggling, and that a collapsed item means that `state.expanded === null`. So this test is setting `state.expanded = null` (by expanding and collapsing an item), and then toggling it again, which is basically the same as the test for `expands an item by default`, which starts with `state.expanded === null` and then toggles an item and asserts that `state.expanded === itemId`.

I'm actually okay with the test, as I tend to like having more tests, but I feel like if I had written this test, and you were reviewing it, you might be saying the above to me about why it's not needed. 😉

kumar303: It might be a little overkill but I added this explicit test to differentiate the starting value from that of `initialState` since technically they are different code paths.

kumar303: To be clear, it is executing a unique code path. It's making sure that the *reducer* resets the state before expanding again.

kumar303:
Thanks for the review!

Issue #667, opened 2019-04-30
Add an AccordionMenu component


kumar303:
Add an AccordionMenu component for https://github.com/mozilla/addons-code-manager/issues/653

Ultimately, this aims to make the file tree easier to use. See https://github.com/mozilla/addons-code-manager/issues/582

PR #670, opened 2019-04-30
Add an AccordionMenu component

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/667

~~⚠️ depends on https://github.com/mozilla/addons-code-manager/issues/668~~

<details>
<summary>Screencast</summary>

</details>

### DIFF
--- a/src/components/AccordionMenu/index.spec.tsx
+++ b/src/components/AccordionMenu/index.spec.tsx
@@ -1,0 +1,172 @@
+import * as React from 'react';
+import { shallow } from 'enzyme';
+import { Store } from 'redux';
+
+import configureStore from '../../configureStore';
+import { actions } from '../../reducers/accordionMenu';
+import { createFakeEvent, shallowUntilTarget, spyOn } from '../../test-helpers';
+import styles from './styles.module.scss';
+
+import AccordionMenu, {
+  AccordionItem,
+  AccordionItemBase,
+  DefaultItemProps,
+  PublicItemProps,
+  makeItemContentId,
+} from '.';
+
+describe(__filename, () => {
+  describe('AccordionMenu', () => {
+    it('renders children', () => {
+      const childClass = 'example-class';
+      const root = shallow(
+        <AccordionMenu>
+          <AccordionItem title="Example">
+            <div className={childClass} />
+          </AccordionItem>
+        </AccordionMenu>,
+      );
+
+      expect(root.find(`.${childClass}`)).toHaveLength(1);
+    });
+  });
+
+  describe('AccordionItem', () => {
+    type RenderParams = { store?: Store } & Partial<PublicItemProps> &
+      Partial<DefaultItemProps>;
+
+    const render = ({
+      store = configureStore(),
+      ...moreProps
+    }: RenderParams) => {
+      const props = {
+        children: <span />,
+        title: 'Example Title',
+        ...moreProps,
+      };
+
+      return shallowUntilTarget(
+        <AccordionItem {...props} />,
+        AccordionItemBase,
+        {
+          shallowOptions: { context: { store } },
+        },
+      );
+    };
+
+    const renderExpanded = ({
+      title = 'Example Title',
+      store = configureStore(),
+      ...props
+    }: RenderParams) => {
+      store.dispatch(actions.toggleItem({ itemId: title }));
+      return render({ title, store, ...props });
+    };
+
+    it('does not expand children by default', () => {
+      const childClass = 'example-class';
+
+      const root = render({
+        children: <div className={childClass} />,
+      });
+
+      const item = root.find(`.${styles.itemContent}`);
+      expect(item).not.toHaveClassName(styles.itemExpanded);
+      expect(item).toHaveProp('aria-expanded', 'false');
+
+      // The child still gets rendered, though:
+      expect(item.find(`.${childClass}`)).toHaveLength(1);
+    });
+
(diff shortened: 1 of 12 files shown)


kumar303:
@bobsilverberg this has been rebased and should be ready for review now

src/components/AccordionMenu/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+  children: AnyReactNode;
+  title: string;
+};
+
+export type DefaultItemProps = {
+  expanded: boolean;
+};
+
+type ItemPropsFromState = {
+  expandedFromState: boolean;
```
bobsilverberg: I find this `expanded` and `expandedFromState` a bit confusing. I understand it now after reviewing all the code, but maybe it would be a bit clearer if `expandedFromState` were renamed to just `expanded` (because it represents the "true" expanded state of the item), and if `expanded` were renamed to `expandInitially`, or something like that, to indicate that it is used only to request that an item be initially expanded and after that isn't used for anything.

src/components/AccordionMenu/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+  return `accordionContent${id}`;
+};
+
+export type PublicItemProps = {
+  children: AnyReactNode;
+  title: string;
+};
+
+export type DefaultItemProps = {
+  expanded: boolean;
```
bobsilverberg: I guess that if one creates an `AccordionMenu` and specifies `expanded = true` on more than one `AccordionItem`, then the last one listed would "win"? Is this an issue, or just what one would expect?

src/components/AccordionMenu/index.spec.tsx
```diff
@@ -0,0 +1,163 @@
...
+  AccordionItem,
+  AccordionItemBase,
+  DefaultItemProps,
+  PublicItemProps,
+  makeItemContentId,
+} from '.';
+
+describe(__filename, () => {
+  describe('AccordionMenu', () => {
+    it('renders children', () => {
```
bobsilverberg: The test is called `renders children`, but only one child is created. Do you think it would be better to have multiple children and assert that they all are rendered?

src/components/AccordionMenu/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+};
+
+type ItemProps = PublicItemProps &
+  DefaultItemProps &
+  ItemPropsFromState &
+  ConnectedReduxProps;
+
+export class AccordionItemBase extends React.Component<ItemProps> {
+  static defaultProps = {
+    expanded: false,
```
bobsilverberg: I don't think there's a basic test for this. I.e., a test that renders an item without specifying a value for the `expanded` property, and asserts that `actions.toggleItem` was *not* dispatched.

src/components/AccordionMenu/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+  render() {
+    const { children, expandedFromState, title } = this.props;
+    const contentId = makeItemContentId(title);
+
+    return (
+      <React.Fragment>
+        <Button
+          aria-controls={contentId}
+          onClick={this.onItemClick}
+          className={makeClassName(styles.item, styles.itemButton)}
```
bobsilverberg: There don't seem to be any tests that assert that the `Button` has a class of `styles.item`. Does that matter, or is that not something we'd usually test for?

src/components/AccordionMenu/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+        >
+          {title}
+        </Button>
+        <div
+          aria-expanded={expandedFromState ? 'true' : 'false'}
+          className={makeClassName(styles.item, styles.itemContent, {
```
bobsilverberg: There don't seem to be any tests that assert that the `div` has a class of `styles.item`. Does that matter, or is that not something we'd usually test for?

src/components/AccordionMenu/index.tsx
kumar303: There isn't really an easy way to fix but I think it's OK. It doesn't seem like a hard thing to catch during development.

src/components/AccordionMenu/index.tsx
kumar303: The `styles.item` class is hard-coded configuration so I don't think this is important to test (it won't change based on any inputs). It's more important to add tests for classes that are dynamically assigned based on inputs.

src/components/AccordionMenu/index.tsx
kumar303: Same answer as above. It doesn't matter because this is configuration that won't change based on inputs. A test would only fail if the developer intentionally removed the configuration which would just annoy the developer at best.

src/components/AccordionMenu/index.tsx
kumar303: Ah, yes, I struggled with this. Good idea.

src/components/AccordionMenu/index.spec.tsx
kumar303: In React, one child and multiple children are both accessed through `this.props.children` just the same -- there's no difference. There would be no additional code path to test.

src/components/AccordionMenu/index.tsx
kumar303: Good catch

kumar303:
@bobsilverberg thanks, this is ready for another look

Issue #673, opened 2019-05-01
Fix static React ref in CodeOverview


kumar303:
The CodeOverview component (https://github.com/mozilla/addons-code-manager/issues/621) has a static ref attribute. This was a mistake because every new instance would share the same ref instance. 

It won't be a problem for CodeOverview but if someone were to copy/paste the ref setup code for a new component then they'd see the problem (I may or may not have been that someone).

PR #674, opened 2019-05-01
Do not use a static React ref in CodeOverview

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/673

I also moved `createFakeRef()` to the helpers file because this will be useful to other components.

### DIFF
--- a/src/components/CodeOverview/index.tsx
+++ b/src/components/CodeOverview/index.tsx
@@ -31,7 +31,7 @@ export type DefaultProps = {
     addEventListener: typeof window.addEventListener;
     removeEventListener: typeof window.removeEventListener;
   };
-  initialOverviewRef: React.RefObject<HTMLDivElement>;
+  createOverviewRef: () => React.RefObject<HTMLDivElement> | null;
   overviewPadding: number;
   rowTopPadding: number;
   rowHeight: number;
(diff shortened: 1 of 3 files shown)


src/components/CodeOverview/index.tsx
```diff
@@ -48,7 +48,7 @@ export class CodeOverviewBase extends React.Component<Props, State> {
   static defaultProps = {
     _debounce: debounce,
     _window: window,
-    initialOverviewRef: React.createRef<HTMLDivElement>(),
```
bobsilverberg: Ah, so creating a ref here means that every instance of this component will share a ref? I would have thought that each one would get its own ref, because instances of components don't share props, do they? Is it something about the nature of a ref, or `React.createRef`?

kumar303: > Ah, so creating a ref here means that every instance of this component will share a ref? 

Yes

> I would have thought that each one would get its own ref, because instances of components don't share props, do they? Is it something about the nature of a ref, or `React.createRef`?

It's because the `static` keyword of `static defaultProps` creates a class property, not an instance property. 

I was experimenting with auto-scrolling content in `AccordionItem` using the same ref setup. With this bug, the last item to be rendered was the only one that got scrolled. This makes sense because they were all pointing to the same ref variable; the last one overwrote the others.

kumar303:
Snyk took a nap. I'm going to merge to master and see what happens 💃

Issue #680, opened 2019-05-03
The file menu does not always show the selected file


kumar303:
After making the file tree "sticky" (https://github.com/mozilla/addons-code-manager/issues/653), selecting a file with the keyboard doesn't always show the selected file.

**For QA**: go to the browse page, scroll until the file menu is almost out of view. Press the 'j' key to jump to a file so that the file list gets hidden. It should not be hidden.

Issue #679, opened 2019-05-03
Stop the version dropdowns on the compare page from scrolling out of view


kumar303:
**For QA:** check that browsing the compare page and switching between versions still works.

The app layout entered a fullscreen grid in https://github.com/mozilla/addons-code-manager/issues/652

The compare page is a bit hard to use (but not harder than before!) because the version dropdowns scroll out of view when you scroll through a diff.

<details>
<summary>Example screencast</summary>

![2019-05-03 08 48 27](https://user-images.githubusercontent.com/55398/57141727-89674a00-6d80-11e9-8327-e246cf6df6a1.gif)

</details>

Let's make the controls sticky to fix this.

AlexandraMoga:
Verified fixed on -dev with Fx66, Win10x64

Browsing the compare page is still working, you can change between files and select various versions to compare

![2019-05-07_13-11-54](https://user-images.githubusercontent.com/31961530/57292144-c194c280-70c9-11e9-8de9-192ffd28d740.gif)

Issue #678, opened 2019-05-03
Decouple KeyboardShortcuts from FileTree


kumar303:
Currently, `KeyboardShortcuts` (https://github.com/mozilla/addons-code-manager/issues/459) is coupled with `FileTree` because it relies on the [`tree.pathList`](https://github.com/mozilla/addons-code-manager/blob/f1672ff648a02551eb4b1b7329524943a35d158c/src/components/FileTree/index.tsx#L175). This makes it difficult to render `KeyboardComponents` *outside* of `FileTree`.

Let's make it easier to render `KeyboardComponents` on its own.

bobsilverberg:
This makes sense, but just FYI, one of the reasons I did it that way is because the keyboard shortcuts do not make any sense outside of the context of a `FileTree`, so I wanted to ensure that they are never mounted/rendered unless a tree exists.

kumar303:
> I wanted to ensure that [keyboard shortcuts] are never mounted/rendered unless a tree exists.

We can still do this while decoupling the component. We could either make a parent component responsible for loading the file tree (and passing it as a prop) or make a file tree provider component that the both of them could use.

Really, I just filed this issue so I can land https://github.com/mozilla/addons-code-manager/issues/653 with a temporary solution. The quickest fix was to just let `KeyboardComponents` look for a tree in state and do nothing when one doesn't exist. That's not ideal, though.

Issue #647, opened 2019-04-26
File metadata (info) is not displayed on the compare page


bobsilverberg:
I don't know if this is intentional, but it seems like a missing feature to me. The "Information about this file" box is shown on the Browse page, but not on the Compare page.

kumar303:
Yeah, we should show file info on the compare page. The info would be related to the head / newer version. Showing a download link and hash all seem like useful things.

ioanarusiczki:
Verified on -dev with FF67(Win10). 
Looks good to me. 
I used the following compare view
https://code.addons-dev.allizom.org/en-US/compare/495710/versions/1541610...1688333/

![Information for file](https://user-images.githubusercontent.com/33448286/60433737-338f1100-9c0e-11e9-9b02-b897ad7906f3.gif)

PR #681, opened 2019-05-03
Use AccordionMenu for files, info, key shortcuts

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/653

**NOTE**: in attempt to keep the patch small (ha), I deferred these tasks: 
- Decouple KeyboardShortcuts: https://github.com/mozilla/addons-code-manager/issues/678
- Fix hidden file menu issues: https://github.com/mozilla/addons-code-manager/issues/680

<details>
<summary>Screencast</summary>

</details>

The compare page works, too, but I think it needs this tweak: https://github.com/mozilla/addons-code-manager/issues/679

### DIFF
--- a/src/components/KeyboardShortcuts/index.spec.tsx
+++ b/src/components/KeyboardShortcuts/index.spec.tsx
@@ -2,12 +2,20 @@ import * as React from 'react';
 import { Store } from 'redux';
 
 import configureStore from '../../configureStore';
-import { RelativePathPosition } from '../../reducers/fileTree';
-import { actions as versionActions } from '../../reducers/versions';
+import {
+  actions as fileTreeActions,
+  RelativePathPosition,
+} from '../../reducers/fileTree';
+import {
+  actions as versionsActions,
+  createInternalVersion,
+} from '../../reducers/versions';
 import {
   CreateKeydownEventParams,
   createKeydownEvent,
   createFakeThunk,
+  fakeVersion,
+  fakeVersionEntry,
   shallowUntilTarget,
   spyOn,
 } from '../../test-helpers';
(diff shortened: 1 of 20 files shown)


src/components/KeyboardShortcuts/index.spec.tsx
```diff
@@ -2,12 +2,20 @@ import * as React from 'react';
...
 
 import configureStore from '../../configureStore';
-import { RelativePathPosition } from '../../reducers/fileTree';
-import { actions as versionActions } from '../../reducers/versions';
+import {
+  actions as fileTreeActions,
+  RelativePathPosition,
+} from '../../reducers/fileTree';
+import {
+  actions as versionActions,
```
bobsilverberg: I know this is a carryover from an error of mine, but I think we want to call this `versionsActions` not `versionActions`.

src/components/VersionFileViewer/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+import Loading from '../Loading';
+import { Version, VersionFile } from '../../reducers/versions';
+import { AnyReactNode } from '../../typeUtils';
+import { gettext } from '../../utils';
+import styles from './styles.module.scss';
+
+export enum ItemTitles {
+  Files = 'Files',
+  Information = 'Information',
+  // eslint-disable-next-line no-shadow
```
bobsilverberg: Could you avoid this by using the key `Shortcuts` instead of `KeyboardShortcuts`? I think that would be equally understandable.

src/components/VersionFileViewer/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+            <KeyboardShortcuts
+              currentPath={version.selectedPath}
+              versionId={version.id}
+            />
+          </AccordionItem>
+        </AccordionMenu>
+      }
+      mainSidePanelClass={styles.mainSidePanel}
+      altSidePanel={
+        file ? <CodeOverview content={file.content} version={version} /> : null
```
bobsilverberg: FYI, I think the "view an image" patch that I'm about to land will cause a conflict with this.

src/pages/Browse/index.tsx
```diff
@@ -130,35 +125,14 @@ export class BrowseBase extends React.Component<Props> {
...
-          ) : null
-        }
+      <VersionFileViewer
+        file={file}
+        onSelectFile={this.viewVersionFile}
+        showFileInfo
+        version={version}
       >
-        {file ? (
+        {version && file ? (
```
bobsilverberg: My patch will cause conflicts with this as well.

src/pages/Compare/index.tsx
```diff
@@ -131,14 +130,11 @@ export class CompareBase extends React.Component<Props> {
...
-          version ? (
-            <FileTree onSelect={this.viewVersionFile} versionId={version.id} />
-          ) : (
-            this.renderLoadingMessageOrError(gettext('Loading file tree...'))
-          )
-        }
+      <VersionFileViewer
+        file={null}
+        onSelectFile={this.viewVersionFile}
+        showFileInfo={false}
```
bobsilverberg: Does it make sense that we don't show file info on the compare page? That doesn't seem right to me. Do we have an issue open to address this?

kumar303: Good idea. I found an issue https://github.com/mozilla/addons-code-manager/issues/647

kumar303:
@bobsilverberg Thanks. Can you check that I resolved the image related conflicts correctly?

PR #684, opened 2019-05-03
Condense version compare dropdowns, make sticky

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/679

Additionally, I associated labels and dropdowns for accessibility.

<details>
<summary>Before</summary>

</details>

<details>
<summary>After</summary>

</details>

### DIFF
--- a/src/components/VersionSelect/index.tsx
+++ b/src/components/VersionSelect/index.tsx
@@ -12,7 +12,7 @@ import styles from './styles.module.scss';
 
 export type PublicProps = {
   className?: string;
-  id: string;
+  controlId: string;
   isLoading: boolean;
   isSelectable: (version: VersionsListItem) => boolean;
   label: string;
@@ -52,7 +52,7 @@ class VersionSelectBase extends React.Component<PublicProps> {
   render() {
     const {
       className,
-      id,
+      controlId,
       isLoading,
       label,
       listedVersions,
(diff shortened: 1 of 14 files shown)


src/reducers/versions.tsx
```diff
@@ -699,6 +699,27 @@ export const getCompareInfoKey = ({
   return [addonId, baseVersionId, headVersionId, path].join('/');
 };
 
+export const createInternalCompareInfo = ({
```
kumar303: I needed this for a previous version of the patch but it's no longer needed. We might as well keep it, though. The common pattern is to always use `createInternal*` functions.

src/components/VersionSelect/index.tsx
```diff
@@ -61,14 +63,14 @@ class VersionSelectBase extends React.Component<PublicProps> {
...
-            <FontAwesomeIcon icon="long-arrow-alt-left" />
-          </div>
-        )}
-
-        <Form.Group as={Col} className={className}>
-          <Form.Label>{label}</Form.Label>
+        <Form.Group
+          as={Col}
+          className={makeClassName(className, styles.formGroup)}
+        >
```
willdurand: You could use the `controlId` prop here, instead of passing it to each component below: https://react-bootstrap.github.io/components/forms/#form-group-props

stories/setup/styles.scss
```diff
@@ -48,6 +48,11 @@ body.fullscreen {
   }
 }
 
+.DiffViewStory-panel {
+  border-radius: $border-radius;
+  border: 1px solid $light-gray;
+}
+
 .CodeView-panel {
```
willdurand: This does not look consistent with the rest. I know it has nothing to do with this patch, though.

src/components/VersionSelect/index.tsx
kumar303: Ooh, fancy. Thanks for the tip.

stories/setup/styles.scss
kumar303: That's true. I filed https://github.com/mozilla/addons-code-manager/issues/692

PR #687, opened 2019-05-03
Scroll a FileTreeNode into view when needed

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/680

~~🌵 Depends on https://github.com/mozilla/addons-code-manager/pull/681~~

<details>
<summary>Before</summary>

</details>

<details>
<summary>After</summary>

</details>

### DIFF
--- a/src/components/CodeOverview/index.spec.tsx
+++ b/src/components/CodeOverview/index.spec.tsx
@@ -201,35 +201,6 @@ describe(__filename, () => {
     expect(root.state('overviewHeight')).toEqual(fakeRef.current.clientHeight);
   });
 
-  it('only sets the overview height for defined refs', () => {
-    const { root, instance } = renderWithInstance({
-      createOverviewRef: () => null,
-    });
-
-    const overviewHeight = 200;
-    root.setState({ overviewHeight });
-
-    instance.setOverviewHeight();
-
-    // Make sure no new height was set.
-    expect(root.state('overviewHeight')).toEqual(overviewHeight);
-  });
-
-  it('only sets the overview height for active refs', () => {
-    const { root, instance } = renderWithInstance({
-      // Set an inactive ref.
-      createOverviewRef: () => React.createRef(),
-    });
-
-    const overviewHeight = 200;
-    root.setState({ overviewHeight });
-
-    instance.setOverviewHeight();
-
-    // Make sure no new height was set.
-    expect(root.state('overviewHeight')).toEqual(overviewHeight);
-  });
-
   it('can reset the overview height', () => {
     const { root, instance } = renderWithInstance();
 
--- a/src/components/FileTreeNode/index.spec.tsx
+++ b/src/components/FileTreeNode/index.spec.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { shallow } from 'enzyme';
+import { ListGroup } from 'react-bootstrap';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 import { TreefoldRenderPropsForFileTree } from '../FileTree';
@@ -12,6 +13,7 @@ import {
 import LinterProvider, { LinterProviderInfo } from '../LinterProvider';
 import {
   createFakeExternalLinterResult,
+  createFakeRef,
   fakeExternalLinterMessage,
   fakeVersion,
   simulateLinterProvider,
@@ -19,6 +21,7 @@ import {
 import styles from './styles.module.scss';
 
 import FileTreeNode, {
+  DefaultProps,
   LINTER_KNOWN_LIBRARY_CODE,
   PublicProps,
   findMostSevereTypeForPath,
@@ -66,7 +69,7 @@ export const getTreefoldRenderProps = ({
 };
 
 describe(__filename, () => {
-  type RenderParams = Partial<PublicProps>;
+  type RenderParams = Partial<PublicProps> & Partial<DefaultProps>;
 
   const render = ({
     version = createInternalVersion(fakeVersion),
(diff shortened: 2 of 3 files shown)


bobsilverberg:
Actually, one quirk I noticed, which seems annoying to me, but maybe it wouldn't be to others, is that when the selected file it at the top of the files viewport, and you press `j`, to move to the next file, rather than selecting the next file in the viewable list, the whole list scrolls so the next file is now at the top. I'm not sure if this is something we can fix, but I find it a bit odd. The same thing happens at the bottom of the viewport.

kumar303:
> when the selected file is at the top of the files viewport, and you press `j` to move to the next file ... the whole list scrolls so the next file is now at the top.

I noticed that too. It's a bit disorienting. The list is still usable, though, since the node you are interested in is in view ... I guess?

I checked on Chrome and `element.scrollIntoView()` does the exact same thing. Huh. I feel like it should not be scrolling for these cases.

I took a closer look and we could maybe fix it by only calling `scrollIntoView()` when we know for sure that the element is out of view. Here is an example of checking for elements out of view:

<details>
<summary>Patch</summary>

```diff
diff --git a/src/components/FileTree/index.tsx b/src/components/FileTree/index.tsx
index a6138c5..62f84f6 100644
--- a/src/components/FileTree/index.tsx
+++ b/src/components/FileTree/index.tsx
@@ -36,6 +36,7 @@ export type PublicProps = {
 
 export type DefaultProps = {
   _log: typeof log;
+  createRef: () => React.RefObject<HTMLDivElement> | null;
 };
 
 type PropsFromState = {
@@ -50,8 +51,11 @@ export class FileTreeBase extends React.Component<Props> {
 
   static defaultProps: DefaultProps = {
     _log: log,
+    createRef: () => React.createRef(),
   };
 
+  private treeShellRef = this.props.createRef();
+
   constructor(props: Props) {
     super(props);
     // Allow dependency injection to test all the ways loadData() gets executed.
@@ -72,6 +76,13 @@ export class FileTreeBase extends React.Component<Props> {
     if (version && !tree) {
       dispatch(fileTreeActions.buildTree({ version }));
     }
+
+    if (this.treeShellRef && this.treeShellRef.current) {
+      console.log(
+        'tree top offset:',
+        this.treeShellRef.current.getBoundingClientRect().top,
+      );
+    }
   };
 
   renderNode = (props: TreefoldRenderPropsForFileTree) => {
@@ -163,7 +174,7 @@ export class FileTreeBase extends React.Component<Props> {
             {gettext('Close all folders')}
           </Button>
         </div>
-        <div className={styles.treeShell}>
+        <div ref={this.treeShellRef} className={styles.treeShell}>
           <Treefold
             nodes={[tree.nodes]}
             render={this.renderNode}
diff --git a/src/components/FileTreeNode/index.tsx b/src/components/FileTreeNode/index.tsx
index 0e6b808..6dde41a 100644
--- a/src/components/FileTreeNode/index.tsx
+++ b/src/components/FileTreeNode/index.tsx
@@ -143,7 +143,10 @@ class FileTreeNodeBase<TreeNodeType> extends React.Component<Props> {
   _scrollIntoViewIfNeeded = () => {
     if (this.isSelected()) {
       if (this.nodeRef && this.nodeRef.current) {
-        this.nodeRef.current.scrollIntoView();
+        console.log(
+          'node top offset:',
+          this.nodeRef.current.getBoundingClientRect().top,
+        );
       } else {
         log.warn(`nodeRef.current was unexpectedly empty: ${this.nodeRef}`);
       }
```

</details>

In the console, here's when the element is in view:

Screenshot 2019-05-06 12 42 53


Here's where it's not:

Screenshot 2019-05-06 12 43 07

As you can see, it's complicated because we'd have to prop drill the position of `FileTree` to `FileTreeNode` somehow.

Do you think it's worth fixing? Maybe could you file an issue just so we have it in the backlog?

Maybe there is an existing collision library for React that we could use.

kumar303:
@bobsilverberg thanks for the early look. This is ready for an official review now.

src/components/CodeOverview/index.spec.tsx
```diff
@@ -201,35 +201,6 @@ describe(__filename, () => {
     expect(root.state('overviewHeight')).toEqual(fakeRef.current.clientHeight);
   });
 
-  it('only sets the overview height for defined refs', () => {
```
bobsilverberg: Why is this patch removing these tests?

bobsilverberg:
FYI, I filed #695

src/components/CodeOverview/index.spec.tsx
kumar303: While looking at this code (for unrelated reasons), I realized that these tests check cases that are already covered by TypeScript so they are pointless [1]. 

The first test covers what TypeScript would cover here:

```
src/components/CodeOverview/index.tsx:85:37 - error TS2531: Object is possibly 'null'.

85     this.setState({ overviewHeight: ref.current.clientHeight });
                                       ~~~
```

And the second one:

```
src/components/CodeOverview/index.tsx:86:39 - error TS2531: Object is possibly 'null'.

86       this.setState({ overviewHeight: ref.current.clientHeight });
                                         ~~~~~~~~~~~
```

[1] One could say that the point of the tests is to boost our coverage score but we don't actually count *if* branches so they don't affect our score anyway.

kumar303:
The missing coverage is this line:

https://github.com/mozilla/addons-code-manager/blob/aece67ff59f97b2300ae760127340b4d528989b8/src/components/FileTreeNode/index.tsx#L212

I don't understand it because there is a test that executes this line. I thought maybe it was saying that the default value of `props` is never evoked but I added a test to do that and it still said missing coverage 🤷‍♀

PR #693, opened 2019-05-06
Format query params sent to the API

### LINKED ISSUE #164: Make query param handling more robust

### LINKED ISSUE #666: Error displaying files with special characters in name
Add-ons can contain files with 'non standard' characters in them, and these can prevent those files from being displayed.
In this case the file name is '<script>alert(1);' - its worth noting that the script is _not_ run :)
amo-bad-files

willdurand:
Fixes #164
Fixes #666 😈 

---

You can try this patch by browsing: http://localhost:3000/en-US/browse/616714/versions/1688036/?path=%3Cscript%3Ealert%281%29%3B

### DIFF
--- a/src/api/index.spec.tsx
+++ b/src/api/index.spec.tsx
@@ -16,6 +16,7 @@ import {
   isErrorResponse,
   logOutFromServer,
   makeApiURL,
+  makeQueryString,
 } from '.';
 
 describe(__filename, () => {
@@ -188,6 +189,22 @@ describe(__filename, () => {
         expect.urlWithTheseParams(query),
         expect.any(Object),
       );
+    });
+
+    it('calls makeQueryString to construct the query string', async () => {
+      const _makeQueryString = jest.fn();
+      const query = { foo: '1', bar: 'abc' };
+
+      await callApiWithDefaultApiState({
+        _makeQueryString,
+        endpoint: '/url',
+        // Here, we disable the lang so that only `query` is passed to
+        // `makeQueryString()`.
+        lang: null,
+        query,
+      });
+
+      expect(_makeQueryString).toHaveBeenCalledWith(query);
     });
 
     it('can include credentials', async () => {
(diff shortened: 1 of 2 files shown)


src/api/index.tsx
```diff
@@ -77,6 +79,20 @@ export const makeApiURL = ({
   return parts.join('');
 };
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const makeQueryString = (query: { [key: string]: any }) => {
```
kumar303: How about this?

```diff
diff --git a/src/api/index.tsx b/src/api/index.tsx
index 5e66083..6a58351 100644
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -79,8 +79,9 @@ export const makeApiURL = ({
   return parts.join('');
 };
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export const makeQueryString = (query: { [key: string]: any }) => {
+export const makeQueryString = (query: {
+  [key: string]: string | number | null | void | boolean;
+}) => {
   const resolvedQuery = { ...query };
   Object.keys(resolvedQuery).forEach((key) => {
     const value = resolvedQuery[key];
```

It would catch passing object values by mistake

src/api/index.spec.tsx
```diff
@@ -489,4 +506,47 @@ describe(__filename, () => {
       }).toThrow(/Either `path` or `url` must be defined/);
     });
   });
+
+  describe('makeQueryString', () => {
+    it('transforms an object to a query string', () => {
+      const query = makeQueryString({ user: 123, addon: 321 });
+      expect(query).toContain('user=123');
```
kumar303: ooh, I see that we have the fancy expect helper already. This test (and the others, where applicable) should use it:

```diff
diff --git a/src/api/index.spec.tsx b/src/api/index.spec.tsx
index 854fe7d..1d03db5 100644
--- a/src/api/index.spec.tsx
+++ b/src/api/index.spec.tsx
@@ -510,8 +510,9 @@ describe(__filename, () => {
   describe('makeQueryString', () => {
     it('transforms an object to a query string', () => {
       const query = makeQueryString({ user: 123, addon: 321 });
-      expect(query).toContain('user=123');
-      expect(query).toContain('addon=321');
+      expect(query).toEqual(
+        expect.urlWithTheseParams({ user: '123', addon: '321' }),
+      );
     });
 
     it('ignores undefined query string values', () => {
```

src/api/index.spec.tsx
```diff
@@ -489,4 +506,46 @@ describe(__filename, () => {
       }).toThrow(/Either `path` or `url` must be defined/);
     });
   });
+
+  describe('makeQueryString', () => {
+    it('transforms an object to a query string', () => {
+      const query = makeQueryString({ user: 123, addon: 321 });
+      expect(query).toEqual(expect.urlWithTheseParams({ addon: '321' }));
```
kumar303: This should also check `user: 123`

willdurand: bwaa it was there, I swear :x

willdurand:
erf... sorry for the mistake

PR #696, opened 2019-05-06
Fix type in makeQueryString

willdurand:


src/api/index.tsx
```diff
@@ -80,7 +80,7 @@ export const makeApiURL = ({
 };
 
 export const makeQueryString = (query: {
-  [key: string]: string | number | null | void | boolean;
+  [key: string]: string | number | null | undefined | boolean;
```
kumar303: Really? Weird

willdurand: Yes. I think `undefined !== void` but more `undefined` is included in `void`. The failure comes from the `url` lib, which explicitly requires `undefined`, not `void`. I will file an issue to investigate this `void` vs `undefined` problem.

PR #701, opened 2019-05-07
Add a new job for `yarn serve` on travis-ci

### LINKED ISSUE #698: Add a travis-ci job to check the server build
#700 is about a TS error in server. Travis-CI did not catch this bug so let's see if we can add an extra check for that.

willdurand:
Fixes #698 

---

I wasnt sure but it works as I want.. First commit is the patch, second commit shows that it works in case of a TS error, third commit makes it green again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,6 @@ jobs:
         # `yarn build` takes some time so let's simulate it instead.
         - mkdir build && touch build/index.html
         - yarn serve &
-        # If `yarn serve` fails, the following commands should not be run.
         - sleep 120 && pkill -f node
     - stage: storybook deployment
       <<: *node-prod
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,14 @@ jobs:
     - name: storybook-smoke-test (node@production)
       <<: *node-prod
       script: yarn storybook-smoke-test
+    - name: yarn serve (node@production)
+      <<: *node-prod
+      script:
+        # `yarn build` takes some time so let's simulate it instead.
+        - mkdir build && touch build/index.html
+        - yarn serve &
+        # If `yarn serve` fails, the following commands should not be run.
+        - sleep 120 && pkill -f node
     - stage: storybook deployment
       <<: *node-prod
       if: branch = master AND type = push AND fork = false
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,10 +49,9 @@ jobs:
     - name: yarn serve (node@production)
       <<: *node-prod
       script:
-        # `yarn build` takes some time so let's simulate it instead.
-        - mkdir build && touch build/index.html
+        - yarn build
         - yarn serve &
-        - sleep 120 && pkill -f node
+        - sleep 30 && kill $! # kill last backgrounded process (a.k.a. the node server)
     - stage: storybook deployment
       <<: *node-prod
       if: branch = master AND type = push AND fork = false
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -80,7 +80,7 @@ export const makeApiURL = ({
 };
 
 export const makeQueryString = (query: {
-  [key: string]: string | number | null | undefined | boolean;
+  [key: string]: string | number | null | void | boolean;
 }) => {
   const resolvedQuery = { ...query };
   Object.keys(resolvedQuery).forEach((key) => {
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -80,7 +80,7 @@ export const makeApiURL = ({
 };
 
 export const makeQueryString = (query: {
-  [key: string]: string | number | null | void | boolean;
+  [key: string]: string | number | null | undefined | boolean;
 }) => {
   const resolvedQuery = { ...query };
   Object.keys(resolvedQuery).forEach((key) => {


kumar303:
Please update this PR title with a real title to make sure the merge/squashed commit gets the right message

willdurand:
> Please update this PR title with a real title to make sure the merge/squashed commit gets the right message

I usually edit those messages before merging, but sure.

.travis.yml
```diff
@@ -46,6 +46,13 @@ jobs:
     - name: storybook-smoke-test (node@production)
       <<: *node-prod
       script: yarn storybook-smoke-test
+    - name: yarn serve (node@production)
+      <<: *node-prod
+      script:
+        # `yarn build` takes some time so let's simulate it instead.
+        - mkdir build && touch build/index.html
+        - yarn serve &
```
kumar303: When I run this locally with the commit in this branch that causes a TS error, I don't see any errors. Do you know why that is? I'm on Node 10.15.0. Why does it cause an error on Travis?

willdurand: Do you run with `&` or without?

Travis runs Node v10.15.3. I definitely get the error locally (running `yarn serve` only)

willdurand: I am on v10.13.0 locally (and v11, sometimes)

kumar303: Hmm. I upgraded to 10.15.3 locally and I still don't see any errors by running:

```
touch build/index.html
yarn serve
```

It doesn't seem to make a difference if I send `yarn serve` to the background or not.

willdurand: Try removing your `node_modules` and getting fresh deps?

kumar303: Whoops! I was on the "revert revert" commit, not the "revert" commit 😎  I can reproduce the error locally now.

.travis.yml
```diff
@@ -46,6 +46,13 @@ jobs:
...
+        - sleep 120 && pkill -f node
```
kumar303: This seems fragile. I can foresee `yarn serve` failing for other reasons, like maybe not binding to the port, etc.

Can't we just simulate the way `ts-node` is trying to compile the source?

I tried adding this file as `tsconfig-node.js`:

```js
{
  "extends": "./tsconfig.json",
  "compilerOptions": {
    "module": "commonjs"
  },
  "files": ["scripts/server.tsx"]
}
```

...and running `tsc -p tsconfig-node.json` but it didn't reproduce the error. There must be some other compiler options it's adding. 

I don't want to get out of sync with how `yarn serve` actually works but I'm just wondering if there is a more direct way to check for compilation errors.

willdurand: Dont you think all reasons that fail `yarn serve` should fail the build?

kumar303: I was thinking more, like, what if the port isn't available on the CI machine. Since we don't care about running a server then, yes, this should not fail the build. I guess the port would most likely be available, though.

.travis.yml
```diff
@@ -46,6 +46,13 @@ jobs:
     - name: storybook-smoke-test (node@production)
       <<: *node-prod
       script: yarn storybook-smoke-test
+    - name: yarn serve (node@production)
+      <<: *node-prod
+      script:
+        # `yarn build` takes some time so let's simulate it instead.
+        - mkdir build && touch build/index.html
```
kumar303: This seems fragile. Even though it might be slower I think this should probably run a real build. This job will run in parallel to the others so it ultimately shouldn't slow us down, I'd think.

.travis.yml
kumar303: Maybe only sleep for 30 seconds? If we catch a failure in the wild where TS hadn't finished yet, we could increase the sleep time.

kumar303: Also, `pkill -f node` seems a bit brutal. Are you sure that will be safe? Is every job guaranteed to get its own VM? Travis doesn't use node for anything? Will it ever?

.travis.yml
willdurand: Ok, that sounds good to me.

.travis.yml
willdurand: 👍

willdurand: I am not sure I can catch the right PID but I will try to do that.

> Are you sure that will be safe?

Yes.

> Is every job guaranteed to get its own VM?

Yes. (technically not a VM anymore but its own system/container, yes)

> Travis doesn't use node for anything?

Nope.

> Will it ever?

Yes. If it does, then it can potentially break builds.

willdurand:
@kumar303 thanks for the review, I changed the job to fully build the app and kill only the node server. Maybe we could merge this job with the existing `build` one?

> One idea is to patch `scripts/server.tsx` to accept a command line option like `--compile-only` but that would be a bigger patch, I guess.

I agree, I thought about that but we need to boot the server to trigger most of the server code. I think it is simpler and more accurate to run `yarn serve` and make sure it serves the app.

kumar303:
> I agree, I thought about that but we need to boot the server to trigger most of the server code. I think it is simpler and more accurate to run `yarn serve` and make sure it serves the app.

We don't need to boot the server to check for compilation errors. The compilation happens before any code runs; it's just static analysis. To test this, I uncommented the call to `app.listen()` in `scripts/server.tsx` and I got the exact same compilation errors. 

I think adding `--compile-only` would be a better solution since it would be faster. In fact, the `yarn build` step isn't even necessary. I just tried it.

What do you think about re-doing this patch using the `--compile-only` approach? Would you rather defer it to a follow-up patch?

kumar303:
> In fact, the `yarn build` step isn't even necessary. I just tried it.

Actually, a successful compilation would probably crash soon after so `build` may still be necessary. I really wish there was a way to tell `ts-node` to *just compile*.

willdurand:
@kumar303 if the build step is necessary, then I'd like to keep the unix approach instead of writing more JS code for that. `ts-node` needs the `--compile-only` flag but not our server code 🤷‍♂

kumar303:
> @kumar303 if the build step is necessary, then I'd like to keep the unix approach instead of writing more JS code for that. `ts-node` needs the `--compile-only` flag but not our server code 🤷‍♂

Our test would be 30 seconds faster if we added a `--compile-only` option to `scripts/server.tsx` but, yeah, more code to write. I'll file a low-priority follow-up.

I also bumped this: https://github.com/TypeStrong/ts-node/issues/326

willdurand:
> Our test would be 30 seconds faster if we added a `--compile-only` option to `scripts/server.tsx` but, yeah, more code to write. I'll file a low-priority follow-up.

We would probably get a boost by merging this new job with the existing `build` job because the job exec time is one thing but the number of concurrent jobs slows down a build too. That's what I suggested above.

> I also bumped this: [TypeStrong/ts-node#326](https://github.com/TypeStrong/ts-node/issues/326)

That sounds like the right approach to me.

willdurand:
> We would probably get a boost by merging this new job with the existing `build` job because the job exec time is one thing but the number of concurrent jobs slows down a build too. That's what I suggested above.

we'll see if that slows us down.

PR #703, opened 2019-05-07
Enhance getRelativeDiff to return a path for the next/previous file with a diff

### LINKED ISSUE #702: Create a getRelativeDiff function to return an anchor or path for the next/previous diff
This supports #615 

The `getRelativeDiffAnchor` function we have will be useful on its own for locating the first anchor in a file, but we also need to be able to navigate between diffs in the current file *and* to be able to navigate to the next/previous file with a diff. 

This issue is about creating a `getRelativeDiff` which will return either an anchor or a path depending on whether the next diff is in the current file or a different file.

This new function can then be used in conjunction with a keyboard shortcut and action to navigate between diffs.

bobsilverberg:
Fixes #702

### DIFF
--- a/src/reducers/fileTree.spec.tsx
+++ b/src/reducers/fileTree.spec.tsx
@@ -15,11 +15,7 @@ import reducer, {
   goToRelativeFile,
   initialState,
 } from './fileTree';
-import {
-  ExternalVersionEntry,
-  createInternalVersion,
-  createInternalVersionEntry,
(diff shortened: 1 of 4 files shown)


bobsilverberg:
Hold off reviewing this one. I think I want to do something a bit different.

bobsilverberg:
I'm going to close this and open a new one.

PR #706, opened 2019-05-07
Create a getRelativeDiff function to return an anchor or path for the next/previous diff

### LINKED ISSUE #702: Create a getRelativeDiff function to return an anchor or path for the next/previous diff

bobsilverberg:
Fixes #702

### DIFF
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -2217,24 +2217,28 @@ describe(__filename, () => {
     const _getRelativeDiff = ({ ...params }) => {
       return getRelativeDiff({
         currentAnchor: '',
-        diff: createFakeDiffWithChanges([[]]),
+        diff: createFakeDiffWithChanges([]),
         pathList: [fakeVersion.file.selected_file],
         version: createInternalVersion(fakeVersion),
         ...params,
       });
     };
 
-    it('calls _getRelativeDiffAnchor to locate the next diff', () => {
-      const _getRelativeDiffAnchor = jest.fn();
+    it('returns an anchor from _getRelativeDiffAnchor', () => {
+      const anchor = 'D1';
+      const path = file1;
+      const _getRelativeDiffAnchor = jest.fn().mockReturnValue(anchor);
       const currentAnchor = '';
       const diff = createFakeDiffWithChanges([[]]);
       const position = RelativePathPosition.next;
+      const { version } = getFakeVersionAndPathList([{ path, status: 'M' }]);
 
-      _getRelativeDiff({
+      const result = _getRelativeDiff({
         _getRelativeDiffAnchor,
         currentAnchor,
         diff,
         position,
+        version,
       });
 
       expect(_getRelativeDiffAnchor).toHaveBeenCalledWith({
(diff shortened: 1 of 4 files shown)


src/reducers/versions.tsx
```diff
@@ -509,6 +513,47 @@ export const getRelativeDiffAnchor = ({
...
+};
+
+type RelativeDiffResult = {
+  anchor: string | null;
+  path: string | null;
+};
+
+export const getRelativeDiff = ({
+  _findRelativePathWithDiff = findRelativePathWithDiff,
+  _getRelativeDiffAnchor = getRelativeDiffAnchor,
```
kumar303: The coverage report says that no tests are executing `getRelativeDiff()` with the real `getRelativeDiffAnchor()` dependency. We have TS to make sure it should work but it actually wouldn't hurt to add an integration test to cover the real usage of `getRelativeDiffAnchor()` if it's possible to do so.

src/reducers/versions.tsx
```diff
@@ -509,6 +513,47 @@ export const getRelativeDiffAnchor = ({
...
+  currentAnchor,
+  diff,
+  pathList,
+  position = RelativePathPosition.next,
+  version,
+}: GetRelativeDiffParams): RelativeDiffResult => {
```
kumar303: Instead of returning `RelativeDiffResult` can't this just return a `link: string`? I'd think you would handle an anchor or path just the same, like `history.push(link)`, no?

src/reducers/versions.spec.tsx
```diff
@@ -2208,4 +2210,92 @@ describe(__filename, () => {
...
     });
   });
+
+  describe('getRelativeDiff', () => {
+    const file1 = 'file1.js';
+
+    const _getRelativeDiff = ({ ...params }) => {
+      return getRelativeDiff({
+        currentAnchor: '',
+        diff: createFakeDiffWithChanges([[]]),
```
kumar303: Since arrays are allowed to have zero items, you could make this:

```diff
diff --git a/src/reducers/versions.spec.tsx b/src/reducers/versions.spec.tsx
index f2e0643..ad19ef7 100644
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -2217,7 +2217,7 @@ describe(__filename, () => {
     const _getRelativeDiff = ({ ...params }) => {
       return getRelativeDiff({
         currentAnchor: '',
-        diff: createFakeDiffWithChanges([[]]),
+        diff: createFakeDiffWithChanges([]),
         pathList: [fakeVersion.file.selected_file],
         version: createInternalVersion(fakeVersion),
         ...params,
```

I think that's a little more straightforward.

src/reducers/versions.spec.tsx
```diff
@@ -2208,4 +2210,92 @@ describe(__filename, () => {
...
+      const path = 'some/path.js';
+      const _findRelativePathWithDiff = jest.fn().mockReturnValue(path);
+      const _getRelativeDiffAnchor = jest.fn().mockReturnValue(null);
+
+      const result = _getRelativeDiff({
+        _findRelativePathWithDiff,
+        _getRelativeDiffAnchor,
+      });
+
+      expect(result).toEqual({ anchor: null, path });
```
kumar303: You could just put the call assertion above this line:

```js
expect(_findRelativePathWithDiff).toHaveBeenCalledWith(...)
```

Putting a call assertion on its own (without checking the return value) isn't such a helpful test because calling a function doesn't mean much. It's more important to check the inputs and outputs.

src/reducers/versions.spec.tsx
```diff
@@ -2208,4 +2210,92 @@ describe(__filename, () => {
...
+      const position = RelativePathPosition.next;
+
+      const result = _getRelativeDiff({
+        _getRelativeDiffAnchor,
+        currentAnchor,
+        diff,
+        position,
+      });
+
+      expect(result).toEqual({ anchor, path: null });
```
kumar303: Same here. You could just put the call assertion above this line:

```js
expect(_getRelativeDiffAnchor).toHaveBeenCalledWith(...)
```

src/reducers/versions.tsx
bobsilverberg: I'm not sure about this. I wouldn't want to have to calculate the full path for the link in here, but I could simply return a string that includes the path with or without an anchor appended. E.g., either `file1.js` or `file1.js#D2`.

My plan, though, was to use this information to construct a new location using data in `router.location`, similar to what we do in [`viewVersionFile`](https://github.com/mozilla/addons-code-manager/blob/4bf9a43cae9cb4cc062e5f00b9a174f3bdd2ba56/src/reducers/versions.tsx#L830-L833), and therefore it seemed like it would be more handy to have the `path`, which ends up being a query parameter, and the `anchor`, which ends up being the `hash`, as separate properties.

Does that make sense?

src/reducers/versions.tsx
```diff
@@ -509,6 +513,47 @@ export const getRelativeDiffAnchor = ({
...
+  path: string | null;
+};
+
+export const getRelativeDiff = ({
+  _findRelativePathWithDiff = findRelativePathWithDiff,
+  _getRelativeDiffAnchor = getRelativeDiffAnchor,
+  currentAnchor,
+  diff,
+  pathList,
+  position = RelativePathPosition.next,
```
kumar303: The coverage report says that the default value for this is not executed by any tests. This makes me think that you don't need to set a default for `position` -- why not just make it required? I think it would be better if the calling code always explicitly set a position.

PR #709, opened 2019-05-08
Add an env to use code-manager with a local olympia instance

### LINKED ISSUE #710: Add config to run code-manager and addons-server locally (side by side)
It'd be handy to be able to use a local addons-server instance with code-manager also running locally. Let's have some config for that.

willdurand:
Fixes #710

---

This is a proposal to add a new env to code-manager that we can use with addons-server. I have been using this env for a while, then it broke because of the API version change (fixed in https://github.com/mozilla/addons-nginx/pull/11) today, so I fixed it and I am proposing it here now.

You can test it by installing addons-server and running `yarn olympia` in code-manager after that. Hopefully the added instructions are clear enough :)

### DIFF
--- a/.env.olympia
+++ b/.env.olympia
@@ -1,0 +1,12 @@
+# This is a special configuration file for running code-manager locally using a
+# local addons-server instance.
+
+REACT_APP_API_HOST=http://olympia.test
+
+# The default port (`3000`) is taken by addons-frontend when we run the whole
+# addons-server stack.
+PORT=4000
+
+REACT_APP_USE_INSECURE_PROXY=true
+REACT_APP_CRA_PORT=3100
+REACT_APP_IS_LOCAL_DEV=true
--- a/.env.olympia
+++ b/.env.olympia
@@ -5,7 +5,7 @@ REACT_APP_API_HOST=http://olympia.test
 
 # The default port (`3000`) is taken by addons-frontend when we run the whole
 # addons-server stack.
-PORT=4000
+PORT=5000
 
 REACT_APP_USE_INSECURE_PROXY=true
 REACT_APP_CRA_PORT=3100
--- a/README.md
+++ b/README.md
@@ -103,6 +103,14 @@ This runs [ESLint][] to discover problems within our codebase without executing 
 
 This runs all the _lint_ commands at once.
 
+### `yarn olympia`
+
+A prerequisite to run this command is to [install addons-server locally](https://addons-server.readthedocs.io/en/latest/topics/install/index.html).
+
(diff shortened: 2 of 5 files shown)


.env.olympia
```diff
@@ -0,0 +1,12 @@
+# This is a special configuration file for running code-manager locally using a
+# local addons-server instance.
+
+REACT_APP_API_HOST=http://olympia.test
+
+# The default port (`3000`) is taken by addons-frontend when we run the whole
+# addons-server stack.
+PORT=4000
```
kumar303: What about using 4001 so it doesn't conflict with a prod build? This could be helpful for cookies and other domain related things in the browser.

README.md
```diff
@@ -103,6 +103,14 @@ This runs [ESLint][] to discover problems within our codebase without executing
 
 This runs all the _lint_ commands at once.
 
+### `yarn olympia`
+
+A prerequisite to run this command is to [install addons-server locally](https://addons-server.readthedocs.io/en/latest/topics/install/index.html).
+
+This runs all development processes in a single column using [stmux][]. Open [http://olympia.dev:4000](http://olympia.dev:4000) to view the development site. **Press CTRL-a-? for help**. The application is configured to use a [local AMO API](https://addons-server.readthedocs.io/).
```
kumar303: These should all be `olympia.test` since that's currently the [preferred domain](https://addons-server.readthedocs.io/en/latest/topics/install/docker.html). The `.dev` one was problematic because it's an official domain.

kumar303:
> This mostly looks good but I can't get it to work locally. 

Hold up. I think I forgot to run `docker-compose pull`. I'm trying to freshen everything up...

willdurand:
> This mostly looks good but I can't get it to work locally. When I clear cookies, press the login button, and get to [http://olympia.test:4000/api/v5/accounts/login/start/?config=code-manager&to=](http://olympia.test:4000/api/v5/accounts/login/start/?config=code-manager&to=/) I see _Webpack is still loading!_ and my addons-frontend in Docker seems to be stuck in a loop. I'm not sure why.

ah that looks like your addons-server setup is not working. Did you pull the latest Docker images? Once you pulled the images, you'll have to run `make update_docker`.

Can you see the homepage on http://olympia.test? Sometimes, the addons-server stack does not work super well and it takes time for addons-frontend to recover. Usually I make sure all services are up and running.

README.md
willdurand: Oops, you're right. Typo..

.env.olympia
willdurand: Good idea!

Issue #588, opened 2019-04-15
Provide keyboard shortcuts for hiding and unhiding the file tree


bobsilverberg:
This is a spin-off from #199.

The shortcut should be:
`h` - hide or unhide tree

It should be a toggle which either hides or unhides the tree based on the current state.

We might want to discuss this one a bit, because at this point just hiding or unhiding the tree won't free up real estate for the file contents because there are other things on the left-hand side. I.e., keyboard shortcuts and file metadata. We probably need to define a new view which allows for more screen real estate for the file contents, but which keeps the keyboard shortcuts available and also the file metadata.

Here are examples from the current app:

With tree unhidden:
![Screenshot 2019-04-15 14 41 36](https://user-images.githubusercontent.com/142755/56157145-c95ebc80-5f8c-11e9-8fbb-46e00873d9e7.png)

With tree hidden:
![Screenshot 2019-04-15 14 41 51](https://user-images.githubusercontent.com/142755/56157135-c237ae80-5f8c-11e9-9e10-d0ca598daea1.png)

With tree hidde, hovering over the folder icon shows the tree:
![Screenshot 2019-04-15 14 42 00](https://user-images.githubusercontent.com/142755/56157120-ba780a00-5f8c-11e9-82d8-d334b69ec95f.png)

@wagnerand / @kumar303 / @willdurand what do you think?

willdurand:
Could we try to build a sidebar _à la_ GitLab, which we could collapse but still keep in the page?

Menu expanded:

<img width="1392" alt="Screen Shot 2019-04-15 at 21 47 18" src="https://user-images.githubusercontent.com/217628/56160696-24fa6b80-5fc8-11e9-9d2d-1017431fd797.png">

Menu collapsed:

<img width="1392" alt="Screen Shot 2019-04-15 at 21 47 20" src="https://user-images.githubusercontent.com/217628/56160697-24fa6b80-5fc8-11e9-99fe-247a3c885bd7.png">

This is a menu here but we could build something similar for the left panel.

bobsilverberg:
> Could we try to build a sidebar _à la_ GitLab, which we could collapse but still keep in the page?

I like that, very nice @willdurand! I think we might also want to make sure the file metadata remains visible in this collapsed state. The old app used a footer, which worked well for this, but one of the issues with a footer is that it gets pushed off the page when the file is too large. Although I suppose the same happens with our current layout when the tree is too large as well.

[...]

PR #721, opened 2019-05-09
[WIP] Toggle the main side panel

willdurand:
This is a POC for #588. Once we have something like this, adding a keyboard shortcut will be trivial. You can't see it but there is a tooltip on the button that says "toggle this panel". In addtion, the "collapse ..." button is added by the `ContentShell`, so we can still put any content in the main side panel, without having to worry about positioning the button (we can see that in the screenshots below).

I'll split this patch if we agree to have such a feature.

## Screenshots

Screen Shot 2019-05-09 at 16 15 25
Screen Shot 2019-05-09 at 16 15 19

---

What do you think?

### DIFF
--- a/src/components/FullscreenGrid/ContentShell.spec.tsx
+++ b/src/components/FullscreenGrid/ContentShell.spec.tsx
@@ -1,0 +1,62 @@
+import * as React from 'react';
+
+import configureStore from '../../configureStore';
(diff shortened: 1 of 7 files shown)


willdurand:
Closing in favor of https://github.com/mozilla/addons-code-manager/pull/735..

PR #723, opened 2019-05-09
Select and scroll to the first/last diff on the compare page

### LINKED ISSUE #722: Allow for DiffView to automatically select the first/last diff in a file
This supports #615 

When a user tries to navigate to the next/previous diff and they are already on the last/first diff for the current file, we need to load the next/previous file with a diff, and jump to the first/last diff in that file.

This issue is about enabling that to happen. I.e., adding the ability to specify a URL param which will indicate that the diff being loaded should have the first/last difference in the file selected.

For QA:

The new URL parameters can be tested by loading a file that contains multiple changes in the compare view. Then add `?scrollTo=firstDiff` to the URL, and the page should reload with the first line of the first change scrolled to the top of the window, and selected. Next, change the URL parameter to `?scrollTo=lastDiff`, and the page should reload with the first line of the last change scrolled to the top of the window, and selected.

bobsilverberg:
Fixes #722

### DIFF
--- a/src/components/DiffView/index.spec.tsx
+++ b/src/components/DiffView/index.spec.tsx
@@ -1,4 +1,5 @@
 /* eslint-disable @typescript-eslint/camelcase */
+import queryString from 'query-string';
 import React from 'react';
 import {
   Diff,
@@ -8,7 +9,7 @@ import {
   getChangeKey,
 } from 'react-diff-view';
 import { ShallowWrapper, shallow } from 'enzyme';
-import { Location } from 'history';
+import { History, Location } from 'history';
 
 import basicDiff from './fixtures/basicDiff';
 import multipleDiff from './fixtures/multipleDiff';
@@ -16,9 +17,10 @@ import diffWithDeletions from './fixtures/diffWithDeletions';
 import LinterMessage from '../LinterMessage';
 import LinterProvider, { LinterProviderInfo } from '../LinterProvider';
 import { getLanguageFromMimeType } from '../../utils';
-import { createInternalVersion } from '../../reducers/versions';
+import { DiffPosition, createInternalVersion } from '../../reducers/versions';
 import {
   createContextWithFakeRouter,
+  createFakeHistory,
   createFakeLinterMessagesByPath,
   createFakeLocation,
   fakeVersion,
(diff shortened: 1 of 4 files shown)


bobsilverberg:
I believe there will be conflicts from #738, so it probably makes sense to review that one first.

bobsilverberg:
This has been rebased and is ready to review.

src/components/DiffView/index.spec.tsx
```diff
@@ -222,6 +225,196 @@ describe(__filename, () => {
...
+    const history = createFakeHistory({ location });
+
+    const queryParams = queryString.parse(location.search);
+    const newParams = { ...queryParams };
+
+    delete newParams.showDiff;
+    const newLocation = {
+      ...location,
+      hash: `#${lastDiffAnchor}`,
+      search: `?${queryString.stringify(newParams)}`,
```
willdurand: I am not sure you need to prepend the stringified query string with a `?`, see: https://github.com/sindresorhus/query-string#usage

src/components/DiffView/index.tsx
```diff
@@ -48,12 +56,57 @@ export type Props = PublicProps & DefaultProps & RouterProps;
...
+      _getDiffAnchors,
+      _getRelativeDiffAnchor,
+      diff,
+      history,
+      location,
+    } = this.props;
+
+    if (diff) {
+      const queryParams = queryString.parse(location.search);
+      const { showDiff } = queryParams;
```
willdurand: `showDiff` makes me think that it's a boolean flag that will load some diff, but it's actually about revealing the anchor of the first diff of the next file/last diff of the previous file.

Could we find a more accurate name for this query parameter? `scrollPosition` maybe? In the future, we might want to navigate between comments, so we could have `firstDiff`, `lastDiff`, and later: `firstComment`, 	`lastComment`. I don't know, what do you think?

src/reducers/versions.tsx
```diff
@@ -470,46 +475,43 @@ export const getDiffAnchors = (diff: DiffInfo): string[] => {
...
-    // Since we aren't looking for an anchor relative to an existing one,
-    // just get the first anchor.
-    newIndex = 0;
-  } else {
-    const currentIndex = anchors.indexOf(currentAnchor);
-    if (currentIndex < 0) {
-      throw new Error(`Could not locate anchor: ${currentAnchor} in the diff.`);
+      newIndex =
+        position === RelativePathPosition.previous
+          ? currentIndex - 1
```
willdurand: What if `currentIndex` is equal to `0`?

src/components/DiffView/index.tsx
bobsilverberg: Good idea. I replaced the URL parameter with `scrollTo`, and replaced the `DiffPosition` `enum` with `ScrollTarget`.

src/reducers/versions.tsx
bobsilverberg: That's dealt with by the `if` on line 509 which checks `if (newIndex >= 0 && newIndex < anchors.length)`. In the case of `currentIndex` being equal to `0`, that `if` evaluates to `false` (because `newIndex` would be `-1`) and therefore we fall out the bottom of the larger `if` and return `null` at line 515.

bobsilverberg: @willdurand I'll hold off merging this so you can take a look and see if that addresses your concern.

PR #727, opened 2019-05-09
Change getDiffAnchors, getRelativeDiffAnchor and getRelativeDiff to accept an array of diffs

### LINKED ISSUE #726: Change API and reducer code to support no array of diffs
Depends on #738 
Depends on https://github.com/mozilla/addons-server/pull/11402

This issue is now just about making the changes to support the fact that the API will return a single diff object and not an array of diffs.

For QA:

Please check for regressions in the Compare feature. Also note that in order for this to work on dev, both this change and the change from https://github.com/mozilla/addons-server/pull/11402 need to have landed. I will take care of landing them at the same time.

bobsilverberg:
Fixes #726

### DIFF
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -2059,13 +2059,13 @@ describe(__filename, () => {
 
   describe('getRelativeDiffAnchor', () => {
     it('returns null if there are no changes in the diff', () => {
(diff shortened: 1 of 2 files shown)


bobsilverberg:
Hold off reviewing this. It may become obsolete.

bobsilverberg:
We're not going to need this, so closing.

Issue #413, opened 2019-03-18
Indicate which files / directories have been added, modified, or deleted


kumar303:
User story:

> As a reviewer, I want to quickly see which files have been added, modified, or deleted in a new version so that I only need to review the relevant files

As an example, the current file viewer highlights the modified files / directories in yellow but we don't need to do it exactly this way. 

Here, the `install.rdf` file was modified and the `META-INF` directory contains modified files:

<img width="1192" alt="Screenshot 2019-03-18 12 19 24" src="https://user-images.githubusercontent.com/55398/54549459-15abe200-4978-11e9-9df1-4ab1a85e81c8.png">

Here, the `META-INF` directory and the `manifest.json` file were deleted:

<img width="1193" alt="Screenshot 2019-03-18 12 21 35" src="https://user-images.githubusercontent.com/55398/54549624-73d8c500-4978-11e9-92a1-1d3a95c7e2c8.png">

https://reviewers.addons-dev.allizom.org/en-US/firefox/files/compare/262229...262228/#top

willdurand:
Has the API landed for that?

kumar303:
Is this blocked? I thought we could use `mode` in the `diff` response of [the compare endpoint](https://addons-server.readthedocs.io/en/latest/topics/api/reviewers.html#compare) to figure out if the file was added / deleted.

willdurand:
We need confirmation from @EnTeQuAk but I remember a discussion about the `file.entries` containing the information about the added/deleted/renamed files.

EnTeQuAk:
> Is this blocked? I thought we could use mode in the diff response of the compare endpoint to figure out if the file was added / deleted.

That is actually the API. The only part that hasn't landed yet is support for detecting copy and renames but that's a nice-to-have and doesn't exist in the current tool.

If you need that information in `files.entries` too, let me know but that feels somehow like duplication to me.

willdurand:
I am confused because I thought the Compare API returned a `diff` for a single file, but it looks like we want the `mode` for all the files here, hence the discussion I still think we had a while ago about having the Compare API returning `file.entries` with all the files for the two compared versions + the `mode` for each file (mode being: added/deleted/renamed).

EnTeQuAk:
Some background: The diff api can very happily return a diff for all files (needs a change but we explicitly strict the diff currently, the diff is always generated for all files in the background).

[...]

Issue #431, opened 2019-03-20
Add storybook for deep file trees, long names, and linter messages


kumar303:
Let's make a storybook out of this file tree: http://localhost:3000/en-US/browse/494431/versions/1532144/

It showcases some important style behaviors:
* A deep tree of files
* Long file names that might need to be truncated
* Icons for linter messages 

Example:

<img width="568" alt="Screenshot 2019-03-20 12 57 59" src="https://user-images.githubusercontent.com/55398/54708034-740bc880-4b10-11e9-8e47-08791e51b8af.png">

AlexandraMoga:
Could you please add a qa:not needed label to this issue?
Thanks

PR #728, opened 2019-05-09
Indicate file diff status in FileTreeNode

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/413

I didn't have time to add a story but we could do that in https://github.com/mozilla/addons-code-manager/issues/431 💳 

Examples, mostly from http://localhost:3000/en-US/compare/616645/versions/1655530...1655531/?path=scripts/background.js

Screenshot 2019-05-17 11 41 50
Screenshot 2019-05-17 11 42 07
Screenshot 2019-05-17 11 42 30
Screenshot 2019-05-17 11 42 51

### DIFF
--- a/src/components/FileTreeNode/index.spec.tsx
+++ b/src/components/FileTreeNode/index.spec.tsx
@@ -5,7 +5,6 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 import configureStore from '../../configureStore';
 import { TreefoldRenderPropsForFileTree } from '../FileTree';
-import { actions as versionsActions } from '../../reducers/versions';
 import {
   ExternalLinterMessage,
   createInternalMessage,
@@ -13,6 +12,11 @@ import {
 } from '../../reducers/linter';
 import LinterProvider, { LinterProviderInfo } from '../LinterProvider';
 import {
+  actions as versionsActions,
+  VersionEntryStatus,
+} from '../../reducers/versions';
+import {
+  createExternalVersionWithEntries,
   createFakeExternalLinterResult,
   createFakeRef,
   createStoreWithVersion,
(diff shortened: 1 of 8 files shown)


willdurand:
I like the UI for that but I am confused by the colors. What's the meaning of each color? (Edit: it's in the code but from the screenshots, I could not get it).

I hardly see the difference between the green and red ones, so I'd like to see another visual indicator or a better color contrast. Maybe we could have the letters `A`, `M` and `D`.

kumar303:
> I like the UI for that but I am confused by the colors. What's the meaning of each color? (Edit: it's in the code but from the screenshots, I could not get it).

I intended to loosely match the colors of an actual diff:

Screenshot 2019-05-15 09 30 58


That is, green for *added* and red for *deleted*. The orange for *modified* doesn't exist in diffs but that was just to indicate that something was both deleted and added. I suppose *modified* could be green.

> 
> I hardly see the difference between the green and red ones, so I'd like to see another visual indicator or a better color contrast. Maybe we could have the letters `A`, `M` and `D`.

Ah, good point. I'll see if I can add the letters to it.

kumar303:
@willdurand does the pink and green line highlights from the screenshot in https://github.com/mozilla/addons-code-manager/pull/728#issuecomment-492681617 have enough contrast for you? I could follow those colors.

I may leave the addition of showing a letter (`A`, `D`) for a follow-up, just to keep it moving.

Even if the contrast between deleted and modified is not high enough for some folks, that's probably OK. It's more important that one can immediately spot a *changed* file / directory in the tree.

willdurand:
@kumar303 yep, these pink and green colors are ok. Maybe we could add html titles at least, and ship the letters later..

willdurand:
I quickly retried this patch and the colors look a bit better, maybe a bit too pale but at least distinguishable.

kumar303:
@willdurand I darkened the colors slightly. What do you think? The contrast still fails ﻿﻿https://color.review/ but it's better than before. If I darken it much more, the hues will become too close on the visible spectrum.

willdurand:
> @willdurand I darkened the colors slightly. What do you think? The contrast still fails ﻿﻿[color.review](https://color.review/) but it's better than before. If I darken it much more, the hues will become too close on the visible spectrum.

The new contrast passed my validator 😛 

I suppose the idea of having letters (or `+/~/-` signs, maybe) in addition to the colors would solve what you mention in your last sentence.

PR #729, opened 2019-05-10
Extract ContentShell component

### LINKED ISSUE #730: Move `ContentShell` component to its own file
Supports #588

---

This is needed because we'll likely connect this component as well as the `FullscreenGrid` component. There will be too much code into this `FullscreenGrid.tsx` file, so let's split it.

**For QA:** there should be no visual regression, i.e. pages should show 3 panels (left panel usually has the accordion menu, center panel has code, right panel has a code overview).

willdurand:
Fixes #730

---

This patch supports what's described in #721. I kept the `ContentShell`
files in `FullscreenGrid/` because it is strongly coupled.

### DIFF
--- a/src/components/FullscreenGrid/index.tsx
+++ b/src/components/FullscreenGrid/index.tsx
@@ -11,7 +11,10 @@ type PublicProps = {
 
 type Props = PublicProps;
 
-export type PanelAttribs = 'altSidePanel' | 'mainSidePanel';
+export enum PanelAttribs {
+  altSidePanel = 'altSidePanel',
+  mainSidePanel = 'mainSidePanel',
+}
 
 export const Header = ({
   children,
--- a/src/components/FullscreenGrid/index.tsx
+++ b/src/components/FullscreenGrid/index.tsx
@@ -2,7 +2,17 @@ import * as React from 'react';
 import makeClassName from 'classnames';
 
 import { AnyReactNode } from '../../typeUtils';
+import ContentShellComponent from './ContentShell';
 import styles from './styles.module.scss';
+
+type PublicProps = {
+  children?: AnyReactNode;
+  className?: string;
+};
+
+type Props = PublicProps;
+
+export type PanelAttribs = 'altSidePanel' | 'mainSidePanel';
 
 export const Header = ({
   children,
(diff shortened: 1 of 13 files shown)


src/components/FullscreenGrid/index.tsx
```diff
@@ -2,8 +2,18 @@ import * as React from 'react';
...
 import styles from './styles.module.scss';
 
+type PublicProps = {
+  children?: AnyReactNode;
+  className?: string;
+};
+
+type Props = PublicProps;
+
+export type PanelAttribs = 'altSidePanel' | 'mainSidePanel';
```
bobsilverberg: I know this is old code, but could this be an `enum` instead?

src/components/FullscreenGrid/index.tsx
```diff
@@ -18,54 +28,14 @@ export const Header = ({
...
+export const ContentShell = ContentShellComponent;
 
-const FullscreenGrid = ({
-  children,
-  className,
-}: {
-  children?: AnyReactNode;
-  className?: string;
-}) => {
+export const FullscreenGridBase = ({ children, className }: Props) => {
```
bobsilverberg: Out of interest, why did this need to be converted into a class component?

src/components/FullscreenGrid/index.tsx
```diff
@@ -18,54 +28,14 @@ export const Header = ({
...
-      <main className={makeClassName(styles.content, className)}>
-        {children}
-      </main>
-      <aside className={makeClassName(styles.altSidePanel, altSidePanelClass)}>
-        {altSidePanel}
-      </aside>
-    </React.Fragment>
-  );
-};
+export const ContentShell = ContentShellComponent;
```
bobsilverberg: I find this a bit odd. Why not just make `ContentShell` a component in its own right, as we've done with basically every other component. I realize that we don't ever imagine using it outside of a `FullscreenGrid`, but it still feels weird to have a component which is only accessible from within another component.

What's the advantage to doing it this way?

src/components/FullscreenGrid/index.tsx
willdurand: It's not a class-based component but I added the `Base` suffix as this is now a connected component. This is needed to retrieve the `fullGrid` reducer data.

src/components/FullscreenGrid/index.tsx
willdurand: mmh, this was mostly for backward compatibility. Let's move this component to `src/components` then.

src/components/FullscreenGrid/index.tsx
willdurand: Yes!

PR #732, opened 2019-05-10
Add fullscreenGrid reducer

### LINKED ISSUE #731: Add a reducer to change the `FullscreenGrid` UI
Supports #588

---

We want to hide the left panel of the browse/compare pages in #588. Let's have a reducer to store the state/perform this action. The reducer will be named `fullscreenGrid` since it controls the `FullscreenGrid` component (similar to the accordion menu logic).

willdurand:
Fixes #731 

---

Extracted from #721

### DIFF
--- a/src/reducers/fullscreenGrid.tsx
+++ b/src/reducers/fullscreenGrid.tsx
@@ -1,0 +1,31 @@
+import { Reducer } from 'redux';
+import { ActionType, createAction, getType } from 'typesafe-actions';
+
+export const actions = {
+  toggleMainSidePanel: createAction('TOGGLE_MAIN_SIDE_PANEL'),
+};
+
+export type FullscreenGridState = {
+  mainSidePanelIsExpanded: boolean;
+};
+
+export const initialState: FullscreenGridState = {
+  mainSidePanelIsExpanded: true,
+};
+
+const reducer: Reducer<FullscreenGridState, ActionType<typeof actions>> = (
+  state = initialState,
+  action,
+): FullscreenGridState => {
+  switch (action.type) {
+    case getType(actions.toggleMainSidePanel):
+      return {
+        ...state,
+        mainSidePanelIsExpanded: !state.mainSidePanelIsExpanded,
+      };
+    default:
+      return state;
+  }
+};
+
+export default reducer;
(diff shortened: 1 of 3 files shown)


src/reducers/fullscreenGrid.tsx
```diff
@@ -0,0 +1,31 @@
+import { Reducer } from 'redux';
+import { ActionType, createAction, getType } from 'typesafe-actions';
+
+export const actions = {
+  toggleMainSidePanel: createAction('TOGGLE_MAIN_SIDE_PANEL'),
```
bobsilverberg: It seems like it might be worth being able to toggle either panel from the beginning. It wouldn't be much more code than you're adding right now and would mean we wouldn't need those changes later.

Maybe a `togglePanel` action which accepts an argument for which panel to toggle?

src/reducers/fullscreenGrid.tsx
```diff
@@ -0,0 +1,31 @@
...
+};
+
+export type FullscreenGridState = {
+  mainSidePanelIsExpanded: boolean;
+};
+
+export const initialState: FullscreenGridState = {
+  mainSidePanelIsExpanded: true,
```
bobsilverberg: I don't think you have explicit coverage for this. The first test does rely on the default value of this being `true`, but that's an assumption of the test. Maybe it would be good to add a third test to check that the initial state is true? Or maybe that would amount to testing configuration? I'm not sure, tbh.

willdurand: ah correct, I'll add a test case to ensure we don't change the default value since we want to have the panels expanded by default

willdurand:
> This looks good @willdurand, thanks for this. I feel like now might be a good time to implement this for both panels, rather than just the main side panel, as it seems like it would be little to add at this point, and more effort to add later (e.g., it would require another review). What do you think?


That works for me. I'll add the missing code for that :) Thanks!

src/reducers/fullscreenGrid.tsx
willdurand: I added a `toggleAltSidePanel` action so that we don't have to deal with (invalid) panel IDs and `undefined` values.

PR #734, opened 2019-05-10
Add ToggleButton

### LINKED ISSUE #733: Create a `ToggleButton` for the `FullscreenGrid`
Supports #588

---

In order to implement #588, we need a `ToggleButton` for the `FullscreenGrid`.

willdurand:
Fixes #733

---

This PR adds a `ToggleButton` and a story for it. This component lives in `FullscreenGrid` because it is highly coupled to it. We could still move it elsewhere later if needed. The button handles left and right directions so that we could toggle the right panel in the (near) future.

The look is basic and could be improved but it works and it is too bad.

Usage might be seen in https://github.com/mozilla/addons-code-manager/pull/735.

## Screenshot

Screen Shot 2019-05-10 at 14 59 29


src/components/FullscreenGrid/ToggleButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+    <Button
+      className={makeClassName(
+        styles.ToggleButton,
+        {
+          [styles.toggleLeft]: toggleLeft,
+        },
+        className,
+      )}
+      onClick={onClick}
+      title={gettext('Toggle this panel')}
```
bobsilverberg: This seems like it could be obscure to a user. Perhaps `title` should default to this value, but be able to be overridden via a prop?

src/components/FullscreenGrid/ToggleButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+    <span className={styles.label}>{label}</span>
+  ) : null;
+  const icon = toggleLeft ? 'angle-double-left' : 'angle-double-right';
+
+  return (
+    <Button
+      className={makeClassName(
+        styles.ToggleButton,
+        {
+          [styles.toggleLeft]: toggleLeft,
```
bobsilverberg: This doesn't seem to be covered by a test.

src/components/FullscreenGrid/ToggleButton.tsx
willdurand: I'll defer that to #735.

willdurand:
> This looks good, nice work @willdurand. I have a similar question to the one I asked about #729: what do we gain by having this component living alongside the `FullscreenGrid` component, as opposed to on its own? As you mention in your description, we may want to use it elsewhere, so why not just put it elsewhere right now?

I moved this component to `src/components/ToggleButton`.

PR #735, opened 2019-05-10
Toggle main side panel

### LINKED ISSUE #736: Toggle the main side panel
Supports #588

---

In order to ship #588, which is about hiding the main side panel using a keyboard shortcut, we need to be able to toggle the main side panel (i.e. opening/closing it). Let's do this here.

**For QA:** a new button will be added to the left panel (a.k.a. main side panel) and it will also users to open/close this panel.

willdurand:
Fixes #736
~~:warning: Depends on #729, #732 and #734~~

---

This patch allows to expand/collapse the main side panel (left panel). We have everything to implement the same feature for the right panel if we want to.

I filed an issue for the style of the buttons already, because I am pretty sure we can do better: https://github.com/mozilla/addons-code-manager/issues/737.

## Gifs

App:

Storybook 1:

ezgif com-resize

Storybook 2:

## Screenshots

Regular screens:

Screen Shot 2019-05-10 at 15 07 15
Screen Shot 2019-05-10 at 15 07 13
Screen Shot 2019-05-10 at 15 07 08
Screen Shot 2019-05-10 at 15 07 05

Smaller screens:

Screen Shot 2019-05-10 at 15 07 51
Screen Shot 2019-05-10 at 15 07 49
Screen Shot 2019-05-10 at 15 07 41
Screen Shot 2019-05-10 at 15 07 38

### DIFF
--- a/src/components/FullscreenGrid/index.spec.tsx
+++ b/src/components/FullscreenGrid/index.spec.tsx
@@ -51,7 +51,7 @@ describe(__filename, () => {
       const root = render(<FullscreenGrid />);
 
       expect(root).toHaveClassName(`${styles.FullscreenGrid}`);
-      expect(root).not.toHaveClassName(`${styles.mainSidePanelIsCollapsed}`);
+      expect(root).not.toHaveClassName(`${styles.hasACollapsedMainSidePanel}`);
     });
 
     it('sets an extra CSS class when the main side panel is collapsed', () => {
@@ -60,7 +60,7 @@ describe(__filename, () => {
 
       const root = render(<FullscreenGrid />, { store });
 
-      expect(root).toHaveClassName(`${styles.mainSidePanelIsCollapsed}`);
+      expect(root).toHaveClassName(`${styles.hasACollapsedMainSidePanel}`);
     });
   });
 
--- a/src/components/FullscreenGrid/index.spec.tsx
+++ b/src/components/FullscreenGrid/index.spec.tsx
@@ -2,14 +2,27 @@ import * as React from 'react';
 import { shallow } from 'enzyme';
 
 import ContentShell from './ContentShell';
+import configureStore from '../../configureStore';
+import { actions as fullscreenGridActions } from '../../reducers/fullscreenGrid';
+import { shallowUntilTarget } from '../../test-helpers';
+import styles from './styles.module.scss';
 
-import FullscreenGrid, { Header } from '.';
+import FullscreenGrid, { Header, FullscreenGridBase } from '.';
 
 describe(__filename, () => {
   describe('FullscreenGrid', () => {
+    const render = (
+      element: JSX.Element,
+      { store = configureStore() } = {},
+    ) => {
+      return shallowUntilTarget(element, FullscreenGridBase, {
+        shallowOptions: { context: { store } },
+      });
+    };
+
     it('accepts a custom className', () => {
       const className = 'MyGrid';
-      const root = shallow(
+      const root = render(
         <FullscreenGrid className={className}>
           <Header />
           <ContentShell />
@@ -22,7 +35,7 @@ describe(__filename, () => {
     it('renders children', () => {
       const childClass = 'ChildExample';
 
-      const root = shallow(
+      const root = render(
         <FullscreenGrid>
           <Header />
           <ContentShell>
@@ -32,6 +45,22 @@ describe(__filename, () => {
       );
 
       expect(root.find(`.${childClass}`)).toHaveLength(1);
+    });
+
+    it('renders with a default CSS class', () => {
+      const root = render(<FullscreenGrid />);
+
+      expect(root).toHaveClassName(`${styles.FullscreenGrid}`);
+      expect(root).not.toHaveClassName(`${styles.mainSidePanelIsCollapsed}`);
+    });
+
+    it('sets an extra CSS class when the main side panel is collapsed', () => {
+      const store = configureStore();
+      store.dispatch(fullscreenGridActions.toggleMainSidePanel());
+
+      const root = render(<FullscreenGrid />, { store });
+
+      expect(root).toHaveClassName(`${styles.mainSidePanelIsCollapsed}`);
     });
   });
 
--- a/src/components/FullscreenGrid/index.spec.tsx
+++ b/src/components/FullscreenGrid/index.spec.tsx
@@ -50,8 +50,8 @@ describe(__filename, () => {
     it('renders with a default CSS class', () => {
       const root = render(<FullscreenGrid />);
 
-      expect(root).toHaveClassName(`${styles.FullscreenGrid}`);
-      expect(root).not.toHaveClassName(`${styles.hasACollapsedMainSidePanel}`);
+      expect(root).toHaveClassName(styles.FullscreenGrid);
+      expect(root).not.toHaveClassName(styles.hasACollapsedMainSidePanel);
     });
 
     it('sets an extra CSS class when the main side panel is collapsed', () => {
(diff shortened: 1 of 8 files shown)


willdurand:
@bobsilverberg this patch has been rebased. I only pushed the changes to toggle the left side panel. I'll open a new PR for the right side panel, which requires a bit more thinking as we'll have to handle different states (depending on the left panel too).

src/components/FullscreenGrid/styles.module.scss
```diff
@@ -1,5 +1,8 @@
 @import '../../scss/variables';
 
+$collapse-button-height: 40px;
```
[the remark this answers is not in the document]
kumar303: The idea with the patch is to always render the collapsed state (the tall button panel) within ContentShell but allow the child content to fully control the layout of the expanded state.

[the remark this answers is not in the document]
kumar303: Oh, hey, you can actually match the button styles just by letting the flex layout calculate everything. Here is a patch with the collapse button matching the accordion buttons exactly:

```diff
diff --git a/src/components/FullscreenGrid/styles.module.scss b/src/components/FullscreenGrid/styles.module.scss
index 7caaca4..3eb4149 100644
--- a/src/components/FullscreenGrid/styles.module.scss
+++ b/src/components/FullscreenGrid/styles.module.scss
@@ -32,7 +32,9 @@ $expand-button-width: 40px;
   margin-left: $default-padding;
 
   .mainSidePanelToggleButton {
-    height: $collapse-button-height;
+    height: auto;
+    margin-top: $default-padding;
+    padding: $default-padding;
     width: 100%;
   }
 
@@ -66,8 +68,13 @@ $expand-button-width: 40px;
   padding: $default-padding;
 }
 
+.mainSidePanel {
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+}
+
 .mainSidePanelContent {
-  height: calc(100% - #{$collapse-button-height + $default-padding});
   overflow-x: auto;
 
   .mainSidePanelIsCollapsed & {
```

(Apparently this problem haunted me over the weekend, ha.)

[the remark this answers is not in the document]
willdurand: I filed https://github.com/mozilla/addons-code-manager/issues/737 because I could not figure out how to get the same style for all buttons. Thanks for the patch!

To answer the very first question, I'd like to keep the in the side panel and not in the accordion so that it is a fullscreen grid feature and it works out of the box, no matter what we put in the side panels.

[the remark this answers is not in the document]
kumar303: > To answer the very first question, I'd like to keep the in the side panel and not in the accordion so that it is a fullscreen grid feature and it works out of the box, no matter what we put in the side panels.

Yes, I agree that this is nice. The second patch I gave you allows the buttons to look the same and also keeps the feature encapsulated in `ContentShell`. Did that work for the other screens, too?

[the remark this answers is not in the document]
willdurand: > Did that work for the other screens, too?

Looks like it works super well!

kumar303:
I keep getting distracted by meetings but it seems that this patch breaks the accordion menu on Chrome only. 

See http://localhost:9001/?path=/story/accordionmenu--long-content and http://localhost:3000/en-US/browse/494431/versions/1532144/ . The page is taller than the viewport. 

It would be nice to figure out why and solve it if we can in case reviewers want to use Chrome. I haven't found anything obvious yet. I'll take a closer look after lunch.

willdurand:
> I keep getting distracted by meetings but it seems that this patch breaks the accordion menu on Chrome only.

Looks like changing the `height` value to `auto` fixed this, right?

src/components/FullscreenGrid/styles.module.scss
```diff
@@ -10,15 +12,45 @@
...
 .mainSidePanel {
+  display: flex;
+  flex-direction: column;
   grid-area: FGMainSidePanel;
+  height: auto;
+  justify-content: space-between;
   margin-left: $default-padding;
+
+  .mainSidePanelToggleButton {
+    height: auto;
```
kumar303: This fixed things in Chrome but there are still a number of quirks that I don't understand. There is at least one bug in Firefox release (not Nightly) that causes unselected accordion items to expand sometimes.

Here's just one example in Nightly (awkward spacing between accordion buttons):

Screenshot 2019-05-22 09 20 43


I tried switching it to a grid and all issues were solved! I see it working great in Firefox (Release, Nightly) and Chrome. I say we use a grid:

```diff
diff --git a/src/components/FullscreenGrid/styles.module.scss b/src/components/FullscreenGrid/styles.module.scss
index 89dd8b2..6156a12 100644
--- a/src/components/FullscreenGrid/styles.module.scss
+++ b/src/components/FullscreenGrid/styles.module.scss
@@ -23,16 +23,16 @@ $expand-button-width: 40px;
 }
 
 .mainSidePanel {
-  display: flex;
+  display: grid;
   flex-direction: column;
   grid-area: FGMainSidePanel;
+  grid-gap: $default-padding;
+  grid-template-rows: 1fr auto;
   height: auto;
-  justify-content: space-between;
   margin-left: $default-padding;
 
   .mainSidePanelToggleButton {
     height: auto;
-    margin-top: $default-padding;
     padding: $default-padding;
     width: 100%;
 
@@ -43,7 +43,7 @@ $expand-button-width: 40px;
 
   &.mainSidePanelIsCollapsed {
     background-color: $light-gray;
-    margin-top: 0;
+    grid-gap: 0;
     padding: 0;
 
     .mainSidePanelToggleButton {

```

src/components/FullscreenGrid/index.spec.tsx
```diff
@@ -33,6 +46,22 @@ describe(__filename, () => {
 
       expect(root.find(`.${childClass}`)).toHaveLength(1);
     });
+
+    it('renders with a default CSS class', () => {
+      const root = render(<FullscreenGrid />);
+
+      expect(root).toHaveClassName(`${styles.FullscreenGrid}`);
```
kumar303: This doesn't need a template string

src/components/FullscreenGrid/index.spec.tsx
```diff
@@ -33,6 +46,22 @@ describe(__filename, () => {
...
+      expect(root).not.toHaveClassName(`${styles.hasACollapsedMainSidePanel}`);
```
kumar303: Same here

src/components/FullscreenGrid/index.spec.tsx
```diff
@@ -33,6 +46,22 @@ describe(__filename, () => {
...
+    });
+
+    it('sets an extra CSS class when the main side panel is collapsed', () => {
+      const store = configureStore();
+      store.dispatch(fullscreenGridActions.toggleMainSidePanel());
+
+      const root = render(<FullscreenGrid />, { store });
+
+      expect(root).toHaveClassName(`${styles.hasACollapsedMainSidePanel}`);
```
kumar303: Same

src/components/FullscreenGrid/styles.module.scss
willdurand: oh hum, I tried with a lot of content in the accordion tabs because I had issues with that, but I did not retry with short contents :(

willdurand: Thanks for the css patch1

willdurand:
@kumar303 I applied your patch and fixed the other things. It works well in FF Nightly and Chrome when playing with code-manager. It mostly works for Storybook stories too, but it looks like one story does not have a full width button for some reason:

Screen Shot 2019-05-22 at 19 10 42

I am not sure why and it only appears in Storybook.

PR #739, opened 2019-05-10
Change the versions reducer to work with a single diff per file rather than an array

### LINKED ISSUE #738: Change the versions reducer to work with a single diff per file rather than an array
This supports #726 and will allow us to land a change immediately, rather than having to wait for the API to change. This involves changing createInternalDiff to just take the first diff in the array and then changing all of the other reducer code to work with a single diff.

For QA:

Please check for regressions in the Compare view.

bobsilverberg:
Fixes #738

### DIFF
--- a/src/components/DiffView/index.spec.tsx
+++ b/src/components/DiffView/index.spec.tsx
@@ -161,12 +161,10 @@ describe(__filename, () => {
 
     // Simulate the interface of <Diff />
     const children = root.find(Diff).prop('children');
-    const diffComponent = shallow(<div>{children(diff.hunks)}</div>);
+    const diffWrapper = shallow(<div>{children(diff.hunks)}</div>);
 
-    expect(diffComponent.find(`.${styles.hunk}`)).toHaveLength(
-      diff.hunks.length,
-    );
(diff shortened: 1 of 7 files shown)


src/components/DiffView/index.spec.tsx
```diff
@@ -168,35 +156,34 @@ describe(__filename, () => {
...
   it('renders hunks with separators', () => {
-    const diffs = parseDiff(diffWithDeletions);
-    const root = renderWithLinterProvider({ diffs });
+    const diff = parseDiff(diffWithDeletions)[0];
+    const root = renderWithLinterProvider({ diff });
 
     // Simulate the interface of <Diff />
     const children = root.find(Diff).prop('children');
-    const diff = shallow(<div>{children(diffs[0].hunks)}</div>);
+    const diffComponent = shallow(<div>{children(diff.hunks)}</div>);
```
willdurand: `diffInstance`? or `wrapper?` or `diffWrapper`?

`Component` makes me think that it's something to instantiate.

bobsilverberg:
> r+wc, thanks!
> 
> There is a server patch ready to be merged if you want to synchronize the server patch and this one, up to you.

I think I'll land this, and then coordinate the server patch with #726, which will now be a much smaller patch.

PR #745, opened 2019-05-13
Update code to deal with API returning a single diff object

### LINKED ISSUE #726: Change API and reducer code to support no array of diffs

bobsilverberg:
Fixes #726 

Note that this should land at the same time as https://github.com/mozilla/addons-server/pull/11402.

### DIFF
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -1220,22 +1220,20 @@ describe(__filename, () => {
         ...fakeVersion,
         file: {
           ...fakeVersion.file,
-          diff: [
-            {
-              ...fakeExternalDiff,
-              hunks: [
-                {
-                  ...fakeExternalDiff.hunks[0],
-                  changes: [
-                    {
-                      ...fakeExternalDiff.hunks[0].changes[0],
-                      ...change,
-                    },
-                  ],
-                },
-              ],
-            },
-          ],
+          diff: {
+            ...fakeExternalDiff,
+            hunks: [
+              {
+                ...fakeExternalDiff.hunks[0],
+                changes: [
+                  {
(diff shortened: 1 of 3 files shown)


src/reducers/versions.spec.tsx
```diff
@@ -1280,14 +1276,14 @@ describe(__filename, () => {
       expect(diff).toHaveProperty('hunks');
     });
 
-    it('returns null if the array of diffs is empty', () => {
+    it('returns null if diff is falsey', () => {
```
muffinresearch: Can the diff be undefined at any point?

EnTeQuAk: It can't, it's either `null` or an empty object in the worst case.

bobsilverberg: @EnTeQuAk Hmm, I'm assuming it's either `null` or ` valid `diff` object. When could it be an empty object? I wonder if I need to deal with that in our code.

bobsilverberg: If it can be an empty object I definitely need to deal with that. So could you confirm if that is in fact possible?

EnTeQuAk: Double-checked, no, you're right, only a valid diff or `null` are valid responses. An empty object won't be the case.

bobsilverberg: Great, thanks Chris!

PR #749, opened 2019-05-14
Provide keyboard navigation for next / previous diff

### LINKED ISSUE #615: Provide keyboard navigation for next / previous diff
User story

> As a reviewer, I want to quickly navigate between the next / previous diff so I can focus on just the changes introduced by a new version.

Keyboard shortcuts:
* `n` for next diff
* `p` for previous diff

See https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/git-add-delete-example and https://reviewers.addons-dev.allizom.org/en-US/firefox/files/compare/376192...376191/ to view this in the old site.

Screenshot 2019-04-18 10 31 03

bobsilverberg:
Fixes #615 

~~Depends on #722~~ 

There are a couple of tests that are not really how I'd like them, but I figured the best way to get feedback was to submit the code for review and see what suggestions come back.

### DIFF
--- a/src/components/KeyboardShortcuts/index.spec.tsx
+++ b/src/components/KeyboardShortcuts/index.spec.tsx
@@ -1,3 +1,5 @@
+import { History } from 'history';
+import queryString from 'query-string';
 import * as React from 'react';
 import { Store } from 'redux';
 
@@ -9,13 +11,19 @@ import {
 import {
   actions as versionsActions,
   createInternalVersion,
+  getCompareInfo,
+  CompareInfo,
 } from '../../reducers/versions';
 import {
   CreateKeydownEventParams,
+  createContextWithFakeRouter,
   createKeydownEvent,
+  createFakeHistory,
+  createFakeLocation,
   createFakeThunk,
   fakeVersion,
   fakeVersionEntry,
+  fakeVersionWithDiff,
   shallowUntilTarget,
   spyOn,
 } from '../../test-helpers';
@@ -28,7 +36,45 @@ import KeyboardShortcuts, {
 } from '.';
 
 describe(__filename, () => {
+  type GetRouteParamsParams = {
+    addonId?: number;
+    baseVersionId?: number;
+    headVersionId?: number;
+  };
+
+  const getRouteParams = ({
+    addonId = 9999,
+    baseVersionId = 1,
+    headVersionId = 1000,
+  }: GetRouteParamsParams = {}) => {
+    return {
+      addonId: String(addonId),
+      baseVersionId: String(baseVersionId),
+      headVersionId: String(headVersionId),
+    };
+  };
+
+  const createFakeRouteComponentProps = ({
+    history = createFakeHistory(),
+    params = getRouteParams(),
+  } = {}) => {
+    return {
+      history,
+      location: history.location,
+      match: {
+        params,
+        isExact: true,
+        path: '/some-path',
+        url: '/some-url',
+      },
+    };
+  };
+
   type RenderParams = {
+    addonId?: string;
+    baseVersionId?: string;
+    headVersionId?: string;
+    history?: History;
     store?: Store;
   } & Partial<PublicProps & DefaultProps>;
 
@@ -63,27 +109,42 @@ describe(__filename, () => {
   };
 
   const render = ({
+    addonId = '999',
+    baseVersionId = '1',
     currentPath = 'file1.js',
-    versionId = 1235,
+    headVersionId = '2',
+    history = createFakeHistory(),
     store,
+    versionId = 1235,
     ...moreProps
   }: RenderParams = {}) => {
     const props = {
-      _goToRelativeFile: jest.fn(),
+      ...createFakeRouteComponentProps({
+        history,
+        params: { addonId, baseVersionId, headVersionId },
+      }),
       currentPath,
       versionId,
       ...moreProps,
+    };
+
+    const contextWithRouter = createContextWithFakeRouter({
+      history,
+      match: props.match,
+    });
+    const context = {
+      ...contextWithRouter,
+      context: {
+        ...contextWithRouter.context,
+        store: store || configureStoreWithFileTree({ versionId }),
+      },
     };
 
     return shallowUntilTarget(
       <KeyboardShortcuts {...props} />,
       KeyboardShortcutsBase,
       {
-        shallowOptions: {
-          context: {
-            store: store || configureStoreWithFileTree({ versionId }),
-          },
-        },
+        shallowOptions: { ...context },
       },
     );
   };
@@ -164,6 +225,205 @@ describe(__filename, () => {
         position,
         versionId,
       });
+    },
+  );
+
+  it.each([
+    ['previous', 'p', RelativePathPosition.previous],
+    ['next', 'n', RelativePathPosition.next],
+  ])(
+    'dispatches goToRelativeDiff with %s when "%s" is pressed',
+    (direction, key, position) => {
+      const addonId = 1;
+      const baseVersionId = 2;
+      const currentPath = fakeVersionWithDiff.file.selected_file;
+      const hash = '#D1';
+      const headVersionId = 3;
+      const pathList = [currentPath];
+      const versionId = headVersionId;
+      const version = {
+        ...fakeVersionWithDiff,
+        id: headVersionId,
+        file: {
+          ...fakeVersionWithDiff.file,
+          // eslint-disable-next-line @typescript-eslint/camelcase
+          selected_file: currentPath,
+        },
+      };
+      const history = createFakeHistory({
+        location: createFakeLocation({
+          search: queryString.stringify({ path: currentPath }),
+          hash,
+        }),
+      });
+
+      const store = configureStoreWithFileTree({ versionId, pathList });
+      store.dispatch(versionsActions.loadVersionInfo({ version }));
+      store.dispatch(
+        versionsActions.loadDiff({
+          addonId,
+          baseVersionId,
+          headVersionId,
+          path: currentPath,
+          version,
+        }),
+      );
+
+      const compareInfo = getCompareInfo(
+        store.getState().versions,
+        addonId,
+        baseVersionId,
+        headVersionId,
+        currentPath,
+      ) as CompareInfo;
+
+      const dispatch = spyOn(store, 'dispatch');
+      const fakeThunk = createFakeThunk();
+      const _goToRelativeDiff = fakeThunk.createThunk;
+
+      renderAndTriggerKeyEvent(
+        { key: key as string },
+        {
+          _goToRelativeDiff,
+          ...getRouteParams({
+            addonId,
+            baseVersionId,
+            headVersionId,
+          }),
+          currentPath,
+          history,
+          store,
+          versionId,
+        },
+      );
+
+      expect(dispatch).toHaveBeenCalledWith(fakeThunk.thunk);
+      expect(_goToRelativeDiff).toHaveBeenCalledWith({
+        currentAnchor: hash.replace(/^#/, ''),
+        diff: compareInfo.diff,
+        pathList,
+        position,
+        versionId,
+      });
+    },
+  );
+
+  it.each(['p', 'n'])(
+    'does not dispatch goToRelativeDiff when "%s" is pressed if expected match params are not present',
+    (key) => {
+      const addonId = 1;
+      const baseVersionId = 2;
+      const currentPath = fakeVersionWithDiff.file.selected_file;
+      const headVersionId = 3;
+      const pathList = [currentPath];
+      const versionId = headVersionId;
+      const version = {
+        ...fakeVersionWithDiff,
+        id: headVersionId,
+        file: {
+          ...fakeVersionWithDiff.file,
+          // eslint-disable-next-line @typescript-eslint/camelcase
+          selected_file: currentPath,
+        },
+      };
+      const history = createFakeHistory({
+        location: createFakeLocation({
+          search: queryString.stringify({ path: currentPath }),
+        }),
+      });
+
+      const store = configureStoreWithFileTree({ versionId, pathList });
+      store.dispatch(versionsActions.loadVersionInfo({ version }));
+      store.dispatch(
+        versionsActions.loadDiff({
+          addonId,
+          baseVersionId,
+          headVersionId,
+          path: currentPath,
+          version,
+        }),
+      );
+
+      const dispatch = spyOn(store, 'dispatch');
+      const fakeThunk = createFakeThunk();
+      const _goToRelativeDiff = fakeThunk.createThunk;
+
+      renderAndTriggerKeyEvent(
+        { key: key as string },
+        {
+          _goToRelativeDiff,
+          addonId: undefined,
+          baseVersionId: undefined,
+          currentPath,
+          headVersionId: undefined,
+          history,
+          store,
+          versionId,
+        },
+      );
+
+      expect(dispatch).not.toHaveBeenCalledWith(fakeThunk.thunk);
+    },
+  );
+
+  it.each(['p', 'n'])(
+    'does not dispatch goToRelativeDiff when "%s" is pressed if the diff is not loaded',
+    (key) => {
+      const addonId = 1;
+      const baseVersionId = 2;
+      const currentPath = fakeVersionWithDiff.file.selected_file;
+      const headVersionId = 3;
+      const pathList = [currentPath];
+      const versionId = headVersionId;
+      const version = {
+        ...fakeVersionWithDiff,
+        id: headVersionId,
+        file: {
+          ...fakeVersionWithDiff.file,
+          // eslint-disable-next-line @typescript-eslint/camelcase
+          selected_file: currentPath,
+        },
+      };
+      const history = createFakeHistory({
+        location: createFakeLocation({
+          search: queryString.stringify({ path: currentPath }),
+        }),
+      });
+
+      const store = configureStoreWithFileTree({ versionId, pathList });
+      store.dispatch(versionsActions.loadVersionInfo({ version }));
+      store.dispatch(
+        versionsActions.loadDiff({
+          // Load a diff that does not match the match params.
+          addonId: addonId + 1,
+          baseVersionId,
+          headVersionId,
+          path: currentPath,
+          version,
+        }),
+      );
+
+      const dispatch = spyOn(store, 'dispatch');
+      const fakeThunk = createFakeThunk();
+      const _goToRelativeDiff = fakeThunk.createThunk;
+
+      renderAndTriggerKeyEvent(
+        { key: key as string },
+        {
+          _goToRelativeDiff,
+          ...getRouteParams({
+            addonId,
+            baseVersionId,
+            headVersionId,
+          }),
+          currentPath,
+          history,
+          store,
+          versionId,
+        },
+      );
+
+      expect(dispatch).not.toHaveBeenCalledWith(fakeThunk.thunk);
     },
   );
 
--- a/src/components/KeyboardShortcuts/index.spec.tsx
+++ b/src/components/KeyboardShortcuts/index.spec.tsx
@@ -1,4 +1,4 @@
-import { History } from 'history';
+import { Location } from 'history';
 import queryString from 'query-string';
 import * as React from 'react';
 import { Store } from 'redux';
@@ -11,16 +11,15 @@ import {
 import {
   actions as versionsActions,
   createInternalVersion,
-  getCompareInfo,
-  CompareInfo,
 } from '../../reducers/versions';
 import {
   CreateKeydownEventParams,
   createContextWithFakeRouter,
   createKeydownEvent,
-  createFakeHistory,
+  createFakeCompareInfo,
   createFakeLocation,
   createFakeThunk,
+  createStoreWithVersion,
   fakeVersion,
   fakeVersionEntry,
   fakeVersionWithDiff,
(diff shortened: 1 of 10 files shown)


bobsilverberg:
I have rebased this after the landing of #723, so it is now ready for review.

src/components/KeyboardShortcuts/index.tsx
```diff
@@ -126,6 +175,14 @@ export class KeyboardShortcutsBase extends React.Component<Props> {
             <kbd>j</kbd>
           </dt>
           <dd>{gettext('Down file')}</dd>
+          <dt>
+            <kbd>p</kbd>
+          </dt>
+          <dd>{gettext('Previous change')}</dd>
```
kumar303: I suggest rendering these in a disabled state when `compareInfo` is `null`. This would help communicate that they aren't applicable to the browse page.

src/components/KeyboardShortcuts/index.tsx
```diff
@@ -142,9 +199,26 @@ export class KeyboardShortcutsBase extends React.Component<Props> {
...
+
+  const addonId = parseInt(match.params.addonId, 10);
+  const baseVersionId = parseInt(match.params.baseVersionId, 10);
+  const headVersionId = parseInt(match.params.headVersionId, 10);
+  const path = getPathFromQueryString(history) || undefined;
+
+  let compareInfo = null;
+
+  if (addonId && baseVersionId && headVersionId) {
+    compareInfo = getCompareInfo(
```
kumar303: Mapping this object here has a few problems:

* It introduces a dependency on `compareInfo` without any way to load it.
* It copies/pastes all the URL matching logic [from `Compare`](https://github.com/mozilla/addons-code-manager/blob/5ab911bda04346c6d5e9583f467992f9051b86ce/src/pages/Compare/index.tsx#L171-L174) which is fragile, especially since it would be unclear that changing those route params will break this component

I know that Redux is often nicer than prop-drilling but I think we should prop-drill for this case unless there is a necessity to use Redux.

The prop-drilling would only need to go two levels deep. You could render `<VersionFileViewer compareInfo={compareInfo} ... />` from [`Compare`](https://github.com/mozilla/addons-code-manager/blob/5ab911bda04346c6d5e9583f467992f9051b86ce/src/pages/Compare/index.tsx#L140) and render `<KeyboardShortcuts compareInfo={compareInfo} ... />` from [`VersionFileViewer`](https://github.com/mozilla/addons-code-manager/blob/5ab911bda04346c6d5e9583f467992f9051b86ce/src/components/VersionFileViewer/index.tsx#L61).

Also, this would simplify your test setup in some cases.

src/components/KeyboardShortcuts/index.tsx
```diff
@@ -97,6 +120,32 @@ export class KeyboardShortcutsBase extends React.Component<Props> {
...
+            dispatch(
+              _goToRelativeDiff({
+                currentAnchor,
+                diff: compareInfo.diff,
+                pathList,
+                position: RelativePathPosition.next,
+                versionId,
+              }),
+            );
+          }
```
kumar303: I'd like to at least see a warning in the console for the `else` branch here. Ideally, it should pop up a message on the screen but we don't have a framework for that. I think it would be confusing when pressing `p` does nothing on the browse page.

src/components/KeyboardShortcuts/index.spec.tsx
```diff
@@ -167,6 +228,205 @@ describe(__filename, () => {
...
+      };
+      const history = createFakeHistory({
+        location: createFakeLocation({
+          search: queryString.stringify({ path: currentPath }),
+          hash,
+        }),
+      });
+
+      const store = configureStoreWithFileTree({ versionId, pathList });
+      store.dispatch(versionsActions.loadVersionInfo({ version }));
```
kumar303: This seems like something you'd want to have happen automatically. There is a new helper you could use:

```diff
diff --git a/src/components/KeyboardShortcuts/index.spec.tsx b/src/components/KeyboardShortcuts/index.spec.tsx
index 8e1dcb7..d99223f 100644
--- a/src/components/KeyboardShortcuts/index.spec.tsx
+++ b/src/components/KeyboardShortcuts/index.spec.tsx
@@ -21,6 +21,7 @@ import {
   createFakeHistory,
   createFakeLocation,
   createFakeThunk,
+  createStoreWithVersion,
   fakeVersion,
   fakeVersionEntry,
   fakeVersionWithDiff,
@@ -80,28 +81,34 @@ describe(__filename, () => {
 
   const configureStoreWithFileTree = ({
     pathList = ['file1.js'],
-    store = configureStore(),
     versionId = 321,
+    externalVersion = {
+      ...fakeVersion,
+      id: versionId,
+      file: {
+        ...fakeVersion.file,
+        entries: pathList.reduce((pathMap, path: string) => {
+          return {
+            ...pathMap,
+            [path]: {
+              ...fakeVersionEntry,
+              filename: path,
+              path,
+            },
+          };
+        }, {}),
+      },
+    },
+    store = createStoreWithVersion(externalVersion),
+  }: {
+    pathList?: string[];
+    versionId?: number;
+    externalVersion?: typeof fakeVersion;
+    store?: Store;
   } = {}) => {
     store.dispatch(
       fileTreeActions.buildTree({
-        version: createInternalVersion({
-          ...fakeVersion,
-          id: versionId,
-          file: {
-            ...fakeVersion.file,
-            entries: pathList.reduce((pathMap, path: string) => {
-              return {
-                ...pathMap,
-                [path]: {
-                  ...fakeVersionEntry,
-                  filename: path,
-                  path,
-                },
-              };
-            }, {}),
-          },
-        }),
+        version: createInternalVersion(externalVersion),
       }),
     );
 
```

src/reducers/versions.spec.tsx
```diff
@@ -1832,6 +1835,32 @@ describe(__filename, () => {
...
     });
 
+    it('appends showDiff if requested', async () => {
+      const location = createFakeLocation({ pathname });
+      const history = createFakeHistory({ location });
+      const diffPosition = ScrollTarget.firstDiff;
+
+      const { dispatch, thunk } = thunkTester({
+        createThunk: () =>
+          viewVersionFile({
```
kumar303: Just a personal preference but anytime an arrow function goes multiline, I like to see it wrapped in curly braces. This makes it easier to avoid a syntax error when you go to edit the return value later on.

src/reducers/versions.spec.tsx
```diff
@@ -2470,4 +2524,206 @@ describe(__filename, () => {
...
+        pathList,
+        position,
+        version,
+      });
+
+      expect(dispatch).toHaveBeenCalledWith(
+        push(
+          // TODO: This is probably not good enough.
+          // We should check the location object as well, but I'm not sure how
+          // to set up a location in the router in the test.
```
kumar303: Did you try [`createFakeLocation()`](https://github.com/mozilla/addons-code-manager/blob/5ab911bda04346c6d5e9583f467992f9051b86ce/src/test-helpers.tsx#L396)?

src/reducers/versions.spec.tsx
```diff
@@ -2470,4 +2524,206 @@ describe(__filename, () => {
...
+        createThunk: () =>
+          _goToRelativeDiff({
+            versionId: versionId + 1,
+          }),
+        store,
+      });
+
+      // TODO: This doesn't work. Maybe we cannot test for an exception in a thunk?
+      expect(async () => {
+        await thunk();
```
kumar303: `await` is just syntactic sugar to work with promises. In this case you could either write your own `try/catch` block around `await thunk()` or deal with the promise directly, like this, which is a bit easier:

```diff
diff --git a/src/reducers/versions.spec.tsx b/src/reducers/versions.spec.tsx
index 76e5fef..0e9e923 100644
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -2692,9 +2692,9 @@ describe(__filename, () => {
       });
 
       // TODO: This doesn't work. Maybe we cannot test for an exception in a thunk?
-      expect(async () => {
-        await thunk();
-      }).toThrow('Cannot go to relative diff without a version loaded');
+      await expect(thunk()).rejects.toThrow(
+        'Cannot go to relative diff without a version loaded',
+      );
     });
 
     it('does not call _getRelativeDiff if the version is not loaded', async () => {

```

src/reducers/versions.spec.tsx
```diff
@@ -2470,4 +2524,206 @@ describe(__filename, () => {
...
+            _getRelativeDiff,
+            versionId: versionId + 1,
+          }),
+        store,
+      });
+
+      // TODO: Without this, the test stops because of the exception,
+      // but this doesn't seem like a good thing to do in a test.
+      try {
+        await thunk();
```
kumar303: I guess you could use `await expect(thunk()).rejects.toThrow()` again but is this test really necessary? A thrown error will stop execution so I don't think you need to worry about rogue dispatches.

src/reducers/versions.tsx
```diff
@@ -917,6 +923,64 @@ export const viewVersionFile = ({
...
+        hash: `#${nextDiffInfo.anchor}`,
+      };
+
+      dispatch(push(newLocation));
+    } else if (nextDiffInfo.path) {
+      dispatch(
+        _viewVersionFile({
+          preserveHash: false,
+          selectedPath: nextDiffInfo.path,
+          showDiff:
```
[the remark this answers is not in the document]
bobsilverberg: I was pretty sure I encountered this as well, and then fixed it. Maybe those changes got lost in the rebase? I'll look into this.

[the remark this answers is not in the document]
bobsilverberg: Yes, it was a problem with a change not being properly applied after a rebase. It's fixed in the next iteration.

src/components/KeyboardShortcuts/index.tsx
bobsilverberg: Good idea. I have created some styles for this. Let me know if they look good, or if you'd like to see them changed.

src/components/KeyboardShortcuts/index.tsx
bobsilverberg: This is a good idea, thanks. I'll implement this.

src/components/KeyboardShortcuts/index.tsx
bobsilverberg: Ok, I added this. Do you think we also need a test for this (which would involve needing to inject a fake logger), or is this the kind of thing we don't need to bother testing as it is just a warning?

src/components/KeyboardShortcuts/index.spec.tsx
bobsilverberg: Nice, thanks!

src/reducers/versions.spec.tsx
bobsilverberg: What I was missing was using `createBrowserHistory` to create a `History` object and then passing that into `configureStore`, but I figured that out. :-)

src/reducers/versions.spec.tsx
bobsilverberg: Ah, I was trying to find a `toReject` function similar to `toThrow`, but I didn't think to look for `rejects`.

src/reducers/versions.spec.tsx
bobsilverberg: Good point. I agree that this test isn't really needed.

bobsilverberg:
Thanks for the review @kumar303. Some really great suggestions! I've addressed everything and this is now ready for another look.

kumar303:
Coverage is complaining because you're only executing this function with mocks (not real objects) but I'm not worried about it. Static typing will catch any problems with calling the dependencies. I've suggested adding an integration test in the past but I'm looking at the type definitions here; I don't think it's important to add extra tests.

Screenshot 2019-05-22 15 51 25

bobsilverberg:
> Coverage is complaining because you're only executing this function with mocks (not real objects) but I'm not worried about it. Static typing will catch any problems with calling the dependencies. I've suggested adding an integration test in the past but I'm looking at the type definitions here; I don't think it's important to add extra tests.
> 
> Screenshot 2019-05-22 15 51 25

Yeah I noticed that and added some `instanbul ignore` comments to address the coverage report issue.

src/components/KeyboardShortcuts/index.tsx
kumar303: I don't see any value in adding a test with a fake logger and asserting it gets called correctly. TS already handles this. Actually, I think we could all stop adding these types of tests.

src/components/KeyboardShortcuts/index.tsx
```diff
@@ -97,6 +114,36 @@ export class KeyboardShortcutsBase extends React.Component<Props> {
...
+              _goToRelativeDiff({
+                currentAnchor,
+                diff: compareInfo.diff,
+                pathList,
+                position: RelativePathPosition.previous,
+                versionId,
+              }),
+            );
+          } else {
+            log.warn('Cannot navigate to next change without diff loaded');
```
kumar303:
```suggestion
            log.warn('Cannot navigate to previous change without diff loaded');
```

src/components/KeyboardShortcuts/index.tsx
```diff
@@ -115,6 +162,11 @@ export class KeyboardShortcutsBase extends React.Component<Props> {
   }
 
   render() {
+    const { compareInfo } = this.props;
+    const disabledClassname = makeClassName({
+      [styles.disabled]: !compareInfo,
```
kumar303: This could be:

```ts
const disabledClassname = !compareInfo ? styles.disabled : '';
```

It would be a little easier to read if it were specific to the item (not the state):

```ts
const diffKeysClassName = !compareInfo ? styles.disabled : '';
```

src/components/KeyboardShortcuts/index.tsx
```diff
@@ -126,6 +178,14 @@ export class KeyboardShortcutsBase extends React.Component<Props> {
             <kbd>j</kbd>
           </dt>
           <dd>{gettext('Down file')}</dd>
+          <dt className={disabledClassname}>
+            <kbd className={disabledClassname}>p</kbd>
```
kumar303: Since you're already adding a nested style for `kbd` then I say don't bother adding a class name here. Just let the `dt` class name do the styling.

src/components/KeyboardShortcuts/styles.module.scss
```diff
@@ -15,3 +15,11 @@
     margin-bottom: 0;
   }
 }
+
+.disabled {
```
kumar303: The styles look good 👍

src/reducers/versions.spec.tsx
```diff
@@ -1832,6 +1836,33 @@ describe(__filename, () => {
...
+        },
+        store: configureStore({ history }),
+      });
+
+      await thunk();
+
+      expect(dispatch).toHaveBeenCalledWith(
+        push({
+          ...location,
+          search: `?path=${selectedPath}&scrollTo=${diffPosition}`,
```
kumar303: Use [`expect.urlWithTheseParams()`](https://github.com/mozilla/addons-code-manager/blob/5ab911bda04346c6d5e9583f467992f9051b86ce/src/jest-matchers/index.tsx#L32) here if you can. Otherwise, you're relying on the order of object serialization which could be flaky one day.

src/test-helpers.tsx
```diff
@@ -393,6 +394,22 @@ export const createFakeLinterMessagesByPath = ({
   return map[path];
 };
 
+export const createFakeCompareInfo = ({
```
kumar303: It will be a bit more verbose but this should hook into `createInternalCompareInfo()` to mimic real world usage as much as possible (patch below). 

We need to be careful about manually constructing internal objects in tests since it could be fragile. It's safer to start with a known external object, i.e. one that would come from our API.

```diff
diff --git a/src/reducers/versions.tsx b/src/reducers/versions.tsx
index c339f6c..990b9f8 100644
--- a/src/reducers/versions.tsx
+++ b/src/reducers/versions.tsx
@@ -166,7 +166,7 @@ export type VersionFile = {
   version: string;
 };
 
-type VersionEntry = {
+export type VersionEntry = {
   depth: number;
   filename: string;
   mimeType: string;
diff --git a/src/test-helpers.tsx b/src/test-helpers.tsx
index edc737c..3009cb5 100644
--- a/src/test-helpers.tsx
+++ b/src/test-helpers.tsx
@@ -26,8 +26,10 @@ import {
   ExternalVersionsList,
   ExternalVersionsListItem,
   Version,
+  VersionEntry,
   VersionEntryType,
   actions as versionsActions,
+  createInternalCompareInfo,
   createInternalDiff,
   createInternalVersion,
   createInternalVersionEntry,
@@ -398,16 +400,24 @@ export const createFakeCompareInfo = ({
   baseVersionId = 1,
   headVersionId = 2,
   mimeType = 'mime/type',
+  entry = createInternalVersionEntry({
+    ...fakeVersionEntry,
+    mimetype: mimeType,
+  }),
   version = fakeVersionWithDiff,
+}: {
+  mimeType?: string;
+  baseVersionId?: number;
+  headVersionId?: number;
+  entry?: VersionEntry;
+  version?: ExternalVersionWithDiff;
 } = {}) => {
-  return {
-    diff: createInternalDiff({
-      baseVersionId,
-      headVersionId,
-      version,
-    }),
-    mimeType,
-  };
+  return createInternalCompareInfo({
+    baseVersionId,
+    entry,
+    headVersionId,
+    version,
+  });
 };
 
 export const createFakeLocation = (props = {}): Location => {
```

src/reducers/versions.spec.tsx
bobsilverberg: Good idea. I am having trouble figuring out the syntax for using it in this situation as we are not asserting about a property of an object that was passed into the call to `dispatch`, but rather about a property of an object that was passed into the call of `push`. I tried this:

```js
        push(
          expect.objectContaining({
            hash: undefined,
            search: expect.urlWithTheseParams({
              path: selectedPath,
              scrollTo: diffPosition,
            }),
          }),
        ),
      );
```

Which yields this failure:

```
  ● /Users/bsilverberg/Documents/gitRepos/addons-code-manager/src/reducers/versions.spec.tsx › viewVersionFile › appends scrollTo if requested

    expect(jest.fn()).toHaveBeenCalledWith(expected)

    Expected mock function to have been called with:
      {"payload": {"args": [ObjectContaining {"hash": undefined, "search": urlWithTheseParams<[object Object]>}], "method": "push"}, "type": "@@router/CALL_HISTORY_METHOD"}
    as argument 1, but it was called with
      {"payload": {"args": [{"key": "some-key", "pathname": "/some/path/to/a/page", "search": "?path=some-path&scrollTo=firstDiff", "state": {}}], "method": "push"}, "type": "@@router/CALL_HISTORY_METHOD"}.

    Difference:

    - Expected
    + Received

      Object {
        "payload": Object {
          "args": Array [
    -       ObjectContaining {
    -         "hash": undefined,
    -         "search": urlWithTheseParams<[object Object]>,
    +       Object {
    +         "key": "some-key",
    +         "pathname": "/some/path/to/a/page",
    +         "search": "?path=some-path&scrollTo=firstDiff",
    +         "state": Object {},
            },
          ],
          "method": "push",
        },
        "type": "@@router/CALL_HISTORY_METHOD",
      }

      {"payload": {"args": [ObjectContaining {"hash": undefined, "search": urlWithTheseParams<[object Object]>}], "method": "push"}, "type": "@@router/CALL_HISTORY_METHOD"}
    as argument 1, but it was called with
      {"error": undefined, "meta": undefined, "payload": {"selectedPath": "some-path", "versionId": 123}, "type": "UPDATE_SELECTED_PATH"}.

    Difference:

    - Expected
    + Received

      Object {
    +   "error": undefined,
    +   "meta": undefined,
        "payload": Object {
    -     "args": Array [
    -       ObjectContaining {
    -         "hash": undefined,
    -         "search": urlWithTheseParams<[object Object]>,
    -       },
    -     ],
    -     "method": "push",
    +     "selectedPath": "some-path",
    +     "versionId": 123,
        },
    -   "type": "@@router/CALL_HISTORY_METHOD",
    +   "type": "UPDATE_SELECTED_PATH",
      }

      1863 |       );
      1864 |
    > 1865 |       expect(dispatch).toHaveBeenCalledWith(
           |                        ^
      1866 |         push(
      1867 |           expect.objectContaining({
      1868 |             hash: undefined,

      at Object.toHaveBeenCalledWith (src/reducers/versions.spec.tsx:1865:24)
```

I'm not sure why it's showing me two match failures, so maybe that's a hint as to what the problem is? Is it because of the expect inside an expect inside an expect?

@kumar303 Do you have any advice?

bobsilverberg:
Thanks for the review @kumar303. I have addressed all of the outstanding issues except for the one raised at https://github.com/mozilla/addons-code-manager/pull/749#discussion_r286693552. If you can help me solve this that would be great, otherwise I can just land the test as is.

src/reducers/versions.spec.tsx
kumar303: You probably need to do it in two calls:

```tsx
        push(
          expect.objectContaining({
            hash: undefined,
          }),
        ),
      );
      expect(dispatch.calls[0].search).toEqual(
            expect.urlWithTheseParams({
              path: selectedPath,
              scrollTo: diffPosition,
            }),
      );
```

... but maybe it would work like this?

```tsx
          expect.objectContaining({
            hash: undefined,
            search: expect.urlWithTheseParams({
              path: selectedPath,
              scrollTo: diffPosition,
            }),
          }),
      );
```

bobsilverberg: It turns out that:

```js
        push(
          expect.objectContaining({
            pathname,
            search: expect.urlWithTheseParams({
              path: selectedPath,
              scrollTo: diffPosition,
            }),
          }),
        ),
      );
```

works, which is what I had originally, except that the difference is that I didn't include `hash: undefined` in the `expect.objectContaining`. That was the problem. Even though `hash` was not present in the actual object, including `hash: undefined` in the `expect.objectContaining` was failing because, I presume, it was looking for `hash: undefined` to be present in the object. Looking at the test, there's actually no reason for it to be checking anything about `hash` in the call anyway, so I removed that and now it's working.

PR #751, opened 2019-05-14
Enable scrolling to a particular linter message

### LINKED ISSUE #750: Return the line number from getRelativeMessage
This supports #520 

In order to be able to move from linter message to linter message, we need to know the line of code that corresponds to the message, so this issue is about returning `line`, in addition to `path` and `uid` from `getRelativeMessage`.

bobsilverberg:
Fixes #750

### DIFF
--- a/src/components/CodeView/index.spec.tsx
+++ b/src/components/CodeView/index.spec.tsx
@@ -262,10 +262,10 @@ describe(__filename, () => {
 
     expect(message).toHaveLength(1);
     expect(message).toHaveProp('message', selectedMessageMap.byLine[line][0]);
-    expect(message).toHaveProp('shouldScrollToMessage', false);
+    expect(message).toHaveProp('_scrollToElement', null);
   });
 
-  it('indicates that a byLine message should be scrolled into view', () => {
+  it('passes scrollToElement to a linter message if it should be scrolled into view', () => {
     const uid = 'some-byLine-message-id';
     const externalMessage = {
       ...fakeExternalLinterMessage,
@@ -283,7 +283,7 @@ describe(__filename, () => {
 
     const message = root.find(LinterMessage);
     expect(message).toHaveLength(1);
-    expect(message).toHaveProp('shouldScrollToMessage', true);
+    expect(message).toHaveProp('_scrollToElement', scrollToElement);
   });
 
   it('renders multiple LinterMessage components on one line', () => {
@@ -339,7 +339,7 @@ describe(__filename, () => {
     const message = root.find(LinterMessage);
     expect(message).toHaveLength(1);
     expect(message).toHaveProp('message', expect.objectContaining({ uid }));
-    expect(message).toHaveProp('shouldScrollToMessage', false);
+    expect(message).toHaveProp('_scrollToElement', null);
   });
 
   it('indicates that a global message should be scrolled into view', () => {
@@ -356,7 +356,7 @@ describe(__filename, () => {
 
     const message = root.find(LinterMessage);
     expect(message).toHaveLength(1);
-    expect(message).toHaveProp('shouldScrollToMessage', true);
+    expect(message).toHaveProp('_scrollToElement', scrollToElement);
   });
 
   it('renders all global LinterMessage components', () => {
--- a/src/components/CodeView/index.spec.tsx
+++ b/src/components/CodeView/index.spec.tsx
@@ -23,7 +23,7 @@ import {
 } from '../../test-helpers';
 import styles from './styles.module.scss';
 
-import CodeView, { CodeViewBase, PublicProps, scrollToSelectedLine } from '.';
+import CodeView, { CodeViewBase, PublicProps, scrollToElement } from '.';
 
 describe(__filename, () => {
   type RenderParams = Partial<PublicProps> & {
@@ -100,9 +100,11 @@ describe(__filename, () => {
   };
 
   const renderWithMessages = ({
+    location = createFakeLocation(),
     messages,
     ...props
   }: {
+    location?: Location<{}>;
     messages: Partial<ExternalLinterMessage>[];
     props?: RenderParams;
   }) => {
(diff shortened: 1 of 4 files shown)


src/components/CodeView/index.spec.tsx
```diff
@@ -100,9 +100,11 @@ describe(__filename, () => {
   };
 
   const renderWithMessages = ({
+    location = createFakeLocation(),
```
bobsilverberg: This seems unnecessary as `location` is in `RenderParams`, but TS was mad about my use of `location` in the new tests where I am calling `renderWithMessages`. I think it's a case of TS getting confused by spreads, which I think we've seen in the past. Without adding this here, I was getting TS errors in the test code when trying to pass `location` into `renderWithMessages`.

bobsilverberg:
Note also that I originally had a prop called `shouldScrollToMessage` in `LinterMessage` to indicate whether the `ref` should be set to `scrollToElement`, but then I realized that we could just pass either `scrollToElement` or `null` into `LinterMessage` from `CodeView` and therefore we don't actually need `shouldScrollToMessage`. It is a bit opaque though, and I suppose if we ever want to be able to do if from a different component then we'd need to do the same thing, so maybe I should put `shouldScrollToMessage` back into `LinterMessage`. You can see the original in the first commit and my removal of it in the second commit.

src/components/LinterMessage/index.tsx
```diff
@@ -59,22 +60,29 @@ const renderDescription = (description: LinterMessage['description']) => {
...
 
   return (
-    <Alert
-      className={makeClassName(styles[variant], { [styles.inline]: inline })}
-      variant={variant}
-    >
-      <Alert.Heading>{decodeHtmlEntities(linterMessage)}</Alert.Heading>
-      <p className={styles.description}>{renderDescription(description)}</p>
-    </Alert>
+    <div className="LinterMessage" ref={_scrollToElement}>
```
kumar303: btw, you can theoretically just assign the `ref` prop to `Alert` but I'm seeing what looks like a bug in the TS definitions (https://github.com/react-bootstrap/react-bootstrap/issues/3805).

src/components/CodeView/index.tsx
```diff
@@ -72,7 +72,17 @@ export class CodeViewBase extends React.Component<Props> {
       <React.Fragment>
         {selectedMessageMap &&
           selectedMessageMap.global.map((message) => {
-            return <LinterMessage key={message.uid} message={message} />;
+            return (
+              <LinterMessage
+                key={message.uid}
+                message={message}
+                _scrollToElement={
```
kumar303: Do you really need to patch `CodeView` for this? This won't work on the compare page since `CodeView` is not rendered there. 

Instead of passing a `_scrollToElement` prop, would it be possible to connect `LinterMessage` to Redux state and have it look in state to know if it should be scrolled into view?

src/components/LinterMessage/index.spec.tsx
```diff
@@ -141,7 +149,23 @@ describe(__filename, () => {
...
 
-    expect(root).toHaveClassName(`.${styles.inline}`);
+    expect(root.find(Alert)).toHaveClassName(`.${styles.inline}`);
+  });
+
+  it('calls _scrollToElement() when requested', () => {
+    const _scrollToElement = jest.fn();
+
+    // We need `mount()` because `ref` is only used in a DOM environment.
+    const root = mount(
```
kumar303: Hmm, I don't think it's a great idea to use `mount` here. It will introduce complications if `LinterMessage` becomes a Redux connected component now or in the future.

Here is an example of a way to test it without `mount`:
https://github.com/mozilla/addons-code-manager/blob/c29a7ddd16ec6b43d84ccc580b415ca3988c259a/src/components/FileTreeNode/index.tsx#L30

https://github.com/mozilla/addons-code-manager/blob/c29a7ddd16ec6b43d84ccc580b415ca3988c259a/src/components/FileTreeNode/index.spec.tsx#L599

Testing it without `mount` isn't perfect because you can't cover the actual `ref=` definition part but maybe that's OK. I guess it depends. Anyway, just throwing this out as an alternative.

src/components/CodeView/index.tsx
```diff
@@ -54,13 +54,13 @@ type Props = PublicProps & RouteComponentProps;
...
   id: string;
-  ref?: typeof scrollToSelectedLine;
+  ref?: typeof scrollToElement;
 };
 
 export class CodeViewBase extends React.Component<Props> {
   renderWithLinterInfo = ({ selectedMessageMap }: LinterProviderInfo) => {
     const {
-      _scrollToSelectedLine = scrollToSelectedLine,
+      _scrollToElement = scrollToElement,
```
kumar303: Why does this callback need to be applied to the `LinterMessage` ref at the same time it's applied to the table row ref? Aren't scrolling to a line and scrolling to a linter message independent actions?

src/components/LinterMessage/index.tsx
bobsilverberg: Yes, I tried that first and got the TS error you mention in that issue, which is why I created the `div`.

src/components/LinterMessage/index.tsx
```diff
@@ -59,22 +60,29 @@ const renderDescription = (description: LinterMessage['description']) => {
 };
 
 type PublicProps = {
+  _scrollToElement: typeof scrollToElement | null;
```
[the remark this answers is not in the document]
bobsilverberg: Good point, although I think I'm going to take your advice from another comment and make this use Redux state rather than a passed prop.

src/components/LinterMessage/index.tsx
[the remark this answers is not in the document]
bobsilverberg: Yes, just the tests. I'll change it.

src/components/CodeView/index.tsx
bobsilverberg: Yes, I'm going to try using Redux state instead.

src/components/CodeView/index.tsx
bobsilverberg: I don't really understand this question, I didn't think I was applying it to both, but I guess that could be possible given the code. In any case, I will probably get rid of this when moving to using Redux state instead.

bobsilverberg:
I am closing this in favour of #781.

Issue #699, opened 2019-05-07
The file tree jumps to the the current selected file when a subfolder is opened


AlexandraMoga:
**STR**

1. Open the browse files page for an addon - https://code.addons-dev.allizom.org/en-US/browse/616666/versions/1655561/
2. Locate a sub-folder in the file tree and click on it

**Actual result:**
The file viewer jumps back to the current open file

**Expected result:**
The file viewer remains in the position where the sub-folder was expanded

**Notes:**
- reproduced on -dev with Fx66, Win10x64
- I could not reproduce in the compare page

![2019-05-07_14-11-03](https://user-images.githubusercontent.com/31961530/57295514-623bb000-70d3-11e9-9167-f2a8e7424f46.gif)

wagnerand:
This looks similar to #642.

bobsilverberg:
This is likely due to https://github.com/mozilla/addons-code-manager/pull/687. I can take a look this morning.

bobsilverberg:
Yes, this was definitely introduced by #687. The problem is that that patch causes the currently selected file to always scroll into view whenever the `FileTreeNode` component mounts or updates, and expanding a folder dispatches `TOGGLE_EXPANDED_PATH`, which seems to cause a number of mounts. This is coupled with the fact that `isSelected` on `FileTreeNode` only cares about which file is currently selected (via `version.selectedPath`), and is unaffected when a user selects a folder to expand. Thus far we have never really cared about when a user selected a folder to expand, and have only used that event to add to `expandedPaths`, but it looks like we now might need to change that, and have a way of knowing which is the currently selected folder.

@kumar303 what are you thoughts on this?

kumar303:
> @kumar303 what are you thoughts on this?

It makes the site unusable so we should fix it ASAP 😀 I'll take a look.

AlexandraMoga:
Verified fixed on -dev with Fx66, Win10x64

- there's no more scroll jumping in the file tree
![2019-05-17_17-11-35](https://user-images.githubusercontent.com/31961530/57934510-8e6fe180-78c8-11e9-9bbb-e6e439f0f71c.gif)

PR #753, opened 2019-05-15
Only scroll/focus a FileTreeNode when necessary

kumar303:
This fixes https://github.com/mozilla/addons-code-manager/issues/699 by keeping track of focused nodes and only scrolling when they first come into focus.

### DIFF
--- a/src/components/FileTree/index.tsx
+++ b/src/components/FileTree/index.tsx
@@ -78,7 +78,9 @@ export class FileTreeBase extends React.Component<Props> {
     const { onSelect, version } = this.props;
 
     if (version) {
-      return <FileTreeNode {...props} onSelect={onSelect} version={version} />;
+      return (
+        <FileTreeNode {...props} onSelect={onSelect} versionId={version.id} />
+      );
     }
     return <Loading message={gettext('Loading version...')} />;
   };
@@ -184,7 +186,10 @@ const mapStateToProps = (
   const version = getVersionInfo(state.versions, versionId);
 
   if (!version) {
-    // This should never happen as we are fetching the version in all of the
+    // TODO: support loading version objects as needed.
+    // https://github.com/mozilla/addons-code-manager/issues/754
+    //
+    // An empty version should never happen as we are fetching the version in all of the
     // parents on this component and only rendering the FileTree if we have a
     // version, but let's log a warning in case we encounter a case where this
     // does happen.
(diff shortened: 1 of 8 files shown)


src/components/FileTree/index.tsx
```diff
@@ -78,7 +78,9 @@ export class FileTreeBase extends React.Component<Props> {
     const { onSelect, version } = this.props;
 
     if (version) {
-      return <FileTreeNode {...props} onSelect={onSelect} version={version} />;
+      return (
+        <FileTreeNode {...props} onSelect={onSelect} versionId={version.id} />
```
willdurand: This and the change in `FileTreeNode` go against https://github.com/mozilla/addons-code-manager/pull/400#discussion_r267058168

src/test-helpers.tsx
```diff
@@ -650,3 +651,12 @@ export const createFakeRef = (
     },
   };
 };
+
+export const createStoreWithVersion = (
```
willdurand: Should we file an issue to use this helper function everywhere else?

src/components/FileTree/index.tsx
kumar303: Yes, it does go against it which is why I filed https://github.com/mozilla/addons-code-manager/issues/754 to fix it. There's a todo comment about it `mapStateToProps()`.

The problem here is that we can't rely on a passed `version` object *and* be guaranteed a re-render when `version.focusedPath` changes. We need to map the object to state to guarantee an update.

src/test-helpers.tsx
kumar303: Yeah, good idea. I did a quick search and it can be used in a lot of places.

kumar303:
> It took me a while to understand how a "focused path" would fix this issue given that the selected path was always visible and therefore looked "focused" enough to me, but I got it.

Well, the selected path was not always visible. In the case where scrolling was needed, the path node was off the screen in the panel overflow; it was invisible.

> 
> I think something mentioning `scroll` (like `lastScrolledPath`, maybe) would make things more obvious given that "focus" is often referred to things visible, but I am not a native speaker 🤷‍♂

Hmm, *scrolling into view* is more of an implementation of making the path node visible where it was previously invisible. I can see how *focused* could be confusing, though, since it implies a real DOM focus. I changed the names to use the word *visible* instead.

Thanks for the review.

kumar303:
The Snyk report won't load for me (times out). Let's see what happens.

Issue #762, opened 2019-05-16
Simulate an error so it can be seen in Sentry


kumar303:
We need to be able to simulate an error so it can be seen in Sentry. This will help us check each environment (dev, stage, prod) to debug https://github.com/mozilla/addons-code-manager/issues/760

PR #768, opened 2019-05-20
Allow error simulation to test Sentry

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/762 with some simple buttons.

### DIFF
--- a/.env
+++ b/.env
@@ -18,3 +18,6 @@ INLINE_RUNTIME_CHUNK=false
 # CRA forces us to use a variable named `PUBLIC_URL` to set the CDN URL...
 # See: https://facebook.github.io/create-react-app/docs/advanced-configuration
 PUBLIC_URL=https://addons-code-manager.cdn.mozilla.net
+
+# When true, allow the app to trigger errors that will be reported to Sentry.
+REACT_APP_UNSAFE_ERROR_SIMULATION=false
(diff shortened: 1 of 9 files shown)


willdurand:
Trying this patch locally, I got a CSP error because of the CSP dev settings. I do believe we don't need to override `connectSrc` in dev. The prod config should be correct given that we're using the insecure proxy locally.

I fixed the CSP issue by removing one line and everything continued to work:

```diff
diff --git a/.env.common-local b/.env.common-local
index dbd9762..47d93e7 100644
--- a/.env.common-local
+++ b/.env.common-local
@@ -9,6 +9,6 @@ REACT_APP_IS_LOCAL_DEV=true
 # This is the DSN for the `addons-code-manager-test` Sentry project.
 # Uncomment the line with the variable below if you want to see (local) errors
 # in: https://sentry.prod.mozaws.net/operations/addons-code-manager-test/.
-# REACT_APP_SENTRY_DSN=https://e4cdb3d53fc84e599b1662d7e51e3fed@sentry.prod.mozaws.net/425
+REACT_APP_SENTRY_DSN=https://e4cdb3d53fc84e599b1662d7e51e3fed@sentry.prod.mozaws.net/425
 
 REACT_APP_UNSAFE_ERROR_SIMULATION=true
diff --git a/src/server/index.tsx b/src/server/index.tsx
index 630ef88..98951dc 100644
--- a/src/server/index.tsx
+++ b/src/server/index.tsx
@@ -252,7 +252,6 @@ export const createServer = ({
           imgSrc: [...prodCSP.imgSrc, "'self'"],
           scriptSrc: ["'self'", "'unsafe-inline'"],
           styleSrc: ["'self'", "'unsafe-inline'"],
-          connectSrc: ["'self'"],
         },
       }),
     );
```

After that, I could see errors in Sentry **but** `.error` calls are not reported (the second button). That would explain why we did not receive any error in prod last time.

kumar303:
Thanks for the patch. I totally forgot that I could test this locally with the test instance 🤦‍ I made the CSP changes and added a new button to test console logging, to deduce if the `logLevel` wrapper was causing a problem. It's not. I'm still digging into the disappearing console messages but at least we have a way to reproduce it with this simulation. 

I'll tackle the logging problem separately.

PR #781, opened 2019-05-23
Return a line from getRelativeMessage

### LINKED ISSUE #750: Return the line number from getRelativeMessage

bobsilverberg:
Fixes #750

### DIFF
--- a/src/reducers/fileTree.spec.tsx
+++ b/src/reducers/fileTree.spec.tsx
@@ -693,7 +693,7 @@ describe(__filename, () => {
           currentMessageUid: '',
           messageMap,
         }),
-      ).toEqual({ path: file1, uid: global1 });
+      ).toEqual({ line: null, path: file1, uid: global1 });
     });
 
     it('returns the first global message from a path with messages if no currentMessageUid exists', () => {
@@ -715,7 +715,7 @@ describe(__filename, () => {
           messageMap,
           pathList,
         }),
-      ).toEqual({ path: file2, uid: global1 });
+      ).toEqual({ line: null, path: file2, uid: global1 });
     });
 
     it('returns the first byLine message if no currentMessageUid exists and no global messages exist', () => {
@@ -730,7 +730,7 @@ describe(__filename, () => {
           currentMessageUid: '',
           messageMap,
         }),
-      ).toEqual({ path: file1, uid: line1FirstUid });
+      ).toEqual({ line: line1, path: file1, uid: line1FirstUid });
     });
 
     describe('messages within a file', () => {
@@ -754,7 +754,7 @@ describe(__filename, () => {
             messageMap,
             position: RelativePathPosition.next,
           }),
-        ).toEqual({ path: file1, uid: global2 });
+        ).toEqual({ line: null, path: file1, uid: global2 });
       });
 
       it('returns the previous global message in the file', () => {
@@ -764,7 +764,7 @@ describe(__filename, () => {
             messageMap,
             position: RelativePathPosition.previous,
           }),
-        ).toEqual({ path: file1, uid: global1 });
+        ).toEqual({ line: null, path: file1, uid: global1 });
       });
 
       it('returns the first byLine message when at the last global message for the file', () => {
@@ -774,7 +774,7 @@ describe(__filename, () => {
             messageMap,
             position: RelativePathPosition.next,
           }),
-        ).toEqual({ path: file1, uid: line1FirstUid });
+        ).toEqual({ line: line1, path: file1, uid: line1FirstUid });
       });
 
       it('returns the last global message when at the first byLine message for the file', () => {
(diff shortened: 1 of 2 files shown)


src/reducers/fileTree.spec.tsx
```diff
@@ -693,7 +693,7 @@ describe(__filename, () => {
           currentMessageUid: '',
           messageMap,
         }),
-      ).toEqual({ path: file1, uid: global1 });
+      ).toEqual({ line: null, path: file1, uid: global1 });
```
kumar303: You shouldn't have had to update this test because all it's doing is checking that the *first* global message is discovered. I suggest changing the assertion to:

```ts
      expect(
        _getRelativeMessage({
          currentMessageUid: '',
          messageMap,
        }),
      ).toEqual(expect.objectContaining({ uid: global1 }));
```

It's not important to add `line` to the test assertions when not relevant because TypeScript covers that logic.

src/reducers/fileTree.spec.tsx
```diff
@@ -715,7 +715,7 @@ describe(__filename, () => {
           messageMap,
           pathList,
         }),
-      ).toEqual({ path: file2, uid: global1 });
+      ).toEqual({ line: null, path: file2, uid: global1 });
```
kumar303: Same here

src/reducers/fileTree.spec.tsx
```diff
@@ -730,7 +730,7 @@ describe(__filename, () => {
           currentMessageUid: '',
           messageMap,
         }),
-      ).toEqual({ path: file1, uid: line1FirstUid });
+      ).toEqual({ line: line1, path: file1, uid: line1FirstUid });
```
kumar303: Same

src/reducers/fileTree.spec.tsx
```diff
@@ -754,7 +754,7 @@ describe(__filename, () => {
             messageMap,
             position: RelativePathPosition.next,
           }),
-        ).toEqual({ path: file1, uid: global2 });
+        ).toEqual({ line: null, path: file1, uid: global2 });
```
kumar303: ditto

src/reducers/fileTree.spec.tsx
```diff
@@ -764,7 +764,7 @@ describe(__filename, () => {
             messageMap,
             position: RelativePathPosition.previous,
           }),
-        ).toEqual({ path: file1, uid: global1 });
+        ).toEqual({ line: null, path: file1, uid: global1 });
```
kumar303: Same again. It would be good to verify that all messages in the setup data have unique uids, though. I won't comment on all of them but I did a quick glance and I think they should all use `expect.objectContaining()`.

src/reducers/fileTree.tsx
```diff
@@ -235,7 +235,8 @@ export const findRelativePathWithDiff = ({
   );
 };
 
-type MessagePathAndUid = {
+type MessagePathLineAndUid = {
```
kumar303: You could pick a generic name, like `RelativeMessageInfo`

src/reducers/fileTree.tsx
```diff
@@ -261,7 +262,8 @@ const findRelativeMessage = (
         position === RelativePathPosition.previous ? msgArray.length - 1 : 0;
 
       if (msgArray[msgIndex]) {
-        return { path: nextPath, uid: msgArray[msgIndex].uid };
+        const message = msgArray[msgIndex];
+        return { line: message.line, path: nextPath, uid: message.uid };
```
kumar303: This (and the next one) would benefit from a shared arrow function you can call, like:

```ts
return makeRelativeMessageInfo({ message, path });
```

```ts
const makeRelativeMessageInfo = ({ message, path }) => {
  return { line: message.line, path, uid: message.uid };
};
```

bobsilverberg:
Thanks for the review @kumar303. I have addressed the issues and added some new tests specifically for line numbers.

src/reducers/fileTree.spec.tsx
```diff
@@ -774,7 +784,27 @@ describe(__filename, () => {
...
           }),
-        ).toEqual({ path: file1, uid: line1FirstUid });
+        ).toEqual(expect.objectContaining({ uid: line1FirstUid }));
+      });
+
+      it('returns the expected line number for a byLine message', () => {
+        // Expect the first message on line 1.
+        expect(
+          _getRelativeMessage({
+            currentMessageUid: global2,
```
kumar303: For future reference, it reduces test readability to rely on a [common setup routine](https://github.com/mozilla/addons-code-manager/blob/2fb604f38607f7fc4593adefb8139e81117ee9d5/src/reducers/fileTree.spec.tsx#L737-L748). It's hard for me to look at this test and understand what the inputs are. 

It's convenient to rely on common setup like this but you can still get the right amount of convenience by calling a setup helper explicitly at the top of every test with just a few lines of code.

This is not a change request, just a comment.

bobsilverberg: I totally understand what you mean, and even thought about it while adding these new tests - the fact that the inputs aren't obvious from the test itself. I'll keep this in mind for the next time I consider doing this, or if we need to do more work on these test cases.

Issue #756, opened 2019-05-16
Show add-on name in top bar


wagnerand:
Instead of showing _addons-code-manager_, we could show the add-on name. Showing the version would also be nice, at least in browse mode. In compare mode, you can see the versions in the compare dropdowns.

kumar303:
https://github.com/mozilla/addons-code-manager/pull/806 added a similar feature but for web browser page titles. Maybe we can re-use that functionality here.

mirefly:
I found it more complicated than I expected... I need to access `versionId` from `Navbar`, so I put `Navbar` in `Route` like the following. I am not sure I am on the right track @kumar303 

```diff 
--- a/src/components/App/index.tsx
+++ b/src/components/App/index.tsx
@@ -157,10 +157,15 @@ export class AppBase extends React.Component<Props> {
 
     return (
       <FullscreenGrid>
-        <Header>
-          {!loading && <Navbar />}
-          {this.renderErrors()}
-        </Header>
+        <Route
+          path={'/:lang/browse/:addonId/versions/:versionId/'}
+          children={() => (
+            <Header>
+              {!loading && <Navbar />}
+              {this.renderErrors()}
+            </Header>
+          )}
+        />
         {this.renderContent()}
       </FullscreenGrid>
     );

```

kumar303:
@mirefly yeah, this one is tricky. Using a route won't work for [compare views](https://code.addons-dev.allizom.org/en-US/compare/502868/versions/1541685...1688187/) since the URL parameters for fetching the version are different. I'm actually trying to come up with a solution for propagating the "current version" to the header in https://github.com/mozilla/addons-code-manager/issues/805 so maybe hold off on this issue for a week-ish?

You might find some interesting issues in the [code quality](https://github.com/mozilla/addons-code-manager/issues?q=is%3Aissue+is%3Aopen+label%3A%22component%3A+code+quality%22) label. However, some are more complex than others.

wagnerand:
This is quite relevant since #789 is fixed.

kumar303:
@mirefly did you want to take another look? It should be easier now that we have https://github.com/mozilla/addons-code-manager/pull/885 . 

[...]

PR #806, opened 2019-06-03
Set page titles for Browse and Compare

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/809


kumar303:
I didn't consider it because we don't need SSR support. Is it still worth using? I guess it would be one less custom component to manage but yet another dependency to manage.

kumar303:
I really want [`useTitle()`](https://streamich.github.io/react-use/?path=/story/side-effects-usetitle--docs) but we're still waiting on Enzyme support. ⏳

willdurand:
> I didn't consider it because we don't need SSR support. Is it still worth using? I guess it would be one less custom component to manage but yet another dependency to manage.

In addons-frontend, it's used by both (client and server) sides. This dependency handles more than page titles and provides a declarative style _à la_ React.

kumar303:
Sure, I can switch to it. Adding something with SSR support and other features we don't need seems a bit overkill but I suppose bundle size isn't a huge concern right now.

kumar303:
It's using `react-helmet` now

src/pages/Browse/index.spec.tsx
```diff
@@ -477,4 +477,33 @@ describe(__filename, () => {
...
+    const name = 'AdBlockPlus';
+    const versionString = '1.0-beta';
+    const version = {
+      ...fakeVersion,
+      id: fakeVersion.id + 1,
+      addon: {
+        ...fakeVersion.addon,
+        name: {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          [(process.env as any).REACT_APP_DEFAULT_API_LANG]: name,
```
bobsilverberg: It's a bit ugly that we have to do this. Is there a way to have a friendlier test helper that would allow us to work with localizedStrings?

kumar303:
> I'm not sure it really addresses the issue at #756.

Hah, that's true! I misread the issue. I still think having page titles is a useful feature so I filed https://github.com/mozilla/addons-code-manager/issues/809 and updated the PR description.

> While I like the fact that we now have a useful HTML `title` for the page, it's still not particularly easy to see. It does appear on the tab, but for me, even with only one tab in a window, the tab is so narrow that I cannot really read the version information (without hovering over it with my mouse).

That's just a limitation of the default tab layout. If you switch to any vertical tab layout (via an extension) then it's much more useful to have a page title for tabs. Also, page titles are important for awesome bar completion, link sharing, and bookmarking.

kumar303:
> So I suggest ... we use the same values instead of the static "addons-code-manager" in the nav bar.

Yeah, let's do that in https://github.com/mozilla/addons-code-manager/issues/756

src/pages/Browse/index.spec.tsx
kumar303: That is true. I pushed a fix, thanks.

kumar303:
Yep, good idea. I added a helper for that.

Issue #682, opened 2019-05-03
After selecting an inline linter message from page overview and scrolling up\down, you can't select it anymore


vcarciu:
Steps to reproduce:
1.Go to https://code.addons-dev.allizom.org/en-US/browse/494431/versions/1532144/?path=messageResponder.js
2.Select the inline linter message from page overview in the right side
3.Scroll down until the linter message is not displayed anymore
4.Select the linter message again from page overview

Expected results:
The linter message and according line of code is displayed again at the top of the page

Actual results:
Nothing happens

Please see a video of the issue : 
![lintermes](https://user-images.githubusercontent.com/1583842/57144767-75483a80-6dca-11e9-8986-4cb518cc9f4d.gif)

vcarciu:
Verified as fixed : 
![lintermess](https://user-images.githubusercontent.com/1583842/58964843-4140aa80-87b8-11e9-8c5d-1dfd568bcb17.gif)

PR #808, opened 2019-06-03
Always scroll to a linter message when clicked

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/682

### DIFF
--- a/src/components/CodeOverview/index.spec.tsx
+++ b/src/components/CodeOverview/index.spec.tsx
@@ -130,6 +130,15 @@ describe(__filename, () => {
     return { innerRoot, root, selectedMessageMap };
   };
 
+  const renderWithFixedHeight = (
+    props: RenderParams = {},
+    { overviewHeight = 400 } = {},
+  ) => {
+    const root = render(props);
+    root.setState({ overviewHeight });
+    return renderWithLinterProvider({ root });
+  };
+
   it('configures LinterProvider', () => {
     const version = createInternalVersion(fakeVersion);
     const root = render({ version });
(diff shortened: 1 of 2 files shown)


src/components/CodeOverview/index.spec.tsx
```diff
@@ -240,6 +246,49 @@ describe(__filename, () => {
...
+      .at(line - 1)
+      .simulate('click');
+
+    expect(_document.querySelector).toHaveBeenCalledWith(
+      getCodeLineAnchor(line),
+    );
+    expect(fakeDOMNode.scrollIntoView).toHaveBeenCalled();
+  });
+
+  it('does not scroll a linter message if its anchor does not exist', () => {
```
bobsilverberg: I get what you're doing here, but I find it a bit odd that the test explicitly says "does not scroll a linter message" but then doesn't actually assert anything about that. What it does is create a situation in which an anchor does not exist, and verifies that it doesn't throw an exception, but it doesn't specifically verify that nothing was scrolled. I understand that we cannot really test that as we're not returning a DOMNode, but rather `null`, but I wonder if we should name the test differently.

src/components/CodeOverview/index.spec.tsx
```diff
@@ -130,6 +130,15 @@ describe(__filename, () => {
     return { innerRoot, root, selectedMessageMap };
   };
 
+  const renderWithFixedHeight = (
```
bobsilverberg: This is a nice enhancement to the test suite. 👍

src/components/CodeOverview/index.spec.tsx
kumar303: I updated the spec so it's more precise

Issue #771, opened 2019-05-21
Make layout a bit more compact


kewisch:
This may be a matter of personal preference, but I've reduced the page size to about 80% so it is somewhat similarly sized to the old file viewer. I think we should consider adjusting the styling so it is a bit more compact and you can see more code on one screen.

kumar303:
> I've reduced the page size to about 80%

@kewisch what do you mean by this? Could you provide the dimensions of your window, roughly?

In the next release, the side panel can be collapsed which might help. Demo: https://github.com/mozilla/addons-code-manager/pull/735

kewisch:
I've used the page zoom features of Firefox to zoom out to 80%. I'm basically saying I feel the whole UI is a bit too big and we could save space by making fonts and such smaller.

I'm running a Macbook Pro with 2880 x 1800 Retina.

kumar303:
Oh, I totally see what you mean now. Yeah, the font size was bigger than how I have my own code editor set up.

I made some adjustments in https://github.com/mozilla/addons-code-manager/pull/810 ; can you check the screenshots? @wagnerand maybe you'd also like to weigh in on these screenshots?

kewisch:
This already looks better, thanks! I'm thinking next it would be good to reduce margins and padding a bit, for example on `.alert` it would be enough to have let's say `padding: .5rem .75rem`. Buttons also could be a bit more dense. What do you think?

kumar303:
Sure, we could tighten it up more. Can you file a separate issue?

kumar303:
@kewisch actually, I ended up making the buttons more compact as part of this change, too. It should be landing on https://code.addons-dev.allizom.org/ shortly.

ioanarusiczki:
@kumar303 I verified this issue on code.addons-dev.allizom.org (on FF67 -Win10) 
The font size is smaller, also the buttons look more compact as shown in https://github.com/mozilla/addons-code-manager/pull/810

Am example 

<details>
<summary>Before - AMO stage </summary>


![AMO stage](https://user-images.githubusercontent.com/33448286/59272967-d7188180-8c5f-11e9-8ed1-bf73752637ee.png)

</details>

<details>
<summary>After - AMO dev </summary>

![AMO dev](https://user-images.githubusercontent.com/33448286/59272990-e4357080-8c5f-11e9-83b4-a3272b42c065.png)

</details>

PR #810, opened 2019-06-04
Reduce font size for a more compact code view

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/771

This should make it easier to read large code files and also to deal with big file trees. I used the old review site as a reference and also my own editor's font size.

Below are some screenshots from a macOS retina screen so you may need to consider that when viewing them at their actual size.

<details>
<summary>Before</summary>

acm-font-fix_before-1
acm-font-fix_before-2
acm-font-fix_before-3
acm-font-fix_before-4

</details>

<details>
<summary>After</summary>

acm-font-fix_after-1
acm-font-fix_after-2
acm-font-fix_after-3
acm-font-fix_after-4

</details>


wagnerand:
I like this change! 👍

src/styles.scss
```diff
@@ -5,3 +5,7 @@ body {
 
 // Import Bootstrap and its default variables
 @import '~bootstrap/scss/bootstrap.scss';
+
+body {
+  font-size: 87%;
```
willdurand: lol, `87`?

src/scss/variables.scss
```diff
@@ -12,7 +12,8 @@ $alert-secondary-bg-color: #e2e3e5;
 
 // Shared style for the CodeView and DiffView components.
 $code-font-family: Consolas, Courier, monospace;
-$code-font-size: 16px;
+$code-font-size: 13px;
```
willdurand: Could we use `em` here given that the body font-size is going to be a percentage value?

src/styles.scss
kumar303: This scales the default browser size (typically `16px`) to roughly `14px` which looked about right to me. Should I use a pixel value instead? The best practice used to be setting a percentage but iirc that was only because browsers wouldn't zoom pixel sized fonts. Now they do.

src/scss/variables.scss
kumar303: `13px` seems important for viewing code. What would be better about using a percentage? Just for consistency with the rest of the site?

src/styles.scss
willdurand: I think we can keep it as is.

src/scss/variables.scss
willdurand: Yep. If one wants to increase the font (using the browser zoom feature), then the font size of the code should also be increased.

kumar303: Browsers now scale pixel sized fonts when zooming. Is there any other reason not to use pixels?

willdurand: If so, then let's switch to pixels for everything then? Including the `body` font size, WDYT?

kumar303: Yeah, good idea.

kumar303:
@willdurand this is ready for another look

src/refractor/prism.css
```diff
@@ -1,4 +1,7 @@
 /*
+
+ Do not edit this file. Edit prism.overrides.scss instead.
```
willdurand: ❤️

kumar303:
For accessibility: I checked that zooming works (it does) and also that [setting a minimum font size](https://support.mozilla.org/en-US/kb/font-size-and-zoom-increase-size-of-web-pages#w_set-a-minimum-text-size) works (it does).

kumar303:
> Large buttons on the left panel have a larger font size than the rest (or at least, it seems off compared to the rest) but great change, thanks!

Hmm, yeah. TIL: `rem` sets font sizes relative to the *default* browser font size, not the body font size (like I thought).

kumar303:
> Hmm, yeah. TIL: `rem` sets font sizes relative to the _default_ browser font size, not the body font size (like I thought).

Oh, duh, it's only because `rem` is relative to the root element. I moved the font-size to `html` and bootstrap is in sync now ✨

willdurand:
> Oh, duh, it's only because `rem` is relative to the root element. I moved the font-size to `html` and bootstrap is in sync now ✨

cool, that works well 🥇

Issue #813, opened 2019-06-05
The Browse web page title contains an '@' before the add-on version string


AlexandraMoga:
Follow up for #809 

**STR:**
1. Open the  browse page for an add-on - see https://code.addons-dev.allizom.org/en-US/browse/385621/versions/1655456/
2. Observe the page title

**Actual result:**
The page title is `Browse DuckDuckGo Plus@2019.1.31`

**Expected result:**
The page title should more likely be `Browse DuckDuckGo Plus:2019.1.31`

**Notes:**
- another suggestion would be to use the version ID instead of the version string
- repoduced on -dev with Fx67, Win10x64

![image](https://user-images.githubusercontent.com/31961530/58963824-849a1980-87b6-11e9-92e6-e2bcdd50d4bb.png)

kumar303:
I used the at sign on purpose to convey *this add-on at this version*. Maybe I'll add a space, though, to make it look more like `Browse DuckDuckGo Plus @ 2019.1.31`

kumar303:
It seems that switching the browse title to a colon is a popular request so I'll go with that 😅

AlexandraMoga:
Verified fixed on -dev
![image](https://user-images.githubusercontent.com/31961530/59110719-5ea47e80-8948-11e9-959d-97188e3740ad.png)

![image](https://user-images.githubusercontent.com/31961530/59110734-6401c900-8948-11e9-8fa9-e403c984a2e0.png)

PR #815, opened 2019-06-05
Adjust Browse / Compare page titles

kumar303:
Closes https://github.com/mozilla/addons-code-manager/issues/813

This makes some minor adjustments to the titles.

Browse example: `Browse DuckDuckGo Plus: 2019.1.31`

Compare example: `Compare Grammarly for Firefox: 123456...234567`

### DIFF
--- a/src/pages/Browse/index.spec.tsx
+++ b/src/pages/Browse/index.spec.tsx
@@ -502,6 +502,6 @@ describe(__filename, () => {
 
     const root = render({ store, versionId: String(version.id) });
 
-    expect(root.find('title')).toHaveText(`Browse ${name}@${versionString}`);
+    expect(root.find('title')).toHaveText(`Browse ${name} @ ${versionString}`);
   });
 });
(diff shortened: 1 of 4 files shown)


src/pages/Browse/index.spec.tsx
```diff
@@ -502,6 +502,6 @@ describe(__filename, () => {
 
     const root = render({ store, versionId: String(version.id) });
 
-    expect(root.find('title')).toHaveText(`Browse ${name}@${versionString}`);
+    expect(root.find('title')).toHaveText(`Browse ${name} @ ${versionString}`);
```
willdurand: You could maybe use `:` since this is the separator used in the compare title page

kumar303: Heh, that's what QA suggested, too. They are conveying different things, though. The browse title is conveying *browse this add-on at this version* whereas the compare title is conveying *compare this add-on between version1 and version2*.

kumar303: Well, I guess it doesn't matter. I can make a colon since that makes more sense to everyone.

PR #818, opened 2019-06-06
Implement goToRelativeMessage

### LINKED ISSUE #782: Create a goToRelativeMessage function
This supports #520 

~~Depends on #750~~

Create a function that can be called by an action which will navigate between linter messages. This function will either push a new location for a message on in current file, or will dispatch `viewVersionFile` for a message in a different file.

bobsilverberg:
Fixes #782

### DIFF
--- a/src/reducers/fileTree.spec.tsx
+++ b/src/reducers/fileTree.spec.tsx
@@ -1,4 +1,7 @@
 /* eslint @typescript-eslint/camelcase: 0 */
+import { push } from 'connected-react-router';
+import queryString from 'query-string';
+
 import reducer, {
   ROOT_PATH,
   DirectoryNode,
@@ -13,12 +16,17 @@ import reducer, {
   getRelativePath,
   getTree,
   goToRelativeFile,
+  goToRelativeMessage,
   initialState,
 } from './fileTree';
 import { createInternalVersion, createInternalVersionEntry } from './versions';
 import { getMessageMap } from './linter';
+import { getCodeLineAnchor } from '../components/CodeView/utils';
+import configureStore from '../configureStore';
 import {
   createFakeExternalLinterResult,
+  createFakeHistory,
+  createFakeLocation,
   createFakeThunk,
   createVersionWithInternalEntries,
   fakeVersion,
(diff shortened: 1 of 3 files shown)


src/reducers/fileTree.tsx
```diff
@@ -372,6 +376,75 @@ export const goToRelativeFile = ({
...
+
+      const queryParams = queryString.parse(location.search);
+      const newParams = { ...queryParams };
+      newParams.path = path;
+      newParams.messageUid = uid;
+
+      // Update the location with the hash for the message uid.
+      const newLocation = {
+        ...location,
+        hash: getCodeLineAnchor(line || 0),
```
kumar303: When `line` is empty, it means you're dealing with a global message so shouldn't this be `getCodeLineAnchor(line || 1)`? Or does `0` scroll to the top?

src/reducers/fileTree.tsx
```diff
@@ -372,6 +376,75 @@ export const goToRelativeFile = ({
...
+
+    if (nextRelativeMessageInfo) {
+      const { line, path, uid } = nextRelativeMessageInfo;
+      const {
+        router: { location },
+      } = getState();
+
+      const queryParams = queryString.parse(location.search);
+      const newParams = { ...queryParams };
+      newParams.path = path;
```
kumar303: It would be helpful to create a constant for `'path'` and share it with this code:

https://github.com/mozilla/addons-code-manager/blob/d821eeb670e7a24b321654684a4f40fe802ff67a/src/utils.tsx#L56

Otherwise, it's hard to grep for `path` and find all the times it's used as a query string parameter. While you're at it, I'd suggest making `messageUid` a constant, too.

src/reducers/fileTree.tsx
```diff
@@ -372,6 +376,75 @@ export const goToRelativeFile = ({
...
+      currentMessageUid,
+      currentPath,
+      messageMap,
+      pathList,
+      position,
+    });
+
+    if (nextRelativeMessageInfo) {
+      const { line, path, uid } = nextRelativeMessageInfo;
+      const {
```
kumar303: This could fit on one line

src/reducers/fileTree.tsx
```diff
@@ -372,6 +376,75 @@ export const goToRelativeFile = ({
...
+        router: { location },
+      } = getState();
+
+      const queryParams = queryString.parse(location.search);
+      const newParams = { ...queryParams };
```
kumar303: It would be helpful to add a test that makes sure existing query parameters are preserved.

src/reducers/fileTree.spec.tsx
```diff
@@ -1065,6 +1073,175 @@ describe(__filename, () => {
...
+          }),
+        store: configureStore({ history }),
+      });
+
+      await thunk();
+
+      expect(dispatch).toHaveBeenCalledWith(
+        push({
+          ...location,
+          search: queryString.stringify({ messageUid: uid, path: currentPath }),
```
kumar303: This needs to use something like [`expect.urlWithTheseParams()`](https://github.com/mozilla/addons-code-manager/blob/d821eeb670e7a24b321654684a4f40fe802ff67a/src/jest-matchers/index.tsx#L32) otherwise the test will rely on the random order of stringifying object keys.

src/reducers/fileTree.tsx
bobsilverberg: Yes, in my next patch I am adding a `div` with an `id` that corresponds to `getCodeLineAnchor(0)` above the global linter messages to provide a destination for scrolling to global messages.

src/reducers/fileTree.tsx
bobsilverberg: I _think_ I understand what you're asking, so I've implemented it. You can let me know if I got it right when I push the next commit.

src/reducers/fileTree.tsx
bobsilverberg: For some reason each time I put that on a single line, `prettier` changes it back to look like this. Not sure why. Do you want to give it a try and see if the same happens to you?

kumar303: ah, ok, then no worries. Prettier is the boss.

src/reducers/fileTree.tsx
kumar303: That sounds good. Adding a test (if there's not one already) to make sure it goes to line 0 would be good.

bobsilverberg:
Thanks for the review, @kumar303. I have addressed all of the issues.

PR #824, opened 2019-06-10
 Enable keyboard navigation between linter messages 

### LINKED ISSUE #520: Enable keyboard navigation between linter messages
User story:

> As a reviewer, I want to quickly see inline linter messages so I can understand their impact when reviewing code.

This supports #199 

Two of the keyboard shortcuts required for #199 are meant to allow a user to jump between linter messages for the extension. 

The "Next message" shortcut would:
- Scroll to the next linter message in the current file, if there is another linter message which has not yet been scrolled to.
- Jump to the next file with a linter message, and scroll it into view, if there are no more linter messages in the current file. 

The "Previous message" shortcut would:
- Scroll to the previous linter message in the current file, if there is another linter message above the current scroll position.
- Jump to the previous file with a linter message, and scroll the last message for that file into view, if there are no previous linter messages in the current file. 

Note that the first case for each shortcut is a requirement (i.e., scrolling to messages in the current file), and that the second case is a "nice-to-have", but it does seem like it would add a lot of value in terms of enabling keyboard navigation for the entire extension's code.

See https://github.com/mozilla/addons-code-manager/issues/523 for addressing this same user story with a visual UI.

bobsilverberg:
Fixes #520

### DIFF
--- a/src/components/CodeView/index.spec.tsx
+++ b/src/components/CodeView/index.spec.tsx
@@ -3,12 +3,17 @@ import PropTypes from 'prop-types';
 import { Link } from 'react-router-dom';
 import { Location } from 'history';
 import { mount } from 'enzyme';
+import { Store } from 'redux';
 
 import configureStore from '../../configureStore';
 import refractor from '../../refractor';
 import LinterMessage from '../LinterMessage';
 import LinterProvider, { LinterProviderInfo } from '../LinterProvider';
-import { ExternalLinterMessage, getMessageMap } from '../../reducers/linter';
+import {
+  ExternalLinterMessage,
+  actions as linterActions,
+  getMessageMap,
+} from '../../reducers/linter';
 import { createInternalVersion } from '../../reducers/versions';
 import { getCodeLineAnchor, getCodeLineAnchorID, mapWithDepth } from './utils';
 import { getLanguageFromMimeType } from '../../utils';
@@ -28,6 +33,7 @@ import CodeView, { CodeViewBase, PublicProps, scrollToSelectedLine } from '.';
 describe(__filename, () => {
   type RenderParams = Partial<PublicProps> & {
     location?: Location<{}>;
+    store?: Store;
   };
 
   const createGlobalExternalMessage = (props = {}) => {
@@ -83,6 +89,7 @@ describe(__filename, () => {
 
   const renderWithMount = ({
     location = createFakeLocation(),
+    store = configureStore(),
     ...otherProps
   }: RenderParams = {}) => {
     const options = createContextWithFakeRouter({ location });
@@ -94,18 +101,19 @@ describe(__filename, () => {
       },
       context: {
         ...options.context,
-        store: configureStore(),
+        store,
       },
     });
+  };
+
+  type RenderWithMessagesParams = RenderParams & {
+    messages: Partial<ExternalLinterMessage>[];
   };
 
   const renderWithMessages = ({
     messages,
     ...props
-  }: {
-    messages: Partial<ExternalLinterMessage>[];
-    props?: RenderParams;
-  }) => {
+  }: RenderWithMessagesParams) => {
     const file = 'scripts/content.js';
     const map = getMessageMap(
       createFakeExternalLinterResult({
@@ -227,7 +235,7 @@ describe(__filename, () => {
     });
     // This is an anchor on the table row. This is a bit confusing here because
     // `#` refers to the ID (CSS) selector and not the hash. The ID value is
-    // `L1`.
+    // `I1`.
     expect(root.find(`#${getCodeLineAnchorID(1)}`)).toHaveLength(1);
   });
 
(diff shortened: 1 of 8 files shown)


src/components/CodeView/index.tsx
```diff
@@ -70,6 +70,11 @@ export class CodeViewBase extends React.Component<Props> {
 
     return (
       <React.Fragment>
+        {selectedMessageMap &&
+          selectedMessageMap.global.length > 0 &&
+          isLineSelected(getCodeLineAnchorID(0), location) && (
+            <div id={getCodeLineAnchorID(0)} ref={_scrollToSelectedLine} />
```
kumar303: Hmm, I guess we need something more generic so that it works for `DiffView`, too:

https://github.com/mozilla/addons-code-manager/blob/76866e1634f73e5b3fc9629ef9095c81b77c9bdb/src/components/DiffView/index.tsx#L229

You could file a follow-up for that.

src/components/CodeView/index.tsx
```diff
@@ -70,6 +70,11 @@ export class CodeViewBase extends React.Component<Props> {
...
+          )}
         {selectedMessageMap &&
           selectedMessageMap.global.map((message) => {
             return <LinterMessage key={message.uid} message={message} />;
```
kumar303: Is there an issue on file about giving linter messages a selected style? I know we talked about it. If not, please file it. I think it will help to show what's going on when there are multiple messages.

src/components/CodeView/index.spec.tsx
```diff
@@ -333,6 +334,68 @@ describe(__filename, () => {
...
+    const location = createFakeLocation({ hash: `#${getCodeLineAnchorID(0)}` });
+    const _scrollToSelectedLine = jest.fn();
+
+    const { root } = renderWithMessages({
+      _scrollToSelectedLine,
+      location,
+      messages: [externalMessage],
+    });
+
+    // TODO: This doesn't work because we need `mount()` as `ref` is only used in a DOM environment.
```
kumar303: Did you try the mock ref approach? https://github.com/mozilla/addons-code-manager/blob/76866e1634f73e5b3fc9629ef9095c81b77c9bdb/src/components/FileTreeNode/index.spec.tsx#L719

src/components/KeyboardShortcuts/index.tsx
```diff
@@ -145,6 +156,34 @@ export class KeyboardShortcutsBase extends React.Component<Props> {
...
+                position: RelativePathPosition.next,
+                versionId,
+              }),
+            );
+          }
+          break;
+        case 'a':
+          if (messageMap) {
+            dispatch(
+              _goToRelativeMessage({
```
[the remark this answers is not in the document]
bobsilverberg: So what you are saying is that when looking for the next/previous message, on the compare view, we should skip any files that don't have any changes?

src/components/CodeView/index.tsx
bobsilverberg: Ah, yes, I missed that. I think when testing manually it looked like it was working because it still loads the file, and the global messages are at the top, but it doesn't scroll to the global messages. I'll file a follow-up.

bobsilverberg: I opened #825

src/components/CodeView/index.tsx
bobsilverberg: I filed #826

src/components/CodeView/index.spec.tsx
bobsilverberg: I want to make sure I understand what's being suggested. From what I can tell, if I want to use the mock ref approach I also need to adopt a bunch of code similar to what is in `FileTreeNode`, correct? I.e., I'm going to need to add code for `componentDidMount` and `componentDidUpdate` which calls a function which evaluates whether I need to scroll the node in question into view?

What we're doing currently, which is simply assigning `_scrollToSelectedLine` to the `ref` property of the element if we know we want to scroll to it - that will no longer be possible if we want to be able to inject a mock ref. Is that accurate? I just want to check before I go and make a bunch of changes.

I assume that if I am making this change I should apply it not just to the `ref` for the `div` for global messages, but also to the `ref` which is potentially added to a table row at https://github.com/mozilla/addons-code-manager/blob/76866e1634f73e5b3fc9629ef9095c81b77c9bdb/src/components/CodeView/index.tsx#L95. Do you agree?

src/components/KeyboardShortcuts/index.tsx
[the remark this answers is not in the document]
kumar303: Yes, I think so. I'm ignoring the fact that not showing the file at all for *no changes* is a little bit weird 🤔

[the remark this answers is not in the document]
kumar303: s/little bit weird/something I'm not sure what reviewers think of/

src/components/CodeView/index.spec.tsx
kumar303: Ah, I understand the problem better now.

Ideally, we would use enzyme to shallow render and check that the `ref` prop was specified only for the selected state. This not supported by enzyme and there are lots of closed issues about it.

On second thought, I don't think switching to a mock ref object is worth the effort. You are right that it would require some redesign -- the addition of new mount/update logic.

The second best thing is to mount the component to check that the `ref` gets set only in the selected state. On my previous look, I missed that code [for mounting](https://github.com/mozilla/addons-code-manager/blob/365b8958bd517c444f4af1008a0eb3beb211cb4b/src/components/CodeView/index.spec.tsx#L84) already exists! Can't you just re-use that function but dispatch a linter result right before calling it? Like this:

https://github.com/mozilla/addons-code-manager/blob/365b8958bd517c444f4af1008a0eb3beb211cb4b/src/components/LinterProvider/index.spec.tsx#L107

bobsilverberg: Yes, I was able to get that to work, thanks!

src/components/KeyboardShortcuts/index.tsx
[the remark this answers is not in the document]
bobsilverberg: It doesn't look like the current app supports navigating through messages (called "notes" in the current app) for the compare view, so maybe this isn't something that is particularly important for reviewers. I still think we should support it, but maybe it's not such a big deal if files without changes are skipped.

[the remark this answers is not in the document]
bobsilverberg: I filed #831 for this.

bobsilverberg:
Thanks for the review @kumar303. I have addressed all of the issues you have raised thus far.

src/components/KeyboardShortcuts/index.tsx
```diff
@@ -215,6 +263,7 @@ const mapStateToProps = (
 
   return {
     currentAnchor: location.hash.replace(/^#/, ''),
+    messageUid: typeof messageUid === 'string' ? messageUid : '',
```
kumar303: Oh, `queryString` returns an array if `messageUid` appears more than once on the query string. That's annoying. I guess this will ignore multiple query string values but that's fine.

src/components/CodeView/index.spec.tsx
```diff
@@ -333,6 +341,82 @@ describe(__filename, () => {
...
+        createGlobalExternalMessage({ uid: firstUid }),
+        createGlobalExternalMessage({ uid: secondUid }),
+      ],
+    });
+
+    expect(root.find(`#${id}`)).toHaveLength(1);
+  });
+
+  it('does not render a scrollTo div at the top of global LinterMessage components, if line 0 is not selected', () => {
+    const uid = 'some-global-message-id';
```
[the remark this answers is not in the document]
bobsilverberg: It's not needed. I removed it.

PR #835, opened 2019-06-12
Add a keyboard shortcut to toggle the main side panel

### LINKED ISSUE #588: Provide keyboard shortcuts for hiding and unhiding the file tree

willdurand:
Fixes #588

---

I reworked the supported keys a bit so that it is easier to know which key does what and adding new keyboard shortcuts should be a matter of adding a new entry to the `supportedKeys` map + add a `case` block in the switch.

I added a bunch of missing test cases too.

Screen Shot 2019-06-12 at 11 56 46

### DIFF
--- a/src/components/KeyboardShortcuts/index.spec.tsx
+++ b/src/components/KeyboardShortcuts/index.spec.tsx
@@ -12,6 +12,7 @@ import {
   actions as versionsActions,
   createInternalVersion,
 } from '../../reducers/versions';
+import { actions as fullscreenGridActions } from '../../reducers/fullscreenGrid';
 import {
   CreateKeydownEventParams,
   createContextWithFakeRouter,
@@ -127,6 +128,31 @@ describe(__filename, () => {
     const root = render();
 
     expect(root.find(`.${styles.KeyboardShortcuts}`)).toHaveLength(1);
+  });
+
+  it('does not render a description for alias keys', () => {
+    const root = render();
+
+    const aliasKeys = Object.keys(supportedKeys).filter(
+      (key) => supportedKeys[key] === null,
+    );
+
+    for (const key of aliasKeys) {
+      // The expression below tries to find all react components with $key as
+      // unique child, which is the case of the `kbd` components.
+      expect(root.find({ children: key })).toHaveLength(0);
+    }
+  });
+
+  it('adds a disabled style to the diff keys when there is no compare info', () => {
+    const root = render({ compareInfo: null });
+
+    expect(root.find(`dt.${styles.disabled}`)).toHaveLength(2);
+    expect(root.find(`dd.${styles.disabled}`)).toHaveLength(2);
+
+    // Make sure we added the disabled style to the diff keys.
+    expect(root.find(`dt.${styles.disabled}`).at(0)).toIncludeText('n');
+    expect(root.find(`dt.${styles.disabled}`).at(1)).toIncludeText('p');
   });
 
   it('binds and unbinds keydown to the listener', () => {
@@ -292,6 +318,22 @@ describe(__filename, () => {
     );
   });
 
+  it('dispatches the toggle side panels actions when "h" is pressed', () => {
+    const versionId = 723;
+
+    const store = configureStoreWithFileTree({ versionId });
+    const dispatch = spyOn(store, 'dispatch');
+
+    renderAndTriggerKeyEvent({ key: 'h' }, { store, versionId });
+
+    expect(dispatch).toHaveBeenCalledWith(
+      fullscreenGridActions.toggleMainSidePanel(),
+    );
+    expect(dispatch).toHaveBeenCalledWith(
+      fullscreenGridActions.toggleAltSidePanel(),
+    );
+  });
+
   it('does not listen to keyboard events without a file tree', () => {
     // Configure an empty store, one without a file tree loaded.
     const store = configureStore();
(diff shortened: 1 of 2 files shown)


src/components/KeyboardShortcuts/index.tsx
```diff
@@ -145,6 +155,10 @@ export class KeyboardShortcutsBase extends React.Component<Props> {
             log.warn('Cannot navigate to previous change without diff loaded');
           }
           break;
+        case 'h':
+          dispatch(fullscreenGridActions.toggleMainSidePanel());
+          dispatch(fullscreenGridActions.toggleAltSidePanel());
```
kumar303: For the first iteration, I think `h` should only toggle the file tree panel, not the code overview panel. The old review tool did not toggle the code overview with `h` and I can see how one would want to keep it open even while focusing on the code -- the code overview is really helpful for large files. I say we launch it with that behavior and get some feedback from reviewers. 

Additionally, if you have one side panel collapsed before pressing `h`, the two panels switch off/on which doesn't seem like a helpful feature.

src/components/KeyboardShortcuts/index.tsx
```diff
@@ -162,37 +176,34 @@ export class KeyboardShortcutsBase extends React.Component<Props> {
     _document.removeEventListener('keydown', this.keydownListener);
   }
 
-  render() {
-    const { compareInfo } = this.props;
-    const diffKeysClassName = !compareInfo ? styles.disabled : '';
+  makeClassNameForKey(key: string) {
+    // diff keys
```
kumar303: It may seem trite but I prefer to see comments written as complete sentences. It's not so easy to look at an incomplete sentence and guess what the meaning was.

src/components/KeyboardShortcuts/index.spec.tsx
```diff
@@ -129,6 +130,31 @@ describe(__filename, () => {
...
 
+  it('does not render a description for alias keys', () => {
+    const root = render();
+
+    const aliasKeys = Object.keys(supportedKeys).filter(
+      (key) => supportedKeys[key] === null,
+    );
+
+    for (const key of aliasKeys) {
+      // The expression below tries to find all react components with $key as
```
kumar303: I found this comment confusing. How about this?

"The expression below will look for any JSX rendered with an alias key, such as a `<kbd>` tag."

src/components/KeyboardShortcuts/index.tsx
```diff
@@ -16,10 +16,20 @@ import {
...
-export const supportedKeys = ['k', 'j', 'e', 'o', 'c', 'n', 'p'];
+export const supportedKeys: { [key: string]: string | null } = {
+  k: gettext('Up file'),
+  j: gettext('Down file'),
+  o: gettext('Open all folders'),
+  e: null, // same as 'o'
+  c: gettext('Close all folders'),
+  n: gettext('Next change'),
+  p: gettext('Previous change'),
+  h: gettext('Toggle side panels'),
```
kumar303: As a follow-up to my other suggestion, this would need to change

src/components/KeyboardShortcuts/index.tsx
willdurand: Ah, yep, why not. The code overview did not exist in the old tool, so this keyboard shortcut was actually displaying the code using the full width, hence both side panels collapsed, but we can start with only hiding the left panel.

src/components/KeyboardShortcuts/index.tsx
willdurand: Ok, I changed for "Hide file tree"

willdurand: mmh, I wonder if we don't want to change to something like "Hide left side panel", because if one has a different accordion item expanded, it's not really the "file tree" that will be collapsed 🤔

src/components/KeyboardShortcuts/index.tsx
```diff
@@ -16,10 +16,20 @@ import {
...
-export const supportedKeys = ['k', 'j', 'e', 'o', 'c', 'n', 'p'];
+export const supportedKeys: { [key: string]: string | null } = {
+  k: gettext('Up file'),
+  j: gettext('Down file'),
+  o: gettext('Open all folders'),
+  e: null, // same as 'o'
+  c: gettext('Close all folders'),
+  n: gettext('Next change'),
+  p: gettext('Previous change'),
+  h: gettext('Hide file tree'),
```
kumar303: It doesn't just hide it and I think "tree" is not super clear for those who don't know the implementation. I suggest "Toggle main side panel." This avoids making it specifically about files and it avoids calling it "left panel" since technically it would be on the right in a hypothetical RTL mode. The word main is a bit ambiguous but I think after pressing the key it would become clear.

willdurand: yep, that's what I thought too https://github.com/mozilla/addons-code-manager/pull/835#discussion_r292968681

PR #839, opened 2019-06-12
Give the "selected" linter message a style

### LINKED ISSUE #826: Give the "selected" linter message a style
#824 introduces navigating between linter messages, and when there are multiple global messages, or multiple messages for a given line, only the line of code is selected/scrolled into view, but we still "navigate" between each message. This can be confusing as it seems like nothing is happening when multiple messages for the same line are navigated.

We should change the styling of the current "selected" message so it becomes apparent that we are moving from message to message.

Additional note:
If there are enough messages that they do not all appear on a single screen, then this could still be an issue because the user wouldn't see that the style is changing for any messages currently off the screen. We could address that by changing our strategy and not scrolling the code line into view, but rather scrolling the actual message into view, but I think we decided against that awhile back. I believe we are considering collapsing multiple messages for a line, and if so maybe we can look at how this can best work with that at that time. That's not to say we shouldn't implement the styling for "selected" messages now.

bobsilverberg:
Fixes #826 

The highlighted style adds a border to a message that uses the appropriate colour for the message but darker.

### DIFF
--- a/src/components/CodeView/index.tsx
+++ b/src/components/CodeView/index.tsx
@@ -1,8 +1,6 @@
-import makeClassName from 'classnames';
-import queryString from 'query-string';
 import * as React from 'react';
-import { connect } from 'react-redux';
 import { withRouter, Link, RouteComponentProps } from 'react-router-dom';
+import makeClassName from 'classnames';
 
 import styles from './styles.module.scss';
 import LinterMessage from '../LinterMessage';
@@ -13,9 +11,7 @@ import {
   mapWithDepth,
 } from './utils';
 import refractor from '../../refractor';
-import { getLanguageFromMimeType, messageUidQueryParam } from '../../utils';
-import { ApplicationState } from '../../reducers';
-import { LinterMessage as LinterMessageType } from '../../reducers/linter';
+import { getLanguageFromMimeType } from '../../utils';
 import { Version } from '../../reducers/versions';
 import LinterProvider, { LinterProviderInfo } from '../LinterProvider';
 
@@ -53,11 +49,7 @@ export type PublicProps = {
   version: Version;
 };
 
-type PropsFromState = {
-  messageUid: LinterMessageType['uid'];
-};
-
-type Props = RouteComponentProps & PropsFromState & PublicProps;
+type Props = PublicProps & RouteComponentProps;
 
 type RowProps = {
   className: string;
@@ -71,7 +63,6 @@ export class CodeViewBase extends React.Component<Props> {
       _scrollToSelectedLine = scrollToSelectedLine,
       content,
       location,
-      messageUid,
       mimeType,
     } = this.props;
 
@@ -86,13 +77,7 @@ export class CodeViewBase extends React.Component<Props> {
           )}
         {selectedMessageMap &&
           selectedMessageMap.global.map((message) => {
-            return (
-              <LinterMessage
-                key={message.uid}
-                message={message}
-                highlight={message.uid === messageUid}
-              />
-            );
+            return <LinterMessage key={message.uid} message={message} />;
           })}
         <div className={styles.CodeView}>
           <table className={styles.table}>
@@ -142,7 +127,6 @@ export class CodeViewBase extends React.Component<Props> {
                           {selectedMessageMap.byLine[line].map((msg) => {
                             return (
                               <LinterMessage
-                                highlight={msg.uid === messageUid}
                                 inline
                                 key={msg.uid}
                                 message={msg}
@@ -177,16 +161,4 @@ export class CodeViewBase extends React.Component<Props> {
   }
 }
 
-const mapStateToProps = (
-  state: ApplicationState,
-  ownProps: RouteComponentProps,
-): PropsFromState => {
-  const { location } = ownProps;
-  const messageUid = queryString.parse(location.search)[messageUidQueryParam];
-
-  return {
-    messageUid: typeof messageUid === 'string' ? messageUid : '',
-  };
-};
-
-export default withRouter(connect(mapStateToProps)(CodeViewBase));
+export default withRouter(CodeViewBase);
--- a/src/components/CodeView/index.tsx
+++ b/src/components/CodeView/index.tsx
@@ -1,6 +1,8 @@
-import * as React from 'react';
-import { withRouter, Link, RouteComponentProps } from 'react-router-dom';
 import makeClassName from 'classnames';
+import queryString from 'query-string';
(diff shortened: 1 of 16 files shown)


src/components/LinterMessage/index.tsx
```diff
@@ -58,18 +58,31 @@ const renderDescription = (description: LinterMessage['description']) => {
...
   return (
     <Alert
-      className={makeClassName(styles[variant], { [styles.inline]: inline })}
+      className={makeClassName(
+        styles[variant],
+        {
+          [styles.inline]: inline,
+        },
+        {
+          [styles.highlight]: highlight,
```
kumar303: Hmm. It would be nice if we could just call `element.focus()` and set `:focused` styles but I can't figure out an easy way to do it in `react-bootstrap`. Oh, well.

src/components/CodeView/index.tsx
```diff
@@ -161,4 +177,16 @@ export class CodeViewBase extends React.Component<Props> {
   }
 }
 
-export default withRouter(CodeViewBase);
+const mapStateToProps = (
+  state: ApplicationState,
+  ownProps: RouteComponentProps,
+): PropsFromState => {
+  const { location } = ownProps;
+  const messageUid = queryString.parse(location.search)[messageUidQueryParam];
```
kumar303: Why does this need to happen in `mapStateToProps`? Can't it just happen in `render()`? Introducing `connect()` just for this seems wasteful since it registers update listeners to the Redux store.

src/components/LinterMessage/index.tsx
```diff
@@ -58,18 +58,31 @@ const renderDescription = (description: LinterMessage['description']) => {
   }, []);
 };
 
-type PublicProps = {
+export type PublicProps = {
+  highlight?: boolean;
```
kumar303: Instead of receiving this prop, it would be simpler to wrap this component in `withRouter()`, parse the `props.location` query string, and compare `messageUid` to `message.uid`. It pushes the logic into `LinterMessage` but I think that's fine. The query string is global so it makes sense to access it anywhere. What do you think?

src/components/LinterMessage/styles.module.scss
```diff
@@ -46,3 +46,7 @@ $top-arrow-height: 20px;
 .description {
   margin-bottom: 0;
 }
+
+.highlight {
```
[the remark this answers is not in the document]
bobsilverberg: Nice, thanks!

src/components/LinterMessage/index.tsx
bobsilverberg: Yes, that seems like a better approach. Thanks!

src/components/CodeView/index.tsx
bobsilverberg: This is a good point. I've eliminated the issue by moving the logic into `LinterMessage` as you suggested in another comment.

bobsilverberg:
Thanks for the very helpful comments, @kumar303. This is ready for another look.

src/components/LinterMessage/index.spec.tsx
```diff
@@ -144,6 +161,32 @@ describe(__filename, () => {
     expect(root).toHaveClassName(`.${styles.inline}`);
   });
 
+  it('can be marked as highlighted', () => {
+    const uid = 'some-uid';
+    const location = createFakeLocation({
+      search: queryString.stringify({ [messageUidQueryParam]: uid }),
```
kumar303: side note: it's really cool that you can link someone directly to a linter message.

src/components/LinterMessage/styles.module.scss
```diff
@@ -46,3 +46,30 @@ $top-arrow-height: 20px;
...
```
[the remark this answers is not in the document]
bobsilverberg: Thanks for pointing this out. I'd added 4 new stories. One for each type and one for multiple messages with one highlighted.

bobsilverberg:
Thanks for the reviews @kumar303. This is ready for another look.

stories/LinterMessage.stories.tsx
```diff
@@ -153,20 +226,74 @@ storiesOf('LinterMessage', module).addWithChapters('all variants', {
...
+                      'malformed start tag, at line 1, column 26',
+                    ],
+                  })}
+                />
+              </React.Fragment>,
+            );
+          },
+        },
+        {
+          title: 'Two inline messages stacked with one highlighted',
```
kumar303: It would be nice to add one more story to show a highlighted inline message, one with a triangle pointer at the top. The reason is because in the previous version with a 2px border, the triangle border wasn't affected so there are some specificity issues there. We'll want to make sure the border color of the triangle gets adjusted (it does).

bobsilverberg:
I fixed the blue buttons by passing `undefined` into `initialEntries` in the `MemoryRouter` instead of an array containing a single `location` created via `createFakeLocation`. I'm still not sure why the latter caused the styles to get messed up, but it did, and this fixes it so I think it's an acceptable approach.

stories/LinterMessage.stories.tsx
```diff
@@ -13,160 +17,303 @@ const createMessage = (attributes = {}) => {
   });
 };
 
+const render = (children: JSX.Element) => {
+  return renderWithStoreAndRouter(children);
+};
+
+const renderWithMessageUid = (children: JSX.Element, messageUid: string) => {
+  return renderWithStoreAndRouter(children, undefined, [
```
kumar303: This call is confusing since you're using `undefined` to bypass optional arguments. I think it's time to change the signature to:

```tsx
renderWithStoreAndRouter(children, { location: createFakeLocation('...') });
```

The other calls could do:


```tsx
renderWithStoreAndRouter(children, { store: configureStore() });
```

bobsilverberg: Why not just make them all object parameters? I.e., 
```tsx
renderWithStoreAndRouter({children, location: createFakeLocation('...') });
```

and 
```tsx
renderWithStoreAndRouter({children, store: configureStore() });
```

Or is that what you meant?

bobsilverberg: @kumar303 You didn't reply, but looking at your suggestion I guess you're thinking that we always need to pass `children` and the other arguments are optional, so an object of optional arguments makes sense. I'll go ahead and make that change to this PR.

kumar303: Yes, I was more concerned about the optional arguments. Making them all object params would be fine, though.

PR #844, opened 2019-06-13
Add a target for global messages to DiffView

### LINKED ISSUE #825: DiffView needs a scrollTo target for global messages
In #824, we introduced a scrollTo target for global messages into `CodeView` so that the keyboard shortcuts for navigating linter messages would be able to scroll them into view. Because the code that generates global messages lives in both `CodeView` (for browse) and `DiffView` (for compare), we also need a target in `DiffView`. We might be able to create a shared function which can be used by both.

I'll update this issue to add a link to the code added to `CodeView` by #824 after that code lands.

bobsilverberg:
Fixes #825

### DIFF
--- a/src/components/DiffView/index.tsx
+++ b/src/components/DiffView/index.tsx
@@ -15,6 +15,7 @@ import {
 import { withRouter, RouteComponentProps } from 'react-router-dom';
 import makeClassName from 'classnames';
 
+import { getCodeLineAnchorID } from '../CodeView/utils';
 import LinterProvider, { LinterProviderInfo } from '../LinterProvider';
 import LinterMessage from '../LinterMessage';
 import refractor from '../../refractor';
@@ -225,7 +226,10 @@ export class DiffViewBase extends React.Component<Props> {
         )}
 
         {globalLinterMessages.length ? (
-          <div className={styles.globalLinterMessages}>
+          <div
+            className={styles.globalLinterMessages}
+            id={getCodeLineAnchorID(0)}
+          >
             {globalLinterMessages}
           </div>
         ) : null}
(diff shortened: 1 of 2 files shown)


src/components/DiffView/index.tsx
```diff
@@ -225,7 +226,10 @@ export class DiffViewBase extends React.Component<Props> {
         )}
 
         {globalLinterMessages.length ? (
-          <div className={styles.globalLinterMessages}>
+          <div
+            className={styles.globalLinterMessages}
+            id={getCodeLineAnchorID(0)}
```
kumar303: If I understand correctly, the [`CodeView` counterpart](https://github.com/mozilla/addons-code-manager/blob/0e7150a6e4932b7afecfc67f5c859ed4f1eb0e63/src/components/CodeView/index.tsx#L76) could be rewritten to look similar to the `DiffView` one, i.e. global linter messages nested in a div. If so, then we should introduce a common component:

```jsx
<GlobalLinterMessages className={'...'}>
  {globalLinterMessages}
</GlobalLinterMessages>
```

I'm not so worried about duplicating this one line of code (the `id` assignment) but I'm more worried that one of these could go out of sync since it's not immediately clear what `getCodeLineAnchorID(0)` is doing.

If that sounds reasonable, could you file a follow-up to use a shared component?

kumar303:
I'm going to merge this in and file a new issue for my follow-up recommendation.

src/components/DiffView/index.tsx
kumar303: Moved to https://github.com/mozilla/addons-code-manager/issues/872

bobsilverberg:
Thanks for taking care of this @kumar303.

Issue #672, opened 2019-04-30
Add CodeOverview support to the compare page


kumar303:
A CodeOverview component was added in https://github.com/mozilla/addons-code-manager/issues/523 to provide an overview of the entire file while in browse mode.

Let's add this feature to the compare mode. Since the compare mode shows the diff in context of the *newer* file then showing an overview of the newer file would be helpful. You could, for example, navigate to linter messages more easily.

ioanarusiczki:
Verified on AMO dev and stage - FF67 -Win10 

![code overview in compare](https://user-images.githubusercontent.com/33448286/60183809-a7f13b00-982f-11e9-97b4-58772190ba41.gif)

PR #865, opened 2019-06-21
Add CodeOverview to the compare page

kumar303:
~~⚠️  Depends on https://github.com/mozilla/addons-code-manager/pull/860~~

This fixes https://github.com/mozilla/addons-code-manager/issues/672 by adding a code overview for the newer file (the head version). The overview won't be of the diff but maybe we could do that as a follow-up. It would be harder to make an overview of the diff.

Screenshot 2019-06-20 17 29 23

### DIFF
--- a/src/components/CodeOverview/index.spec.tsx
+++ b/src/components/CodeOverview/index.spec.tsx
@@ -298,9 +298,9 @@ describe(__filename, () => {
       }),
     );
 
-    expect(fakeGetCodeLineAnchor).toHaveBeenCalled();
-    expect(fakeGetCodeLineAnchor.mock.calls[lineIndex]).toEqual([line]);
-    expect(fakeGetCodeLineAnchor.mock.calls[lineIndex - 1]).toEqual([line - 1]);
+    // Make sure subsequent line numbers are passed in order.
+    expect(fakeGetCodeLineAnchor).toHaveBeenNthCalledWith(line, line);
+    expect(fakeGetCodeLineAnchor).toHaveBeenNthCalledWith(line - 1, line - 1);
   });
 
   it('scrolls a linter message into view when clicking its link', () => {
--- a/src/components/CodeOverview/index.spec.tsx
+++ b/src/components/CodeOverview/index.spec.tsx
@@ -276,6 +276,33 @@ describe(__filename, () => {
     expect(link).toHaveProp('title', `Jump to line ${line}`);
   });
 
+  it('renders links for a code line using a custom getCodeLineAnchor', () => {
+    const customAnchor = 'example-anchor';
+    const fakeGetCodeLineAnchor = jest.fn(() => customAnchor);
+
+    const location = createFakeLocation();
+    const root = renderWithFixedHeight({
+      getCodeLineAnchor: fakeGetCodeLineAnchor,
+      content: generateFileLines({ count: 3 }).join('\n'),
+      location,
+    });
+
+    const line = 3;
+    const lineIndex = line - 1;
+    const link = root.find(Link).at(lineIndex);
+
+    expect(link).toHaveProp(
+      'to',
+      expect.objectContaining({
+        hash: customAnchor,
+      }),
+    );
+
+    expect(fakeGetCodeLineAnchor).toHaveBeenCalled();
+    expect(fakeGetCodeLineAnchor.mock.calls[lineIndex]).toEqual([line]);
+    expect(fakeGetCodeLineAnchor.mock.calls[lineIndex - 1]).toEqual([line - 1]);
+  });
+
   it('scrolls a linter message into view when clicking its link', () => {
     const fakeDOMNode = { scrollIntoView: jest.fn() };
     const _document = {
--- a/src/components/CodeOverview/index.tsx
+++ b/src/components/CodeOverview/index.tsx
@@ -24,8 +24,6 @@ import {
   generateLineShapes,
 } from '../CodeLineShapes/utils';
 
-export type GetCodeLineAnchor = (line: number) => string;
-
 export type PublicProps = {
   content: string;
   version: Version;
@@ -39,7 +37,7 @@ export type DefaultProps = {
     removeEventListener: typeof window.removeEventListener;
   };
   createOverviewRef: () => React.RefObject<HTMLDivElement> | null;
-  getCodeLineAnchor: GetCodeLineAnchor;
+  getCodeLineAnchor: typeof _getCodeLineAnchor;
   overviewPadding: number;
   rowTopPadding: number;
   rowHeight: number;
--- a/src/components/CodeOverview/index.tsx
+++ b/src/components/CodeOverview/index.tsx
@@ -5,7 +5,10 @@ import makeClassName from 'classnames';
 import chunk from 'lodash.chunk';
 import debounce from 'lodash.debounce';
 
-import { getCodeLineAnchor, getLines } from '../CodeView/utils';
+import {
+  getCodeLineAnchor as _getCodeLineAnchor,
+  getLines,
+} from '../CodeView/utils';
 import {
   LinterMessage as LinterMessageType,
   findMostSevereType,
@@ -21,6 +24,8 @@ import {
   generateLineShapes,
 } from '../CodeLineShapes/utils';
 
+export type GetCodeLineAnchor = (line: number) => string;
+
 export type PublicProps = {
   content: string;
   version: Version;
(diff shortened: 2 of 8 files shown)


willdurand:
>  security/snyk - package.json (Add-ons Team) — 12984 high severity vulnerable dependency paths 

😂

kumar303:
> Looks good, thanks. I only have one minor comment after I read #865.

@willdurand which comment did you have? You mean about `.bind(map)`, the one we discussed?

src/components/CodeOverview/index.tsx
```diff
@@ -53,6 +59,7 @@ export class CodeOverviewBase extends React.Component<Props, State> {
     _window: window,
     createOverviewRef: () => React.createRef<HTMLDivElement>(),
     // This is the padding of the overview container.
+    getCodeLineAnchor: _getCodeLineAnchor,
```
willdurand: We tend to use underscore variables for the prop names, no? (leaving the default imported names unchanged).

src/components/CodeOverview/index.tsx
```diff
@@ -34,6 +39,7 @@ export type DefaultProps = {
     removeEventListener: typeof window.removeEventListener;
   };
   createOverviewRef: () => React.RefObject<HTMLDivElement> | null;
+  getCodeLineAnchor: GetCodeLineAnchor;
```
willdurand: Can't we use `typeof getCodeLineAnchor` here?

src/pages/Compare/index.tsx
```diff
@@ -138,10 +160,23 @@ export class CompareBase extends React.Component<Props> {
...
     const { baseVersionId, headVersionId } = match.params;
 
+    let getCodeLineAnchor;
+    if (compareInfo && compareInfo.diff) {
+      const map = new ForwardComparisonMap(compareInfo.diff);
+      getCodeLineAnchor = map.createCodeLineAnchorGetter();
+    }
+
     // TODO: set showFileInfo=true when we can:
     // https://github.com/mozilla/addons-code-manager/issues/647
```
willdurand: This patch should allow us to fix this TODO, no? If so, can you leave a comment in the linked issue?

src/components/CodeOverview/index.spec.tsx
```diff
@@ -276,6 +276,33 @@ describe(__filename, () => {
...
+    expect(link).toHaveProp(
+      'to',
+      expect.objectContaining({
+        hash: customAnchor,
+      }),
+    );
+
+    expect(fakeGetCodeLineAnchor).toHaveBeenCalled();
+    expect(fakeGetCodeLineAnchor.mock.calls[lineIndex]).toEqual([line]);
+    expect(fakeGetCodeLineAnchor.mock.calls[lineIndex - 1]).toEqual([line - 1]);
```
willdurand: Are you looking for https://jestjs.io/docs/en/expect.html#tohavebeennthcalledwithnthcall-arg1-arg2- or https://jestjs.io/docs/en/expect.html#tohavebeenlastcalledwitharg1-arg2-? We should try to not direct access `mock.calls`.

src/components/CodeOverview/index.tsx
```diff
@@ -21,6 +24,8 @@ import {
   generateLineShapes,
 } from '../CodeLineShapes/utils';
 
+export type GetCodeLineAnchor = (line: number) => string;
```
willdurand: If this is related to `getCodeLineAnchor`, shouldn't this type be defined in `CodeView/utils.tsx`?

willdurand: Also, what would be the difference between this type and `typeof getCodeLineAnchor`? Or could we define this type as follows:

```
type GetCodeLineAnchor = typeof getCodeLineAnchor;
```

src/pages/Compare/index.spec.tsx
```diff
@@ -324,6 +356,106 @@ describe(__filename, () => {
...
+    store.dispatch(
+      versionsActions.loadDiff({
+        addonId,
+        baseVersionId,
+        headVersionId,
+        version,
+      }),
+    );
+
+    // Simulate the beginning of a fetch.
```
willdurand: This comment does not seem needed as it repeats the name of the action dispatched below it. I'd rather see a comment saying that the action results in the file being loading, if any.

src/pages/Compare/index.tsx
```diff
@@ -159,7 +194,8 @@ export class CompareBase extends React.Component<Props> {
         </Helmet>
         <VersionFileViewer
           compareInfo={compareInfo}
-          file={null}
+          file={compareInfo && compareInfo.diff ? versionFile : null}
```
willdurand: Is there a test case for when `compareInfo` is not defined? I only see a test case for when `compareInfo.diff` is `null` in this patch.

src/reducers/versions.tsx
```diff
@@ -721,6 +726,24 @@ type CreateInternalDiffParams = {
   headVersionId: number;
 };
 
+export const createInternalChangeInfo = (
```
willdurand: Nice change but is this needed for this patch?

src/pages/Compare/index.tsx
kumar303: No, that's separate issue. It's about showing the file information (sha hash, etc)

src/reducers/versions.tsx
kumar303: This part of the nested object was not type checked against `DiffInfo` for some reason but when I moved it to get returned from `createInternalChangeInfo()`, it was. It's not necessary for anything in this patch but it was helping me get everything working so it seems worth keeping.

kumar303:
@willdurand 

>  let me know if you want a second look depending on the UI bug mentioned above.

Yeah, could you file a separate issue? I think it's happening before `VersionFileViewer` renders the column so it's probably not related to this patch other than how nothing was rendered in that column before.

src/pages/Compare/index.tsx
willdurand: Well, yeah... I thought that it was blocked because no version file was passed to `VersionFileViewer`, which is solved in this patch.

src/components/CodeOverview/index.tsx
willdurand: I hardly get the difference. We use underscore in names of injected props all the time and not exclusively for testing purpose, so it'd be nice to keep things consistent. If you insist on having a non-underscore name as a prop name, then I'd suggest to rename `_getCodeLineAnchor` to `defaultGetCodeLineAnchor`.

src/components/CodeOverview/index.spec.tsx
kumar303: Cool, I didn't know that existed. Thanks.

src/components/CodeOverview/index.tsx
kumar303: > ```
> type GetCodeLineAnchor = typeof getCodeLineAnchor;
> ```

That works well. Thanks.

willdurand:
> Yeah, could you file a separate issue? I think it's happening before `VersionFileViewer` renders the column so it's probably not related to this patch other than how nothing was rendered in that column before.

Done: https://github.com/mozilla/addons-code-manager/issues/873

src/pages/Compare/index.tsx
kumar303: 13 tests fail if I remove `compareInfo &&` but, more importantly, there is a type error which points directly to the problem. I think it's fine to rely on the type checking for this.

src/components/CodeOverview/index.tsx
kumar303: > We use underscore in names of injected props all the time and not exclusively for testing purpose

Where do we use an underscore prefixed injected prop for something other than testing? That would be confusing and I'd rather not do that anywhere. I guess we adopted the underscore thing from Python since it makes pseudo-private accessors.

> ...then I'd suggest to rename `_getCodeLineAnchor` to `defaultGetCodeLineAnchor`.

I could do that but since we don't do that elsewhere it would introduce an inconsistency. I prefer this style, anyway, as it was the style of `web-ext` 😛 but I don't want to go against the flow.

kumar303: I switched it to a `default*` style but I just made the name more readable as a sentence: `defaultCodeLineAnchorGetter`.

src/pages/Compare/index.tsx
kumar303: oh, I see! I'll defer it to https://github.com/mozilla/addons-code-manager/issues/647 so that it can get its own QA. We may need to change how we display the file name to make it obvious that it's the "head" file but I'm not sure.

Issue #879, opened 2019-06-25
Store linter messages that apply to the entire add-on


kumar303:
**For QA**: please check for any regressions in the features relating to linter messages

To support https://github.com/mozilla/addons-code-manager/issues/805 we need to first store linter messages that apply to the entire add-on, i.e. messages that apply to all files.

ioanarusiczki:
@kumar303 
Verified on -dev. FF67(Win10)
https://github.com/mozilla/addons-code-manager/issues/887 is still reproducible.
I'll update https://github.com/mozilla/addons-code-manager/issues/402 with the examples I worked with.

PR #880, opened 2019-06-25
Store linter messages that apply to entire add-on

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/879

### DIFF
--- a/src/components/FileTreeNode/index.spec.tsx
+++ b/src/components/FileTreeNode/index.spec.tsx
@@ -656,12 +656,14 @@ describe(__filename, () => {
           file: path,
           // Make this a global message.
           line: null,
+          uid: 'global-uid-example',
         },
         {
           ...fakeExternalLinterMessage,
           file: path,
           // Make this a line message.
           line: 123,
+          uid: 'line-uid-example',
         },
       ] as ExternalLinterMessage[];
       const map = _getMessageMap(messages);
@@ -673,6 +675,21 @@ describe(__filename, () => {
         createInternalMessage(messages[0]),
         createInternalMessage(messages[1]),
       ]);
+    });
+
+    it('does not check universal messages', () => {
+      const messages = [
+        {
+          ...fakeExternalLinterMessage,
+          // Make this a universal message.
+          file: null,
+          line: null,
+          uid: 'universal-uid-example',
+        },
+      ] as ExternalLinterMessage[];
+      const map = _getMessageMap(messages);
+
+      expect(findMostSevereTypeForPath(map, 'src/file-1.js')).toEqual(null);
     });
   });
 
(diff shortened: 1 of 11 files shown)


src/components/FileTreeNode/index.spec.tsx
```diff
@@ -674,6 +676,21 @@ describe(__filename, () => {
         createInternalMessage(messages[1]),
       ]);
     });
+
+    it('does not check universal messages', () => {
```
willdurand: What are "universal" messages? Same as "global" messages?

kumar303: They are bigger than global messages 🌎 🌟  *Global* was the name we used to describe [messages that apply to an entire file](https://github.com/mozilla/addons-code-manager/blob/6b83d050a04694ccfde099005d58f4a4fb832455/src/reducers/linter.tsx#L73). What I'm describing are messages that apply to the entire add-on (all files). I'm open to other naming ideas. I couldn't think of anything as concise as *universal*.

kumar303:
Thanks for taking a look. Let me know if you have any other questions

kumar303:
> I don't like the term `universal` because I find it very confusing but I am unable to suggest anything else... I wish we could use `global` for those messages and change the `global` per file by something else (e.g., `general` so that we would use `map.byPath['some/path'].general` and `map.byPath['some/path'].byLine`).
> 
> r+ anyway because it works and I suppose my proposal is a much bigger change.

Yeah, I understand. Unfortunately, we have quite a bit of nomenclature [invested in the concept](https://github.com/mozilla/addons-code-manager/search?q=global&unscoped_q=global) of *global* messages.

On the bright side, if we need to later introduce messages that apply to all add-ons, we could call them GALACTIC.

bobsilverberg:
> On the bright side, if we need to later introduce messages that apply to all add-ons, we could call them GALACTIC.

It seems to me that GALACTIC should be smaller than Universal, not larger, so we'd either need to change Universal to Galactic at that time, or the "all add-ons" messages could be multiversal. :wink

src/reducers/linter.tsx
```diff
@@ -79,6 +80,7 @@ export type LinterMessageMap = {
     // The 'path' key matches the key of ExternalVersionFile['entries']
     [path: string]: LinterMessagesByPath;
   };
+  // Universal messages apply to the entire add-on
```
willdurand: The period is missing.

kumar303:
I can see that *universal* could be confused with the SSR terminology. How about *overall* to indicate that these are overall messages? i.e. those that apply to the entire add-on.

willdurand:
> I can see that _universal_ could be mistaken for SSR terminology. How about _overall_ to indicate that these are overall messages? i.e. those that apply to the entire add-on.

I feel like `general` would be slightly less confusing at that point.

kumar303:
I considered general but I thought *overall* in the context of linter messages was more specific. However, if general was what came to mind for you (as it did for me) then that's valuable feedback. I can switch it to general.

Issue #884, opened 2019-06-26
Store the current version


kumar303:
For https://github.com/mozilla/addons-code-manager/issues/805 and other things, it would be helpful to fetch the current version from state when it's available.

On the `Browse` page, *current version* would be the one you are browsing.

On the `Compare` page, *current version* would be newer version in the comparison view.

Issue #805, opened 2019-06-03
Display linter messages that apply to the entire add-on


kumar303:
### Describe the problem and steps to reproduce it:

Load https://code.addons-dev.allizom.org/en-US/browse/616915/versions/1688319/?path=background.js which produces the message *Validation generated too many errors/warnings so 2 messages were truncated*

### What happened?

The page does not display the *too many errors* message. 

At the time of filing this issue, an error could be seen in the console. Here is a similar production error:

Unexpectedly received a message not mapped to a file: Validation generated too many errors/warnings so 163 messages were truncated. After addressing the visible messages, you'll be able to see the others.

https://sentry.prod.mozaws.net/operations/addons-code-manager-prod/issues/5884330/?query=is:unresolved

### What did you expect to happen?

The page should display the *too many errors* message.

### Anything else we should know?

Example:

```js
"messages": [
      {
        "tier": 1,
        "type": "warning",
        "id": [
          "validation",
          "messages",
          "truncated"
        ],
        "message": "Validation generated too many errors/warnings so 163 messages were truncated. After addressing the visible messages, you&#39;ll be able to see the others.",
        "description": [],
        "compatibility_type": null
      }
]
```

Complete response: [validation-gen-too-many-warnings.json.zip](https://github.com/mozilla/addons-code-manager/files/3248130/validation-gen-too-many-warnings.json.zip)

kumar303:
I added the security label because the UI fails to show a validation error which is bad

kumar303:
How common is it to see top-level, general linter messages? @EnTeQuAk gave me some detailed info to suggest that it won't be very common.

Relevant code: https://github.com/mozilla/addons-server/blob/master/src/olympia/devhub/file_validation_annotations.py

Top-level messages are primarily related to:

* Legacy add-ons restrictions (these are in 99.9% errors, so won't end up in the code-manager)
* WebExt incompatibilities (As those are upgrades from legacy-addons they won't appear in recent add-ons either)
* Akismet related errors: https://github.com/mozilla/addons-server/blob/master/src/olympia/devhub/file_validation_annotations.py#L107
  * Same, they're errors and won't end up in the code-manager
* Search plugin validation - same, all errors, none will end up in code-manager

In `devhub.tasks` are three more validation errors, one for "api key in XPI detected", another for "unsupported file type" (e.g if the uploaded file isn't a XPI, ZIP etc) and
"Invalid or corrupt add-on file.".

For addons-linter…

[...]

Issue #754, opened 2019-05-15
Self-load versions for components that need to


kumar303:
**For QA**: please check for regressions in *TBA*

Components like `FileTree` and `FileTreeNode` (as of https://github.com/mozilla/addons-code-manager/pull/753) accept a `versionId` parameter but they are not responsible for loading the mapped `version` objects that they depend on. Example:

https://github.com/mozilla/addons-code-manager/blob/8fc513313df7e89f3affa1230f39334e060e1c59/src/components/FileTree/index.tsx#L186-L192

Let's give them the ability to load versions themselves, as needed. 

This feature isn't strictly needed but it will give these components the freedom to live in different component configurations. To put it another way, a developer may move these components to a different component tree and get bitten by a nasty bug: versions will be missing.

We could implement this with a `useEffect` hook (easier!) or a render prop wrapper component (harder!).

kumar303:
Our use of hooks is blocked by Enzyme support: https://github.com/airbnb/enzyme/issues/2011

kumar303:
Hmm. It's possible that the new "current version" storage proposed in https://github.com/mozilla/addons-code-manager/issues/884 will make this a non-issue. In other words, if all components can *read* a current version, they probably don't need to also be responsible for loading it.

PR #885, opened 2019-06-26
Store the current version

kumar303:
This fixes https://github.com/mozilla/addons-code-manager/issues/884 by creating Redux storage for a *current version* and setting it while fetching versions. Other components can use this to access the current version or know when it's still loading.

Historically, we've used prop-drilling to pass a version object to those that need it. This created a problem for `FileTree` because of how React/Redux render caching works. We had to prop drill a `versionId` and create an [awkward mapping](https://github.com/mozilla/addons-code-manager/blob/dc18550b066dbb741cef5712b9a46b9611ecb1cf/src/components/FileTree/index.tsx#L186-L197).

To display general linter messages (https://github.com/mozilla/addons-code-manager/issues/805) and display the add-on name (https://github.com/mozilla/addons-code-manager/issues/756) we need access to the current version in the app header. Prop drilling is not even an option since the app header is higher up in the component tree.

This patches introduces a solution for that but it could also be used for other components that might need it, like `FileTree` and maybe `LinterProvider`, too.

What do you think?

Example usage: https://github.com/mozilla/addons-code-manager/pull/892

### DIFF
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -36,6 +36,7 @@ import reducer, {
   initialState,
   isCompareInfoLoading,
   isFileLoading,
+  selectCurrentVersionInfo,
   viewVersionFile,
 } from './versions';
 import { ROOT_PATH, RelativePathPosition } from './fileTree';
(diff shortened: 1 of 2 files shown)


kumar303:
I have some ideas to make this more robust. Stay tuned...

src/reducers/versions.spec.tsx
```diff
@@ -2837,4 +2880,60 @@ describe(__filename, () => {
...
+    it('returns null when there was a fetching error', () => {
+      const version = { ...fakeVersion, id: 42 };
+      let state;
+      state = reducer(
+        state,
+        actions.abortFetchVersion({ versionId: version.id }),
+      );
+      state = reducer(
+        state,
+        actions.setCurrentVersionId({ versionId: version.id }),
```
bobsilverberg: Instead of calling `setCurrentVersionId` here, which doesn't really reflect what would happen in the lifecycle of the app, could you instead call `loadVersionInfo` first, which would cause `currentVersionId` to be set, and then call `abortFetchVersion`? That seems like a more natural sequence of events to emulate a fetching error.

kumar303: I made the action dispatching more realistic -- good idea. In this case of a real fetching error, `loadVersionInfo` would never get called, though.

kumar303:
> At first I was confused about why we were setting `currentVersionId` before the version was fetched and loaded, but I quickly figured out what was going on. 

Yeah, dispatching this before the version gets loaded is important so I added a comment to explain why.

> I'm not sure this really addresses the issue of `FileTree`, because that component is still going to have to worry about whether a version is loaded or not, but it does seem like a step in the right direction.

I've been thinking more about https://github.com/mozilla/addons-code-manager/issues/754 (w.r.t. FileTree) and I'm not sure that FileTree *needs* to be responsible for loading the current version even though originally I had pointed out that it should.

We have a similar "read only" pattern on addons-frontend with `state.users.currentUser`. Many components read this value but they are not responsible for loading the user.

I'll think about this more. The part that still troubles me is you couldn't render `FileTree` in complete isolation because it won't know how to load the version it depends on.

PR #891, opened 2019-06-27
Create a shared getCodeLineAnchorGetter function

### LINKED ISSUE #889: Create a shared getCodeLineAnchorGetter function
This supports #887 

https://github.com/mozilla/addons-code-manager/commit/9010bcb01320e58a5304905067cb6ef90e6c72b0 introduced a special [`getCodeLineAnchor`](https://github.com/mozilla/addons-code-manager/blob/9010bcb01320e58a5304905067cb6ef90e6c72b0/src/pages/Compare/utils.tsx#L77) function which can be used for the `Compare` page when a `diff` is present. There already existed a [`getCodeLineAnchor`](https://github.com/mozilla/addons-code-manager/blob/9010bcb01320e58a5304905067cb6ef90e6c72b0/src/components/CodeView/utils.tsx#L44) function for use on the `Browse` page.

We need this same dual functionality in the [`goToRelativeMessage`](https://github.com/mozilla/addons-code-manager/blob/9010bcb01320e58a5304905067cb6ef90e6c72b0/src/reducers/fileTree.tsx#L391) function in order to generate a correct `hash` to add to the `location`. I initially just copied the [logic](https://github.com/mozilla/addons-code-manager/blob/7f5dced331ff5cd21cfd9073db1dc3c1b418519f/src/pages/Compare/index.tsx#L174-L178) from the `Compare` page into the `KeyboardShortcuts` component, but it got complicated when trying to write tests, so I figured it would be better to extract the logic out of both `Compare` and `CodeView` and create a shared function, which could potentially be mocked in tests.

I have started working on this. @kumar303 what do you think?

bobsilverberg:
Fixes #889 

The code changes are quite simple, but I found the tests confusing. I'm still not sure if the logic is actually where it needs to be in the tests, and it took me ages to figure out how to mock `ForwardComparisonMap`, which I'm now not even sure if we want/need to mock. I couldn't figure out how to mock it without making it the default export from `/pages/Compare/utils`, and the fact that we have to mock it at the top of the test file also seems strange to me.

I look forward to any and all feedback on this patch, as it's messing with my brain.

### DIFF
--- a/src/pages/Compare/index.spec.tsx
+++ b/src/pages/Compare/index.spec.tsx
@@ -25,7 +25,7 @@ import {
 import DiffView from '../../components/DiffView';
 import Loading from '../../components/Loading';
 import VersionFileViewer from '../../components/VersionFileViewer';
-import { getCodeLineAnchorGetter } from '../../utils';
+import { createCodeLineAnchorGetter } from '../../utils';
 import styles from './styles.module.scss';
 
 import Compare, { CompareBase, PublicProps, mapStateToProps } from '.';
@@ -812,11 +812,13 @@ describe(__filename, () => {
   });
 
   it('configures VersionFileViewer with a file', () => {
+    const addonId = 1;
     const baseVersionId = 1;
     const version = { ...fakeVersionWithDiff, id: baseVersionId + 1 };
     const headVersionId = version.id;
 
-    const { root } = loadDiffAndRender({
+    const { store, root } = loadDiffAndRender({
+      addonId,
       baseVersionId,
       headVersionId,
       version,
@@ -832,13 +834,14 @@ describe(__filename, () => {
       }),
     );
 
-    const diff = createInternalDiff({ baseVersionId, headVersionId, version });
-    if (!diff) {
-      throw new Error('diff was unexpectedly empty');
-    }
-    const getterFromFactory = getCodeLineAnchorGetter({
-      compareInfo: { diff, mimeType: 'test' },
-    });
+    const getterFromFactory = createCodeLineAnchorGetter(
+      getCompareInfo(
+        store.getState().versions,
+        addonId,
+        baseVersionId,
+        headVersionId,
+      ),
+    );
 
     const getCodeLineAnchor = viewer.prop('getCodeLineAnchor');
     if (!getCodeLineAnchor) {
--- a/src/pages/Compare/index.spec.tsx
+++ b/src/pages/Compare/index.spec.tsx
@@ -25,8 +25,8 @@ import {
 import DiffView from '../../components/DiffView';
 import Loading from '../../components/Loading';
 import VersionFileViewer from '../../components/VersionFileViewer';
+import { getCodeLineAnchorGetter } from '../../utils';
 import styles from './styles.module.scss';
-import { ForwardComparisonMap } from './utils';
 
 import Compare, { CompareBase, PublicProps, mapStateToProps } from '.';
 
@@ -836,7 +836,9 @@ describe(__filename, () => {
     if (!diff) {
       throw new Error('diff was unexpectedly empty');
     }
-    const map = new ForwardComparisonMap(diff);
+    const getterFromFactory = getCodeLineAnchorGetter({
+      compareInfo: { diff, mimeType: 'test' },
+    });
 
     const getCodeLineAnchor = viewer.prop('getCodeLineAnchor');
     if (!getCodeLineAnchor) {
@@ -845,7 +847,7 @@ describe(__filename, () => {
 
     // As a sanity check, call the configured getter to see if we get
     // the same result as one returned by a simulated getter.
-    expect(getCodeLineAnchor(1)).toEqual(map.getCodeLineAnchor(1));
+    expect(getCodeLineAnchor(1)).toEqual(getterFromFactory(1));
   });
 
   it('configures VersionFileViewer with a file even for empty diffs', () => {
--- a/src/pages/Compare/index.spec.tsx
+++ b/src/pages/Compare/index.spec.tsx
@@ -834,14 +834,14 @@ describe(__filename, () => {
       }),
     );
 
-    const getterFromFactory = createCodeLineAnchorGetter(
-      getCompareInfo(
+    const getterFromFactory = createCodeLineAnchorGetter({
+      compareInfo: getCompareInfo(
         store.getState().versions,
         addonId,
         baseVersionId,
         headVersionId,
       ),
-    );
+    });
 
     const getCodeLineAnchor = viewer.prop('getCodeLineAnchor');
     if (!getCodeLineAnchor) {
--- a/src/pages/Compare/utils.tsx
+++ b/src/pages/Compare/utils.tsx
@@ -24,7 +24,7 @@ const mergeChanges = (
 
 // This is a map of an old vs. new file comparison, giving preference to the
 // "forward" file, i.e. the newer file.
-export default class ForwardComparisonMap {
+export class ForwardComparisonMap {
   private changeMap: ForwardChangeMap;
 
   constructor(diff: DiffInfo) {
--- a/src/pages/Compare/utils.tsx
+++ b/src/pages/Compare/utils.tsx
@@ -24,7 +24,7 @@ const mergeChanges = (
 
 // This is a map of an old vs. new file comparison, giving preference to the
 // "forward" file, i.e. the newer file.
-export class ForwardComparisonMap {
+export default class ForwardComparisonMap {
   private changeMap: ForwardChangeMap;
 
   constructor(diff: DiffInfo) {
--- a/src/utils.spec.tsx
+++ b/src/utils.spec.tsx
@@ -3,13 +3,31 @@ import queryString from 'query-string';
 
 import {
   formatFilesize,
+  getCodeLineAnchorGetter,
   getLanguageFromMimeType,
   getLocalizedString,
   getPathFromQueryString,
   nl2br,
   sanitizeHTML,
 } from './utils';
-import { createFakeHistory, createFakeLocation } from './test-helpers';
+import {
+  createFakeCompareInfo,
+  createFakeHistory,
+  createFakeLocation,
+} from './test-helpers';
+
+// If we don't create the mock for ForwardComparisonMap here it doesn't work!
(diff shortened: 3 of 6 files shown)


src/utils.spec.tsx
```diff
@@ -3,13 +3,31 @@ import queryString from 'query-string';
...
+  createFakeHistory,
+  createFakeLocation,
+} from './test-helpers';
+
+// If we don't create the mock for ForwardComparisonMap here it doesn't work!
+const mockGetterFromForwardComparisonMap = jest.fn();
+const mockCreateCodeLineAnchorGetter = jest
+  .fn()
+  .mockReturnValue(mockGetterFromForwardComparisonMap);
+jest.mock('./pages/Compare/utils', () => {
```
kumar303: This will be very confusing for the future person who has to debug it. It suggests to me that the class is being instantiated incorrectly or that you need dependency injection or maybe something else.

src/utils.tsx
```diff
@@ -61,3 +65,19 @@ export const getPathFromQueryString = (history: History) => {
 
   return typeof path === 'string' && path.length ? path : null;
 };
+
+type GetCodeLineAnchorGetterParams = {
+  _getCodeLineAnchor?: typeof getCodeLineAnchor;
+  compareInfo?: CompareInfo | null | void;
```
kumar303: It would be a mistake if the caller *omitted* this parameter and doing so would trigger incorrect behavior (without triggering any type errors). I suggest making this non-optional so that the caller is forced to do this:

```tsx
getCodeLineAnchorGetter({
  // This makes it clear that passing an undefined object was intentional.
  compareInfo: undefined
});
```

src/utils.tsx
```diff
@@ -61,3 +65,19 @@ export const getPathFromQueryString = (history: History) => {
...
+};
+
+export const getCodeLineAnchorGetter = ({
```
kumar303: I suggest calling this `createCodeLineAnchorGetter()` so that, at a glance, you could immediately see it's different from `getCodeLineAnchor()`

src/pages/Compare/utils.tsx
```diff
@@ -24,7 +24,7 @@ const mergeChanges = (
 
 // This is a map of an old vs. new file comparison, giving preference to the
 // "forward" file, i.e. the newer file.
-export class ForwardComparisonMap {
+export default class ForwardComparisonMap {
```
kumar303: Is there any reason besides the mocking problem to make this a default export? If so, I suggest changing the module name to `forwardComparisonMap.js` to avoid confusion.

src/utils.spec.tsx
```diff
@@ -188,4 +206,38 @@ describe(__filename, () => {
       expect(getPathFromQueryString(history)).toEqual(null);
     });
   });
+
+  describe('getCodeLineAnchorGetter', () => {
+    it('returns getCodeLineAnchor if compareInfo is null', () => {
+      const _getCodeLineAnchor = jest.fn();
+      expect(
+        getCodeLineAnchorGetter({ _getCodeLineAnchor, compareInfo: null }),
+      ).toEqual(_getCodeLineAnchor);
```
kumar303: Can't this just be `getCodeLineAnchor`? I.E. the real function, not a mock? Since you aren't executing the function, I don't think you need to mock it.

src/utils.spec.tsx
```diff
@@ -188,4 +206,38 @@ describe(__filename, () => {
...
+        }),
+      ).toEqual(_getCodeLineAnchor);
+    });
+
+    it('returns getCodeLineAnchor from ForwardComparisonMap is compareInfo.diff exists', () => {
+      expect(
+        getCodeLineAnchorGetter({
+          compareInfo: createFakeCompareInfo(),
+        }),
+      ).toEqual(mockGetterFromForwardComparisonMap);
```
kumar303: Could you just inject a stub class, like `getCodeLineAnchorGetter({ _ForwardComparisonMap: ... })`, to solve the problem up above?

src/pages/Compare/index.spec.tsx
```diff
@@ -836,7 +836,9 @@ describe(__filename, () => {
     if (!diff) {
       throw new Error('diff was unexpectedly empty');
     }
-    const map = new ForwardComparisonMap(diff);
+    const getterFromFactory = getCodeLineAnchorGetter({
+      compareInfo: { diff, mimeType: 'test' },
```
kumar303: This looks fragile. I think you can just get a `compareInfo` object using the same mapper:

```diff
diff --git a/src/pages/Compare/index.spec.tsx b/src/pages/Compare/index.spec.tsx
index de2254f..f6b4ee9 100644
--- a/src/pages/Compare/index.spec.tsx
+++ b/src/pages/Compare/index.spec.tsx
@@ -812,11 +812,13 @@ describe(__filename, () => {
   });
 
   it('configures VersionFileViewer with a file', () => {
+    const addonId = 1;
     const baseVersionId = 1;
     const version = { ...fakeVersionWithDiff, id: baseVersionId + 1 };
     const headVersionId = version.id;
 
-    const { root } = loadDiffAndRender({
+    const { store, root } = loadDiffAndRender({
+      addonId,
       baseVersionId,
       headVersionId,
       version,
@@ -832,12 +834,13 @@ describe(__filename, () => {
       }),
     );
 
-    const diff = createInternalDiff({ baseVersionId, headVersionId, version });
-    if (!diff) {
-      throw new Error('diff was unexpectedly empty');
-    }
     const getterFromFactory = getCodeLineAnchorGetter({
-      compareInfo: { diff, mimeType: 'test' },
+      compareInfo: getCompareInfo(
+        store.getState().versions,
+        addonId,
+        baseVersionId,
+        headVersionId,
+      ),
     });
 
     const getCodeLineAnchor = viewer.prop('getCodeLineAnchor');
```

bobsilverberg: Nice, thanks!

src/utils.spec.tsx
bobsilverberg: I tried injecting a mock class, but couldn't get it to work. In any case I think it makes sense to remove all the mocking so that should address most of the comments in this PR.

src/pages/Compare/utils.tsx
bobsilverberg: It was just because of the mocking. I'll change it back

src/utils.spec.tsx
bobsilverberg: I've removed all the mocking.

src/utils.tsx
bobsilverberg: Good point. This has been resolved.

bobsilverberg:
I've addressed all of the issues, mostly by getting rid of all the mocking. I'm not totally sure about the last test I added for `createCodeLineAnchorGetter` because I wasn't able to figure out how to get an `equals` assertion to work, but maybe I just don't know the correct incantation. This is ready for another look.

src/utils.spec.tsx
```diff
@@ -188,4 +195,46 @@ describe(__filename, () => {
...
+      const getterFromFactoryWithoutCompareInfo = createCodeLineAnchorGetter(
+        null,
+      );
+
+      // We cannot use `toEqual` for the two getters above because they are
+      // different instances, so we check that they return the same thing.
+      expect(getterFromFactory(1)).toEqual(getterFromMap(1));
+      // And verify that a getter from the factory without compareInfo does not
+      // return the same thing.
+      expect(getterFromFactoryWithoutCompareInfo(1)).not.toEqual(
```
kumar303: I think it would be fine to just check that the getter functions are different. It accomplishes the same thing and is slightly more readable to me.

```diff
diff --git a/src/utils.spec.tsx b/src/utils.spec.tsx
index 2ff167c..f56efa4 100644
--- a/src/utils.spec.tsx
+++ b/src/utils.spec.tsx
@@ -232,9 +232,7 @@ describe(__filename, () => {
       expect(getterFromFactory(1)).toEqual(getterFromMap(1));
       // And verify that a getter from the factory without compareInfo does not
       // return the same thing.
-      expect(getterFromFactoryWithoutCompareInfo(1)).not.toEqual(
-        getterFromMap(1),
-      );
+      expect(getterFromFactory).not.toEqual(getterFromFactoryWithoutCompareInfo);
     });
   });
 });
```

src/utils.spec.tsx
```diff
@@ -188,4 +195,46 @@ describe(__filename, () => {
...
+
+      // Generate a getter without compareInfo to verify that it returns a
+      // different result.
+      const getterFromFactoryWithoutCompareInfo = createCodeLineAnchorGetter(
+        null,
+      );
+
+      // We cannot use `toEqual` for the two getters above because they are
+      // different instances, so we check that they return the same thing.
+      expect(getterFromFactory(1)).toEqual(getterFromMap(1));
```
kumar303: I also tried to create a manual mock of the class object but couldn't figure it out. Oh, well. Your approach here seems fine.

src/utils.tsx
```diff
@@ -61,3 +65,13 @@ export const getPathFromQueryString = (history: History) => {
 
   return typeof path === 'string' && path.length ? path : null;
 };
+
+export const createCodeLineAnchorGetter = (
+  compareInfo: CompareInfo | null | void,
```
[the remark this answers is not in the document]
bobsilverberg: Sure, I can do that. I guess I'm just inclined to not use an object when there's only a single argument, but the fact that it can be `undefined` is valid.

src/utils.spec.tsx
bobsilverberg: I don't think that will work for the same reason I couldn't use an `isEqual`. Two objects from the factory are never equal to each other, so saying that `getterFromFactory` is not equal to `getterFromFactoryWithoutCompareInfo` does not prove anything. I tested this by creating a `getterFromFactory2` using the exact same call that generates `getterFromFactory`, and those two getters fail an `isEqual` assertion.

kumar303: The test I proposed is just to make sure that [this line](https://github.com/mozilla/addons-code-manager/blob/df3b815b5d07b16790696bdb31d4fda8ff96bc4c/src/utils.tsx#L76) never gets hit. I still think that would be a sufficient test here. Either way works, though.

bobsilverberg: Ah, right, I was thinking about it backwards. That makes sense now. I'll change it.

PR #904, opened 2019-07-02
Temporarily ignore lodash prototype pollution

kumar303:
The affected packages aren't handling our incoming JSON so this seems safe to ignore. Can someone double check that? I don't see an upstream patch yet but I set the expiry low so we can resolve this soonish.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -14,4 +14,8 @@ ignore:
     - '*':
         reason: We use http-proxy-middleware in development only and react-scripts does not accept user input to trigger an exploit
         expires: 2019-10-31T00:00:00.000Z
+  SNYK-JS-LODASH-450202:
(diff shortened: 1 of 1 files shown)


willdurand:
I went ahead and merged this so that other PRs can be green again (hopefully)

willdurand:
>  I don't see an upstream patch yet but I set the expiry low so we can resolve this soonish.

Here it is: https://github.com/lodash/lodash/pull/4336

willdurand:
Also: I double checked the use of `_.merge`, `_.mergeWith` and `_.defaultsDeep` in the node modules. Those functions don't appear to be used so there should be no problem with ignoring this issue.

kumar303:
Thanks

PR #921, opened 2019-07-08
Return the anchor for the first diff from getRelativeDiffAnchor when a currentAnchor exists that doesn't point to a diff block

### LINKED ISSUE #914: The keyboard navigation is not responsive after clicking a line from the overview
STR:
1. Load https://code.addons-dev.allizom.org/en-US/compare/616913/versions/1688316...1688317/?path=META-INF%2Fmanifest.mf#D5 and click on the overview on one of the lines from 1 to 4
2. Use `n` to navigate to the next change 

Expected result:
The line with the changes is highlighted. 

Actual result:
The line clicked in the overview stays highlighted at step 1. 

Notes:
Also reproducible with linter messages. 
Reproducible on FF67(Win10). 

Clicking a line from the overview then `n` 
Also reproducible if I click a line after all the changes from the page (using `n` should take me to the next file which contains changes).

stays at line3

bobsilverberg:
Fixes #914 

Note that this is the initial "quick fix", which causes the navigation to always go to the first diff in the file if the currently selected line is not at the top of a diff block. The resolves the error in #914, but is only a partial solution. I will file a follow-up for option 2 as discussed at https://github.com/mozilla/addons-code-manager/issues/914#issuecomment-508554395.

### DIFF
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -2309,18 +2309,24 @@ describe(__filename, () => {
       },
     );
 
-    it('throws an error if the currentAnchor cannot be found', () => {
+    it('returns the first anchor if the currentAnchor cannot be found', () => {
       const diff = createFakeDiffWithChanges([
-        [{ lineNumber: 1, type: 'insert' }],
+        [
+          { lineNumber: 1, type: 'normal' },
+          { lineNumber: 2, type: 'delete' },
+          { lineNumber: 3, type: 'insert' },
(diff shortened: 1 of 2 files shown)


bobsilverberg:
It turns out this is not the approach we want to take, and a "quick fix" to just ignore the keystroke and not throw an error turned out to be less quick, so I'm going to close this and instead devote some time to working on option 2 at https://github.com/mozilla/addons-code-manager/issues/914#issuecomment-508554395 instead.

PR #922, opened 2019-07-08
Allow getRelativeDiffAnchor to work with an anchor that isn't a diff anchor

### LINKED ISSUE #914: The keyboard navigation is not responsive after clicking a line from the overview

bobsilverberg:
Fixes #914 

This enables option 2 from https://github.com/mozilla/addons-code-manager/issues/914#issuecomment-508554395, allowing any line in the file to be selected and navigating to the next/previous closest diff in the file.

### DIFF
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -2309,18 +2309,74 @@ describe(__filename, () => {
       },
     );
 
-    it('throws an error if the currentAnchor cannot be found', () => {
+    it('returns the next anchor in the diff given a non-diff anchor', () => {
       const diff = createFakeDiffWithChanges([
-        [{ lineNumber: 1, type: 'insert' }],
+        [
+          { lineNumber: 1, type: 'delete' },
+          { lineNumber: 2, type: 'insert' },
+          { lineNumber: 3, type: 'normal' },
+          { lineNumber: 4, type: 'insert' },
+        ],
       ]);
-      const currentAnchor = 'D99';
-      expect(() => {
+      expect(
         getRelativeDiffAnchor({
-          currentAnchor,
+          currentAnchor: 'N3',
           diff,
           position: RelativePathPosition.next,
-        });
-      }).toThrow(`Could not locate anchor: ${currentAnchor} in the diff.`);
+        }),
+      ).toEqual('I4');
+    });
+
+    it('returns null if there is no next anchor in the diff given a non-diff anchor', () => {
+      const diff = createFakeDiffWithChanges([
+        [
+          { lineNumber: 1, type: 'delete' },
+          { lineNumber: 2, type: 'insert' },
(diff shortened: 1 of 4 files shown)


bobsilverberg:
> 
> Any idea why?

Yes, my logic is wrong and I didn't have enough data in the tests. I'm working on fixing it up now.

bobsilverberg:
Thanks for catching that @kumar303. I believe it's working now.

src/reducers/versions.tsx
```diff
@@ -549,37 +549,58 @@ export type GetRelativeDiffAnchorParams = {
...
+      }
+      // The currentAnchor is the only anchor in the diff.
+      return null;
     }
 
-    if (newIndex >= 0 && newIndex < anchors.length) {
-      return anchors[newIndex];
+    // We have a currentAnchor, but is doesn't match anything in anchors, so
+    // we need to try to find the correct anchor closest to the currentAnchor.
+    const currentAnchorNumber = extractNumber(currentAnchor);
```
kumar303: Hmm, this seems like it wouldn't work in some edge cases where the normal line number might be higher than the inserted line number. I'd think it would be more robust to make a line-based map of all changes (maybe with [`ForwardComparisonMap`](https://github.com/mozilla/addons-code-manager/blob/d2aa422453107fde413c5a34ec741d5b699bb68d/src/pages/Compare/utils.tsx#L27)) and use a pointer to jump around within that map.

However, I can't find any real world diff that will demonstrate that kind of problem 🤷‍♀ 

Ship it. If we run into a real world problem we could revisit this logic.

src/reducers/versions.spec.tsx
```diff
@@ -2301,26 +2301,85 @@ describe(__filename, () => {
...
+      const diff = createFakeDiffWithChanges([
+        [
+          { lineNumber: 1, type: 'delete' },
+          { lineNumber: 2, type: 'insert' },
+          { lineNumber: 3, type: 'normal' },
+        ],
+      ]);
+      expect(
+        getRelativeDiffAnchor({
+          currentAnchor: 'I2',
```
kumar303: A comment explaining that the delete took precedence over this insert might help to show how this is a non-diff anchor.

Issue #896, opened 2019-07-01
Clicking the overview for cose.sig or mozilla.rsa leads to layout issues in compare view


ioanarusiczki:
STR:
1. I went to 
https://code.addons-dev.allizom.org/en-US/compare/616913/versions/1688316...1688317/?path=META-INF%2Fcose.sig on dev
https://code.addons-dev.allizom.org/en-US/compare/616913/versions/1688316...1688317/?path=META-INF%2Fcose.sig on dev
or 
https://code.addons.allizom.org/en-US/compare/1002795/versions/2474830...2474841/?path=META-INF%2Fmozilla.rsa on stage
2. Click in the left panel somewhere on the overview 
3. Observe the page

Expected result:
Page is displayed.

Actual result:
Page won't reload correctly. Stays blank. 

Notes:
Reproducible with FF67 -Win10
Could be related to https://github.com/mozilla/addons-code-manager/issues/402 

![overview click](https://user-images.githubusercontent.com/33448286/60438069-7eae2180-9c18-11e9-8e34-f25f5db6d2df.gif)

bobsilverberg:
I've done some investigation into this. My observations are:

The contents of the file `cose.sig` is displayed in browse mode: http://localhost:3000/en-US/compare/616913/versions/1688316...1688317/?path=META-INF/cose.sig

![Screenshot 2019-07-02 10 10 38](https://user-images.githubusercontent.com/142755/60519490-ab624780-9cb1-11e9-80f9-be45134e1336.png)

but not in compare mode: http://localhost:3000/en-US/compare/616913/versions/1688316...1688317/?path=META-INF/cose.sig

![Screenshot 2019-07-02 10 11 14](https://user-images.githubusercontent.com/142755/60519533-bfa64480-9cb1-11e9-945e-2a1f7f5b57c6.png)

yet the Code Overview is shown in both cases. 

The problem seems to stem from the fact that we are not getting any `hunks` in the `diff` property returned by the Compare API for a binary file, but we are getting the `content` of a binary file returned by the Browse API.

[...]

PR #923, opened 2019-07-08
Return an empty array from getLines when content is empty

### LINKED ISSUE #909: Display nothing rather than a single blank line for binary files
STR:

1. View any binary file in browse mode, e.g., https://code.addons-dev.allizom.org/en-US/browse/494431/versions/1532144/?path=META-INF%2Fcose.sig.

Expected result:

The file content area is empty.

Actual result:

There is a single blank line (labelled with a `1`) displayed in the content area.

Screenshot:

Screenshot 2019-07-16 11 16 52

bobsilverberg:
Fixes #909 

Before:

Screenshot 2019-07-15 14 31 44

After:

Screenshot 2019-07-15 14 32 17

### DIFF
--- a/src/components/CodeView/index.spec.tsx
+++ b/src/components/CodeView/index.spec.tsx
@@ -170,7 +170,7 @@ describe(__filename, () => {
   it('handles empty content', () => {
     const root = renderWithLinterProvider({ content: '' });
 
-    expect(root.find('.tableBody')).toHaveProp('children', []);
+    expect(root.find(`.${styles.tableBody}`)).toHaveProp('children', []);
   });
 
   it('renders multiple lines of code', () => {
--- a/src/components/CodeView/index.spec.tsx
+++ b/src/components/CodeView/index.spec.tsx
@@ -168,11 +168,9 @@ describe(__filename, () => {
   });
 
   it('handles empty content', () => {
-    const content = '';
-    const mimeType = 'text/css';
-    const root = renderWithLinterProvider({ mimeType, content });
+    const root = renderWithLinterProvider({ content: '' });
 
-    expect(root.find('.language-css')).toHaveProp('children', content);
+    expect(root.find('.tableBody')).toHaveProp('children', []);
   });
 
   it('renders multiple lines of code', () => {
--- a/src/components/CodeView/utils.tsx
+++ b/src/components/CodeView/utils.tsx
@@ -34,7 +34,7 @@ export const mapWithDepth = (depth: number, _mapChild = mapChild) => {
 };
 
 export const getLines = (content: string) => {
-  return content.replace('\r\n', '\n').split('\n');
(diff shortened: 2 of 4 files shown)


bobsilverberg:
> How about waiting until [mozilla/addons-server#11782](https://github.com/mozilla/addons-server/issues/11782) lands first?
> 
> I find the description of #909 confusing and I can still reproduce the bug in #896 so I'm not sure what to look for here.

Sure, that makes sense. Once https://github.com/mozilla/addons-server/issues/11782 lands, the bug will appear at which point one can confirm that this fixes it.

bobsilverberg:
@kumar303 https://github.com/mozilla/addons-server/issues/11782 has landed and I have updated the description for #909, as well as provided before and after screenshots of viewing a binary file. I think this should be good to review now.

src/components/CodeView/utils.tsx
```diff
@@ -34,7 +34,7 @@ export const mapWithDepth = (depth: number, _mapChild = mapChild) => {
 };
 
 export const getLines = (content: string) => {
-  return content.replace('\r\n', '\n').split('\n');
+  return content ? content.replace('\r\n', '\n').split('\n') : [];
```
kumar303: Hmm. I don't think we need this patch anymore. When I load https://code.addons-dev.allizom.org/en-US/compare/616913/versions/1688316...1688317/?path=META-INF%2Fcose.sig (from master), everything looks OK. Am I missing anything?

kumar303: Aha, I see that we do need the patch (for the browse page)

src/components/CodeView/index.spec.tsx
```diff
@@ -168,11 +168,9 @@ describe(__filename, () => {
   });
 
   it('handles empty content', () => {
-    const content = '';
-    const mimeType = 'text/css';
-    const root = renderWithLinterProvider({ mimeType, content });
+    const root = renderWithLinterProvider({ content: '' });
 
-    expect(root.find('.language-css')).toHaveProp('children', content);
+    expect(root.find('.tableBody')).toHaveProp('children', []);
```
kumar303: I guess this works because internally the lib adds this global class name. To make it more robust, this should use the `className` that we're explicitly setting:

```diff
diff --git a/src/components/CodeView/index.spec.tsx b/src/components/CodeView/index.spec.tsx
index 385b029..583de61 100644
--- a/src/components/CodeView/index.spec.tsx
+++ b/src/components/CodeView/index.spec.tsx
@@ -170,7 +170,7 @@ describe(__filename, () => {
   it('handles empty content', () => {
     const root = renderWithLinterProvider({ content: '' });
 
-    expect(root.find('.tableBody')).toHaveProp('children', []);
+    expect(root.find(`.${styles.tableBody}`)).toHaveProp('children', []);
   });
 
   it('renders multiple lines of code', () => {
```

Issue #924, opened 2019-07-08
Building a comparison map is slow for large files


kumar303:
**For QA**: please check for regressions in other compare pages. The fix for this may not make a huge impact -- i.e. the file will still load "slowly."

### Describe the problem and steps to reproduce it:

- Sign in
- Load https://code.addons-dev.allizom.org/en-US/compare/495710/versions/1541610...1688333/
- Click on `package-lock.json` to view it

(Also see https://github.com/mozilla/addons-code-manager/issues/779)

### What happened?

It loads very slowly

### What did you expect to happen?

It should load quickly

### Anything else we should know?

Most of the problem in loading this file (which has ~8,000 lines) is the tokenization for highlighting but there is some low hanging fruit in creating a comparison map. Speeding up the comparison map won't make a huge difference but it's easy to do!

AlexandraMoga:
I did not find regressions in the compare view. 

As it was anticipated, this patch did not fix the slow loading issue,or the browser freeze issue mentioned in #779

PR #925, opened 2019-07-08
Build a comparison map faster for large files

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/924

This doesn't make too big of a dent in the slow file loading problem but I noticed building the comparison map was slow and since it's an easy fix, why not?

Here's how I created the benchmark below
- Made a production build with `start-local-dev`
- Manually [turned off highlighting](https://github.com/mozilla/addons-code-manager/blob/d07fbc30e7571dd8d72b8db449de926fac59c2fc/src/components/DiffView/index.tsx#L202)
- In Chrome, profiled while loading `package-lock.json` from https://code.addons-dev.allizom.org/en-US/compare/495710/versions/1541610...1688333/ 
- Looked at the timing of [`createCodeLineAnchorGetter`](https://github.com/mozilla/addons-code-manager/blob/d07fbc30e7571dd8d72b8db449de926fac59c2fc/src/utils.tsx#L69)

Before (1.79 seconds)

Screenshot 2019-07-08 15 19 40

After (18.17 milleseconds)

Screenshot 2019-07-08 15 23 32

### DIFF
--- a/src/pages/Compare/utils.tsx
+++ b/src/pages/Compare/utils.tsx
@@ -12,11 +12,13 @@ type ForwardChangeMap = {
   [line: string]: ChangeInfo[];
 };
 
-const mergeChanges = (
+const mergeChangesInPlace = (
   changeMap: ForwardChangeMap,
   line: string,
   change: ChangeInfo,
 ) => {
+  // This destructively merges changes into each line's ChangeInfo array
+  // when it exists.
   const lineChanges = changeMap[line] || [];
   lineChanges.push(change);
   return lineChanges;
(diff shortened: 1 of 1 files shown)


src/pages/Compare/utils.tsx
```diff
@@ -12,11 +12,13 @@ type ForwardChangeMap = {
   [line: string]: ChangeInfo[];
 };
 
-const mergeChanges = (
+const mergeChangesInPlace = (
```
bobsilverberg: I'm trying to understand how this is "merging". It seems like it's just appending a change to a array of changes. This is probably fine in terms of what it does, but the name is confusing to me, as I assumed merging meant possibly dealing with duplicates as well.

kumar303:
> Looks good to me, I did not reproduce the benchmark but nothing is broken.

I ran it again and got the same results. I updated the PR to remove the manual editing step and also to show the exact commits each was built from.

Issue #926, opened 2019-07-08
Building DiffView widgets is slow for large files


kumar303:
**For QA**: please check for regressions in other compare pages. The fix for this may not make a huge impact -- i.e. the file will still load "slowly."

### Describe the problem and steps to reproduce it:

- Sign in
- Load https://code.addons-dev.allizom.org/en-US/compare/495710/versions/1541610...1688333/
- Click on `package-lock.json` to view it

(Also see https://github.com/mozilla/addons-code-manager/issues/779)

### What happened?

It loads very slowly

### What did you expect to happen?

It should load quickly

### Anything else we should know?

Part of rendering a diff on a compare page involves building a map of "widgets," like inserting an inline linter message, for example. This takes a significant amount of time on large files (those with many lines).

AlexandraMoga:
No regressions found in the compare view.
The slowness is still persisting though, but this was anticipated.

PR #927, opened 2019-07-08
Speed up getWidgets() in DiffView for large files

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/926

A significant amount of time is spend in `getWidgets()` for large files.

Here's how I created the benchmark below
- Made a production build with `start-local-dev`
- In Chrome, profiled while loading `package-lock.json` from https://code.addons-dev.allizom.org/en-US/compare/495710/versions/1541610...1688333/ 
- Looked at the timing of `getWidgets()`

Before (7.13 seconds)

Screenshot 2019-07-08 15 45 47

After (6.12 milleseconds)

Screenshot 2019-07-08 15 49 01

### DIFF
--- a/src/components/DiffView/index.tsx
+++ b/src/components/DiffView/index.tsx
@@ -9,6 +9,7 @@ import {
   Hunk,
   Hunks,
   HunkInfo,
+  WidgetMap,
   getChangeKey,
   tokenize,
 } from 'react-diff-view';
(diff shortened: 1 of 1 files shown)


kumar303:
I think the previous URL was incorrect in the benchmark, so I updated it with a different diff URL and updated the numbers.

PR #934, opened 2019-07-09
Use download_url to display images rather than the file content

### LINKED ISSUE #858: Use download_url to display images rather than the file content
This is a spin-off from #856.

In order to allow the API to not encode and send the content of large image files, we should use the `download_url` of the file to display the image in the browser. This may require an update to our CSP rules.

For QA:

Please verify that all image types that should be viewable in Firefox are still working and displaying images.

bobsilverberg:
Fixes #858 

Note that this is not actually working. There are a couple of issues, one with CSP and the other with auth (it seems).

STR:

With this patch in place (minus the temporary changes to CSP that I added to `/server/index.tsx), visit  http://localhost:3000/en-US/browse/494431/versions/1532144/?path=icons%2Fabp-16-notification-critical.png.
- Note that the `src` of the `img` tag is `/en-US/reviewers/download-git-file/1532144/icons/abp-16-notification-critical.png/`. This is generated via `makeApiURL({ url: file.downloadURL })`, which we also use to display the download link in the Information panel.
I get a number of CSP warnings on this page as follows:
- The first warning is for `https://reviewers.addons-dev.allizom.org/en-US/reviewers/download-git-file/1532144/icons/abp-16-notification-critical.png/`, which I believe is where the actual `src` that is set redirects to.
- If I whitelist `https://reviewers.addons-dev.allizom.org` for `img-src`, then I get a warning about `https://oauth-stable.dev.lcip.org/v1/authorization?client_id=...&redirect_url=https%3A%2F%2Faddons-dev.allizom.org%2Fapi%2Fv3%2Faccounts%2Fauthenticate%2F&scope=profile&state=...&action=signin`.
- If I also whitelist `https://oauth-stable.dev.lcip.org` for `img-src`, then I get a warning about `https://stable.dev.lcip.org/authorization?client_id=...&redirect_url=https%3A%2F%2Faddons-dev.allizom.org%2Fapi%2Fv3%2Faccounts%2Fauthenticate%2F&scope=profile&state=...&action=signin`.
- If I also whitelist `https://stable.dev.lcip.org` for `img-src`, then I get no more CSP warnings, but a broken `img` still appears.
- I have been unable to do anything to make an actual image appear.

Note that if I use `file.downloadURL` directly for `src` rather than `makeApiURL({ url: file.downloadURL })`, I get an initial CSP warning about `https://addons-dev.allizom.org/en-US/reviewers/download-git-file/1532144/icons/abp-16-notification-critical.png/`, and if I whitelist `https://addons-dev.allizom.org` for `img-src` then after that I get the same sequence of 3 CSP warnings, so I'm not sure whether we need to use `makeApiURL({ url: file.downloadURL })` or not, but in either case I cannot get an image to appear.

I'm not sure why we're seeing those requests for `https://*.dev.lcip.org`, but it looks like some sort of auth is being attempted, which might explain why we never see an image.

@kumar303 @willdurand If either of you have any suggestions about what might be going on here, and what we should do about it, I'd appreciate hearing them.


src/server/index.tsx
```diff
@@ -123,7 +123,11 @@ export const createServer = ({
     formAction: ["'none'"],
     frameAncestors: ["'none'"],
     frameSrc: ["'none'"],
-    imgSrc: [staticSrc, 'data:'],
+    imgSrc: [
+      staticSrc,
+      'https://reviewers.addons-dev.allizom.org',
```
bobsilverberg: Note that these are just temporary changes, left here to illustrate where I got to in terms of struggling with CSP. I understand that we would never hardcode values like this.

willdurand:
So, `makeApiURL` is used to change the URL in dev so that it is served by our server (_via_ a proxy). We should have it to avoid authentication issues, in theory.

IIRC we already allow download URLs in the proxy so it should be possible to retrieve files in dev and prod. It's not clear to me why there are CSP errors in dev given that we explicitly whitelist `'self'`.

I'll take a closer look.

willdurand:
@bobsilverberg I investigated and here is a patch that works (on top of this PR):

```diff
diff --git a/.env b/.env
index 91085c0..9b6f388 100644
--- a/.env
+++ b/.env
@@ -21,3 +21,6 @@ PUBLIC_URL=https://addons-code-manager.cdn.mozilla.net
 
 # When true, allow the app to trigger errors that will be reported to Sentry.
 REACT_APP_UNSAFE_ERROR_SIMULATION=false
+
+# This is the reviewers host for images.
+REACT_APP_REVIEWERS_HOST=https://reviewers.addons.mozilla.org
diff --git a/.env.dev b/.env.dev
index a0a778e..df2b86a 100644
--- a/.env.dev
+++ b/.env.dev
@@ -3,3 +3,5 @@
 REACT_APP_API_HOST=https://addons-dev.allizom.org
 # CDN URL
 PUBLIC_URL=https://addons-code-manager-dev-cdn.allizom.org
+# This is the reviewers host for images.
+REACT_APP_REVIEWERS_HOST=https://reviewers.addons-dev.allizom.org
diff --git a/.env.stage b/.env.stage
index b82f273..a50e8b7 100644
--- a/.env.stage
+++ b/.env.stage
@@ -3,3 +3,5 @@
 REACT_APP_API_HOST=https://addons.allizom.org
 # CDN URL
 PUBLIC_URL=https://addons-code-manager-stage-cdn.allizom.org
+# This is the reviewers host for images.
+REACT_APP_REVIEWERS_HOST=https://reviewers.addons.allizom.org
diff --git a/src/pages/Browse/index.tsx b/src/pages/Browse/index.tsx
index 93a075c..fc96536 100644
--- a/src/pages/Browse/index.tsx
+++ b/src/pages/Browse/index.tsx
@@ -138,7 +138,6 @@ export class BrowseBase extends React.Component<Props> {
       return (
         <div className={styles.Image}>
           <img
-            alt=""
             src={makeApiURL({ url: file.downloadURL })}
             className={styles.responsive}
           />
diff --git a/src/server/index.tsx b/src/server/index.tsx
index 1a68a4f..867ee09 100644
--- a/src/server/index.tsx
+++ b/src/server/index.tsx
@@ -25,6 +25,7 @@ export type ServerEnvVars = {
   REACT_APP_AUTHENTICATION_COOKIE: string;
   REACT_APP_AUTH_TOKEN_PLACEHOLDER: string;
   REACT_APP_CRA_PORT: number;
+  REACT_APP_REVIEWERS_HOST: string;
   REACT_APP_SENTRY_DSN: string;
   REACT_APP_USE_INSECURE_PROXY: string;
   SERVER_HOST: string;
@@ -125,8 +126,7 @@ export const createServer = ({
     frameSrc: ["'none'"],
     imgSrc: [
       staticSrc,
-      'https://reviewers.addons-dev.allizom.org',
-      'https://*.dev.lcip.org',
+      `${env.REACT_APP_REVIEWERS_HOST}/en-US/reviewers/download-git-file/`,
     ],
     manifestSrc: ["'none'"],
     mediaSrc: ["'none'"],
@@ -173,33 +173,37 @@ export const createServer = ({
     }
 
     app.use(
-      proxy(
-        [
-          '/api/**',
-          '/**/validation.json',
-          '/**/reviewers/download-git-file/**/',
-        ],
-        {
-          target: env.REACT_APP_API_HOST,
-          autoRewrite: true,
-          changeOrigin: true,
-          cookieDomainRewrite: '',
-          protocolRewrite: 'http',
-          secure: false,
-          onProxyRes: (proxyResponse: http.IncomingMessage) => {
-            // We make cookies unsecure because local development uses http and
-            // not https. Without this change, the server code would not be able
-            // to read the cookie that stores the authentication token.
-            if (proxyResponse.headers['set-cookie']) {
-              const cookies = proxyResponse.headers['set-cookie'].map(
-                (cookie: string) => cookie.replace(/;\s*?(Secure)/i, ''),
-              );
-              // eslint-disable-next-line no-param-reassign
-              proxyResponse.headers['set-cookie'] = cookies;
-            }
-          },
+      proxy(['/api/**', '/**/validation.json'], {
+        target: env.REACT_APP_API_HOST,
+        autoRewrite: true,
+        changeOrigin: true,
+        cookieDomainRewrite: '',
+        protocolRewrite: 'http',
+        secure: false,
+        onProxyRes: (proxyResponse: http.IncomingMessage) => {
+          // We make cookies unsecure because local development uses http and
+          // not https. Without this change, the server code would not be able
+          // to read the cookie that stores the authentication token.
+          if (proxyResponse.headers['set-cookie']) {
+            const cookies = proxyResponse.headers['set-cookie'].map(
+              (cookie: string) => cookie.replace(/;\s*?(Secure)/i, ''),
+            );
+            // eslint-disable-next-line no-param-reassign
+            proxyResponse.headers['set-cookie'] = cookies;
+          }
         },
-      ),
+      }),
+    );
+
+    app.use(
+      proxy(['/**/reviewers/download-git-file/**/'], {
+        target: env.REACT_APP_REVIEWERS_HOST,
+        autoRewrite: true,
+        changeOrigin: true,
+        cookieDomainRewrite: '',
+        protocolRewrite: 'http',
+        secure: false,
+      }),
     );
   }
 
```

It needs some more work but you get the idea hopefully :)

⚠️ **Important:** you have to write a `makeReviewersURL` that is similar to `makeApiURL` (but simpler). All we need to do is removing the host in local dev. We should not be reusing `makeApiURL`.

bobsilverberg:
Thanks for the advice, @willdurand. This is now ready to be reviewed.

src/pages/Browse/index.tsx
```diff
@@ -133,8 +134,16 @@ export class BrowseBase extends React.Component<Props> {
     if (!file || !version) {
       return <Loading message={gettext('Loading content...')} />;
     }
-    if (file.type === 'image') {
-      return <ImageView content={file.content} mimeType={file.mimeType} />;
+    if (file.type === 'image' && file.downloadURL) {
+      return (
+        <div className={styles.Image}>
+          <img
+            alt=""
```
willdurand: I believe this is needed because of ESLint. Could you set it to `null` so that it does not appear in the generated HTML?

src/server/index.spec.tsx
```diff
@@ -303,17 +304,25 @@ describe(__filename, () => {
       });
 
       it('tightens style-src, script-src and img-src if CDN URL is unset', async () => {
+        const fakeEnv = {
+          ...prodEnv,
+          REACT_APP_REVIEWERS_HOST: 'https://reviewers.addons-dev.allizom.org',
```
willdurand: Why not adding this env variable to the `prodEnv` object?

src/server/index.spec.tsx
```diff
@@ -376,29 +394,38 @@ describe(__filename, () => {
...
             console.error('proxy error', error);
           });
+
+          // This Express app is used to simulate the reviewers service.
+          reviewersApp = express();
+          reviewersApp.get('*', (req, res) => {
+            res.send(reviewersResponseBody);
+          });
+
+          fakeReviewersServer = reviewersApp.listen(reviewersPort);
```
willdurand: We should probably add a listener on `error` like it's done for the `fakeApiServer`.

src/utils.spec.tsx
```diff
@@ -244,6 +245,44 @@ describe(__filename, () => {
       );
     });
   });
+
+  describe('makeReviewersURL', () => {
+    it.each(['apiHost', 'reviewersHost'])(
```
willdurand: Why would `apiHost` be involved in `makeReviewersURL`?

src/utils.tsx
```diff
@@ -85,3 +85,24 @@ export const extractNumber = (text: string): number | null => {
...
 };
+
+type MakReviewersURLParams = {
+  apiHost?: string | null;
+  reviewersHost?: string | null;
+  url: string;
+  useInsecureProxy?: boolean;
+};
+
+export const makeReviewersURL = ({
```
willdurand: That function is confusing. I guess it is because the API returns a download URL with the API host instead of the reviewers host?

I would suggest to not do that because we should not involve the `apiHost` in this function I believe. I think we should simply make the download URL relative when `useInsecureProxy` is set to `true` and prepend this relative URL with the `reviewersHost` when `useInsecureProxy` is `false`.

That seems simpler and more future-proof (if, for instance, the API starts to return download URLs with the reviewers host, would this function continue to do its job right?).

WDYT?

src/pages/Browse/index.tsx
bobsilverberg: Yes, that's why it's needed. Sure, I can set it to `null`.

src/server/index.spec.tsx
bobsilverberg: Good idea!

src/server/index.spec.tsx
bobsilverberg: Why do we need one? "Just in case"?

src/utils.spec.tsx
bobsilverberg: Because `makeReviewersURL` replaces `apiHost` with `reviewersHost`, but I see your suggestion below to change that, and changing that will remove the need to worry about `apiHost` for the tests.

src/utils.tsx
bobsilverberg: Yes, that sounds like a good idea.

src/server/index.spec.tsx
willdurand: yep, in case the server fails to start for example.

src/pages/Browse/index.tsx
bobsilverberg: Nope, trying to set it to `null` still yields the ESLint error. It's either `""` or add a comment to disable the error.

bobsilverberg:
Thanks for the review @willdurand. This is now ready for another look.

src/utils.spec.tsx
```diff
@@ -244,6 +245,40 @@ describe(__filename, () => {
...
+
+      expect(
+        makeReviewersURL({
+          reviewersHost,
+          url,
+          useInsecureProxy: false,
+        }),
+      ).toEqual(expectedUrl);
+    });
+  });
```
willdurand: Should we also consider the query string?

bobsilverberg: I added a temporary test which did include a querystring, just to make sure it all works. I can add that back in as a permanent test.

willdurand: If the test case passes, sure that'd be useful.

PR #936, opened 2019-07-09
Converts common border styles to a Sass mixin

### LINKED ISSUE #910: Convert common border styles to a Sass mixin
**For QA**: Please check for regressions in the visual styling of all panel borders

Code manager uses these two common css rules for all borders:

```css
.example {
  border: 1px solid $light-gray;
}
```

Let's move that to a [Sass mixin](https://sass-lang.com/documentation/at-rules/mixin) so that it can be replaced by:

```css
@include border-style();
```

Here is a real world example of [defining a mixin](https://github.com/mozilla/addons-frontend/blob/0541478615fab000c50fa703271e378d8fb7cfcb/src/core/css/inc/mixins.scss#L3) and [using it](https://github.com/mozilla/addons-frontend/blob/e1d2c4001131c9105ae349796558fce60124e09a/src/amo/pages/CategoriesPage/styles.scss#L4).

championshuttler:
Fixes #910

### DIFF
--- a/src/components/Button/styles.module.scss
+++ b/src/components/Button/styles.module.scss
@@ -1,8 +1,9 @@
 @import '../../scss/variables';
+@import '../../scss/mixins';
 
 .button {
   background-color: $light-gray;
-  text-algin: center;
+  text-align: center;
   overflow: hidden;
   padding: $default-padding;
   width: 100%;
@@ -15,8 +16,7 @@
   &:global(.btn-primary:not(:disabled):not(.disabled):active),
   &:hover {
     color: $black;
-    border-radius: $border-radius;
-    border: 1px solid $light-gray;
+    @include border-style();
   }
 
   &:active,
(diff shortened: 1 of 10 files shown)


src/components/Button/styles.module.scss
```diff
@@ -1,8 +1,9 @@
 @import '../../scss/variables';
+@import '../../scss/mixins';
 
 .button {
   background-color: $light-gray;
-  text-algin: center;
+  text-align: center;
```
championshuttler: How was it working before or was not in use 😛

willdurand: good catch!

src/components/FileTree/styles.module.scss
```diff
@@ -7,8 +8,7 @@
 }
 
 .treeShell {
-  border-radius: $border-radius;
-  border: 1px solid $light-gray;
+  @include border-style();
```
willdurand: can you add an empty line below the `@include` block and the regular CSS rules?

Issue #942, opened 2019-07-11
Speed up large diffs by disabling syntax highlighting


kumar303:
### Describe the problem and steps to reproduce it:

- Sign in
- Load a diff that has very long lines (such as a minified library): https://code.addons-dev.allizom.org/en-US/compare/502955/versions/1541798...1541798/?path=popup%2Fpopup.js
- Or load a diff that has many lines: https://code.addons-dev.allizom.org/en-US/compare/495710/versions/1541610...1688333/?path=package-lock.json

### What happened?

The diff loads too slowly and sometimes freezes the browser

### What did you expect to happen?

The diff should load quickly without syntax highlighting

### Anything else we should know?

Turning off syntax highlighting will be a temporary measure until we can speed up the tokenization process (the slow part of highlighting): https://github.com/mozilla/addons-code-manager/issues/928

kumar303:
For QA: please let me know if you still find any slow loading diffs after this.

AlexandraMoga:
@kumar303 I have a few examples of slow loading diffs on -dev:

1. https://code.addons-dev.allizom.org/en-US/compare/385621/versions/1655456...1688374/?path=public%2Fjs%2Fbackground.js
2. https://code.addons-dev.allizom.org/en-US/compare/502868/versions/1541685...1688187/?path=src%2Fjs%2FGrammarly-bg.js
3. https://code.addons-dev.allizom.org/en-US/compare/502868/versions/1541685...1688187/?path=src%2Fjs%2Feditor-popup.js

It is worth noting that all these diffs have ~20k lines

I have another example of an add-on with many contents for which navigating throughout all the files is quite slow: https://code.addons-dev.allizom.org/en-US/compare/1865/versions/1529553...1688304/

On the plus side, the browser no longer freezes when I'm trying to open either of these files.

kumar303:
Thanks. I added https://github.com/mozilla/addons-code-manager/issues/946 to track the long file diff problem.

kumar303:
> I have another example of an add-on with many contents for which navigating throughout all the files is quite slow: https://code.addons-dev.allizom.org/en-US/compare/1865/versions/1529553...1688304/

Oof, good find. This is an API issue so I filed https://github.com/mozilla/addons-server/issues/11838

AlexandraMoga:
marking this as verified fixed then, since the issues discovered are all filed as followups.

PR #944, opened 2019-07-11
Disable highlighting for long or wide diffs

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/942

~~I chose not to add any UI feedback about the disabled highlighting only because this is probably a temporary feature.~~ (there is UI feedback now)

Screenshot 2019-07-11 12 01 09
Screenshot 2019-07-11 12 01 40

### DIFF
--- a/src/components/DiffView/index.spec.tsx
+++ b/src/components/DiffView/index.spec.tsx
@@ -618,7 +618,7 @@ describe(__filename, () => {
     expect(getWidgetNodes(widgets).length).toEqual(0);
   });
 
-  it('generates and passes in tokens for a diff that can be highlighted', () => {
+  it('enables syntax highlighting for diffs when possible', () => {
     const _tokenize = jest.fn(tokenize);
     const root = renderWithLinterProvider({
       _diffCanBeHighlighted: jest.fn(() => true),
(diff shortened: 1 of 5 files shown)


willdurand:
> I chose not to add any UI feedback about the disabled highlighting only because this is probably a temporary feature.

Do you think that we'll always be able to highlight code with a web worker? I am not sure and I think a simple message somewhere would be very useful.

kumar303:
ok, I can add a UI message. I guess maybe we'll always disable highlighting for minified diffs but I'm not sure.

src/components/DiffView/index.spec.tsx
```diff
@@ -572,6 +618,44 @@ describe(__filename, () => {
     expect(getWidgetNodes(widgets).length).toEqual(0);
   });
 
+  it('enables syntax highlighting for diffs when possible', () => {
+    const _tokenize = jest.fn(tokenize);
+    const root = renderWithLinterProvider({
+      _diffCanBeHighlighted: jest.fn(() => true),
```
willdurand: that or `jest.fn().mockReturnValue(true)`

kumar303: It hurts my fingers to type all that 😆 jk

Issue #949, opened 2019-07-16
Function for adding params to a query string


kumar303:
To fix https://github.com/mozilla/addons-code-manager/issues/946 I need a function that takes an existing `location` and adds (or overrides) query params.

PR #950, opened 2019-07-16
Function for adding params to a query string

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/949

This will be used as the *Show original diff* link in https://github.com/mozilla/addons-code-manager/pull/951 which will reload the page with a new query param.


kumar303:
@willdurand good idea. I found a couple -- could you take another look?

Issue #959, opened 2019-07-18
Add a FadableContent component to convey trimmed code


kumar303:
Both the `CodeView` (https://github.com/mozilla/addons-code-manager/issues/947) and `DiffView` (https://github.com/mozilla/addons-code-manager/issues/946) need to render a state where code has been trimmed for performance. Let's add a shared component they can use to fade out the content.

PR #960, opened 2019-07-18
Add FadableContent to convey trimmed code

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/959

Example from the storybook:

Screenshot 2019-07-18 11 33 06

### DIFF
--- a/src/components/FadableContent/styles.module.scss
+++ b/src/components/FadableContent/styles.module.scss
@@ -1,0 +1,16 @@
+@import '../../scss/variables';
+
+.shell {
+  position: relative;
+
+  &::after {
+    // This uses rgba() instead of "transparent" because Safari
+    // doesn't handle that keyword well.
+    background-image: linear-gradient(rgba(255, 255, 255, 0), $white);
+    bottom: 0;
+    content: '';
+    height: 50vh;
+    position: absolute;
+    width: 100%;
+  }
+}
--- a/src/components/FadableContent/index.spec.tsx
+++ b/src/components/FadableContent/index.spec.tsx
@@ -1,0 +1,35 @@
+import { shallow } from 'enzyme';
+import * as React from 'react';
+
+import styles from './styles.module.scss';
+
+import FadableContent, { PublicProps } from '.';
+
+describe(__filename, () => {
+  const render = (props: PublicProps) => {
+    return shallow(<FadableContent {...props} />);
+  };
+
+  it('returns children without a shell when fade=false', () => {
+    const exampleClass = 'example';
+    const root = render({
+      fade: false,
+      children: <div className={exampleClass} />,
+    });
+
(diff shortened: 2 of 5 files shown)


kumar303:
I noticed a bug here that I need to fix. Stay tuned...

src/components/FadableContent/styles.module.scss
```diff
@@ -0,0 +1,14 @@
+@import '../../scss/variables';
+
+.shell {
+  position: relative;
+
+  &::after {
+    background-image: linear-gradient(rgba(255, 255, 255, 0), $white);
```
willdurand: I think using `transparent` would be a little more explicit than `rgba(...)`.

src/components/FadableContent/styles.module.scss
```diff
@@ -0,0 +1,14 @@
...
+    bottom: 0;
+    content: '';
+    height: 400px;
```
willdurand: Why is this enforced in this component? Could we configure its value somehow?

src/components/FadableContent/styles.module.scss
kumar303: Safari handles "transparent" in a wacko way. See the first answer here: https://stackoverflow.com/questions/38391457/linear-gradient-to-transparent-bug-in-latest-safari . That's really the only reason I avoided it. I guess Safari doesn't matter? I always like to aim for cross-compat when possible.

src/components/FadableContent/styles.module.scss
kumar303: I don't think anything would need to configure it but if that day comes we could easily add it to the component props.

src/components/FadableContent/styles.module.scss
kumar303: This is what using `transparent` looks like in today's Safari on macOS Mojave

Screenshot 2019-07-19 10 12 32

kumar303: I can add a comment to the css about it

willdurand: oh mmh, ok then let's keep `rgba` and document it.

src/components/FadableContent/styles.module.scss
willdurand: Why is it set to `400px` only then? Looks ridiculously small on my screen for example. Usually, we tend to consider the height of the window, no?

kumar303: Hmm. Could you send a screenshot? `400px` looked fine when viewing actual truncated code in `DiffView` or `CodeView`. I could switch it to `vh`, I suppose.

kumar303: I switched it to `50vh`

kumar303:
Thanks for the review

Issue #962, opened 2019-07-18
Add a SlowPageAlert to control optimizations


kumar303:
Both the `CodeView` (https://github.com/mozilla/addons-code-manager/issues/947) and `DiffView` (https://github.com/mozilla/addons-code-manager/issues/946) need to show an alert when the contents have been altered for performance reasons. This alert should let them override the alteration.

PR #963, opened 2019-07-18
Add a SlowPageAlert to control optimizations

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/962

Example:

Screenshot 2019-07-19 16 16 16

Also see https://github.com/mozilla/addons-code-manager/pull/951 for more realistic usage.

### DIFF
--- a/src/components/SlowPageAlert/index.tsx
+++ b/src/components/SlowPageAlert/index.tsx
@@ -1,0 +1,43 @@
+import { Location } from 'history';
+import * as React from 'react';
+import { Alert } from 'react-bootstrap';
+
+import {
+  allowSlowPagesParam,
+  createAdjustedQueryString,
+  shouldAllowSlowPages,
+} from '../../utils';
+
+type GetTextForSlowState = (allowSlowPages: boolean) => string;
+
+export type PublicProps = {
+  _shouldAllowSlowPages?: (location: Location) => boolean;
+  getLinkText: GetTextForSlowState;
+  getMessage: GetTextForSlowState;
+  location: Location;
+};
+
+const SlowPageAlertBase = ({
+  _shouldAllowSlowPages = shouldAllowSlowPages,
+  getLinkText,
+  getMessage,
+  location,
+}: PublicProps) => {
+  const allowSlowPages = _shouldAllowSlowPages(location);
+
+  const newLocation = `${location.pathname}${createAdjustedQueryString(
+    location,
+    {
+      [allowSlowPagesParam]: !allowSlowPages,
+    },
+  )}`;
+
+  return (
+    <Alert variant="danger">
+      {getMessage(allowSlowPages)}{' '}
+      <Alert.Link href={newLocation}>{getLinkText(allowSlowPages)}</Alert.Link>
+    </Alert>
+  );
+};
+
+export default SlowPageAlertBase;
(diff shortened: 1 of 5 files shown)


kumar303:
@willdurand would you maybe have time to look at this one? https://github.com/mozilla/addons-code-manager/pull/951 shows how it would be used in practice. The *caveats* section in that PR explains why I'm using `Alert.Link` (for now).

willdurand:
> @willdurand would you maybe have time to look at this one? #951 shows how it would be used in practice. The _caveats_ section in that PR explains why I'm using `Alert.Link` (for now).

Yup. I am totally fine with the full reload.

kumar303:
@willdurand I simplified the storybook as we discussed.

stories/SlowPageAlert.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+    ...otherProps,
+  };
+  return <SlowPageAlert {...props} />;
+};
+
+storiesOf('SlowPageAlert', module).add('default', () => {
+  return render({
+    getMessage: (allowSlowPages: boolean) =>
+      'This file has been shortened for performance.',
+    getLinkText: (allowSlowPages: boolean) => 'Show the original file.',
```
willdurand: I think these two messages could be the default messages in `render()` so that you could simply `return render()`.

src/components/SlowPageAlert/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  createAdjustedQueryString,
+  shouldAllowSlowPages,
+} from '../../utils';
+
+type GetTextForSlowState = (allowSlowPages: boolean) => string;
+
+export type PublicProps = {
+  _shouldAllowSlowPages?: (location: Location) => boolean;
+  getLinkText: GetTextForSlowState;
+  getMessage: GetTextForSlowState;
```
willdurand: I don't think I understand why this component needs those two getters.

Given that it will know about whether to allow/not allow slow pages _via_ `_shouldAllowSlowPages`, why can't the component decide on the messages? I looked at #951 and these messages are static, so I guess it would simplify everything if a `SlowPageAlert` was capable of choosing the right messages.

src/utils.tsx
```diff
@@ -118,3 +119,7 @@ export const makeReviewersURL = ({
 
   return path;
 };
+
+export const shouldAllowSlowPages = (location: Location) => {
```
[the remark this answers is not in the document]
kumar303: Yes; it will be used to decide whether or not to show an alert at the bottom of the code. When the code is not trimmed (due to allowing slow pages) then it's not necessary to show an alert at the bottom.

Example: https://github.com/mozilla/addons-code-manager/pull/951/files#diff-615ca4e2515a61d7ee97fb1634e87fb7R310

src/components/SlowPageAlert/index.tsx
kumar303: The messages are going to be different for `<CodeView>` -- it won't use the word *diff*.

kumar303:
@willdurand thanks for taking a look so late on a Friday 😅 I think I answered your questions but let me know.

src/components/SlowPageAlert/index.tsx
willdurand: ah, interesting. I still think it's a bit convoluted for what it does but it does not hurt I guess.

PR #976, opened 2019-07-27
Converted parameters to single object in createStoreWithVersion()

### LINKED ISSUE #965: Convert createStoreWithVersion() to use object parameters
In https://github.com/mozilla/addons-code-manager/pull/937/ , [`createStoreWithVersion()`](https://github.com/mozilla/addons-code-manager/blob/cf403fafad6a0ecc3ef783a4b1e036cd6be09b2d/src/test-helpers.tsx#L716) grew an extra option but it was added as a positional argument. Since the first positional argument is also optional, let's convert all arguments to a single object.

For example, after converting it to use object parameters, it can be called like this:

```js
// This passes in all options:
createStoreWithVersion({ version: '...', makeCurrent: true });

// This will create a store with the default version:
createStoreWithVersion({ makeCurrent: true });

// This will use the default for makeCurrent when creating a version.
createStoreWithVersion({ version: '...' });
```

Shubhamchinda:
Fixes #965 

Hey @kumar303 , I had to make some changes in the `createStoreWithVersion()` in order to use object parameters,

Please let me know if I approached it right or if any changes are required.

Thanks!

### DIFF
--- a/src/components/FileMetadata/index.spec.tsx
+++ b/src/components/FileMetadata/index.spec.tsx
@@ -16,7 +16,7 @@ import FileMetadata from '.';
 describe(__filename, () => {
   const _getVersionFile = (props = {}) => {
     const version = fakeVersion;
-    const store = createStoreWithVersion({ makeCurrent: true });
+    const store = createStoreWithVersion({ version });
     store.dispatch(
       versionActions.loadVersionFile({
         path: version.file.selected_file,
--- a/src/components/FileMetadata/index.spec.tsx
+++ b/src/components/FileMetadata/index.spec.tsx
@@ -16,7 +16,7 @@ import FileMetadata from '.';
 describe(__filename, () => {
   const _getVersionFile = (props = {}) => {
     const version = fakeVersion;
-    const store = createStoreWithVersion(version);
+    const store = createStoreWithVersion({ makeCurrent: true });
     store.dispatch(
       versionActions.loadVersionFile({
         path: version.file.selected_file,
--- a/src/components/FileTreeNode/index.spec.tsx
+++ b/src/components/FileTreeNode/index.spec.tsx
@@ -131,7 +131,7 @@ describe(__filename, () => {
     treefoldRenderProps = {},
   }) => {
     const externalVersion = fakeVersion;
-    const store = createStoreWithVersion({ makeCurrent: true });
+    const store = createStoreWithVersion({ version: externalVersion });
     const renderProps = getTreefoldRenderProps({
       id: externalVersion.file.selected_file,
       ...treefoldRenderProps,
@@ -315,7 +315,7 @@ describe(__filename, () => {
 
   it('marks a file node as selected', () => {
     const externalVersion = fakeVersion;
-    const store = createStoreWithVersion({ makeCurrent: true });
+    const store = createStoreWithVersion({ version: externalVersion });
 
     const root = renderWithLinterProvider({
       store,
@@ -699,7 +699,7 @@ describe(__filename, () => {
 
   it('configures LinterProvider', () => {
     const externalVersion = fakeVersion;
-    const store = createStoreWithVersion({ makeCurrent: true });
+    const store = createStoreWithVersion({ version: externalVersion });
     const root = render({ store, versionId: externalVersion.id });
 
     const provider = root.find(LinterProvider);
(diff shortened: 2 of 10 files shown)


src/components/FileTreeNode/index.spec.tsx
```diff
@@ -129,7 +131,7 @@ describe(__filename, () => {
     treefoldRenderProps = {},
   }) => {
     const externalVersion = fakeVersion;
-    const store = createStoreWithVersion(externalVersion);
+    const store = createStoreWithVersion({ makeCurrent: true });
```
kumar303: This needs to still pass `version: externalVersion` since it has an explicit dependency on `externalVersion`. Otherwise, the test relies on the default.

src/components/FileTreeNode/index.spec.tsx
```diff
@@ -311,7 +315,7 @@ describe(__filename, () => {
 
   it('marks a file node as selected', () => {
     const externalVersion = fakeVersion;
-    const store = createStoreWithVersion(externalVersion);
+    const store = createStoreWithVersion({ makeCurrent: true });
```
kumar303: Same here

src/components/FileTreeNode/index.spec.tsx
```diff
@@ -695,7 +699,7 @@ describe(__filename, () => {
 
   it('configures LinterProvider', () => {
     const externalVersion = fakeVersion;
-    const store = createStoreWithVersion(externalVersion);
+    const store = createStoreWithVersion({ makeCurrent: true });
```
kumar303: Same here.

src/components/FileTreeNode/index.spec.tsx
```diff
@@ -732,7 +736,7 @@ describe(__filename, () => {
     const externalVersion = fakeVersion;
     const nodeId = externalVersion.file.selected_file;
 
-    const store = createStoreWithVersion(externalVersion);
+    const store = createStoreWithVersion({ makeCurrent: true });
```
kumar303: Same here. I'm commenting on each one just so we don't miss any in the next review.

src/components/FileTreeNode/index.spec.tsx
```diff
@@ -784,7 +790,7 @@ describe(__filename, () => {
 
   it('does not focus a node when selected and already in focus', () => {
     const externalVersion = fakeVersion;
-    const store = createStoreWithVersion(externalVersion);
+    const store = createStoreWithVersion({ makeCurrent: true });
```
kumar303: And here

src/components/FileTreeNode/index.spec.tsx
Shubhamchinda: I thought since `externalVersion = fakeVersion` , So it would be fine to use `makeCurrent : true` here. Anyways Thanks for explaining, will update the PR.

src/components/FileMetadata/index.spec.tsx
```diff
@@ -16,7 +16,7 @@ import FileMetadata from '.';
 describe(__filename, () => {
   const _getVersionFile = (props = {}) => {
     const version = fakeVersion;
-    const store = createStoreWithVersion(version);
+    const store = createStoreWithVersion({ makeCurrent: true });
```
Shubhamchinda: @kumar303 Do we need to change it here as well?

Shubhamchinda:
@kumar303 Updated ✅

src/components/FileMetadata/index.spec.tsx
kumar303: Yes, I missed this one. It should change, too.

Imagine changing the default version object (changing an id, a name, a description). This should not break any tests. If the test is relying implicitly on a name or id in the default version, it might start failing.

Shubhamchinda: okay @kumar303 .I updated the patch.

Thanks! :)

stories/FileTreeNode.stories.tsx
```diff
@@ -82,8 +84,11 @@ storiesOf('FileTreeNode', module).addWithChapters('all variants', {
           sectionFn: () => {
             const versionId = 921;
             const store = createStoreWithVersion({
-              ...fakeVersion,
-              id: versionId,
+              version: {
+                ...fakeVersion,
+                id: versionId,
+              },
+              makeCurrent: true,
```
kumar303: Why does this have `makeCurrent` set to `true`?

kumar303: The old code was not doing that

Shubhamchinda: > The old code was not doing that

I must have missed it. 
Updated it now! :)

Shubhamchinda:
> This looks great. Thanks for all the fix-ups.

You welcome. 

Thanks to you too for helping out! :)

Issue #987, opened 2019-08-06
Add a draft comments reducer


kumar303:
In order to begin entering a comment on a specific line of a file (https://github.com/mozilla/addons-code-manager/issues/112), we need a reducer.

PR #1006, opened 2019-08-09
Adds a reducer for comments

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/987

### DIFF
--- a/src/reducers/comments.spec.tsx
+++ b/src/reducers/comments.spec.tsx
@@ -1,0 +1,168 @@
+import reducer, {
+  CommentInfo,
+  actions,
+  createCommentsKey,
+  createEmptyCommentInfo,
+  createInternalComment,
+} from './comments';
+import { createInternalVersion } from './versions';
+import { createFakeExternalComment, fakeVersion } from '../test-helpers';
+
+describe(__filename, () => {
+  const createCommentInfo = (info: Partial<CommentInfo> = {}): CommentInfo => {
+    return {
+      ...createEmptyCommentInfo(),
+      ...info,
+    };
+  };
+
+  describe('beginComment', () => {
+    it('begins a comment by key', () => {
+      const fileName = 'manifest.json';
+      const line = 123;
+      const versionId = 1;
+
+      const state = reducer(
+        undefined,
+        actions.beginComment({ versionId, fileName, line }),
+      );
+
+      expect(
+        state.byKey[createCommentsKey({ versionId, fileName, line })],
+      ).toEqual(createCommentInfo({ beginNewComment: true }));
+    });
+  });
+
+  describe('finishComment', () => {
+    it('finishes a comment by key', () => {
+      const fileName = 'manifest.json';
+      const line = 123;
+      const versionId = 1;
+
+      const state = reducer(
+        undefined,
+        actions.finishComment({ fileName, line, versionId }),
+      );
+
+      expect(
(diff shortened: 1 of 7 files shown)


src/reducers/comments.tsx
```diff
@@ -0,0 +1,179 @@
...
+            ...info,
+            beginNewComment: true,
+          },
+        },
+      };
+    }
+    case getType(actions.finishComment): {
+      const { versionId, fileName, line } = action.payload;
+
+      const key = createCommentsKey({ fileName, line, versionId });
```
bobsilverberg: Same as 120 and 122, above, this could just be:

`const key = createCommentsKey(action.payload);`

src/reducers/comments.tsx
```diff
@@ -0,0 +1,179 @@
...
+            ...info,
+            beginNewComment: false,
+          },
+        },
+      };
+    }
+    case getType(actions.setComment): {
+      const { comment, fileName, line, versionId } = action.payload;
+
+      const key = createCommentsKey({ fileName, line, versionId });
```
bobsilverberg: The above suggestion won't work here because the `payload` includes `comment`.

src/reducers/comments.spec.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  describe('beginComment', () => {
+    it('begins a comment by key', () => {
+      const fileName = 'manifest.json';
+      const line = 123;
+      const versionId = 1;
+
+      const state = reducer(
+        undefined,
+        actions.beginComment({ versionId, fileName, line }),
```
bobsilverberg: Nit: This isn't a big deal (hence the `Nit:`) but I notice that the order of the properties in the calls to the various actions, as well as in the calls to `createCommentsKey` are inconsistent throughout the file. Sometimes `versionId` is first and sometimes `fileName` is first. It really doesn't matter, but I personally try to keep things consistent by always putting them in alphabetical order. I'm just calling this out to bring it to your attention in case you didn't realize and would like to fix it, but if you don't think it's worth fixing I'm fine with that too.

src/reducers/comments.spec.tsx
```diff
@@ -0,0 +1,168 @@
...
+        state,
+        actions.setComment({ comment: comment2, fileName, line, versionId }),
+      );
+
+      expect(
+        state.byKey[createCommentsKey({ fileName, line, versionId })],
+      ).toMatchObject({
+        commentIds: [comment1.id, comment2.id],
+      });
+    });
```
bobsilverberg: I know you've already (mostly) tested it above in `'sets a comment for a key'`, but might it be worth checking to see that both actual comments exist, as opposed to just checking that `commentIds` is correct? This would ensure that calling `setComment` doesn't accidentally overwrite an existing comment.

src/reducers/comments.spec.tsx
```diff
@@ -0,0 +1,168 @@
...
+      const versionId = 1;
+      const fileName = 'manifest.json';
+      const line = 321;
+
+      expect(createCommentsKey({ versionId, fileName, line })).toEqual(
+        `version:${versionId};file:${fileName};line:${line}`,
+      );
+    });
+
+    it('creates a key from versionId, fileName', () => {
```
bobsilverberg: There is no explicit test for just `versionId` and `line`, which is a legal combination. Perhaps this is enough to cover it, as it does use the same implementation logic, but it might be worth adding anyway.

PR #1024, opened 2019-08-28
update eslint-utils

willdurand:


willdurand:
Snyk was having connectivity issues (again), I could not browse the web page and it did not send a reply to the github webhook, so I merged this patch.

Issue #1030, opened 2019-08-30
Do not trigger keyboard shortcuts from the comment box


kumar303:
While typing a comment on a line of code (forthcoming in https://github.com/mozilla/addons-code-manager/issues/112), the `keydown` event is getting captured by `<KeyboardShortcuts>`. This will interrupt the UI by possibly navigating to different files and destroying the unsaved comment.

PR #1031, opened 2019-08-30
Don't propagate keydown from Comment textarea

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/1030

### DIFF
--- a/src/components/Comment/index.spec.tsx
+++ b/src/components/Comment/index.spec.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { shallow } from 'enzyme';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
-import { createFakeComment } from '../../test-helpers';
+import { createFakeComment, createFakeEvent } from '../../test-helpers';
 import styles from './styles.module.scss';
 
 import Comment, { PublicProps } from '.';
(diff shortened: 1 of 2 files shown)


src/components/Comment/index.spec.tsx
```diff
@@ -55,4 +57,64 @@ describe(__filename, () => {
...
+
+    it('does not add/remove listeners with an undefined ref', () => {
+      // Make sure this doesn't throw an error.
+      const root = render({ createTextareaRef: () => undefined });
+      root.unmount();
+    });
+
+    it('does not add/remove listeners with a null DOM node', () => {
+      // Make sure this doesn't throw an error.
+      const root = render({ createTextareaRef: () => React.createRef() });
```
bobsilverberg: I'm a bit confused by this test. This seems to be doing exactly the same thing as the default behaviour when `createTextareaRef` is undefined, i.e., it returns `React.createRef()` (which is the same as what's done on line 21 of `Comment/index.tsx`), so why is this equivalent to "with a null DOM node"?

kumar303: It's different because this test will define a `this.textareaRef` object with `current` set to `null` (i.e. a `null` DOM node). It will exercise the second part of this if statement:

https://github.com/mozilla/addons-code-manager/blob/63ef0cffa1bf0b1a06c3430db6d6576548424539/src/components/Comment/index.tsx#L39

If you meant "this is the same as the default rendering state" then that's true but it could be risky to rely on other tests to implicitly cover this behavior (they *probably* will but they might not). I wanted to add an explicit test for it so that any breakage might be more obvious.

Issue #988, opened 2019-08-06
Add a thunk for creating / editing a comment


kumar303:
In order to save a comment on a specific line of a file (https://github.com/mozilla/addons-code-manager/issues/112), we need a thunk.

PR #1041, opened 2019-09-04
Adds a thunk for creating / updating a comment

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/988

~~⚠️ This depends on https://github.com/mozilla/addons-code-manager/pull/1040~~

@bobsilverberg I removed the type safety around accepting `comment` or `cannedResponseId` because I ran into errors. I'd need to switch the params to [discriminated unions](https://basarat.gitbooks.io/typescript/docs/types/discriminated-unions.html) to fix it; I don't think that's worth it. For example, I'd need to add a literal `type` field so that one call has `{ type: 'COMMENT' }` and the other has `{ type: 'CANNED_COMMENT' }`.

### DIFF
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -293,26 +293,6 @@ type CommentRequest = {
   lineno?: number | null;
 };
 
-type CreateOrUpdateCommentParamsBase = {
-  _callApi?: typeof callApi;
-  addonId: number;
-  apiState: ApiState;
-  commentId: number | void;
-  fileName: string | null;
-  line: number | null;
-  versionId: number;
-};
-
-type CreateOrUpdateCommentParams = CreateOrUpdateCommentParamsBase & {
-  cannedResponseId?: undefined;
-  comment: string;
-};
-
-type CreateOrUpdateCannedCommentParams = CreateOrUpdateCommentParamsBase & {
-  cannedResponseId: number;
-  comment?: undefined;
-};
-
 export const createOrUpdateComment = async ({
   /* istanbul ignore next */
   _callApi = callApi,
@@ -324,7 +304,17 @@ export const createOrUpdateComment = async ({
   fileName,
   line,
   versionId,
-}: CreateOrUpdateCommentParams | CreateOrUpdateCannedCommentParams) => {
+}: {
+  _callApi?: typeof callApi;
+  addonId: number;
+  apiState: ApiState;
+  cannedResponseId?: number;
+  comment?: string;
+  commentId: number | void;
+  fileName: string | null;
+  line: number | null;
+  versionId: number;
+}) => {
   if (cannedResponseId === undefined && comment === undefined) {
     throw new Error('Either cannedResponseId or comment must be specified');
   }
--- a/src/reducers/comments.spec.tsx
+++ b/src/reducers/comments.spec.tsx
@@ -1,13 +1,19 @@
 /* eslint @typescript-eslint/camelcase: 0 */
+import { actions as errorsActions } from './errors';
 import reducer, {
   CommentInfo,
   actions,
   createCommentKey,
   createEmptyCommentInfo,
   createInternalComment,
+  manageComment,
 } from './comments';
 import { createInternalVersion } from './versions';
-import { createFakeExternalComment, fakeVersion } from '../test-helpers';
+import {
+  createFakeExternalComment,
+  fakeVersion,
+  thunkTester,
+} from '../test-helpers';
 
 describe(__filename, () => {
   const createCommentInfo = (info: Partial<CommentInfo> = {}): CommentInfo => {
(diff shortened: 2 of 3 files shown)


src/api/index.tsx
```diff
@@ -293,26 +293,6 @@ type CommentRequest = {
   lineno?: number | null;
 };
 
-type CreateOrUpdateCommentParamsBase = {
```
bobsilverberg: I saw in your comment that you had to remove this, but I'm not sure I understand why. AFAIU this *should* work. What exactly was the problem with it?

src/reducers/comments.tsx
```diff
@@ -141,6 +200,37 @@ const reducer: Reducer<CommentsState, ActionType<typeof actions>> = (
           [key]: {
             ...info,
             beginNewComment: true,
+            savingComment: false,
+          },
+        },
+      };
+    }
+    case getType(actions.beginSaveComment): {
+      const key = createCommentKey(action.payload);
```
bobsilverberg: These two lines are repeated throughout this file. Might it be worthwhile to create a helper which executes these two lines are returns `{key, info}`?

src/reducers/comments.spec.tsx
```diff
@@ -19,35 +25,80 @@ describe(__filename, () => {
 
   describe('beginComment', () => {
     it('begins a comment by key', () => {
-      const fileName = 'manifest.json';
-      const line = 123;
-      const versionId = 1;
+      const params = { versionId: 1, fileName: 'manifest.json', line: 123 };
```
bobsilverberg: Nit: You could hoist the definition of `params` into the `describe('beginComment'` block and then not have to repeat them in these two tests. I think this would be okay because you never actually care about the specific values in `params` are really just referencing the `params` object in the tests. You might even be able to hoist them up to the top `describe(__filename` block, as it seems the same is true for most of the tests.

src/reducers/comments.spec.tsx
```diff
@@ -19,35 +25,80 @@ describe(__filename, () => {
...
 
-      const state = reducer(
-        undefined,
-        actions.beginComment({ versionId, fileName, line }),
+      const state = reducer(undefined, actions.beginComment(params));
+
+      expect(state.byKey[createCommentKey(params)]).toEqual(
+        createCommentInfo({ beginNewComment: true }),
```
bobsilverberg: Would there be any value in switching this to use `expect.objectContaining({ beginNewComment: true })`, rather than `createCommentInfo({ beginNewComment: true })`? I suppose the latter is less implementation-specific, so it seems better in that case, but I notice we do often use `expect.objectContaining` in the tests.

src/reducers/comments.spec.tsx
```diff
@@ -19,35 +25,80 @@ describe(__filename, () => {
...
+
+      expect(state.byKey[createCommentKey(params)]).toEqual(
+        createCommentInfo({ beginNewComment: false }),
       );
+    });
 
-      expect(
-        state.byKey[createCommentKey({ versionId, fileName, line })],
-      ).toEqual(createCommentInfo({ beginNewComment: false }));
+    it('aborts saving a comment', () => {
```
bobsilverberg: Should this say "finishes saving a comment"? We're not actually aborting the save here.

src/reducers/comments.spec.tsx
```diff
@@ -19,35 +25,80 @@ describe(__filename, () => {
...
+      const params = { versionId: 1, fileName: 'manifest.json', line: 123 };
+
+      let state;
+
+      state = reducer(state, actions.beginSaveComment(params));
+      state = reducer(state, actions.finishComment(params));
+
+      expect(state.byKey[createCommentKey(params)]).toMatchObject({
```
bobsilverberg: Oh, I see you do use `toMatchObject` here, rather than `toEqual(createCommentInfo())`, which is similar to what I asked about regarding `expect.objectContaining()`. Why did you decide to do that here?

src/reducers/comments.spec.tsx
```diff
@@ -176,4 +227,127 @@ describe(__filename, () => {
...
+        comment,
+        commentId,
+        fileName,
+        line,
+        versionId,
+      });
+    });
+
+    it('dispatches beginSaveComment()', async () => {
+      const versionId = 1;
```
bobsilverberg: Could you combine these into a `params` object, like you did for some of the other tests in this file, and then just call `_manageComment(params)` and `toHaveBeenCalledWith(actions.beginSaveComment(params))`?

src/api/index.tsx
kumar303: It's only possible if I switch to [discriminated unions](https://basarat.gitbooks.io/typescript/docs/types/discriminated-unions.html) (i.e. each type will need a unique `type` field) and I don't think that's worth it.

kumar303: I didn't notice this before because I guess I didn't have any code that tried to call the function in all possible ways.

src/reducers/comments.tsx
kumar303: Sure

src/reducers/comments.spec.tsx
kumar303: The first test you commented on only calls one action so it's safe to use `createCommentInfo()`, the common initializer for the state. I used `expect.objectContaining()` here, instead, because multiple actions are involved; I don't want to couple the test to unrelated state changes triggered by those other actions now or in the future.

Issue #1044, opened 2019-09-06
Add creation abilities to the Comment component


kumar303:
Before fully hooking up a comment flow in https://github.com/mozilla/addons-code-manager/issues/112 , the `Comment` component needs to support form submission.

Issue #715, opened 2019-05-08
Edit a comment on a specific line of a file


kumar303:
User story:

> After adding a comment (https://github.com/mozilla/addons-code-manager/issues/113, https://github.com/mozilla/addons-code-manager/issues/112) as a reviewer, decide you want to edit it before submitting to the developer (https://github.com/mozilla/addons-code-manager/issues/114).

kumar303:
This depends on an API change: https://github.com/mozilla/addons-server/issues/11379

EnTeQuAk:
mozilla/addons-server#11379 has been landed, so I think the `needs: api` label should be safe to remove. Let me know if you need anything specific on top of that though.

mirefly:
@kumar303 Could I work on this?

kumar303:
@mirefly yes, for sure! Thanks

ioanarusiczki:
@mirefly I verified on dev and the Edit button is available. Text can be edited and saved.
Should the comments allow some HTML text such and for bold <b>,<italic> or even links?

mirefly:
@ioanarusiczki Thanks for verifying it. I am not sure if the reviewers uses HTML formatting comments. @wagnerand @bobsilverberg do we want to support it?

bobsilverberg:
> @ioanarusiczki Thanks for verifying it. I am not sure if the reviewers uses HTML formatting comments. @wagnerand @bobsilverberg do we want to support it?

Considering that the current workflow is to copy the comments from the Summary and then paste them into a form in the older reviewer tools, it doesn't seem like having the ability to add HTML formatting would help much. Perhaps down the road when we are able to have a tighter integration it will make sense, but for now I don't think it's worth adding support for that.

@wagnerand What do you think?

wagnerand:
I agree with Bob. No need to add HTML support in the foreseeable future.

PR #1045, opened 2019-09-06
Add creation abilities to the Comment component

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/1044

⚠️ Depends on https://github.com/mozilla/addons-code-manager/pull/1041

Keep in mind when reviewing this patch that it only adds comment creation. There are things like `initialComment` which will probably change when I add editing in https://github.com/mozilla/addons-code-manager/issues/715


kumar303:
@bobsilverberg This is ready for review now. It's getting a bit hard to manage multiple patches (I have more waiting) but if you'd like me to make this smaller I could probably extract the reducer changes.

src/components/Comment/index.spec.tsx
```diff
@@ -1,82 +1,309 @@
...
+    const root = render({ commentId: comment.id, store, readOnly: true });
 
     expect(root.find(`.${styles.form}`)).toHaveLength(0);
     expect(root.find(`.${styles.comment}`)).toHaveLength(1);
-    expect(root.find(`.${styles.comment}`).html()).toContain(comment.content);
+    expect(root.find(`.${styles.comment}`).html()).toContain(content);
     expect(root.find(FontAwesomeIcon)).toHaveLength(1);
   });
 
   it('sanitizes the content of a comment', () => {
```
bobsilverberg: Nit: This only applies to `readOnly: true`. I wonder if it would be worthwhile to add that to the test description, or perhaps add a `describe` block for the `readOnly` tests.

src/components/Comment/index.spec.tsx
```diff
@@ -1,82 +1,309 @@
...
 
-  it('renders a form to edit a comment when the readOnly prop is false', () => {
-    const comment = createFakeComment();
-    const root = render({ comment, readOnly: false });
+  it('renders a form to edit a comment when readOnly=false', () => {
+    const comment = createFakeExternalComment({ comment: 'Example' });
+    const { store } = dispatchComment({ comment });
+    const root = render({ commentId: comment.id, store, readOnly: false });
 
     expect(root.find(`.${styles.form}`)).toHaveLength(1);
```
bobsilverberg: In the test for `renders a comment when readOnly=true`, you assert both that `styles.comment` is present and that `styles.form` is not, yet in this test you only assert that `styles.form` is present. Should there be consistency between the tests?

src/components/Comment/index.tsx
```diff
@@ -58,38 +136,64 @@ export class CommentBase extends React.Component<PublicProps> {
 
         <div
           // eslint-disable-next-line react/no-danger
-          dangerouslySetInnerHTML={sanitizeHTML(nl2br(comment.content), ['br'])}
+          dangerouslySetInnerHTML={sanitizeHTML(nl2br(initialComment.content), [
```
bobsilverberg: I see that there's a test that covers `sanitizeHTML`, but no tests for either `nl2br`, nor the fact that `br` is being allowed via `sanitizeHTML`. Do you think we should have coverage for both of those?

src/components/Comment/index.tsx
```diff
@@ -44,9 +85,46 @@ export class CommentBase extends React.Component<PublicProps> {
...
+    const {
+      _manageComment,
+      addonId,
+      dispatch,
+      fileName,
+      initialComment,
+      line,
+      versionId,
+    } = this.props;
+    event.preventDefault();
```
bobsilverberg: This doesn't seem to be covered by a test. Should it be?

src/components/Comment/index.spec.tsx
```diff
@@ -1,82 +1,309 @@
...
+    const textarea = root.find(`.${styles.textarea}`);
+    expect(textarea).toHaveProp('disabled', true);
+    expect(textarea).toHaveProp('value', pendingCommentText);
+
+    const button = root.find(Button);
+    expect(button).toHaveProp('disabled', true);
+    expect(button).toHaveText('Saving…');
+  });
+
+  it('seeds the form with an initial comment', () => {
```
bobsilverberg: I find a bit of a disconnect between this test's description and what it actually checks. From the description I would have expected it to check the contents of the textarea, but what it actually does is check what is submitted when the form is submitted. Would the former be a more accurate test for this, or should the description of the test perhaps be changed?

src/components/Comment/index.spec.tsx
```diff
@@ -1,82 +1,309 @@
...
+    expect(dispatchSpy).toHaveBeenCalledWith(fakeThunk.thunk);
+    expect(_manageComment).toHaveBeenCalledWith(
+      expect.objectContaining({
+        comment: commentText,
+        commentId: comment.id,
+      }),
+    );
+  });
+
+  it('seeds the form with initialCommentText', () => {
```
bobsilverberg: Similar to my comment above, I find the name of the test to not exactly match what the test does. Maybe it's just me though.

bobsilverberg: Also, I don't see a test where both existing comment is dispatched, and also `beginSaveComment` is dispatched with a value for `pendingCommentText`. I.e., the case where a form is opened for an existing comment, and then it is submitted with a changed value. It seems like that case should be covered.

src/components/Comment/index.spec.tsx
```diff
@@ -1,82 +1,309 @@
...
+
+    const fakeThunk = createFakeThunk();
+    const _manageComment = fakeThunk.createThunk;
+
+    const root = render({
+      _manageComment,
+      commentId,
+      store,
+    });
+
```
bobsilverberg: Would it be worth having an assertion that the current contents of the `textarea` is the value from the existing comment? The earlier test for `seeds the form with an initial comment` doesn't check that either, so it seems like there's no coverage for that.

src/components/Comment/index.tsx
```diff
@@ -1,21 +1,61 @@
...
+
+export type DefaultProps = {
+  _manageComment: typeof manageComment;
+  createTextareaRef?: () => TextareaRef;
 };
 
-export class CommentBase extends React.Component<PublicProps> {
+type PropsFromState = {
+  initialComment: Comment | null;
+  initialCommentText: string | null;
```
bobsilverberg: TBH I find this whole `initialComment` and `initialCommentText` thing to be quite confusing. After reading all the code and tests I'm still not sure why we need `initialCommentText`. I was hoping the tests would make it clear, but they do not (to me anyway), which suggests that maybe there is some coverage missing. I did make a comment about a specific case not being covered, but I am concerned by how confused I am about this.

From what I understand, `initialCommentText` is about populating the form with data from `pendingCommentText` in the redux store, but that only seems to be relevant after a user has submitted the form. `initialCommentText` is only accessed in the `constructor`, so I assume it will only ever be used when the component is initially created, and won't ever be used on an update/re-render. So I don't really understand the flow here: when the form is submitted, which will cause `pendingCommentText` to be updated in the redux store, will that cause the `constructor` to rerun? My assumption is "no", but maybe I don't understand the lifecycle of the component. If I am correct that the `constructor` won't rerun, then I don't see how having access to `initialCommentText` in `props` is useful at that point. 

The tests show that you can do an initial render of the component with a value for `pendingCommentText` in the store, and it will use it, but this seems like it's not the way the component will actually behave. I assume that the component will be initially rendered with either an existing comment, or with no `pendingCommentText` in the store, and that only after the form is submitted will a value for `pendingCommentText` be added to the store.

src/components/Comment/index.tsx
```diff
@@ -44,9 +85,46 @@ export class CommentBase extends React.Component<PublicProps> {
     }
   }
 
-  renderComment() {
-    const comment = this.props.comment as Comment;
+  onCommentChange = (event: React.ChangeEvent<HTMLTextAreaElement>) => {
+    event.preventDefault();
```
bobsilverberg: This doesn't seem to be covered by a test. Should it be?

src/components/Comment/index.spec.tsx
```diff
@@ -1,82 +1,309 @@
...
-    return shallow(<Comment {...props} />);
+    return shallowUntilTarget(<Comment {...props} />, CommentBase, {
+      shallowOptions: { context: { store } },
+    });
   };
 
-  it('renders a form when there is no comment', () => {
-    const root = render({ comment: null });
+  const dispatchComment = ({
+    store = configureStore(),
```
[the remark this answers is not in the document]
kumar303: Yeah, it could be useful for future test cases but, more importantly, adding this flexibility is not harder than doing without it. In other words, these two approaches would require the same lines of code:

```ts
  const dispatchComment = ({
    store = configureStore(),
    comment = createFakeExternalComment(),
    fileName = null,
    line = null,
    versionId = 1,
  } = {}) => {
    store.dispatch(actions.setComment({ fileName, line, versionId, comment }));
    return { store };
  };
```

```ts
  const dispatchComment = ({
    comment = createFakeExternalComment(),
    fileName = null,
    line = null,
    versionId = 1,
  } = {}) => {
    store.dispatch(actions.setComment({ fileName, line, versionId, comment }));
    return { store };
  };
```

...only one would be less flexible.

src/components/Comment/index.spec.tsx
kumar303: I don't think it's important to add `readOnly: true` to the spec because it's not a parameter that the feature depends on. To put it another way: the sanitizer feature doesn't depend on the `readOnly` prop for anything.

src/components/Comment/index.spec.tsx
kumar303: whoops, this one needs an assertion that the comment is not present.

src/components/Comment/index.tsx
kumar303: Good catch. I added a test for `nl2br`.

src/components/Comment/index.tsx
kumar303: Personally I don't think it's such a useful thing to add a regression test for. When reading this code, it's clear that the intent is to prevent the default, so, why would a future developer change it? I also can't see how a refactor would accidentally remove this. 

Tests exist primarily to prevent accidental breakage by developers. Tests are not good at preventing *intentional* breakage 😅 

I see `event.preventDefault()` as something similar to putting `color="red"` in the JSX of a component. It's configuration. I'm open to other ideas about it, though.

As a counterpoint, let's say this method had to be called in an if statement, like:

```ts
onClick(event) {
  if (!this.props.allowDefault) {
    event.preventDefault();
  } 
}
```

In that case I'd want to see a regression test because it would be easy to accidentally break that in a refactor.

src/components/Comment/index.spec.tsx
kumar303: I didn't check the textarea value because it seemed like I was reaching into the implementation too much. However, a textarea is how you make a form, heh. I guess it's fine to check the value. The test was taking a previously saved value and checking that it was submitted as part of the form to imply that this value was used in the textarea.

src/components/Comment/index.spec.tsx
kumar303: The "it lets you update a previously saved comment" [test](https://github.com/mozilla/addons-code-manager/pull/1045/files#diff-ae5df7d4b15cf0f3adbcf02ae01c5e85R259) implies that the situation you described will work. However, I have since updated the "it seeds..." tests which I think makes it more clear.

src/components/Comment/index.spec.tsx
kumar303: Yeah, that will be helpful to prove that the textarea is part of the update process.

src/components/Comment/index.tsx
kumar303: > when the form is submitted, which will cause `pendingCommentText` to be updated in the redux store, will that cause the `constructor` to rerun? My assumption is "no", but maybe I don't understand the lifecycle of the component. 

Yes, it will cause the constructor to run because this:

https://github.com/mozilla/addons-code-manager/blob/0b12496b0e2d78b6f792a5b801554e1e93de379e/src/reducers/comments.tsx#L162

...causes a re-render of `CodeView` and that creates a new instance of `Comment`. I'm not exactly sure *why* but it's probably [this render prop](https://github.com/mozilla/addons-code-manager/blob/0b12496b0e2d78b6f792a5b801554e1e93de379e/src/components/LinterProvider/index.tsx#L87).

I had to add this feature precisely because of this behavior, which is surprising. If an update to `pendingCommentText` were to instead update an existing component instance then we wouldn't need the feature at all. The edited comment would be held in state and the "pending" text would already be there. For this same reason, I did not implement a prop update of `initialComment` or `initialCommentText` because that should not be allowed (the internal state should take precedence).

This will probably become more clear when I land the next patch that integrates everything into `CodeView`. I can always do follow-ups later once the big picture comes into focus.

kumar303:
@bobsilverberg thanks, good catches. It's ready for another look and I responded inline.

Issue #1059, opened 2019-09-17
Add features to <Commentable> needed by <CodeView>


kumar303:
In order to integrate the `Commentable` component (see https://github.com/mozilla/addons-code-manager/issues/1007) with `CodeView` for file commenting (https://github.com/mozilla/addons-code-manager/issues/112), we need some enhancements:

- `Commentable` needs to accept a React ref for scroll-to-anchor behavior
- It should be clear that `Commentable` requires `AddComment` in its children. This isn't strictly necessary for `CodeView` but it did come up in the review of the original patch and now that I've done some integration work I think it will be helpful.

PR #1060, opened 2019-09-17
Add features to Commentable needed by CodeView

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/1059

### DIFF
--- a/src/components/Commentable/index.spec.tsx
+++ b/src/components/Commentable/index.spec.tsx
@@ -1,13 +1,36 @@
 import * as React from 'react';
-import { shallow } from 'enzyme';
+import { mount, shallow } from 'enzyme';
+import { Store } from 'redux';
 
+import configureStore from '../../configureStore';
+import AddComment from './AddComment';
 import styles from './styles.module.scss';
 
 import Commentable, { PublicProps } from '.';
 
 describe(__filename, () => {
+  const getRenderProps = ({
+    as = 'span',
+    children = () => <div />,
+    fileName = null,
+    line = null,
+    versionId = 1,
+    ...moreProps
+  }: Partial<PublicProps> = {}) => {
+    return { as, children, fileName, line, versionId, ...moreProps };
+  };
+
   const render = (moreProps: Partial<PublicProps> = {}) => {
-    return shallow(<Commentable as="span" {...moreProps} />);
+    return shallow(<Commentable {...getRenderProps(moreProps)} />);
+  };
+
+  const renderWithMount = ({
+    store = configureStore(),
+    ...moreProps
+  }: { store?: Store } & Partial<PublicProps> = {}) => {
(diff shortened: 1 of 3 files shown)


src/components/Commentable/index.spec.tsx
```diff
@@ -25,14 +48,50 @@ describe(__filename, () => {
...
+    expect(button).toHaveClassName(addCommentClassName);
+  });
+
+  it('can add a ref to the outer DOM node', () => {
+    const shellRef = jest.fn();
+
+    // We need `mount()` because `ref` is only used in a DOM environment.
+    const root = renderWithMount({ shellRef });
+
+    expect(shellRef).toHaveBeenCalledWith(root.getDOMNode());
```
bobsilverberg: I take it that this is something that is done automatically when you set the `ref` of `Commentable` to `shellRef`?

src/components/Commentable/index.tsx
```diff
@@ -1,29 +1,45 @@
...
 }: PublicProps) => {
   return (
     <AsComponent
       className={makeClassName(styles.commentable, className)}
       id={id}
+      ref={shellRef}
     >
-      {children}
+      {children(
+        <AddComment
```
bobsilverberg: This is nice the way you got `Commentable` to include `AddComment`. Well done! 👍

src/components/Commentable/index.tsx
```diff
@@ -1,29 +1,45 @@
...
 
+import AddComment, { PublicProps as AddCommentProps } from './AddComment';
 import { AnyReactNode } from '../../typeUtils';
 import styles from './styles.module.scss';
 
 export type PublicProps = {
+  addCommentClassName?: string;
   as: React.ReactType;
-  children?: AnyReactNode;
+  children: (addComment: JSX.Element) => AnyReactNode;
```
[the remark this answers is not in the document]
kumar303: As far as I can tell it's not possible, at least not with current [TypeScript support for JSX](https://www.typescriptlang.org/docs/handbook/jsx.html#the-jsx-result-type). I guess that makes sense since you can't manipulate JSX from React components anyway (i.e. it doesn't matter what component it's an instance of).

src/components/Commentable/index.spec.tsx
kumar303: Yeah. If you delete the `ref` assignment within `Commentable`, this test fails.

src/components/Commentable/index.tsx
kumar303: Well, the downside is you can't do normal shallow rendering when testing a component that uses `Commentable`. This is pretty major and it's why I originally didn't want to do it. However, we know how to write test helpers to reach in and evoke callbacks so, yeah, I guess it's OK. The more I thought about it, it felt like the *right* thing to do.

kumar303:
Thanks for the quick review

PR #1064, opened 2019-09-18
Actually document local configuration

kumar303:
I forgot how our special local configuration works so we should document this 🤦‍♂

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ When developing a new component, always add a story for it. If you were creating
 You can configure the app by defining environment variables in `.env` files, the standard way to [configure Create React App](https://facebook.github.io/create-react-app/docs/adding-custom-environment-variables#adding-development-environment-variables-in-env), but read on below because **there are some differences**.
 
 - How to override an environment variable for local development
-  - Define it in `env.common-local`. **This differs** from how Create React App wants you to do it. If you put an environment variable in `env.local`, it will get erased.
+  - Define it in `.env.common-local`. **This differs** from how Create React App wants you to do it. If you put an environment variable in `.env.local`, it will get erased.
 - How to define a new environment variable
   - Add it to `.env` with the `REACT_APP_` prefix.
 - How to override an environment variable for a hosted site
--- a/README.md
+++ b/README.md
@@ -67,9 +67,11 @@ When developing a new component, always add a story for it. If you were creating
 
 ## Configuration
 
-You can configure the app by defining environment variables in `.env` files, the standard way to [configure Create React App](https://facebook.github.io/create-react-app/docs/adding-custom-environment-variables#adding-development-environment-variables-in-env).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -67,9 +67,11 @@ When developing a new component, always add a story for it. If you were creating
 
 ## Configuration
 
-You can configure the app by defining environment variables in `.env` files, the standard way to [configure Create React App](https://facebook.github.io/create-react-app/docs/adding-custom-environment-variables#adding-development-environment-variables-in-env).
+You can configure the app by defining environment variables in `.env` files, the standard way to [configure Create React App](https://facebook.github.io/create-react-app/docs/adding-custom-environment-variables#adding-development-environment-variables-in-env), but read on below because **there are some differences**.
 
-- How to add a new environment variable
+- How to override an environment variable for local development
+  - Define it in `.env.common-local`. **This differs** from how Create React App wants you to do it. If you put an environment variable in `.env.local`, it will get erased.
```
willdurand:
```suggestion
  - Define it in `.env.common-local`. **This differs** from how Create React App wants you to do it. If you put an environment variable in `.env.local`, it will get erased because we are generating this file when the app starts.
```

README.md
```diff
@@ -67,9 +67,11 @@ When developing a new component, always add a story for it. If you were creating
 
 ## Configuration
 
-You can configure the app by defining environment variables in `.env` files, the standard way to [configure Create React App](https://facebook.github.io/create-react-app/docs/adding-custom-environment-variables#adding-development-environment-variables-in-env).
+You can configure the app by defining environment variables in `.env` files, the standard way to [configure Create React App](https://facebook.github.io/create-react-app/docs/adding-custom-environment-variables#adding-development-environment-variables-in-env), but read on below because **there are some differences**.
 
-- How to add a new environment variable
+- How to override an environment variable for local development
```
willdurand:
```suggestion
- How to override an environment variable for local development?
```

README.md
```diff
@@ -67,9 +67,11 @@ When developing a new component, always add a story for it. If you were creating
...
+  - Define it in `.env.common-local`. **This differs** from how Create React App wants you to do it. If you put an environment variable in `.env.local`, it will get erased.
+- How to define a new environment variable
```
willdurand:
```suggestion
- How to define a new environment variable?
```

README.md
```diff
@@ -67,9 +67,11 @@ When developing a new component, always add a story for it. If you were creating
...
   - Add it to `.env` with the `REACT_APP_` prefix.
 - How to override an environment variable for a hosted site
```
willdurand:
```suggestion
- How to override an environment variable for a hosted site?
```

README.md
kumar303: Heh. I had that in there but I thought it would be overwhelming (too much information). I'll put it in there since you thought of it, too.

README.md
kumar303: I think it's fine as a statement. If it were a question, the grammar would need to change, anyway. One example: "How do you define a new environment variable?" ... or ... "How would you define ...?"

PR #1066, opened 2019-09-19
Support commenting on a line of a file

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/112

This is the first patch where I can land an actual UI for adding a comment 🎸 🥁 🎹 🔥 🎵 

It will require a config flag which is off everywhere except the local development server. (**TODO**: file a PR to turn it on for the hosted dev site.)

This will only allow you to enter a comment and see it on the page. It *is* saved via the API but if you refresh the page you won't see the comment (that will happen later).

<details>
<summary>Screencast</summary>

</details>

### DIFF
--- a/src/components/CodeView/index.spec.tsx
+++ b/src/components/CodeView/index.spec.tsx
@@ -1,12 +1,9 @@
 import queryString from 'query-string';
 import * as React from 'react';
-import PropTypes from 'prop-types';
 import { Link } from 'react-router-dom';
 import { Location } from 'history';
-import { mount } from 'enzyme';
 import { Store } from 'redux';
 
-import configureStore from '../../configureStore';
 import refractor from '../../refractor';
 import FadableContent from '../FadableContent';
 import LinterMessage from '../LinterMessage';
@@ -18,12 +15,16 @@ import { createInternalVersion } from '../../reducers/versions';
 import { getCodeLineAnchor, getCodeLineAnchorID, mapWithDepth } from './utils';
 import { allowSlowPagesParam, getLanguageFromMimeType } from '../../utils';
 import {
+  SimulateCommentListParams,
+  SimulateCommentableParams,
   createContextWithFakeRouter,
   createFakeExternalLinterResult,
   createFakeLocation,
   fakeExternalLinterMessage,
   fakeVersion,
   shallowUntilTarget,
+  simulateCommentList,
+  simulateCommentable,
   simulateLinterProvider,
 } from '../../test-helpers';
 import styles from './styles.module.scss';
@@ -52,27 +53,20 @@ describe(__filename, () => {
     };
   };
 
-  const getProps = (otherProps = {}) => {
-    return {
+  const render = ({
+    location = createFakeLocation(),
+    ...otherProps
+  }: RenderParams = {}) => {
+    const props = {
       content: 'some content',
       linterMessagesByLine: undefined,
       mimeType: 'mime/type',
       version: createInternalVersion(fakeVersion),
       ...otherProps,
     };
-  };
-
-  const render = ({
-    location = createFakeLocation(),
-    ...otherProps
-  }: RenderParams = {}) => {
-    return shallowUntilTarget(
-      <CodeView {...getProps(otherProps)} />,
-      CodeViewBase,
-      {
-        shallowOptions: createContextWithFakeRouter({ location }),
-      },
-    );
+    return shallowUntilTarget(<CodeView {...props} />, CodeViewBase, {
+      shallowOptions: createContextWithFakeRouter({ location }),
+    });
   };
 
   type RenderWithLinterProviderParams = Partial<LinterProviderInfo> &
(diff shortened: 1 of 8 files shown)


src/components/CodeView/index.tsx
```diff
@@ -134,39 +132,50 @@ export class CodeViewBase extends React.Component<Props> {
...
-                    };
+                  const id = getCodeLineAnchorID(line);
+
+                  let className = styles.line;
+                  let shellRef;
+
+                  if (isLineSelected(id, location)) {
+                    className = makeClassName(
+                      styles.line,
+                      className,
```
bobsilverberg: I'm confused. You set `className` to `styles.line` on line 137, and here you're using both `className` and `styles.line`, so that seems redundant. Can `className` be something other than `styles.line` at this point?

src/components/CodeView/index.spec.tsx
```diff
@@ -93,23 +87,28 @@ describe(__filename, () => {
...
-      },
-      context: {
-        ...options.context,
-        store,
-      },
-    });
+  const simulateCommentableLine = ({
+    addCommentButton,
+    ...props
+  }: Pick<SimulateCommentableParams, 'addCommentButton'> &
```
bobsilverberg: I've not seen `Pick` before. That's a handy feature!

src/components/CodeView/index.spec.tsx
```diff
@@ -170,9 +169,12 @@ describe(__filename, () => {
 
   it('renders plain text code when mime type is not supported', () => {
     const mimeType = 'mime/type';
-    const root = renderWithLinterProvider({ mimeType });
+    const { provider, renderContent } = simulateCommentableLine({
+      mimeType,
+    });
+    const root = renderContent();
```
bobsilverberg: So `provider` will be what we previously had in `root` (i.e., before this patch), and `root` will be a rendered line of content? Maybe these could be named in a way that makes this more obvious?

src/components/CodeView/index.spec.tsx
```diff
@@ -250,34 +261,47 @@ describe(__filename, () => {
 
   it('renders a link for each line number', () => {
     const location = createFakeLocation();
-    const root = renderWithLinterProvider({ content: 'single line', location });
+    const { renderContent, shell } = simulateCommentableLine({
```
bobsilverberg: I'm finding the name `shell` to be a bit confusing. I've gotten used to it seeing it throughout the file, but I wonder if we could come up with a better naming. What if instead of `shell` you used `root`, because `root` is really what we used to call this (and call a similar element in other tests), and then, instead of using `root` to represent a rendered line of code, call that `line` instead. I see you did that at line 224 of this file. I think those names would be easier to understand.

I realize that changing that wholesale means having to replace some instances of `root` with `line` in some of the earlier tests, yielding a larger diff, but I think that would be worth it.

src/components/CodeView/index.spec.tsx
```diff
@@ -471,6 +495,82 @@ describe(__filename, () => {
...
+      });
+
+      expect(shell).toHaveLength(0);
+    });
+
+    it('renders comment list content for each line', () => {
+      const CommentListResult = () => <div />;
+
+      const { shell, renderContent } = simulateInlineCommentList({
+        commentList: <CommentListResult />,
```
bobsilverberg: So `simulateInlineCommentList` automatically puts `commentList` onto each line of content?

This is actually making me a bit worried about this approach with the simulators. Maybe I'm just not getting the abstraction, but it seems like we're setting up data in a way that is totally different from how the app actually works, and then testing the component to see if it looks the way we expect it to. This feels like we could be missing something in terms of how everything works together, but I'm guessing you feel that the tests that already exist for `CommentList` are good enough to cover this?

src/components/CodeView/index.tsx
kumar303: Thanks. It was just a refactoring mistake.

src/components/CodeView/index.spec.tsx
kumar303: Sure, I can do that

src/components/CodeView/index.spec.tsx
kumar303: Disclaimer: render props are really annoying for shallow rendering. They are an exception to how JSX flows and each one requires a special test helper.

> ... it seems like we're setting up data in a way that is totally different from how the app actually works

My intention was to use static types to enforce this. However, I took a closer look and was able to improve / simplify the type safety so that any edit to a `children` prop will nag the developer to fix the test code.

> This feels like we could be missing something in terms of how everything works together, but I'm guessing you feel that the tests that already exist for `CommentList` are good enough to cover this?

Yeah, my philosophy here is that we only need to test that `CommentList` gets *called* correctly. If it gets called correctly then we can assume things will work because it already has its own tests. 

Typically we'd do this by 1) finding a sub-component and asserting its props are correct and 2) relying on static analysis to catch mismatched props. 

In the case of render props, we have to go one extra step and *simulate* how the render prop will be called, based on its signature. That last part about the signature is important. We use static types to enforce it.

kumar303:
Thanks for the review!

Issue #1091, opened 2019-10-01
Add API and thunk to fetch comments


kumar303:
We need an API function and a thunk to fetch comments so they can be rendered (https://github.com/mozilla/addons-code-manager/issues/713)

PR #1092, opened 2019-10-01
Add API and thunk for fetching comments

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/1091

### DIFF
--- a/src/reducers/comments.spec.tsx
+++ b/src/reducers/comments.spec.tsx
@@ -1,11 +1,14 @@
 /* eslint @typescript-eslint/camelcase: 0 */
 import { actions as errorsActions } from './errors';
+import configureStore from '../configureStore';
 import reducer, {
   CommentInfo,
+  ExternalComment,
   actions,
   createCommentKey,
   createEmptyCommentInfo,
   createInternalComment,
+  fetchAndLoadComments,
   initialState,
   manageComment,
   selectCommentInfo,
@@ -14,6 +17,7 @@ import reducer, {
 } from './comments';
 import { createInternalVersion } from './versions';
 import {
+  createFakeCommentsResponse,
   createFakeExternalComment,
   fakeVersion,
   thunkTester,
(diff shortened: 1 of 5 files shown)


willdurand:
>  The diff coverage is `90.56%`.

This is surprisingly low.

src/reducers/comments.spec.tsx
```diff
@@ -637,4 +641,287 @@ describe(__filename, () => {
       });
     });
   });
+
+  describe('abortLoadVersionComments', () => {
+    it('sets isLoading to false', () => {
+      const versionId = 1;
+      let state;
+      state = reducer(state, actions.beginLoadVersionComments({ versionId }));
```
willdurand: It might be helpful to check `isLoading` between these two calls (although I am sure one test case somewhere makes sure `beginLoadVersionComments()` sets `isLoading` to `true`).

src/reducers/comments.spec.tsx
```diff
@@ -637,4 +641,287 @@ describe(__filename, () => {
...
+      const commentId2 = 2;
+      const commentId3 = 2;
+      const commentId4 = 4;
+      const commentId5 = 5;
+      const commentId6 = 6;
+
+      const fileName = 'manifest.json';
+      const line = 321;
+
+      const extComment = ({
```
willdurand: I suppose `ext` means `external` but that's not really obvious. Could you rename this function by chance? `createExternalComment` would work well, I guess.

src/reducers/comments.spec.tsx
```diff
@@ -637,4 +641,287 @@ describe(__filename, () => {
...
+
+      const lineComments = [commentId5, commentId6].map((id) => {
+        return extComment({ id });
+      });
+
+      const _getComments = jest
+        .fn()
+        .mockResolvedValue(
+          createFakeCommentsResponse(
+            versionComments.concat(fileComments).concat(lineComments),
```
willdurand: Could you [`concat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat#Syntax) everything with one call?

```
versionComments.concat(fileComments, lineComments)
```

src/reducers/comments.tsx
```diff
@@ -110,12 +112,24 @@ export const createCommentKey = ({ fileName, line }: CommentKeyParams) => {
 type CommonPayload = CommentKeyParams & { versionId: number };
 
 export const actions = {
+  abortLoadVersionComments: createAction(
+    'ABORT_FETCH_VERSION_COMMENTS',
```
willdurand: I don't understand why the action name is different than the constant.

src/reducers/comments.tsx
```diff
@@ -110,12 +112,24 @@ export const createCommentKey = ({ fileName, line }: CommentKeyParams) => {
...
+      return (payload: { versionId: number }) => resolve(payload);
+    },
+  ),
   abortSaveComment: createAction('ABORT_SAVE_COMMENT', (resolve) => {
     return (payload: CommonPayload) => resolve(payload);
   }),
   beginComment: createAction('BEGIN_COMMENT', (resolve) => {
     return (payload: CommonPayload) => resolve(payload);
   }),
+  beginLoadVersionComments: createAction(
```
willdurand: Same here.

src/reducers/comments.tsx
```diff
@@ -156,6 +170,73 @@ export const selectVersionHasComments = ({
...
+}: {
+  _getComments?: typeof getComments;
+  addonId: number;
+  versionId: number;
+}): ThunkActionCreator => {
+  return async (dispatch, getState) => {
+    const { api: apiState, comments } = getState();
+
+    if (comments.isLoading) {
+      return;
```
willdurand: Would it be worth logging a debug message here?

src/test-helpers.tsx
```diff
@@ -907,3 +908,28 @@ export const dispatchComment = ({
...
+
+export const createFakeApiPage = ({
+  next = null,
+  previous = null,
+  page_size = 25,
+  results = [],
+  count = results.length,
+  page_count = count > page_size ? Math.ceil(count / page_size) : 1,
+}: // eslint-disable-next-line @typescript-eslint/no-explicit-any
+Partial<PaginatedResponse<any>> = {}): PaginatedResponse<any> => {
```
willdurand: Instead of `any`, could you pass a generic type so that in `createFakeCommentsResponse` you can call `createFakeApiPage<ExternalComment>(...)`?

src/reducers/comments.tsx
```diff
@@ -156,6 +170,73 @@ export const selectVersionHasComments = ({
...
+    for (const key of Object.keys(map)) {
+      let line = null;
+      let fileName = null;
+      if (key) {
+        const parts = key.split(':');
+        fileName = parts[0];
+        if (parts[1]) {
+          line = parseInt(parts[1], 10);
+        }
+      }
```
willdurand: I don't understand this part to be honest, probably because I am not familiar with the API response. Maybe add a comment to explain why we want to build a map before calling `setComments()`. Is it because comments are not sorted in the response and we want to batch set comments by filename and line?

willdurand: One more thing: instead of using the key to retrieve `fileName` and `line`, how about changing `map` to:

```js
const map: { [key: string]: {
  fileName: string,
  line: number | null,
  comments: ExternalComment[],
 } } = {};
```

kumar303:
> > The diff coverage is `90.56%`.
> 
> This is surprisingly low.

Every missed line is a default assignment. I used to slap coverage ignore comments on each one but codecov (or something) doesn't honor those anymore (a bug?). The only missed line that wasn't a default assignment was [this helper](https://github.com/mozilla/addons-code-manager/blob/9894659f29c778c1bac97a277f77135ba635ff68/src/test-helpers.tsx#L178) which isn't used anymore, I guess -- not because of my patch, though.

src/reducers/comments.spec.tsx
kumar303: That's already covered by this test: https://github.com/mozilla/addons-code-manager/pull/1092/files/945c2d0f5e8d33a145b2a2db7fc0369da42c3780#diff-680427ffbd1269509fdb6a0d3a6346b1R664

src/reducers/comments.spec.tsx
kumar303: I think a longer name would actually reduce the readability of this since the reader has to parse more words. Since the function is only used for a few lines, I wanted to keep it short.

kumar303: I'll remove the `ext` part

src/reducers/comments.spec.tsx
kumar303: whoops, I can ✨

src/reducers/comments.tsx
kumar303: A refactoring mistake 😎

src/reducers/comments.tsx
kumar303: I usually prefer it that way but in the current implementation (in a forthcoming patch), the debug message would get printed once for each line in the file. So, for a 1000 line file it would be printed 1000 times. This makes me hesitant to log anything.

src/reducers/comments.tsx
kumar303: I'll add a comment. I wanted to minimize the number of dispatches since every Redux dispatch hits all the reducers. The API returns a list of comment objects and what this mapping code does is sort them into version comments, file comments, and line comments to reduce the number of dispatches.

src/test-helpers.tsx
kumar303: Because of the JSX `extends` workaround needed, I'm not sure what to use as the base. I feel like it might work with the JSX workaround. Any ideas?

First attempt:

<details>
<summary>Patch</summary>

```diff
diff --git a/src/api/index.tsx b/src/api/index.tsx
index 19facf0..4399d43 100644
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -352,7 +352,7 @@ export const createOrUpdateComment = async ({
   });
 };
 
-type GetCommentsResponse = PaginatedResponse<ExternalComment[]>;
+export type GetCommentsResponse = PaginatedResponse<ExternalComment[]>;
 
 export const getComments = async ({
   _callApi = callApi,
diff --git a/src/test-helpers.tsx b/src/test-helpers.tsx
index ce37a66..8926bb0 100644
--- a/src/test-helpers.tsx
+++ b/src/test-helpers.tsx
@@ -9,7 +9,7 @@ import { Store } from 'redux';
 import log from 'loglevel';
 import { createAction } from 'typesafe-actions';
 
-import { PaginatedResponse } from './api';
+import { GetCommentsResponse, PaginatedResponse } from './api';
 import configureStore, { ThunkActionCreator } from './configureStore';
 import { ApplicationState } from './reducers';
 import {
@@ -909,15 +909,14 @@ export const dispatchComment = ({
   return dispatchComments({ comments: [comment], ...params });
 };
 
-export const createFakeApiPage = ({
+export const createFakeApiPage = <PaginatedResponseType extends {}>({
   next = null,
   previous = null,
   page_size = 25,
   results = [],
   count = results.length,
   page_count = count > page_size ? Math.ceil(count / page_size) : 1,
-}: // eslint-disable-next-line @typescript-eslint/no-explicit-any
-Partial<PaginatedResponse<any>> = {}): PaginatedResponse<any> => {
+}: Partial<PaginatedResponseType> = {}): PaginatedResponseType => {
   return {
     count,
     next,
@@ -931,5 +930,5 @@ Partial<PaginatedResponse<any>> = {}): PaginatedResponse<any> => {
 export const createFakeCommentsResponse = (
   results = [createFakeExternalComment()],
 ) => {
-  return createFakeApiPage({ results });
+  return createFakeApiPage<GetCommentsResponse>({ results });
 };
```

</details>

<details>
<summary>Error</summary>

```
src/test-helpers.tsx:913:3 - error TS2339: Property 'next' does not exist on type 'Partial<PaginatedResponseType>'.

913   next = null,
      ~~~~

src/test-helpers.tsx:914:3 - error TS2339: Property 'previous' does not exist on type 'Partial<PaginatedResponseType>'.

914   previous = null,
      ~~~~~~~~

src/test-helpers.tsx:915:3 - error TS2339: Property 'page_size' does not exist on type 'Partial<PaginatedResponseType>'.

915   page_size = 25,
      ~~~~~~~~~

src/test-helpers.tsx:916:3 - error TS2339: Property 'results' does not exist on type 'Partial<PaginatedResponseType>'.

916   results = [],
      ~~~~~~~

src/test-helpers.tsx:917:3 - error TS2339: Property 'count' does not exist on type 'Partial<PaginatedResponseType>'.

917   count = results.length,
      ~~~~~

src/test-helpers.tsx:918:3 - error TS2339: Property 'page_count' does not exist on type 'Partial<PaginatedResponseType>'.

918   page_count = count > page_size ? Math.ceil(count / page_size) : 1,
      ~~~~~~~~~~

src/test-helpers.tsx:920:3 - error TS2322: Type '{ count: any; next: any; page_count: any; page_size: any; previous: any; results: any; }' is not assignable to type 'PaginatedResponseType'.
  '{ count: any; next: any; page_count: any; page_size: any; previous: any; results: any; }' is assignable to the constraint of type 'PaginatedResponseType', but 'PaginatedResponseType' could be instantiated with a different subtype of constraint '{}'.

920   return {
      ~~~~~~~~
921     count,
    ~~~~~~~~~~
...
926     results,
    ~~~~~~~~~~~~
927   };
    ~~~~

[10:45:12 AM] Found 7 errors. Watching for file changes.
```

</details>

Second attempt:

<details>
<summary>Patch</summary>

```diff
diff --git a/src/api/index.tsx b/src/api/index.tsx
index 19facf0..4bd4086 100644
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -352,7 +352,8 @@ export const createOrUpdateComment = async ({
   });
 };
 
-type GetCommentsResponse = PaginatedResponse<ExternalComment[]>;
+export type GetCommentsResultsType = ExternalComment[];
+type GetCommentsResponse = PaginatedResponse<GetCommentsResultsType>;
 
 export const getComments = async ({
   _callApi = callApi,
diff --git a/src/test-helpers.tsx b/src/test-helpers.tsx
index ce37a66..58010e9 100644
--- a/src/test-helpers.tsx
+++ b/src/test-helpers.tsx
@@ -9,7 +9,7 @@ import { Store } from 'redux';
 import log from 'loglevel';
 import { createAction } from 'typesafe-actions';
 
-import { PaginatedResponse } from './api';
+import { GetCommentsResultsType, PaginatedResponse } from './api';
 import configureStore, { ThunkActionCreator } from './configureStore';
 import { ApplicationState } from './reducers';
 import {
@@ -909,15 +909,14 @@ export const dispatchComment = ({
   return dispatchComments({ comments: [comment], ...params });
 };
 
-export const createFakeApiPage = ({
+export const createFakeApiPage = <ResultsType extends any[]>({
   next = null,
   previous = null,
   page_size = 25,
   results = [],
   count = results.length,
   page_count = count > page_size ? Math.ceil(count / page_size) : 1,
-}: // eslint-disable-next-line @typescript-eslint/no-explicit-any
-Partial<PaginatedResponse<any>> = {}): PaginatedResponse<any> => {
+}: Partial<PaginatedResponse<ResultsType>> = {}): PaginatedResponse<ResultsType> => {
   return {
     count,
     next,
@@ -931,5 +930,5 @@ Partial<PaginatedResponse<any>> = {}): PaginatedResponse<any> => {
 export const createFakeCommentsResponse = (
   results = [createFakeExternalComment()],
 ) => {
-  return createFakeApiPage({ results });
+  return createFakeApiPage<GetCommentsResultsType>({ results });
 };
```

</details>

<details>
<summary>Errors</summary>

```
src/test-helpers.tsx:916:3 - error TS2322: Type 'never[]' is not assignable to type 'ResultsType'.
  'never[]' is assignable to the constraint of type 'ResultsType', but 'ResultsType' could be instantiated with a different subtype of constraint 'any[]'.

916   results = [],
      ~~~~~~~

[10:51:35 AM] Found 1 error. Watching for file changes.
```

</details>

kumar303: Aha, I figured out a way with a type assertion 🤷‍♂  

See https://github.com/mozilla/addons-code-manager/pull/1092/commits/745d7ce78169ed935732843cc1f3a5141577f9a5

Anyway, it adds essential type safety which I was testing with this:

```diff
diff --git a/src/test-helpers.tsx b/src/test-helpers.tsx
index 9299903..169b9a5 100644
--- a/src/test-helpers.tsx
+++ b/src/test-helpers.tsx
@@ -934,3 +934,5 @@ export const createFakeCommentsResponse = (
 ) => {
   return createFakeApiPage<GetCommentsResponse>({ results });
 };
+
+console.log(createFakeCommentsResponse().results[0].nope);

```

Thanks for the reminder on this. I meant to revisit it 😬

src/reducers/comments.tsx
kumar303: > One more thing: instead of using the key to retrieve `fileName` and `line`, how about changing `map` to:
> 
> ```js
> const map: { [key: string]: {
>   fileName: string,
>   line: number | null,
>   comments: ExternalComment[],
>  } } = {};
> ```

Thanks, this is a lot easier to deal with.

kumar303:
Oh, I have an idea for fixing the `setComments()` action which should let me remove all of the mapping code in `fetchAndLoadComments`. I'll do it in a follow-up.

Issue #1096, opened 2019-10-02
Simplify how comments are fetched and loaded


kumar303:
A thunk to fetch and load comments was added in https://github.com/mozilla/addons-code-manager/issues/1091

This contained some mapping code to minimize the number of redux dispatches when fetching comments. I think we can make changes to the `setComments` action so that this mapping isn't needed.

This supports rendering comments: https://github.com/mozilla/addons-code-manager/issues/713

ioanarusiczki:
@kumar303 I tested yesterday the comment text box for https://github.com/mozilla/addons-code-manager/issues/1082
If there's anything else I can check for this issue please let me know.

kumar303:
Ah, yes, I forgot to mention that QA for this is just regression testing. What you checked already should cover it. Thanks!

PR #1097, opened 2019-10-02
Make setComments() easier in fetchAndLoadComments

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/1096


kumar303:
@willdurand this new `setComments()` signature makes way more sense -- there's no need to create a map of comments anymore.

Issue #713, opened 2019-05-08
Render saved draft comments for files, inline


kumar303:
### Describe the problem and steps to reproduce it:

Consider this imaginary UI flow:

- Sign in as a reviewer
- Open a version browse page
- Click on the line number of a file, enter a comment, press save
- Refresh the browser page

### What happened?

n/a

### What did you expect to happen?

The saved draft comments should be displayed

### Anything else we should know?

This depends on https://github.com/mozilla/addons-server/issues/11374

EnTeQuAk:
With mozilla/addons-server#11374 being fixed, I'm going to remove the `needs: api` label for now. Let me know if you need anything specific on top of that.

ioanarusiczki:
@kumar303 I noticed that on manifest.json this works and I could even send with Postman comments, to modify or delete them.
The problem is with the other files, for the add-on I tested with on .txt or .html files after page refresh I did not see the comments posted - so I should investigate it further with another addon and other file types. 

I will get back with more detailed testing results tomorrow.

kumar303:
ok, let me know what the results are. Maybe there were some API errors in the network console? Also, let me know what add-ons you had this trouble with because I'm not totally sure if commenting works on all versions (maybe the version has to be in a specific state).

ioanarusiczki:
@kumar303 

first example: https://code.addons-dev.allizom.org/en-US/browse/617545/versions/1688861/?path=scripts%2Fshare.js. 

I can see after refresh the comments saved for manifest.json but on the rest of the files they won't show up anymore. 

second example: https://code.addons-dev.allizom.org/en-US/browse/491860/versions/1528357/?path=manifest.json 

I can't see the the saved comments anywhere (not even in manifest.json)

third example: https://code.addons-dev.allizom.org/en-US/browse/494431/versions/1532144/?path=devtools.html

For this add-on I see the comments after refreshing the page on all files I tried with. 

I checked on these addons and I see they all have one version in approved state.

ioanarusiczki:
@kumar303
In the Network I did not find errors 
The only errors shown by Consoles are the two below: 

![the only errors](https://user-images.githubusercontent.com/33448286/66652848-597d7600-ec3f-11e9-8f28-6f7f849ca6e1.png)

kumar303:
[...]

PR #1102, opened 2019-10-03
Render saved comments in Browse mode

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/713

The `CommentList` component gets rendered for every single line in a file. If that line doesn't have comments, it renders `null`. `CommentList` will lazily dispatch a loader (a thunk) *once*. I think performance-wise it's fine because nothing gets dispatched until the thunk starts running and the thunk returns early in case a load is in progress. 

I chose to put the loader here only because this is the component that depends on comments. In the future, more components might depend on loaded comments so we could move it to a shared `<WithLoadedComments>` provider or something.

### DIFF
--- a/src/components/CommentList/index.spec.tsx
+++ b/src/components/CommentList/index.spec.tsx
@@ -9,16 +9,21 @@ import {
 } from '../../reducers/comments';
 import {
   createFakeExternalComment,
+  createFakeThunk,
   dispatchComments,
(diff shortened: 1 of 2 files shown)


kumar303:
@willdurand not a rush but if you have some time to look, I'd like to get this in the tag tomorrow.

Issue #1114, opened 2019-10-09
Add API support for deleting a comment


kumar303:
Add API changes needed to delete a comment. See https://github.com/mozilla/addons-code-manager/issues/714 for the UI.

**For QA**: this change will affect all API calls so please do a rough sanity check on any feature involving API calls (browsing, comparing, viewing linter messages, log in, log out)

ioanarusiczki:
@kumar303 
I verified on dev with FF69(Win10)
Browse
https://code.addons-dev.allizom.org/en-US/browse/492486/versions/1529211/
Compare
https://code.addons-dev.allizom.org/en-US/compare/492486/versions/1529211...1688395/
- no errors in the network for the APIs while navigating through files in Browse and Compare, files load and I did not find anything unusual while browsing except for https://github.com/mozilla/addons-code-manager/issues/1132 which I do not think it is a regression caused by this issue.  In case it's a dupe sorry for re-filing( because it's somehow familiar).

PR #1115, opened 2019-10-09
Add deleteComment() API and support DELETE responses better

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/1114

### DIFF
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -142,6 +142,11 @@ type GetResource<SuccessfulResponse> = {
   successfulResponse: SuccessfulResponse;
 };
 
+type EmptyRequestAndResponse = {
+  requestData: undefined;
+  successfulResponse: string;
+};
+
 export const callApi = async <
   T extends {
     requestData: undefined | {};
@@ -202,7 +207,13 @@ export const callApi = async <
       );
     }
 
-    return await response.json();
+    const contentType = (
+      response.headers.get('content-type') || ''
+    ).toLowerCase();
+
+    return await (contentType.startsWith('application/json')
+      ? response.json()
+      : response.text());
   } catch (error) {
     log.debug('Error caught in callApi():', error);
 
@@ -255,7 +266,10 @@ export const getVersionsList = async ({
 };
 
 export const logOutFromServer = async (apiState: ApiState) => {
-  return callApi<GetResource<{}>>({
+  return callApi<{
+    requestData: undefined;
+    successfulResponse: { ok: boolean };
+  }>({
     apiState,
     // We need to send the credentials (cookies) because the API will return
     // new `Set-Cookie` headers to clear the cookies in the client. Without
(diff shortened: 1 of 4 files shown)


src/api/index.tsx
```diff
@@ -142,6 +142,11 @@ type GetResource<SuccessfulResponse> = {
   successfulResponse: SuccessfulResponse;
 };
 
+type EmptyRequestAndResponse = {
+  requestData: undefined;
+  successfulResponse: string;
```
willdurand: If this represents an empty request and response, why is `successfulResponse` declared as a `string`?

src/api/index.tsx
```diff
@@ -255,7 +266,10 @@ export const getVersionsList = async ({
 };
 
 export const logOutFromServer = async (apiState: ApiState) => {
-  return callApi<GetResource<{}>>({
+  return callApi<{
+    requestData: undefined;
+    successfulResponse: { ok: boolean };
+  }>({
```
willdurand: Shouldn't this be `GetResource<{ok: boolean}>`?

src/api/index.tsx
```diff
@@ -202,7 +207,13 @@ export const callApi = async <
       );
     }
 
-    return await response.json();
+    const contentType = (
+      response.headers.get('content-type') || ''
+    ).toLowerCase();
+
+    return await (contentType.startsWith('application/json')
```
willdurand: Given that we're calling a JSON API, I think it'd be handy to mention that this is to handle `204` responses, as described in https://github.com/whatwg/fetch/issues/113

src/api/index.tsx
kumar303: Because the successful response is an empty string. I guess maybe it could be a string literal? I could try that.

src/api/index.tsx
kumar303: I guess it could be then since it's not a `GET` I had second thoughts ... then I couldn't think of a better name for `GetResource`. `EmptyRequestResponse` seems weird, heh. Maybe I'll think of one.

src/api/index.tsx
kumar303: Well, it's really just any text response, not necessary a 204. Also, I agree with the closure of that issue: it's just not valid json when it's text so an error is expected.

kumar303: Sure, I could add a comment about why the API isn't returning json sometimes

src/api/index.tsx
willdurand: Maybe `GetResource` could be renamed to `Resource`?

kumar303: Hmm, seems sort of vague. I went with `ResponseOnly`, let me know what you think.

src/api/index.tsx
kumar303: Yep, that works ✨

src/api/index.tsx
willdurand: Works for me

willdurand: Work for me

willdurand: Works for me

src/api/index.tsx
willdurand: thanks

kumar303:
Thanks for the quick review

Issue #1121, opened 2019-10-10
Add thunk / reducer for deleting a comment


kumar303:
Add a thunk / reducer for deleting a comment. See https://github.com/mozilla/addons-code-manager/issues/714 for the UI.

PR #1122, opened 2019-10-10
Add thunk / reducer for deleting a comment

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/1121

### DIFF
--- a/src/reducers/comments.tsx
+++ b/src/reducers/comments.tsx
@@ -10,7 +10,7 @@ import {
 } from './versions';
 import {
   createOrUpdateComment,
-  deleteComment as requestCommentDeletion,
+  deleteComment as apiDeleteComment,
   getComments,
   isErrorResponse,
 } from '../api';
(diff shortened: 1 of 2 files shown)


src/reducers/comments.tsx
```diff
@@ -124,6 +136,9 @@ export const actions = {
   beginComment: createAction('BEGIN_COMMENT', (resolve) => {
     return (payload: CommonPayload) => resolve(payload);
   }),
+  beginDeleteComment: createAction('BEGIN_DELETE_COMMENT', (resolve) => {
```
kumar303: This means that the *actual* delete API request has begun.

src/reducers/comments.tsx
```diff
@@ -134,13 +149,19 @@ export const actions = {
     return (payload: CommonPayload & { pendingCommentText: string | null }) =>
       resolve(payload);
   }),
+  considerDeleteComment: createAction('CONSIDER_DELETE_COMMENT', (resolve) => {
```
kumar303: This means that the UI is asking *Do you really want to delete ...?*

src/reducers/comments.tsx
```diff
@@ -209,6 +240,38 @@ export const fetchAndLoadComments = ({
...
+  addonId: number;
+  commentId: number;
+  versionId: number;
+}): ThunkActionCreator => {
+  return async (dispatch, getState) => {
+    const { api: apiState } = getState();
+
+    dispatch(actions.beginDeleteComment({ commentId }));
+
+    const response = await _requestCommentDeletion({
```
willdurand: Nit: I understand why it is named like this, but I find this new name "far" from the original name.

kumar303: And that makes the code harder to follow? Or something else? I changed the alias to `apiDeleteComment` which is closer to the original name but the code is a bit less readable to me. In any case, no biggie -- I don't mind changing it.

Issue #1123, opened 2019-10-11
Send Redux actions to Sentry on error


kumar303:
I was debugging [this error](https://github.com/mozilla/addons-code-manager/issues/713#issuecomment-541081491) but I'm not sure what caused it and I'd really like to see all Redux actions.

This Redux/Sentry middleware looks easy enough to configure: https://github.com/vidit-sh/redux-sentry-middleware

PR #1124, opened 2019-10-11
Send Redux actions to Sentry on error

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/1123

This adds Redux actions to the breadcrumbs and also logs state. I enabled it for production because, hey, why not. We can always disable it.

Example:

Screenshot 2019-10-11 11 55 25

Here is a full example of some actions in breadcrumbs but take note that this includes local development logging that won't be on the hosted sites:

<details>
<summary>Sentry screenshot</summary>

Screenshot_2019-10-11 Error This is a simulation of a thrown error

</details>


kumar303:
We already send logs there as breadcrumbs and it's our own hosted instance (not the service).

willdurand:
> it's our own hosted instance (not the service).

Ah right, I forgot. That's a good point.

I looked at the recent entries in `addons-code-manager-test`. The api token has been filtered out by Sentry, but it means we're sending it over the network. We might want to filter it out before it gets sent.

kumar303:
Yeah, true, we should filter the state for sensitive things. I actually assumed that `redux-sentry-middleware` had filtered `authToken` by default but their docs make no mention of default filtering. I'll try to see if I can do that in a quick an easy way.

willdurand:
> I actually assumed that `redux-sentry-middleware` had filtered `authToken` by default but their docs make no mention of default filtering.

I grep'ed their code and could not find anything either :x

src/configureStore.tsx
```diff
@@ -35,6 +37,53 @@ export type ConnectedReduxProps<A extends Action = AnyAction> = {
...
+    payload: action.payload ? flattenObject(action.payload) : undefined,
+  };
+};
+
+export const redactStateForSentry = (state: ApplicationState) => {
+  // When adding a new state entry to this object, consider the
+  // implication of sending it to Sentry and redact data if necessary.
+  return {
+    accordionMenu: state.accordionMenu,
+    api: { authToken: '[redacted]' },
```
willdurand: Should we spread `state.api` before? (to show how to redact a state entry partially)

kumar303: No. I consciously didn't want to do that. Maybe I will add a comment saying not to do that.

Imagine that someone adds a *new* key to `api` called `api.authTokenSalt` or something. In this situation, I don't want the redacting code to accidentally leak the data. In my patch, the author of such a change will see [this test fail](https://github.com/mozilla/addons-code-manager/blob/390cefc345a74c30c48642859e8c662dc3f82b32/src/configureStore.spec.tsx#L84)  because it's checking that the redacted state at least has the same keys as the real state.

For practical reasons, I didn't choose to explicitly map each item within the other redacted reducers because that would create a maintenance burden. The maintenance burden on the `api` reducers seems reasonable since it will force the author to think about redaction as the `api` shape evolves.

Issue #1152, opened 2019-10-23
Create CommentSummary component


kumar303:
In order to get a basic UI working for finalizing a review (https://github.com/mozilla/addons-code-manager/issues/114), we need a CommentSummary component. This will allow the reviewer to generate some text output of all the comments they've made so far.

PR #1156, opened 2019-10-24
Add a CommentSummary component

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/1152

This ultimately will be rendered in an overlay, shown when the user clicks a button.

<details>
<summary>Example</summary>

Screenshot 2019-10-24 12 11 51

</details>

<details>
<summary>Example in the final overlay (forthcoming patch)</summary>

Screenshot 2019-10-24 12 12 05

</details>


stories/CommentSummary.stories.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { createFakeExternalComment } from '../src/test-helpers';
+import { loremIpsum } from './utils';
+
+const render = ({ comments = [createFakeExternalComment()] } = {}) => {
+  return (
+    <CommentSummary comments={comments.map((c) => createInternalComment(c))} />
+  );
+};
+
+export const createOneLineComment = () => [
```
kumar303: I made this function and the one below because the next story for `CommentSummaryButton` is going to use them.

src/components/CommentSummary/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+        addedHeader = true;
+      }
+      addComment(comment);
+    }
+  }
+
+  const id = 'CommentSummary-textarea';
+
+  return (
+    <form className={className}>
```
bobsilverberg: Just wondering, why did you decide to use a `form` and `textarea` for this, rather than just a `div`? Is it to facilitate copy/paste in the future?

src/components/CommentSummary/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    groups[key].push(comment);
+  }
+
+  for (const key of Object.keys(groups)) {
+    let addedHeader = false;
+
+    for (const comment of groups[key]) {
+      if (!addedHeader) {
```
bobsilverberg: This will repeat the "File:" heading for every line in a given file, which seems like maybe not what we'd want. It seems like we only need the "File:" heading for the first comment in a given file, and only really need to repeat the line which has `filename#lineno` for each line's comment(s).

stories/CommentSummary.stories.tsx
```diff
@@ -0,0 +1,130 @@
...
+      comments: [
+        createFakeExternalComment({
+          filename: null,
+          lineno: null,
+          comment: 'Your add-on looks pretty good.',
+        }),
+      ],
+    });
+  })
+  .add('mixed version and file comments', () => {
```
bobsilverberg: Would it be worthwhile to include a file comment, which is not for a specific line number in here (or in another story)? This one just has all-version comments and a comment for a file on a specific line.

src/components/CommentSummary/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+    });
+
+    if (!groups[key]) {
+      groups[key] = [];
+    }
+
+    groups[key].push(comment);
+  }
+
+  for (const key of Object.keys(groups)) {
```
bobsilverberg: Would it be worth sorting the file comments first by `filename` and then by `lineno` at some point, to guarantee that they render in sequential order, or is it assumed that the `comments` property will always contain comments in the exact order in which we wish to render them?

src/components/CommentSummary/index.spec.tsx
```diff
@@ -0,0 +1,183 @@
...
+        Overall comment(s):
+
+        ${comment1}
+
+        ${comment2}
+      `),
+    );
+  });
+
+  it('renders overall comments mixed with file comments', () => {
```
bobsilverberg: Similar to what I mentioned for the Story, there doesn't seem to be a test which includes overall comments, file comments, and file line comments. I think one of those might be worthwhile.

src/components/CommentSummary/index.tsx
kumar303: The comments will be ordered by how they were entered by the reviewer. This is how github does it and I think it's probably what a reviewer would expect. It's probably a good place to start. We can get some feedback from reviewers to see what they think of it after using it.

src/components/CommentSummary/index.tsx
kumar303: No reason. I have an old school habit of always wrapping form controls in `form` but I researched it just now and it's not necessary for anything. All the reasons for still doing it (server side fallbacks, automatic mobile *Go* buttons, serialization, reset behavior...) don't apply here.

src/components/CommentSummary/index.tsx
kumar303: Jorge's [notes](https://docs.google.com/document/d/1y4crrHxhGejb3eZ6HpjtBdDqONJakfwtfjaOqwq-qZo/edit#heading=h.28bv6iqbbwof) kind of sketched it out this way. I think it might be a little confusing without this header since filenames are not always obvious. In any case, I think this is a good starting point. We can get feedback from reviewers after they use it.

kumar303:
Thanks @bobsilverberg. There are some things about the output I'm not sure of but I think we should get some feedback from reviewers after using it. This is ready for another look.

Issue #114, opened 2019-02-01
Prepare to send review comments to a developer


kumar303:
User story:

> As a code reviewer, I want my code comments to be copied to my review message to the developer.

Reviewers already use a screen like this to enter a review: https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/remote-script-coinhive

The new file viewer (hosted in this repo) will allow them to prepare a review response by commenting directly on the files (https://github.com/mozilla/addons-code-manager/issues/112, https://github.com/mozilla/addons-code-manager/issues/113). The reviewer needs to finalize their review which should either send directly to the developer or save a draft that they can complete on the reviewer tools page.

Step 2 of [these meeting notes](https://docs.google.com/document/d/1y4crrHxhGejb3eZ6HpjtBdDqONJakfwtfjaOqwq-qZo/edit) suggest a flow for submitting comments to a developer.

ioanarusiczki:
@kumar303 I checked this on dev with FF69(Win10):

**Test results**

Functionality:
- the `Comment Summary` button can be opened and the content is displayed as[ here](https://user-images.githubusercontent.com/55398/67706756-5399cb80-f987-11e9-911e-fe3d9e60faed.png) 
- I could close it using `Close` button, `esc` key or clicking outside the box or another action such as clicking on a code line from the code overview or collapsing the side panels. 
- the number of comments is displayed between the add-ons name and `Comment Summary` button. If I delete or add a comment the number is updated correctly. 

The UI:
-  the font type for `Comment Summary` blue button looks a bit different and better in the screenshot https://user-images.githubusercontent.com/55398/67706756-5399cb80-f987-11e9-911e-fe3d9e60faed.png. Perhaps it's just the size different but both ways, that could be adjusted. 

![the button](https://user-images.githubusercontent.com/33448286/67783948-af418300-fa73-11e9-81bb-61e135c40438.png)

- The [screenshot]( https://user-images.githubusercontent.com/55398/67706756-5399cb80-f987-11e9-911e-fe3d9e60faed.png) displays a blue border that I did not see present on [dev](https://code.addons-dev.allizom.org/en-US/browse/390151/versions/1688302/?messageUid=0f2fa4cbd8354a6fb1020f5f7ec2369e&path=mozilla-recommendation.json), it would look nicer that way.

![no blue border](https://user-images.githubusercontent.com/33448286/67784032-cf714200-fa73-11e9-95e1-ef7299f9b6b3.png)

[...]

PR #1170, opened 2019-10-28
Add Navbar actions to summarize all comments

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/114

Notes
- This is a quick implementation of enabling reviewers to finish a review. They will have to copy / paste the text but at least it's a start.
- This is not hidden behind a config flag because comments will [never get loaded](https://github.com/mozilla/addons-code-manager/blob/6c62fb989785ee0802f41a961f3cf27e784b7f74/src/components/CodeView/index.tsx#L194-L207) unless the config flag is enabled.

<details>
<summary>Example screenshots</summary>

Screenshot 2019-10-28 13 26 58
Screenshot 2019-10-28 13 27 07

</details>

### DIFF
--- a/src/components/Navbar/index.tsx
+++ b/src/components/Navbar/index.tsx
@@ -3,8 +3,10 @@ import { Button, Navbar } from 'react-bootstrap';
 import { connect } from 'react-redux';
 
 import { gettext, getLocalizedString } from '../../utils';
+import CommentSummaryButton from '../CommentSummaryButton';
 import LoginButton from '../LoginButton';
 import { ApplicationState } from '../../reducers';
+import { Comment, selectVersionComments } from '../../reducers/comments';
 import { Version, selectCurrentVersionInfo } from '../../reducers/versions';
 import { ConnectedReduxProps } from '../../configureStore';
 import { User, selectCurrentUser, requestLogOut } from '../../reducers/users';
@@ -15,6 +17,7 @@ type PublicProps = {
 };
 
 type PropsFromState = {
+  comments: Comment[] | undefined;
   user: User | null;
   currentVersion: Version | null | undefined | false;
 };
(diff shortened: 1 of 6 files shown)


src/components/Navbar/index.tsx
```diff
@@ -32,17 +35,40 @@ export class NavbarBase extends React.Component<Props> {
...
+      <Navbar className={styles.Navbar} expand="lg" variant="dark">
         <Navbar.Brand className={styles.brand}>
-          {currentVersion && (
-            <span className={styles.addonName}>
-              {getLocalizedString(currentVersion.addon.name)}
-            </span>
-          )}
+          <div className={styles.info}>
+            {currentVersion && (
+              <div className={styles.infoItem}>
```
bobsilverberg: I notice in the test that you used to check for the text of `.${styles.addonName}`, which would only include the add-on's name, but now you're checking for the text of `.${styles.info}` which could include more than the name. I know you're controlling that in the test by making sure there aren't also some comments, but do you think there is any value in still assigning a unique `className` to this `div`? I'm not sure there is, but figured I'd raise it.

kumar303: I try to avoid adding class names purely for testing so that's my one hesitation. If we ever show the comments navbar for zero or undefined comments then this test would break but we could cross this bridge when we come to it.

kumar303:
> I love how little code it takes to make this change to the Navbar!

Heh, well, there were a lot of patches leading up to it.

Thanks for the quick review.

bobsilverberg:
> > I love how little code it takes to make this change to the Navbar!
> 
> Heh, well, there were a lot of patches leading up to it.
> 
> Thanks for the quick review.

I didn't mean to imply that you didn't do a ton of work leading up to this - I know that's why the patch was so small. I was trying to compliment you on that fact. 👍

kumar303:
Oh, I know. I was just trying to say that it wasn't a small patch, unfortunately! React / Redux requires a lot of code and that's something I've been thinking about lately.

Issue #1174, opened 2019-10-29
Indicate when a file or folder has comments


kumar303:
### Describe the problem and steps to reproduce it:

* Add a lot of comments in multiple files
* Begin finishing your review (https://github.com/mozilla/addons-code-manager/issues/114)

### What happened?

If you want to revisit / delete / edit some comments, it's hard to know which files and folders have comments.

### What did you expect to happen?

Just like for linter messages, the file tree should provide some icons as clues. Example:

<img width="551" alt="Screenshot 2019-10-29 15 30 55" src="https://user-images.githubusercontent.com/55398/67806780-8dd79b80-fa61-11e9-8a89-90577c109321.png">

ioanarusiczki:
@kumar303 
Looks good to me. The only UI thing which would be that the message icon is a bit unaligned next to a linter warning but it doesn't bother me.

![a bit unaligned](https://user-images.githubusercontent.com/33448286/68127245-28caec80-ff1e-11e9-8035-cbdfcda8d1c3.png)

as functionality:
- it is available when a message is available for a file. When all messages for a particular file are deleted the message icon disappears.
- clicking on a directory message icon will open/close the files below.
- the tooltip when hovering over the icon for a directory is "This directory contains files with comments".  For a file: "This file has comments". 

Verified on dev with FF70(Win10).

kumar303:
> ... The only UI thing would be that the message icon is a bit unaligned next to a linter warning but it doesn't bother me.

Hmm, yeah. I noticed this, too. The icons are indeed perfectly center-aligned but because the comment icon is a box with a triangle at the bottom, the box appears off center.

PR #1177, opened 2019-10-29
Indicate when a file or folder has comments

kumar303:
~~⚠️  Depends on https://github.com/mozilla/addons-code-manager/pull/1176~~

Fixes https://github.com/mozilla/addons-code-manager/issues/1174

<details>
<summary>Screenshots</summary>

Screenshot 2019-10-29 15 47 34
Screenshot 2019-10-29 15 47 45

</details>

### DIFF
--- a/src/components/FileTreeNode/index.spec.tsx
+++ b/src/components/FileTreeNode/index.spec.tsx
@@ -5,6 +5,7 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 import configureStore from '../../configureStore';
 import { TreefoldRenderPropsForFileTree } from '../FileTree';
+import { actions as commentsActions } from '../../reducers/comments';
 import {
   ExternalLinterMessage,
   createInternalMessage,
@@ -17,6 +18,7 @@ import {
 } from '../../reducers/versions';
 import {
   createExternalVersionWithEntries,
+  createFakeExternalComment,
   createFakeExternalLinterResult,
   createFakeRef,
   createStoreWithVersion,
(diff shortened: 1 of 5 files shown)


src/components/FileTreeNode/index.spec.tsx
```diff
@@ -403,6 +445,57 @@ describe(__filename, () => {
...
+      ...getTreefoldRenderProps({ id: dir }),
+    });
+
+    const icons = root.find(`.${styles.nodeIcons}`).find(FontAwesomeIcon);
+    expect(icons).toHaveLength(1);
+
+    expect(icons.at(0).key()).toEqual('comments-icon');
+  });
+
+  it('adds a linter and comment icon for a file', () => {
```
bobsilverberg: Should there also be a test for "adds a linter and comment icon for a folder", or is that unnecessary?

src/components/FileTreeNode/index.tsx
```diff
@@ -204,10 +207,26 @@ export class FileTreeNodeBase<TreeNodeType> extends React.Component<Props> {
...
+
+    if (hasComments) {
+      nodeIcons.push(
+        <FontAwesomeIcon
+          fixedWidth
+          key="comments-icon"
+          flip="horizontal"
+          icon={['fas', 'comment-alt']}
+          size="1x"
+          title={gettext('This file has comments')}
```
bobsilverberg: This ends up putting this title on folders as well as files. For linter notices we have a `getTitleForType` function which ensures that files are titled as files and folders are titled as folders, so maybe we need something similar here? Or maybe just a ternary right here would do the trick?

kumar303:
Thanks for the review

Issue #1093, opened 2019-10-01
Show more than 25 saved comments


kumar303:
### Describe the problem and steps to reproduce it:

- Sign in
- Add more than 25 comments to a file
- Refresh the page

### What happened?

Only the first 25 saved comments are shown

### What did you expect to happen?

All saved comments should be shown

### Anything else we should know?

Rendering saved comments was added in https://github.com/mozilla/addons-code-manager/issues/713

alexkayabula:
Hello @kumar303, I am trying to reproduce this issue but when I run the local app and sign in I keep getting this error `Unexpected status for GET /reviewers/addon/494431/versions/1532144/?lang=en-US: 403`.
It is associated with `{"detail":"You do not have permission to perform this action."}`.
Any help is much appreciated.

ioanarusiczki:
@kumar303 

Frontend testing:

I verified it with https://code.addons-dev.allizom.org/en-US/browse/618280/versions/1689601/?path=popup%2Fcss%2Fbootstrap.min.css

I don't know if there's a limit, but I went up to 50 comments on different files. The number is updated correctly and comments are displayed after a page refresh. 

I did some backend testing too :

I posted a comment using the [API ](https://addons-dev.allizom.org/api/v4/reviewers/addon/618280/versions/1689601/draft_comments/)which was created with a 201 status. The frontend on dev counted it too as being available (the 49th comment made) yet it is not displayed because the `lineno`:8 does not exist for the file
https://code.addons-dev.allizom.org/en-US/browse/618280/versions/1689601/?path=popup%2Fcss%2Fbootstrap.min.css
POST 
`{   "comment":"testing the plus 49 comments",
	"filename":"popup/css/bootstrap.min.css",
	"lineno":8
}`

The same thing happens if I don't specify the path of the file correctly.

`{   "comment":"testing the plus 51 comments",
	"filename":"bootstrap.min.css",
	"lineno":8
}`

![Comments51](https://user-images.githubusercontent.com/33448286/68132163-e35eed00-ff26-11e9-8e21-9e162f3c734b.png)

I think this is a known API limitation (it won't detect if code line exists or not, same for a file).

kumar303:
Thanks.

[...]

PR #1190, opened 2019-10-31
Load all API pages when fetching saved comments

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/1093

Example of loading more than 25 saved comments:

Screenshot 2019-10-31 15 04 39

### DIFF
--- a/src/api/index.spec.tsx
+++ b/src/api/index.spec.tsx
@@ -48,6 +48,23 @@ describe(__filename, () => {
     return api;
   };
 
+  const setUpFetchAllPagesMock = () => {
+    const _fetchAllPages = jest.fn();
+
+    type GetNextResponse = Parameters<typeof fetchAllPages>[0];
+
+    // This can be used by a test to simulate the interface of
+    // fetchAllPages(getNext) whereby it invokes getNext(nextPageUrl).
+    const getNextResponse: GetNextResponse = (nextPageUrl) => {
+      expect(_fetchAllPages).toHaveBeenCalled();
+
+      const getNext = _fetchAllPages.mock.calls[0][0] as GetNextResponse;
+      return getNext(nextPageUrl);
+    };
+
+    return { _fetchAllPages, getNextResponse };
+  };
+
   describe('callApi', () => {
     const callApiWithDefaultApiState = (params = {}) => {
       return callApi({ apiState: defaultApiState, endpoint: '/', ...params });
@@ -869,7 +886,7 @@ describe(__filename, () => {
   describe('getAllComments', () => {
     it('calls fetchAllPages', async () => {
       const _getComments = jest.fn();
-      const _fetchAllPages = jest.fn();
+      const { _fetchAllPages, getNextResponse } = setUpFetchAllPagesMock();
       const addonId = nextUniqueId();
       const apiState = defaultApiState;
       const versionId = nextUniqueId();
@@ -882,13 +899,8 @@ describe(__filename, () => {
         versionId,
       });
 
-      expect(_fetchAllPages).toHaveBeenCalled();
-
-      // Simulate the interface of fetchAllPages whereby it invokes
-      // getNext(nextPageUrl).
       const nextPageUrl = 'endpoint/?page=2';
-      const getNext = _fetchAllPages.mock.calls[0][0];
-      getNext(nextPageUrl);
+      getNextResponse(nextPageUrl);
 
       expect(_getComments).toHaveBeenCalledWith({
         addonId,
--- a/src/api/index.spec.tsx
+++ b/src/api/index.spec.tsx
@@ -884,15 +884,16 @@ describe(__filename, () => {
 
       expect(_fetchAllPages).toHaveBeenCalled();
 
+      // Simulate the interface of fetchAllPages whereby it invokes
+      // getNext(nextPageUrl).
+      const nextPageUrl = 'endpoint/?page=2';
       const getNext = _fetchAllPages.mock.calls[0][0];
-
-      const nextUrl = 'endpoint/?page=2';
-      getNext(nextUrl);
+      getNext(nextPageUrl);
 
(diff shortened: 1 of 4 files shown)


src/api/index.tsx
```diff
@@ -184,17 +178,36 @@ export const callApi = async <
     body = JSON.stringify(bodyData);
   }
 
-  type QueryWithLang = {
-    [key: string]: string;
-  };
+  if (endpointUrl && endpoint) {
```
bobsilverberg: I feel like we've looked at a similar thing before, but isn't there a way to define this with TypeScript?

src/api/index.spec.tsx
```diff
@@ -773,9 +845,57 @@ describe(__filename, () => {
...
+        apiState,
+        versionId,
+      });
+
+      expect(_fetchAllPages).toHaveBeenCalled();
+
+      const getNext = _fetchAllPages.mock.calls[0][0];
+
+      const nextUrl = 'endpoint/?page=2';
+      getNext(nextUrl);
```
bobsilverberg: This feels a bit odd to me. IIUC, `getNext` is the function that would have been called by `_fetchAllPages`? So what you're doing is taking that function and calling it with an expected value so that you can then assert that it did what you expected it to do (i.e., call `_getComments` with the expected value). This all seems to make sense, except for the fact that you're manually calling `getNext`. Shouldn't `_fetchAllPages` have already called `getNext`, and if so wouldn't it be nicer if we could just assert about what happened at that time, rather than having to manually call `getNext` ourselves?

I gather this is being done because of the complicated way this all works under the covers which makes it tricky to test, but I thought I'd raise it to see if it's possible to do it the way I'm suggesting above.

src/api/index.tsx
```diff
@@ -364,26 +435,52 @@ export const createOrUpdateComment = async ({
...
+  let endpoint;
+
+  if (!endpointUrl) {
+    endpoint = `reviewers/addon/${addonId}/versions/${versionId}/draft_comments`;
+  }
+
   return _callApi<ResponseOnly<GetCommentsResponse>>({
     apiState,
-    endpoint: `reviewers/addon/${addonId}/versions/${versionId}/draft_comments`,
+    endpointUrl: endpointUrl || undefined,
```
bobsilverberg: I'm having a bit of trouble understanding why `getComments` needs to have special handling for `endpointUrl`, and why we cannnot just handle that in `callApi`. Couldn't `callApi` decide what to do based on what it receives for both `endpoint` and `endpointUrl`? I.e., if there is an `endpointUrl` use that, otherwise use `endpoint`?

kumar303: > I'm having a bit of trouble understanding why `getComments` needs to have special handling for `endpointUrl`, and why we cannnot just handle that in `callApi`. Couldn't `callApi` decide what to do based on what it receives for both `endpoint` and `endpointUrl`? I.e., if there is an `endpointUrl` use that, otherwise use `endpoint`?

`callApi()` could make that decision but how would we know that it's the right decision? It could be wrong and that would introduce subtle bugs. The caller should be the one to decide.

Just as an exercise, I committed https://github.com/mozilla/addons-code-manager/pull/1190/commits/fb15d6004eb812577fcc85426f3e9fd595ec8642 to introduce the behavior you described but it feels more risky and I don't think it's helpful. It allowed me to remove an if statement from `getComments()` but overall the diff adds more code than it takes away.

src/api/index.spec.tsx
kumar303: I added a comment to make this test code more clear.

It's treating `fetchAllPages` as an external dependency with a known interface: a function taking a callback parameter. It is safe to mock this because TS will enforce adherence to the interface.

The test then simulates the interface which makes the test a lot simpler. It doesn't have to set up a mock response that returns a next page URL and doesn't have to worry about avoiding infinite fetching loops. If the implementation of `fetchAllPages` were to later change, this test would not break as long as the interface didn't change.

kumar303: 💡 I moved this code to a helper since it's likely that other API tests would need it in the future. This should make the test code a little easier to understand, as well. I added more type safety to it so that the test code will produce an error like this for incorrect usage:

```
src/api/index.spec.tsx:906:7 - error TS2554: Expected 1 arguments, but got 0.

906       getNextResponse();
          ~~~~~~~~~~~~~~~~~

  src/api/index.tsx:264:5
    264     nextUrl: PaginatedResponse<ResultsType[]>['next'],
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    An argument for 'nextUrl' was not provided.
```

src/api/index.tsx
kumar303: I thought we tried before but abandoned it because it required a [discriminated union](https://basarat.gitbooks.io/typescript/docs/types/discriminated-unions.html) which was too complicated? We could maybe do it with [function overloading](https://howtodoinjava.com/typescript/function-overloading/) but our code style enforces anonymous functions which don't support that.

I dunno, adding an if statement and one test seems like a good way to solve it.

kumar303: Well, huh, I can't find the discussion we had on this. If you had come up with a typed solution for this that I'm forgetting about, let me know.

kumar303:
Thanks for all the feedback. I can do follow-ups if needed.

src/api/index.tsx
bobsilverberg: I don't have the exact syntax handy, and I think you're right that maybe we looked at using a discriminated union, but I'm not sure why we'd need one. The basic idea is this: create two types, each of which look like `CallApiParams` except that one (which we'll call `TypeA`) only has an `endpoint` property and the other one (which we'll call `TypeB`) only has an `endpointUrl` property. Then in place of `CallApiParams` we use `TypeA | TypeB`. Will that not work?

Issue #1209, opened 2019-11-06
Always use the correct TypeScript version in VSCode


kumar303:
VSCode will try to use its own version of TypeScript by default. Let's make sure it always uses the one from `node_modules` to keep all developers in sync.

PR #1210, opened 2019-11-06
Use the correct TypeScript version in VSCode

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/1209

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,8 @@ typings/
 
 # Storybook
 storybook-static/
+
+# IDE files
+# VSCode does not read local project files that might execute binaries by
+# default due to security so this config wouldn't have an effect, anyway.
+.vscode
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    // Always use the locally installed version of TypeScript.
+    "typescript.tsdk": "node_modules/typescript/lib"
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,0 @@
-{
-    // Always use the locally installed version of TypeScript.
-    "typescript.tsdk": "node_modules/typescript/lib"
-}
--- a/README.md
+++ b/README.md
@@ -93,6 +93,14 @@ Here are some tips for solving performance problems in addition to what's alread
 - When a React component is taking a long time to render and you don't see any other components underneath it, it's time to switch away from the _User Timing_ tab of the profiler to the _Main_ tab so you can look at actual function executions. Clicking on a function will give you information about its source.
 - You can try using the [React devtool extension](https://reactjs.org/docs/optimizing-performance.html#profiling-components-with-the-devtools-profiler) for profiling but it doesn't provide a great timeline so it's hard to visualize overall slowness.
 
+## Setting up VSCode
+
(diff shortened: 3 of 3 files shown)


kumar303:
If I solved it just for myself then someone else could get tripped up by it. It's important that everyone is developing against the same version of TypeScript. Thus, it seems to make more sense as a shared config, no? If someone is *not* developing in VSCode then the file has no effect.

kumar303:
I view it like the local `.gitignore`. We *could* tell everyone to edit their global `'~/.gitignore` file but, why? We can just include a local one for the project and it will work the same on all machines.

bobsilverberg:
FWIW, I think it's fine to include this file (and I do use VSCode).

willdurand:
The way I see it is that we should only have config files related to the project itself. We have a `.gitignore` for this project, excluding artifacts generated by the project, but we don't exclude all kind of weird files (like swap files, vim backup files, etc).

As for VSCode, it seems to me that what needs to be fixed is VSCode itself, not the project. That looks like something no one wants ever. I can run the project successfully with a different IDE, without having to fix non-senses 😛 

Next time, we'll have to add config files for the Jetbrains IDE (WebStorm?), then Eclipse, then X, then Z, etc. I would like to avoid that, ideally. If that's easier for you, then I guess it is fine. We have too many (hidden) files in the root dir already anyway..

kumar303:
OMG waaat. Even by committing this local config file, it wouldn't get applied on a different machine because of [securitay](https://github.com/microsoft/vscode/issues/30069#issuecomment-312732928). OK, I guess we should just ignore the local config then! I also added docs because this is a pretty bad gotcha and I dunno VSCode seems popular.

.gitignore
```diff
@@ -73,3 +73,8 @@ typings/
 
 # Storybook
 storybook-static/
+
+# IDE files
+# VSCode does not read local project files that might execute binaries by
+# default due to security so this config wouldn't have an effect, anyway.
+.vscode
```
bobsilverberg: Why do we need this if no `.vscode` file is being added? Or does the procedure added to the docs create one? I didn't see any changes to a file in `.vscode` locally, but I did already have a file there.

kumar303: To fix the local TS problem, one needs to create a `.vscode` directory but this shouldn't ever get checked in, which is why I ignored it.

bobsilverberg: Do you really? I thought is was persisted via localstorage.

kumar303: I'm not sure what you mean. Is what persisted in local storage?

bobsilverberg: My understanding from reading the issue is that when you use the UI in VSCode to indicate you want to use the project's version of TS, that choice is stored in localstorage, so I'm not sure what the file in `.vscode` is for.

kumar303: When I used the UI to make the choice, VSCode saved the configuration to a local `.vscode/settings.json` file. The [security issue](https://github.com/microsoft/vscode/issues/30069#issuecomment-312732928) was saying that local storage is used to know if a user has interacted with the UI rather than opened some project with an existing `.vscode/settings.json` file, I think.

willdurand: I would argue that this folder should be ignored in your global gitignore: https://help.github.com/en/github/using-git/ignoring-files#create-a-global-gitignore 🙂

Issue #1106, opened 2019-10-07
Continuous " Saving" state is displayed when saving an empty comment 


ioanarusiczki:
Followup for https://github.com/mozilla/addons-code-manager/issues/1082

STR:
1. Load a file for an add-on in code-manager such as https://code.addons-dev.allizom.org/en-US/browse/617545/versions/1688861/
2. Submit an empty comment (just click "Save")

Expected result:
An empty comment cannot be submitted - the field requires at least one char and then the "Save" button becomes enabled.

Actual result:
A "saving" state. 

Note:
This issue is reproducible on dev with FF69(Win10). 

![issues2](https://user-images.githubusercontent.com/33448286/66318352-d1882b00-e923-11e9-9ee9-4b1f4a2d5b9d.png)

Please note that I can create with Postman using the [API](https://addons-server.readthedocs.io/en/latest/topics/api/reviewers.html#drafting-comments) an empty comment.
![API response](https://user-images.githubusercontent.com/33448286/66317860-fa5bf080-e922-11e9-9d48-fce118e4c894.png)

kumar303:
The cause of this is `Error: Either cannedResponseId or comment must be specified` which results in an unhandled promise rejection, coming from this:

https://github.com/mozilla/addons-code-manager/blob/980d88413d4457f0b545d2dd413d9195934da6b4/src/api/index.tsx#L411

The error is thrown from the [`manageComment`](https://github.com/mozilla/addons-code-manager/blob/980d88413d4457f0b545d2dd413d9195934da6b4/src/reducers/comments.tsx#L364) thunk.

kumar303:
I filed https://github.com/mozilla/addons-code-manager/issues/1239 to make this unexpected error more visible.

ioanarusiczki:
@kumar303 Verified on dev with FF70(Win10). 
- an empty comment can be saved until the fix of https://github.com/mozilla/addons-server/issues/12865

![empty comment](https://user-images.githubusercontent.com/33448286/68941431-8973e800-07ae-11ea-9c01-11a6c39be16b.gif)

PR #1238, opened 2019-11-13
Pass empty comments to the API

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/1106 by not throwing [this error](https://github.com/mozilla/addons-code-manager/blob/980d88413d4457f0b545d2dd413d9195934da6b4/src/api/index.tsx#L411).

After this patch, the UI will let you submit an empty comment until https://github.com/mozilla/addons-server/issues/12865 is fixed. Once that API issue is fixed, the error will show up at the top like other errors. I also confirmed that the comment form will remain open and editable when the error is displayed, allowing the user to correct the mistake and try again.

### DIFF
--- a/src/components/Comment/index.spec.tsx
+++ b/src/components/Comment/index.spec.tsx
@@ -182,6 +182,28 @@ describe(__filename, () => {
     });
   });
 
+  it('can submit an empty comment', () => {
+    const store = configureStore();
+    const dispatchSpy = spyOn(store, 'dispatch');
+
+    const fakeThunk = createFakeThunk();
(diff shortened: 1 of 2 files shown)


bobsilverberg:
Have you considered disallowing submission of an empty comment instead of taking this approach? Why should the app even allow a user to submit an empty comment?

kumar303:
I considered it but it would involve adding a lot of custom code. We already have code to deal with the API error and present it in a standard way so why bother? I don't think it's worth it. We took a similar approach on `addons-frontend` (i.e. let the API report errors) and I think it worked well.

PR #1240, opened 2019-11-14
Don't trim diffs by default

### LINKED ISSUE #1230: Disable trimming of large (slow) diffs

* Load a large diff (see https://github.com/mozilla/addons-code-manager/issues/971) in compare mode

### What happened?

The diff should not be truncated

The diff is currently truncated as an attempt to speed up the page

We'd like to show the entire diff as a workaround to https://github.com/mozilla/addons-code-manager/issues/971 which is about broken navigation due to the trimmed file.

bobsilverberg:
Fixes #1230

### DIFF
--- a/src/components/SlowPageAlert/index.tsx
+++ b/src/components/SlowPageAlert/index.tsx
@@ -11,7 +11,8 @@ import {
 type GetTextForSlowState = (allowSlowPages: boolean) => string;
 
 export type PublicProps = {
-  _shouldAllowSlowPages?: (location: Location) => boolean;
+  _shouldAllowSlowPages?: typeof shouldAllowSlowPages;
+  defaultAllowSlowPagesToTrue?: boolean;
   getLinkText: GetTextForSlowState;
   getMessage: GetTextForSlowState;
   location: Location;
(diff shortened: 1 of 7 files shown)


src/components/SlowPageAlert/index.tsx
```diff
@@ -11,19 +11,24 @@ import {
 type GetTextForSlowState = (allowSlowPages: boolean) => string;
 
 export type PublicProps = {
-  _shouldAllowSlowPages?: (location: Location) => boolean;
+  _shouldAllowSlowPages?: typeof shouldAllowSlowPages;
+  defaultAllowSlowPagesToTrue?: boolean;
```
kumar303: How about `allowSlowPagesByDefault`?

src/utils.tsx
```diff
@@ -120,6 +120,17 @@ export const makeReviewersURL = ({
   return path;
 };
 
-export const shouldAllowSlowPages = (location: Location) => {
-  return queryString.parse(location.search)[allowSlowPagesParam] === 'true';
+export const shouldAllowSlowPages = ({
+  defaultToTrue = false,
```
kumar303: how about `allowByDefault`?

bobsilverberg:
Good suggestions, thanks @kumar303.

PR #1244, opened 2019-11-15
Add a link back to Reviewer Tools

### LINKED ISSUE #984: Add Reviewer Tools back link on Browse and Compare pages
STR:
1. Load dev and go to a Browse page and then a Compare page such as 
https://code.addons-dev.allizom.org/en-US/browse/616911/versions/1688344/
https://code.addons-dev.allizom.org/en-US/compare/616909/versions/1688312...1688345/
2. Observe the top of the page (the dark grey area)

Expected result:
There is a link to the reviewer's page such as https://reviewers.addons-dev.allizom.org/en-US/reviewers/review/webext-csp78

Actual result:
The link is not available. 

Notes:
reproducible on -dev and -stage with FF68(Win10).

bobsilverberg:
Fixes #984 

I wasn't sure whether we wanted a link *added* to the Navbar, or whether we wanted to use the existing label for the add-on name *as* the link back to the reviewer tools, so I did the former. It's easy enough to change this so no new link is added and we simply linkify the existing label. Because I wasn't sure about this, I didn't apply any styling to the new link, which is why it's currently blue. I will fix that if we decide to keep the new link.

Screenshot:

Screenshot 2019-11-15 12 12 08

### DIFF
--- a/src/components/Navbar/index.spec.tsx
+++ b/src/components/Navbar/index.spec.tsx
@@ -23,18 +23,23 @@ import Navbar from '.';
 describe(__filename, () => {
   type RenderParams = {
     _requestLogOut?: typeof requestLogOut;
+    reviewersHost?: string;
     store?: Store;
   };
 
   const render = ({
     _requestLogOut = jest.fn(),
+    reviewersHost,
     store = configureStore(),
   }: RenderParams = {}) => {
     // TODO: Use shallowUntilTarget()
     // https://github.com/mozilla/addons-code-manager/issues/15
-    const root = shallow(<Navbar _requestLogOut={_requestLogOut} />, {
-      context: { store },
-    }).shallow();
+    const root = shallow(
+      <Navbar _requestLogOut={_requestLogOut} reviewersHost={reviewersHost} />,
+      {
+        context: { store },
+      },
+    ).shallow();
 
     return root;
   };
@@ -58,7 +63,25 @@ describe(__filename, () => {
       });
       const root = render({ store });
 
-      expect(root.find(`.${styles.info}`)).toHaveText(addonName);
+      expect(root.find(`.${styles.addonName}`)).toHaveText(addonName);
+    });
+
+    it('renders a link to reviewer tools', () => {
+      const reviewersHost = 'https://example.com';
+      const slug = 'some-slug';
+      const store = createStoreWithVersion({
+        version: {
+          ...fakeVersion,
+          addon: { ...fakeVersionAddon, slug },
+        },
+        makeCurrent: true,
+      });
+      const root = render({ reviewersHost, store });
+
+      expect(root.find(`.${styles.reviewerToolsLink}`)).toHaveProp(
+        'href',
+        `${reviewersHost}/reviewers/review/${slug}`,
+      );
     });
   });
 
@@ -66,7 +89,13 @@ describe(__filename, () => {
     it('does not render addon name', () => {
       const root = render();
 
-      expect(root.find(`.${styles.info}`)).toHaveText('');
+      expect(root.find(`.${styles.addonName}`)).toHaveLength(0);
+    });
+
+    it('does not render a link to reviewer tools', () => {
+      const root = render();
+
+      expect(root.find(`.${styles.reviewerToolsLink}`)).toHaveLength(0);
     });
   });
 
--- a/src/components/Navbar/index.spec.tsx
+++ b/src/components/Navbar/index.spec.tsx
@@ -1,4 +1,3 @@
-import { shallow } from 'enzyme';
 import React from 'react';
 import { Store } from 'redux';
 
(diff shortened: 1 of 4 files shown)


src/components/Navbar/index.tsx
```diff
@@ -56,16 +59,30 @@ export class NavbarBase extends React.Component<Props> {
...
-              <div className={styles.infoItem}>
-                {getLocalizedString(currentVersion.addon.name)}
-              </div>
+              <>
+                <div
+                  className={makeClassName(styles.infoItem, styles.addonName)}
+                >
+                  {getLocalizedString(currentVersion.addon.name)}
+                </div>
+                <div className={styles.infoItem}>
```
kumar303: It would make more sense to me if this were first in the list, maybe?

src/components/Navbar/index.tsx
```diff
@@ -56,16 +59,30 @@ export class NavbarBase extends React.Component<Props> {
...
+                  <a
+                    className={styles.reviewerToolsLink}
```
kumar303: The blue link is hard to read against the dark gray. I suggest making the text white and underlining it to make it clear that it's a link.

src/components/Navbar/index.tsx
```diff
@@ -56,16 +59,30 @@ export class NavbarBase extends React.Component<Props> {
...
+                    href={`${reviewersHost}/reviewers/review/${currentVersion.addon.slug}`}
+                    rel="noopener noreferrer"
+                    target="_blank"
+                  >
+                    {gettext('Back to reviewer tools')}
```
kumar303: How about just Reviewer Tools? That should be clear enough.

src/components/Navbar/index.tsx
```diff
@@ -27,6 +29,7 @@ type Props = PublicProps & PropsFromState & ConnectedReduxProps;
 export class NavbarBase extends React.Component<Props> {
   static defaultProps = {
     _requestLogOut: requestLogOut,
+    reviewersHost: process.env.REACT_APP_REVIEWERS_HOST,
```
kumar303: `.env.common-local` needs to get `REACT_APP_REVIEWERS_HOST=https://reviewers.addons-dev.allizom.org`.

Currently, it's using the prod setting.

src/components/Navbar/index.tsx
kumar303: You could add an icon in there to make the shortened *Reviewer Tools* work better:

```diff
diff --git a/src/components/Navbar/index.tsx b/src/components/Navbar/index.tsx
index d676be6..55a440a 100644
--- a/src/components/Navbar/index.tsx
+++ b/src/components/Navbar/index.tsx
@@ -2,6 +2,7 @@ import makeClassName from 'classnames';
 import * as React from 'react';
 import { Button, Navbar } from 'react-bootstrap';
 import { connect } from 'react-redux';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 import { gettext, getLocalizedString } from '../../utils';
 import CommentSummaryButton from '../CommentSummaryButton';
@@ -79,6 +80,7 @@ export class NavbarBase extends React.Component<Props> {
                     rel="noopener noreferrer"
                     target="_blank"
                   >
+                    <FontAwesomeIcon icon={['fas', 'arrow-left']} />
                     {gettext('Back to reviewer tools')}
                   </a>
                 </div>
```

bobsilverberg:
Thanks for the review @kumar303. Here's what the newer version of the screen looks like:

Screenshot 2019-11-18 13 43 56

src/components/Navbar/styles.module.scss
```diff
@@ -59,3 +60,16 @@ $bg-color: $dark-gray;
...
+
+.reviewerToolsLink {
+  &,
+  &:active,
+  &:hover {
+    color: $white;
+    text-decoration: underline;
+  }
+
+  svg {
```
kumar303: We had a rule on `addons-frontend` to always use a css class when selecting elements. I think it's a good rule to follow because subtle performance problems can creep in without it.

src/components/Navbar/styles.module.scss
```diff
@@ -1,4 +1,5 @@
 @import '../../scss/variables';
+@import '../../scss/mixins';
```
kumar303: I don't see any new mixins used. Is this still needed?

src/components/Navbar/index.spec.tsx
```diff
@@ -23,18 +23,23 @@ import Navbar from '.';
 describe(__filename, () => {
   type RenderParams = {
     _requestLogOut?: typeof requestLogOut;
+    reviewersHost?: string;
     store?: Store;
   };
 
   const render = ({
     _requestLogOut = jest.fn(),
+    reviewersHost,
```
kumar303: Since this is optional, you could pass it in as `...props`. That would also make this `render()` more consistent with most others.

src/components/Navbar/index.spec.tsx
```diff
@@ -23,18 +23,23 @@ import Navbar from '.';
 describe(__filename, () => {
   type RenderParams = {
     _requestLogOut?: typeof requestLogOut;
+    reviewersHost?: string;
```
kumar303: This must be an older test file. I suggest changing it to `RenderParams = Partial<PublicProps> & Partial<DefaultProps>`.

src/components/Navbar/index.spec.tsx
```diff
@@ -23,18 +23,23 @@ import Navbar from '.';
...
     _requestLogOut = jest.fn(),
+    reviewersHost,
     store = configureStore(),
   }: RenderParams = {}) => {
     // TODO: Use shallowUntilTarget()
     // https://github.com/mozilla/addons-code-manager/issues/15
-    const root = shallow(<Navbar _requestLogOut={_requestLogOut} />, {
-      context: { store },
-    }).shallow();
+    const root = shallow(
```
kumar303: It looks like you can switch this to `shallowUntilTarget()` and remove the comment up above.

src/components/Navbar/styles.module.scss
bobsilverberg: Ah, no, I removed the mixin that I added.

Issue #1249, opened 2019-11-18
Extract overlay / popover from comment summary


kumar303:
I'd like to extract and re-use the comment summary overlay for putting compare-version controls in the nav bar (https://github.com/mozilla/addons-code-manager/issues/758).

**For QA**: Please check for regressions in the comment summary button and overlay

ioanarusiczki:
@kumar303 Checking withhttps://code.addons-dev.allizom.org/en-US/browse/502724/versions/1541510/?allowSlowPages=true&messageUid=c94a6ffe04da45bfac4a17ebf1dc3557&path=content_scripts%2Fcheck-inventory.js I did not see anything unusual. 
Verified on dev with FF70(Win10).

PR #1250, opened 2019-11-18
Extract PopoverButton from comment summary

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/1249


kumar303:
I've noticed that before but it's inconsistent for me. Sometimes it works for me the first time 🤷‍♂ In any case, it seems like something to do with storybook and probably the complexity of doing overlays (the detached dom node, etc). It doesn't happen in the actual app.

Issue #1247, opened 2019-11-18
Selecting a file in the navigation scrolls to line of previous file


wagnerand:
### Describe the problem and steps to reproduce it:

I noticed that selecting a file in the navigation scrolls to the same line of the previous file.

1. Open a file
2. Scroll down
3. Click on a different file in the navigation

### What happened?

The window is scrolling to the same line as before.

### What did you expect to happen?

Stay at the top of the file.

kumar303:
Hmm. I think maybe it always behaved like this but pre-loading makes it happen all the time now. To confirm, this only applies to navigating to a pre-loaded file, like the next in list, right?

wagnerand:
> To confirm, this only applies to navigating to a pre-loaded file, like the next in list, right?

Yes, I can confirm that.

AlexandraMoga:
Verified fixed on -dev both for the browse and compare view.

When another file is selected from the file tree (also applies to keyboard navigation), the top of the file is being displayed:

![2019-11-21_13-20-16](https://user-images.githubusercontent.com/31961530/69333920-50ca8780-0c62-11ea-9641-7a4b8b22d9bc.gif)

PR #1253, opened 2019-11-19
Recreate the content DOM node on location changes

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/1247

### DIFF
--- a/src/components/FullscreenGrid/ContentShell.spec.tsx
+++ b/src/components/FullscreenGrid/ContentShell.spec.tsx
@@ -1,22 +1,36 @@
+import { Location } from 'history';
 import * as React from 'react';
 import { Store } from 'redux';
 
 import configureStore from '../../configureStore';
 import { actions as fullscreenGridActions } from '../../reducers/fullscreenGrid';
-import { spyOn, shallowUntilTarget } from '../../test-helpers';
+import {
+  createContextWithFakeRouter,
+  createFakeLocation,
+  spyOn,
+  shallowUntilTarget,
+} from '../../test-helpers';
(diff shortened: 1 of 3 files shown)


src/components/FullscreenGrid/ContentShell.spec.tsx
```diff
@@ -182,4 +196,11 @@ describe(__filename, () => {
       fullscreenGridActions.toggleAltSidePanel(),
     );
   });
+
+  it('renders the main content panel with a location key', () => {
+    const key = 'locationKey';
+    const root = render({ location: createFakeLocation({ key }) });
+
+    expect(root.find(`.${styles.content}`).key()).toEqual(key);
```
bobsilverberg: I didn't even know `key()` was a thing! A guess it's rare to need to assert on a `key` in a test.

kumar303:
> Looks good. I'm not sure how `location.key` is calculated (a quick browse of the source didn't make anything obvious), but I assume you're confident that it will be different for each location navigated to.

`location.key` is not [documented](https://reacttraining.com/react-router/web/api/location) well but as far as I can tell this key represents the current location, including the query string and hash. This will change when navigating to different files.

src/components/FullscreenGrid/ContentShell.spec.tsx
kumar303: Heh, yeah, I learned about it recently too. This kind of test relies on trusting the React API but key behavior seems stable enough. It's features like that where having some functional test coverage would be nice.

kumar303:
Thanks for the quick review

PR #1255, opened 2019-11-20
Fixes #1193: Add helper for creating an error response

### LINKED ISSUE #1193: Add helper for creating an error response
Some API code deals with an [`ErrorResponseObject`](https://github.com/mozilla/addons-code-manager/blob/master/src/api/index.tsx#L120). Here's how we write tests to simulate this:

```js
const errorResponse = { 
  error: new Error('API Error') 
};
const getSomethingFromApi = jest.fn()
  .mockResolvedValue(errorResponse);
```

We should replace this with a helper similar to [`createFakeApiPage()`](https://github.com/mozilla/addons-code-manager/blob/ceb6e9158b68d2dec1e667a99f6ea0ae25148804/src/test-helpers.tsx#L943). Maybe something like:

```js
const errorResponse = createErrorResponse({ 
  error: new Error('API Error') 
});
```

This will help us evolve the `ErrorResponseObject` in the future if we need to. Currently, it would be hard to evolve this type because any time `jest.fn()` is used, it returns the type `unknown` which stops TypeScript from type checking anything further.

ragnarokatz:
reference issue #1193

**description:**
Refactor the creation of all {error: Error} objects with the createErrorResponse helper function for all the test cases that use it. Prettier-ed and linted.

**tests checked & passed:**
yarn lint, yarn eslint, yarn typecheck, yarn type-coverage, yarn test

- [x] This PR relates to an existing open issue and there are no existing PRs open for the same issue.
- [x] Add a description of the changes introduced in this PR.
- [x] The change has been successfully run locally.

### DIFF
--- a/src/test-helpers.tsx
+++ b/src/test-helpers.tsx
@@ -10,7 +10,11 @@ import { Store } from 'redux';
 import log from 'loglevel';
 import { createAction } from 'typesafe-actions';
 
-import { GetCommentsResponse, PaginatedResponse } from './api';
+import {
+  GetCommentsResponse,
+  PaginatedResponse,
+  ErrorResponseType,
+} from './api';
 import configureStore, { ThunkActionCreator } from './configureStore';
 import { ApplicationState } from './reducers';
 import {
@@ -965,6 +969,14 @@ export const createFakeCommentsResponse = (
   results = [createFakeExternalComment()],
 ) => {
   return createFakeApiPage<GetCommentsResponse>({ results });
+};
+
+export const createErrorResponse = (
+  errorResponse = {
+    error: new Error('Example error'),
+  },
+): ErrorResponseType => {
+  return errorResponse;
 };
 
 export const setMockFetchResponseJSON = (
--- a/src/api/index.spec.tsx
+++ b/src/api/index.spec.tsx
@@ -9,6 +9,7 @@ import {
   createFakeApiPage,
   createFakeCommentsResponse,
   createFakeExternalComment,
+  createErrorResponse,
   nextUniqueId,
   setMockFetchResponseJSON,
 } from '../test-helpers';
@@ -384,7 +385,9 @@ describe(__filename, () => {
 
   describe('isErrorResponse', () => {
     it('returns true if a response object is an error', () => {
-      const response = { error: 'this is an error' };
+      const response = createErrorResponse({
+        error: new Error('API Error'),
+      });
 
       expect(isErrorResponse(response)).toEqual(true);
     });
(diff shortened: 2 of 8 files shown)


src/test-helpers.tsx
```diff
@@ -967,6 +971,10 @@ export const createFakeCommentsResponse = (
   return createFakeApiPage<GetCommentsResponse>({ results });
 };
 
+export const createErrorResponse = (errorResponse: { error: Error }) => {
+  return errorResponse as ErrorResponseType;
```
kumar303: We try to avoid type assertions because they bypass some type checking. In this case, it works just the same to declare a return type of the function so I suggest doing that instead:

```diff
diff --git a/src/test-helpers.tsx b/src/test-helpers.tsx
index 17ce432..9da3169 100644
--- a/src/test-helpers.tsx
+++ b/src/test-helpers.tsx
@@ -971,8 +971,10 @@ export const createFakeCommentsResponse = (
   return createFakeApiPage<GetCommentsResponse>({ results });
 };
 
-export const createErrorResponse = (errorResponse: { error: Error }) => {
-  return errorResponse as ErrorResponseType;
+export const createErrorResponse = (errorResponse: {
+  error: Error;
+}): ErrorResponseType => {
+  return errorResponse;
 };
 
 export const setMockFetchResponseJSON = (
```

src/test-helpers.tsx
```diff
@@ -967,6 +971,10 @@ export const createFakeCommentsResponse = (
   return createFakeApiPage<GetCommentsResponse>({ results });
 };
 
+export const createErrorResponse = (errorResponse: { error: Error }) => {
```
kumar303: It would be more convenient if this helper returned a default response with a pre-filled error. This would be helpful in cases where the test does not need to make assertions about the actual error. For example:

```diff
diff --git a/src/test-helpers.tsx b/src/test-helpers.tsx
index 17ce432..def1f93 100644
--- a/src/test-helpers.tsx
+++ b/src/test-helpers.tsx
@@ -971,7 +971,11 @@ export const createFakeCommentsResponse = (
   return createFakeApiPage<GetCommentsResponse>({ results });
 };
 
-export const createErrorResponse = (errorResponse: { error: Error }) => {
+export const createErrorResponse = (
+  errorResponse = {
+    error: new Error('Example error'),
+  },
+) => {
   return errorResponse as ErrorResponseType;
 };
 
```

ragnarokatz: Fixed, added default value for parameter.

src/test-helpers.tsx
ragnarokatz: Fixed, declared strong type checking for the return type.

caitmuenster:
Thanks so much for the patch, @ragnarokatz! 🎉 Your contribution has been added to our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#November_2019). 

Would you be interested in creating an account on [mozillians.org](https://mozillians.org)? I'd be happy to vouch for you!

ragnarokatz:
@caitmuenster Yes, that would be great. I have created my account here:
image

caitmuenster:
@ragnarokatz Yay! You are vouched. ✨

Welcome onboard! I look forward to seeing you around the project.

Issue #1256, opened 2019-11-20
Expand capabilities of PopoverButton


kumar303:
The version comparison screen needed for https://github.com/mozilla/addons-code-manager/issues/758 will benefit from a popover since it will be accessible via NavBar.

The goal of this issue is:

* Expand the capabilities of `PopoverButton` so that it can be used for showing a version comparison popover
* Move sharable logic from `CommentSummaryButton` into `PopoverButton`

**For QA**: Please check for regressions in the *Comment Summary* button and its popover content (summary of comments, close button, etc).

ioanarusiczki:
@kumar303 Moving it to the next milestone because there will be no push this week.

ioanarusiczki:
@kumar303 No issues found. 
Verified on dev with FF70(Win10).

PR #1257, opened 2019-11-20
Move Redux control into PopoverButton

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/1256

### DIFF
--- a/src/components/CommentSummaryButton/index.tsx
+++ b/src/components/CommentSummaryButton/index.tsx
@@ -4,11 +4,8 @@ import { connect } from 'react-redux';
 
 import { gettext } from '../../utils';
 import { ApplicationState } from '../../reducers';
-import {
-  Comment,
-  actions as commentsActions,
-  selectVersionComments,
-} from '../../reducers/comments';
+import { Comment, selectVersionComments } from '../../reducers/comments';
+import { actions as popoverActions } from '../../reducers/popover';
 import { selectCurrentVersionInfo } from '../../reducers/versions';
 import { ConnectedReduxProps } from '../../configureStore';
 import CommentSummary from '../CommentSummary';
@@ -21,18 +18,15 @@ export type DefaultProps = {};
 
 type PropsFromState = {
   comments: Comment[] | undefined;
-  showOverlay: boolean;
 };
 
 type Props = PublicProps & DefaultProps & PropsFromState & ConnectedReduxProps;
 
-export const CommentSummaryButtonBase = ({
-  comments,
-  dispatch,
-  showOverlay,
-}: Props) => {
+export const CommentSummaryButtonBase = ({ comments, dispatch }: Props) => {
+  const id = 'COMMENTS_SUMMARY';
   return (
     <PopoverButton
+      id={id}
       content={
         <>
           <Popover.Title as="h3">{gettext('Comments')}</Popover.Title>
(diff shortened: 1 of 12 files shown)


src/components/CommentSummaryButton/index.tsx
```diff
@@ -52,14 +46,7 @@ export const CommentSummaryButtonBase = ({
           </Popover.Content>
         </>
       }
-      onHide={() => {
-        dispatch(commentsActions.hideSummaryOverlay());
-      }}
-      onOpen={() => {
-        dispatch(commentsActions.toggleSummaryOverlay());
```
kumar303: I didn't port over this *toggle* action because technically it's not necessary. The button still works as expected because of how `Overlay` keeps track of its state internally and triggers the right callback.

src/components/PopoverButton/index.spec.tsx
```diff
@@ -1,53 +1,72 @@
-import { shallow } from 'enzyme';
 import * as React from 'react';
+import { Store } from 'redux';
 import { Button, Overlay, OverlayChildrenParams } from 'react-bootstrap';
 
-import { createFakeRef } from '../../test-helpers';
+import { createFakeRef, shallowUntilTarget, spyOn } from '../../test-helpers';
```
bobsilverberg: Nit: I try to keep the imports in alphabetical order (by source), but it's not a big deal.

src/reducers/popover.tsx
```diff
@@ -0,0 +1,53 @@
+import { Reducer } from 'redux';
+import { ActionType, createAction, getType } from 'typesafe-actions';
+
+export type PopoverIdType = 'COMMENTS_SUMMARY' | 'COMPARE_VERSIONS';
+
+export type PopoverState = {
+  current: PopoverIdType | undefined;
```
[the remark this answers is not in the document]
kumar303: Yeah, this was my intention. I wanted to use this to add a second popover button to the navbar. When clicking, it should close any other open navbar popover. I can't envision a future where we'd want it to work a different way but if we did then I guess we could move back to the controlled `Popover` and make a wrapper like `ControlledPopover` for the Redux version. I didn't do that because YAGNI 🤷‍♂

src/reducers/popover.spec.tsx
```diff
@@ -0,0 +1,71 @@
...
+  actions,
+  selectPopoverIsVisible,
+  initialState,
+} from './popover';
+
+describe(__filename, () => {
+  it('can show a popover', () => {
+    const id = 'COMMENTS_SUMMARY';
+    let state;
+    state = reducer(state, actions.hide(id));
```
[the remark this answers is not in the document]
kumar303: I didn't want the test to rely on how the default state is hidden. There is another test *selectPopoverIsVisible...returns false by default* which covers the functionality of being hidden by default.

src/components/PopoverButton/index.spec.tsx
kumar303: What part of it isn't alphabetical?

kumar303: Do you mean that this line should come after `../../reducers...`?

kumar303: I could make that change but I don't see how it's very helpful because if you're looking for something, you're more likely looking for an exported object rather than the file itself.

bobsilverberg: I would see them in this order, alphabetically:

```
import { actions as popoverActions } from '../../reducers/popover';
import { createFakeRef, shallowUntilTarget, spyOn } from '../../test-helpers';
```

because `c` comes before `r` comes before `t`. But as I said, it's not a big deal.

kumar303:
Thanks for the review

PR #1260, opened 2019-11-21
Comment on a line of a diff

### LINKED ISSUE #989: Comment on a specific line of a diff
User story:

> As a code reviewer, I want to be able to enter custom comments for specific code lines in the file and diff viewer.

When viewing a diff, the reviewer can enter a comment a specific line. This will be similar to adding a comment to a file (https://github.com/mozilla/addons-code-manager/issues/112).

The suggested UI flow per [meeting notes](https://docs.google.com/document/d/1y4crrHxhGejb3eZ6HpjtBdDqONJakfwtfjaOqwq-qZo/edit) is roughly this:
- click on a line of a diff to begin entering a comment
- when presented with a list of canned responses, click *other*
- begin entering text
- press save

It's possible that it will be easy to hook into `DiffView`. I originally thought it would be tricky and filed https://github.com/otakustay/react-diff-view/issues/59 which led to https://github.com/otakustay/react-diff-view/#customize-gutter . That approach may not be necessary anymore, though.

bobsilverberg:
Fixes #989 

This allows for commenting on a normal or inserted line, but not a deleted line as there are issues that need to be ironed out on the server side to enable that. An "add comment" button will appear when hovering over the line number, and any comments added will appear underneath the line, just as they do in the browse view.

Screenshot:

### DIFF
--- a/src/components/DiffView/index.spec.tsx
+++ b/src/components/DiffView/index.spec.tsx
@@ -17,6 +17,7 @@ import { History, Location } from 'history';
 import basicDiff from './fixtures/basicDiff';
 import multipleDiff from './fixtures/multipleDiff';
 import diffWithDeletions from './fixtures/diffWithDeletions';
+import CommentList from '../CommentList';
 import FadableContent from '../FadableContent';
 import LinterMessage from '../LinterMessage';
 import LinterProvider, { LinterProviderInfo } from '../LinterProvider';
@@ -38,6 +39,7 @@ import {
   createFakeLocation,
   fakeExternalDiff,
   fakeVersion,
+  nextUniqueId,
   shallowUntilTarget,
   simulateLinterProvider,
 } from '../../test-helpers';
@@ -573,23 +575,44 @@ describe(__filename, () => {
     );
   });
 
-  it('renders just the right amount of widgets', () => {
+  it('renders just the right amount of widgets with messages', () => {
+    const firstLine = nextUniqueId();
+    const secondLine = nextUniqueId();
+    const thirdLine = nextUniqueId();
+    const lines = [firstLine, secondLine, thirdLine];
     const externalMessages = [
-      { line: 1, uid: 'first' },
-      { line: 2, uid: 'second' },
+      { line: firstLine, uid: 'first' },
+      { line: secondLine, uid: 'second' },
     ];
-
+    const version = createInternalVersion({
+      ...fakeVersion,
+      id: nextUniqueId(),
+    });
+    const diff = createDiffWithHunks([
+      createHunkWithChanges([
+        { type: 'insert', new_line_number: firstLine },
+        { type: 'insert', new_line_number: secondLine },
+        { type: 'insert', new_line_number: thirdLine },
+      ]),
+    ]);
     const widgets = renderAndGetWidgets({
-      diff: parseDiff(diffWithDeletions)[0],
+      diff,
+      enableCommenting: true,
       selectedMessageMap: createFakeLinterMessagesByPath({
         messages: externalMessages,
       }),
+      version,
     });
 
-    // As a sanity check on how the widgets are mapped,
-    // make sure we have truthy React nodes for exactly the same
-    // number of lines containing messages.
-    expect(getWidgetNodes(widgets).length).toEqual(externalMessages.length);
+    // We should have one widget per line.
+    expect(getWidgetNodes(widgets).length).toEqual(lines.length);
+
+    // We should have one widget with a message per message.
+    const widgetsWithMessages = lines.filter(
+      (line) =>
+        renderWidget(diff.hunks, widgets, line).find(LinterMessage).length,
+    );
+    expect(widgetsWithMessages.length).toEqual(externalMessages.length);
   });
 
   it('renders multiple inline messages on the same line', () => {
@@ -627,12 +650,80 @@ describe(__filename, () => {
     );
   });
 
-  it('does not render widgets without linter messages', () => {
-    const widgets = renderAndGetWidgets({
-      diff: parseDiff(diffWithDeletions)[0],
-      selectedMessageMap: null,
+  describe('comment list', () => {
+    it('renders inline comments for inserted lines', () => {
+      const firstLine = nextUniqueId();
+      const secondLine = nextUniqueId();
+      const version = createInternalVersion({
+        ...fakeVersion,
+        id: nextUniqueId(),
+      });
+      const diff = createDiffWithHunks([
+        createHunkWithChanges([
+          { type: 'insert', new_line_number: firstLine },
+          { type: 'insert', new_line_number: secondLine },
+        ]),
+      ]);
+      const widgets = renderAndGetWidgets({
+        diff,
+        enableCommenting: true,
+        version,
+      });
+
+      for (const line of [firstLine, secondLine]) {
+        const commentList = renderWidget(diff.hunks, widgets, line).find(
+          CommentList,
+        );
+        expect(commentList).toHaveProp('addonId', version.addon.id);
+        expect(commentList).toHaveProp('fileName', version.selectedPath);
+        expect(commentList).toHaveProp('line', line);
+        expect(commentList).toHaveProp('versionId', version.id);
+      }
     });
-    expect(getWidgetNodes(widgets).length).toEqual(0);
+
+    it('does not render inline comments for normal or deleted lines', () => {
+      const firstLine = nextUniqueId();
+      const secondLine = nextUniqueId();
+      const version = createInternalVersion({
+        ...fakeVersion,
+        id: nextUniqueId(),
+      });
+      const diff = createDiffWithHunks([
+        createHunkWithChanges([
+          { type: 'delete', old_line_number: firstLine },
+          { type: 'normal', old_line_number: secondLine },
+        ]),
+      ]);
+      const widgets = renderAndGetWidgets({
+        diff,
+        enableCommenting: true,
+        version,
+      });
+
+      for (const line of [firstLine, secondLine]) {
+        const widget = renderWidget(diff.hunks, widgets, line);
+        expect(widget.find(CommentList)).toHaveLength(0);
+      }
+    });
+
+    it('does not render inline comments when the feature is disabled', () => {
+      const line = nextUniqueId();
+      const version = createInternalVersion({
+        ...fakeVersion,
+        id: nextUniqueId(),
+      });
+      const diff = createDiffWithHunks([
+        createHunkWithChanges([{ type: 'insert', new_line_number: line }]),
+      ]);
+      const widgets = renderAndGetWidgets({
+        diff,
+        enableCommenting: false,
+        version,
+      });
+
+      const widget = renderWidget(diff.hunks, widgets, line);
+      expect(widget.find(CommentList)).toHaveLength(0);
+    });
   });
 
   it('enables syntax highlighting for diffs when possible', () => {
--- a/src/components/DiffView/index.spec.tsx
+++ b/src/components/DiffView/index.spec.tsx
@@ -18,7 +18,7 @@ import { History, Location } from 'history';
 import basicDiff from './fixtures/basicDiff';
 import multipleDiff from './fixtures/multipleDiff';
 import diffWithDeletions from './fixtures/diffWithDeletions';
-import CommentableBase from '../Commentable';
+import Commentable from '../Commentable';
 import CommentList from '../CommentList';
 import FadableContent from '../FadableContent';
 import LinterMessage from '../LinterMessage';
@@ -44,6 +44,8 @@ import {
   fakeVersion,
   nextUniqueId,
   shallowUntilTarget,
+  simulateCommentable,
+  simulateCommentList,
   simulateLinterProvider,
 } from '../../test-helpers';
 import styles from './styles.module.scss';
@@ -63,6 +65,7 @@ describe(__filename, () => {
   >;
 
   const render = ({
+    enableCommenting = true,
     history = createFakeHistory(),
     location = createFakeLocation(),
     ...props
@@ -72,6 +75,7 @@ describe(__filename, () => {
     return shallowUntilTarget(
       <DiffView
         diff={parseDiff(basicDiff)[0]}
+        enableCommenting={enableCommenting}
         mimeType="text/plain"
         version={createInternalVersion(fakeVersion)}
         {...props}
@@ -600,7 +604,6 @@ describe(__filename, () => {
     ]);
     const widgets = renderAndGetWidgets({
       diff,
-      enableCommenting: true,
       selectedMessageMap: createFakeLinterMessagesByPath({
         messages: externalMessages,
       }),
@@ -654,7 +657,8 @@ describe(__filename, () => {
   });
 
   describe('comment list', () => {
-    it('renders inline comments for inserted lines', () => {
+    it('renders inline comments for normal and inserted lines', () => {
+      const CommentListResult = () => <div />;
       const firstLine = nextUniqueId();
       const secondLine = nextUniqueId();
       const version = createInternalVersion({
@@ -663,20 +667,30 @@ describe(__filename, () => {
       });
       const diff = createDiffWithHunks([
         createHunkWithChanges([
-          { type: 'insert', new_line_number: firstLine },
+          {
+            type: 'normal',
+            new_line_number: firstLine,
+            old_line_number: firstLine,
+          },
           { type: 'insert', new_line_number: secondLine },
         ]),
       ]);
       const widgets = renderAndGetWidgets({
         diff,
-        enableCommenting: true,
         version,
       });
 
       for (const line of [firstLine, secondLine]) {
-        const commentList = renderWidget(diff.hunks, widgets, line).find(
-          CommentList,
-        );
+        const widget = renderWidget(diff.hunks, widgets, line);
+
+        const { shell, renderContent } = simulateCommentList({
+          commentList: <CommentListResult />,
+          root: widget,
+        });
+        const commentWidget = renderContent(shell.at(0));
+        expect(commentWidget.find(CommentListResult)).toHaveLength(1);
+
+        const commentList = widget.find(CommentList);
         expect(commentList).toHaveProp('addonId', version.addon.id);
         expect(commentList).toHaveProp('fileName', version.selectedPath);
         expect(commentList).toHaveProp('line', line);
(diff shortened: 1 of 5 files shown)


bobsilverberg:
@kumar303 Can you please take a look at this and let me know what you think, both of the code and the UX?

src/components/DiffView/styles.module.scss
```diff
@@ -44,6 +44,10 @@
   .diff-line .diff-code-selected {
     background-color: $light-gray;
   }
+
+  .diff-gutter:hover button.btn {
```
bobsilverberg: I had to do this hacky thing in order to target the button that is inside the gutter. The button is defined in a different file. Maybe there's a better way of doing this?

src/components/DiffView/index.tsx
```diff
@@ -265,6 +284,39 @@ export class DiffViewBase extends React.Component<Props> {
...
+
+    const { isInsert, lineNumber } = change;
+    const changeKey = getChangeKey(change);
+
+    let gutter = wrapInAnchor(renderDefault());
+    if (enableCommenting && isInsert && lineNumber && side === 'new') {
+      gutter = (
+        <div className={styles.gutter}>
+          {gutter}
+          <Commentable
```
kumar303: This approach seems to work but it would be nicer if one could hover over a line (not just the gutter) and see the comment button appear. This would convey that it's possible to add comments since that may not be obvious. 

To do that, we'd need to replace [the whole `<tr>`](https://github.com/otakustay/react-diff-view/blob/7a97efa2b1ed11c79cd755ba2fb0475de82b09a7/src/Hunk/UnifiedHunk/UnifiedChange.js#L80) with `<Commentable as="tr" />`. I don't think that's possible but it might be worth filing a feature request for it. They are super fast about fixing things.

But, yeah, the `renderGutter()` approach still works and I think it's worth pursuing that as a quick fix.

src/components/DiffView/styles.module.scss
kumar303: I think you just needed to wrap the whole thing (including the line number) in `Commentable`. This is working for me:

<details>
<summary>Patch</summary>

```diff
diff --git a/src/components/DiffView/index.tsx b/src/components/DiffView/index.tsx
index 3922d61..f7db532 100644
--- a/src/components/DiffView/index.tsx
+++ b/src/components/DiffView/index.tsx
@@ -295,22 +295,25 @@ export class DiffViewBase extends React.Component<Props> {
     const { isInsert, lineNumber } = change;
     const changeKey = getChangeKey(change);
 
-    let gutter = wrapInAnchor(renderDefault());
+    const defaultGutter = wrapInAnchor(renderDefault());
+    let gutter = defaultGutter;
     if (enableCommenting && isInsert && lineNumber && side === 'new') {
       gutter = (
-        <div className={styles.gutter}>
-          {gutter}
-          <Commentable
-            as="span"
-            id={changeKey}
-            className={styles.commentButton}
-            line={lineNumber}
-            fileName={version.selectedPath}
-            versionId={version.id}
-          >
-            {(addCommentButton) => <>{addCommentButton}</>}
-          </Commentable>
-        </div>
+        <Commentable
+          as="div"
+          id={changeKey}
+          className={styles.gutter}
+          line={lineNumber}
+          fileName={version.selectedPath}
+          versionId={version.id}
+        >
+          {(addCommentButton) => (
+            <>
+              {defaultGutter}
+              <span className={styles.commentButton}>{addCommentButton}</span>
+            </>
+          )}
+        </Commentable>
       );
     }
 
diff --git a/src/components/DiffView/styles.module.scss b/src/components/DiffView/styles.module.scss
index bae2ff5..04a8900 100644
--- a/src/components/DiffView/styles.module.scss
+++ b/src/components/DiffView/styles.module.scss
@@ -44,10 +44,6 @@
   .diff-line .diff-code-selected {
     background-color: $light-gray;
   }
-
-  .diff-gutter:hover button.btn {
-    visibility: visible;
-  }
 }
 
 .extraMessages {
```

</details>

Ideally, we would wrap as much as possible in `Commentable` but I don't know if it's possible to do much more.

src/components/DiffView/index.tsx
```diff
@@ -197,17 +204,29 @@ export class DiffViewBase extends React.Component<Props> {
...
       }
 
-      let widget = null;
+      let widget = (
+        <>
+          {enableCommenting && change.isInsert && (
+            <CommentList
+              addonId={version.addon.id}
+              fileName={version.selectedPath}
+              line={line || null}
```
kumar303: I guess TS wanted you to fall back to `null`? This should not be set to `null` so maybe you need to throw an error instead.

src/components/DiffView/index.tsx
```diff
@@ -197,17 +204,29 @@ export class DiffViewBase extends React.Component<Props> {
...
+              versionId={version.id}
+            >
+              {(commentList) => <div>{commentList}</div>}
+            </CommentList>
+          )}
+        </>
+      );
       if (messages && messages.length) {
-        widget = (
+        widget = widget && (
```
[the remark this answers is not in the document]
bobsilverberg: I thought what I did worked, but I've changed it to join them together in a react fragment as per your comment.

src/components/DiffView/index.tsx
bobsilverberg: I missed this one when I was adjusting the one below. I only have to gate this on `line` and all is fine.

src/components/DiffView/index.tsx
bobsilverberg: Good suggestion. I opened https://github.com/otakustay/react-diff-view/issues/70.

kumar303: @bobsilverberg thanks for opening an issue. Rather than "wrap" the `<tr>` we really want to provider our own `<tr>` tag with custom props, like a class name. Maybe that's obvious enough in your issue description but I wanted to point it out since it's not possible to wrap `<tr>` tags, it will break the table layout.

bobsilverberg: He's given me some suggestions of how to make this work. I think I'm going to try to land it first the way it is, with the button in the gutter, as we're still going to need that even if I also use `renderToken` to allow for hovering over the code area. I will file a follow-up to look into adding `renderToken` as well.

bobsilverberg:
Thanks for your help thus far, @kumar303. This is ready for another look. I'm not entirely sure of the approach I took for testing `renderGutter`, I had to do some weird things.

src/components/DiffView/index.tsx
```diff
@@ -197,17 +204,32 @@ export class DiffViewBase extends React.Component<Props> {
         messages = selectedMessageMap.byLine[line];
       }
 
-      let widget = null;
+      let widget = (
+        <>
+          {enableCommenting && line && change.isInsert && (
```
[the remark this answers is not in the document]
bobsilverberg: I wasn't sure if we might end up with duplicates for normal lines, but I guess that's impossible? Duplicates would only be an insert + a delete?

src/components/DiffView/index.tsx
```diff
@@ -265,6 +287,42 @@ export class DiffViewBase extends React.Component<Props> {
...
+    wrapInAnchor,
+  }: RenderGutterParams) => {
+    const { enableCommenting, version } = this.props;
+
+    const { isInsert, lineNumber } = change;
+    const changeKey = getChangeKey(change);
+
+    const defaultGutter = wrapInAnchor(renderDefault());
+    let gutter = defaultGutter;
+    if (enableCommenting && isInsert && lineNumber && side === 'new') {
```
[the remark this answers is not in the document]
bobsilverberg: It shows the form for me, with the change, so I'm not sure what the problem was, but it seems okay on my end. Thanks for the suggestion!

src/components/DiffView/styles.module.scss
```diff
@@ -57,3 +57,12 @@
 .inlineLinterMessages {
   padding: 1rem 1rem 0 1rem;
 }
+
+.gutter {
```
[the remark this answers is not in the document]
bobsilverberg: Nice, thanks!

src/components/DiffView/index.spec.tsx
```diff
@@ -573,23 +578,44 @@ describe(__filename, () => {
...
+      createHunkWithChanges([
+        { type: 'insert', new_line_number: firstLine },
+        { type: 'insert', new_line_number: secondLine },
+        { type: 'insert', new_line_number: thirdLine },
+      ]),
+    ]);
     const widgets = renderAndGetWidgets({
-      diff: parseDiff(diffWithDeletions)[0],
+      diff,
+      enableCommenting: true,
```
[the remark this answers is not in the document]
bobsilverberg: I was trying to be explicit about what data is being used to turn it on, but yeah, I guess that's not really necessary if the default test state is to have comments enabled.

src/components/DiffView/index.spec.tsx
```diff
@@ -627,12 +653,80 @@ describe(__filename, () => {
     );
   });
 
-  it('does not render widgets without linter messages', () => {
```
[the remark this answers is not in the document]
bobsilverberg: Not with the current implementation. It renders a widget with a `CommentList` for every line, even if there are no comments. This allows the `CommentList` to be the only thing that cares about whether there are comments or not, but I guess that could be non-performant. Do you think we should be checking in `DiffView` to see if there are comments? I haven't checked yet to see how much (if any) extra work that would be.

src/components/DiffView/index.spec.tsx
```diff
@@ -910,4 +1004,141 @@ describe(__filename, () => {
...
+          change,
+          renderDefault,
+          inHoverState: false,
+          side,
+          wrapInAnchor,
+        });
+        if (shallowRender) {
+          return shallow(<div>{gutter}</div>);
+        }
+        return gutter;
```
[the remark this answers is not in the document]
bobsilverberg: I saw `simulateCommentable` when I was writing the tests, but TBH I was confused about what it did and how it worked. I'll have to take another look, but you're right, that simple change makes it all work!

src/components/DiffView/index.tsx
```diff
@@ -197,17 +204,32 @@ export class DiffViewBase extends React.Component<Props> {
         messages = selectedMessageMap.byLine[line];
       }
 
-      let widget = null;
+      let widget = (
+        <>
```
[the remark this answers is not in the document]
bobsilverberg: I originally had it as:

```tsx
widget = widget && (
  <div className={styles.inlineLinterMessages}>
    {messages.map((msg) => {
      return <LinterMessage key={msg.uid} message={msg} inline />;
    })}
  </div>
);
```

but I recall that you asked it to be wrapped in a fragment. Maybe I misunderstood what you wanted. If neither the above is okay, nor what's in my current patch, what should the syntax be?

src/components/DiffView/index.tsx
```diff
@@ -197,17 +204,32 @@ export class DiffViewBase extends React.Component<Props> {
...
+          {enableCommenting && line && change.isInsert && (
+            <CommentList
+              addonId={version.addon.id}
+              fileName={version.selectedPath}
+              line={line}
+              versionId={version.id}
+            >
+              {(commentList) => <div>{commentList}</div>}
```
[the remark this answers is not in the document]
bobsilverberg: Ok, I did this, and it works, but I find it pretty confusing to read. Maybe my test can be improved, or maybe we're stuck with the fact that we need to call `renderContent` using the `shell`, which just comes across as very cryptic in the test.

src/components/DiffView/index.tsx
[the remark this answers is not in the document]
bobsilverberg: Actually, that syntax above doesn't work. But I can't just remove the fragment. Is there a better way to do it?

bobsilverberg:
Thanks for the review and all of the suggestions, @kumar303. I've addressed everything except for the two comments that remain unresolved for which I've added comments/questions. It's ready for another look.

src/components/DiffView/index.spec.tsx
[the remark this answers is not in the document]
kumar303: I didn't do any performance testing but I think it should be fine 😬  This was by design to make it easy to lazy load comments (not saying it's a great idea, heh).

src/components/DiffView/index.tsx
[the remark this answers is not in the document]
kumar303: Why can't you just remove the fragment? I only see one component rendered inside the fragment. Removing it compiles OK:

```diff
diff --git a/src/components/DiffView/index.tsx b/src/components/DiffView/index.tsx
index 37b464a..d626909 100644
--- a/src/components/DiffView/index.tsx
+++ b/src/components/DiffView/index.tsx
@@ -204,20 +204,17 @@ export class DiffViewBase extends React.Component<Props> {
         messages = selectedMessageMap.byLine[line];
       }
 
-      let widget = (
-        <>
-          {enableCommenting && line && !change.isDelete && (
-            <CommentList
-              addonId={version.addon.id}
-              fileName={version.selectedPath}
-              line={line}
-              versionId={version.id}
-            >
-              {(commentList) => <div>{commentList}</div>}
-            </CommentList>
-          )}
-        </>
-      );
+      let widget =
+        enableCommenting && line && !change.isDelete ? (
+          <CommentList
+            addonId={version.addon.id}
+            fileName={version.selectedPath}
+            line={line}
+            versionId={version.id}
+          >
+            {(commentList) => <div>{commentList}</div>}
+          </CommentList>
+        ) : null;
       if (messages && messages.length) {
         widget = (
           <>

```

src/components/DiffView/index.tsx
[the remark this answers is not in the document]
kumar303: Hmm, I'm not sure. In any case, the normal view (when there's no "diff") should show comments.

src/components/DiffView/index.tsx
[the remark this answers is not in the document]
bobsilverberg: Doh! I was looking at the *wrong* fragment. Although it was clear from the line number, for some reason I thought you were talking about the fragment on line 223, not line 208. My bad. I'll remove the one on 208.

bobsilverberg:
@kumar303 In case this has fallen off your radar, it's ready for another look.

kumar303:
Yeah, it's in my queue, thanks. Could you also update the description when you get a chance?

src/components/DiffView/index.spec.tsx
```diff
@@ -627,12 +656,105 @@ describe(__filename, () => {
...
+      });
+
+      for (const line of [firstLine, secondLine]) {
+        const widget = renderWidget(diff.hunks, widgets, line);
+
+        const { shell, renderContent } = simulateCommentList({
+          commentList: <CommentListResult />,
+          root: widget,
+        });
+        const commentWidget = renderContent(shell.at(0));
```
kumar303: Since there will only be one `<CommentList>` in the widget, you can simulate it with the default params:

```diff
diff --git a/src/components/DiffView/index.spec.tsx b/src/components/DiffView/index.spec.tsx
index 5ef8bf3..d5b93a6 100644
--- a/src/components/DiffView/index.spec.tsx
+++ b/src/components/DiffView/index.spec.tsx
@@ -683,11 +683,11 @@ describe(__filename, () => {
       for (const line of [firstLine, secondLine]) {
         const widget = renderWidget(diff.hunks, widgets, line);
 
-        const { shell, renderContent } = simulateCommentList({
+        const { renderContent } = simulateCommentList({
           commentList: <CommentListResult />,
           root: widget,
         });
-        const commentWidget = renderContent(shell.at(0));
+        const commentWidget = renderContent();
         expect(commentWidget.find(CommentListResult)).toHaveLength(1);
 
         const commentList = widget.find(CommentList);
```

src/components/DiffView/index.tsx
```diff
@@ -197,17 +204,29 @@ export class DiffViewBase extends React.Component<Props> {
...
 
-      let widget = null;
+      let widget = enableCommenting && line && !change.isDelete && (
+        <CommentList
+          addonId={version.addon.id}
+          fileName={version.selectedPath}
+          line={line}
+          versionId={version.id}
+        >
+          {(commentList) => <div>{commentList}</div>}
```
kumar303: This can be `{(allComments) => allComments}`. That is, there's no need for a div and in fact it's already [wrapped in one](https://github.com/mozilla/addons-code-manager/blob/83eb299a13464e5f25a1bbd47a4fae917eac3f6b/src/components/CommentList/index.tsx#L106). 

It may seem weird to need a render prop for this but if you consider wanting to render a container element around the comments but only when they exist, there's actually no other way to do that.

src/components/DiffView/index.spec.tsx
```diff
@@ -910,4 +1032,146 @@ describe(__filename, () => {
...
+      defaultGutter = <div />,
+      enableCommenting = true,
+      renderDefault = jest.fn(),
+      side = 'new',
+      version = createInternalVersion(fakeVersion),
+      wrapInAnchor = jest.fn().mockReturnValue(defaultGutter),
+    }: Partial<RenderGutterParams> &
+      RenderParams & {
+        defaultGutter?: React.ReactElement;
+        shallowRender?: boolean;
```
kumar303: This isn't used anymore

src/components/DiffView/index.tsx
```diff
@@ -197,17 +204,29 @@ export class DiffViewBase extends React.Component<Props> {
         messages = selectedMessageMap.byLine[line];
       }
 
-      let widget = null;
+      let widget = enableCommenting && line && !change.isDelete && (
```
kumar303: In the rare (and probably unlikely) case that line is `0` then the fragment below will put a zero in the HTML which is not what you want. I suggest using a ternary and setting it to `null` when false.

src/components/DiffView/index.tsx
```diff
@@ -265,6 +284,42 @@ export class DiffViewBase extends React.Component<Props> {
...
 
+  renderGutter = ({
+    change,
+    renderDefault,
+    side,
+    wrapInAnchor,
+  }: RenderGutterParams) => {
+    const { enableCommenting, version } = this.props;
+
+    const { isDelete, lineNumber } = change;
```
kumar303: This [part of the docs](https://github.com/otakustay/react-diff-view/#key-of-change) suggests maybe you should be using `oldLineNumber` for normal changes. Could you try to comment on some normal diffs and compare it in browse mode to make sure they appear on the right lines? It might be fine; I have a vague recollection that `lineNumber` and `oldLineNumber` are identical....maybe.

src/components/DiffView/index.tsx
```diff
@@ -265,6 +284,42 @@ export class DiffViewBase extends React.Component<Props> {
...
+    const changeKey = getChangeKey(change);
+
+    const defaultGutter = wrapInAnchor(renderDefault());
+    let gutter = defaultGutter;
+    if (enableCommenting && !isDelete && lineNumber && side === 'new') {
+      gutter = (
+        <Commentable
+          as="div"
+          id={changeKey}
```
kumar303: You don't have to define an ID here. It's only there in case the wrapped element (in this case `div`) needs an `id` attribute.

src/components/DiffView/index.tsx
bobsilverberg: I checked, and a comment added in Compare mode on a normal line does also appear in Browse mode for the same file and version, so I guess the current code is okay.

Issue #1263, opened 2019-11-22
Show version comparison range in header


kumar303:
### Describe the problem and steps to reproduce it:

* Log in
* Load any version comparison page

### What happened?

The navbar header displays the add-on name

### What did you expect to happen?

It should also display the version comparison range

### Anything else we should know?

The version select boxes *currently* show an indication of the version comparison range but those are going away after https://github.com/mozilla/addons-code-manager/issues/757 (i.e. they will move into a hidden popover menu).

This is a necessary step towards implementing https://github.com/mozilla/addons-code-manager/issues/757

ioanarusiczki:
@kumar303 
I verified on dev and the version number is available now in the header.

Browse
 
![Browse](https://user-images.githubusercontent.com/33448286/69873151-61838a80-126c-11ea-9b2a-0e07d69df2a1.png)

Compare 

![Compare](https://user-images.githubusercontent.com/33448286/69873186-76f8b480-126c-11ea-9659-8c2c54edf984.png)

PR #1264, opened 2019-11-22
Show base / current version indicators in NavBar

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/1263

I noticed a few things in the tests that needed cleanup. I tried to exclude those changes but `git add -p` wasn't making it easy. Let me know if you'd rather I postpone the cleanup until later but I don't think it will make the patch much smaller.

Before:

Screenshot 2019-11-22 12 33 17

After (while browsing):

Screenshot 2019-11-22 12 33 48

After (while comparing versions):

Screenshot 2019-11-22 12 34 11

### DIFF
--- a/src/components/Navbar/index.spec.tsx
+++ b/src/components/Navbar/index.spec.tsx
@@ -1,7 +1,8 @@
+import { ShallowWrapper } from 'enzyme';
 import React from 'react';
 import { Store } from 'redux';
 
-import configureStore from '../../configureStore';
+import configureStore, { ConnectedReduxProps } from '../../configureStore';
 import CommentSummaryButton from '../CommentSummaryButton';
 import LoginButton from '../LoginButton';
 import styles from './styles.module.scss';
@@ -13,21 +14,25 @@ import {
   fakeUser,
   fakeVersion,
   fakeVersionAddon,
+  nextUniqueId,
   shallowUntilTarget,
   spyOn,
 } from '../../test-helpers';
 import { actions as userActions } from '../../reducers/users';
+import { actions as versionsActions } from '../../reducers/versions';
 
-import Navbar, { NavbarBase, PublicProps } from '.';
+import Navbar, { NavbarBase, PublicProps, mapStateToProps } from '.';
 
 describe(__filename, () => {
-  type RenderParams = Partial<PublicProps> & { store?: Store };
+  type RenderParams = Partial<PublicProps> &
+    Partial<ConnectedReduxProps> & { store?: Store };
 
   const render = ({
     store = configureStore(),
     ...moreProps
   }: RenderParams = {}) => {
     const props = {
+      _fetchVersion: jest.fn(),
       _requestLogOut: jest.fn(),
       ...moreProps,
     };
@@ -44,6 +49,22 @@ describe(__filename, () => {
     return store;
   };
 
+  const createStoreWithCurrentVersion = ({
+    addonId = nextUniqueId(),
+    id = nextUniqueId(),
+    versionString = '2.0.0',
+  } = {}) => {
+    return createStoreWithVersion({
+      version: {
+        ...fakeVersion,
+        id,
+        version: versionString,
+        addon: { ...fakeVersion.addon, id: addonId },
+      },
+      makeCurrent: true,
+    });
+  };
+
   describe('when a version is loaded', () => {
     it('renders addon name', () => {
       const addonName = 'addon name example';
@@ -56,7 +77,7 @@ describe(__filename, () => {
       });
       const root = render({ store });
 
-      expect(root.find(`.${styles.addonName}`)).toHaveText(addonName);
+      expect(root.find(`.${styles.addonName}`).text()).toContain(addonName);
     });
 
     it('renders a link to reviewer tools', () => {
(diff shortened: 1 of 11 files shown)


src/pages/Index/index.tsx
```diff
@@ -72,6 +72,7 @@ export class IndexBase extends React.Component<Props> {
 
   componentDidMount() {
     const { dispatch } = this.props;
+    dispatch(versionsActions.unsetCurrentBaseVersionId());
```
bobsilverberg: I guess it might not be worth it if this is the only place we're going to need to call both, but I was wondering if it would be worth having an action like `unsetCurrentVersions` which does both `unsetCurrentBaseVersionId` and `unsetCurrentVersionId`?

src/components/Navbar/index.tsx
```diff
@@ -9,30 +9,99 @@ import CommentSummaryButton from '../CommentSummaryButton';
...
+      currentVersion,
+      currentBaseVersion,
+      currentBaseVersionId,
+      dispatch,
+    } = this.props;
+    const { nextBaseVersionImprint } = this.state;
+
+    if (
+      currentBaseVersion &&
+      nextBaseVersionImprint !== currentBaseVersion.version
```
bobsilverberg: I know that this has been around pretty much forever, but looking at this code I am suddenly realizing that using `version` as the property name for the "version name/number" is actually quite confusing. A quick glance at this had me wondering why we were extracting a `version` from a `Version` object. I don't know if this is worth thinking about addressing in a follow-up by renaming this property.

src/components/Navbar/index.tsx
```diff
@@ -9,30 +9,99 @@ import CommentSummaryButton from '../CommentSummaryButton';
...
+  fetchVersion,
+  getVersionInfo,
+  selectCurrentVersionInfo,
+} from '../../reducers/versions';
 import { ConnectedReduxProps } from '../../configureStore';
 import { User, selectCurrentUser, requestLogOut } from '../../reducers/users';
 import styles from './styles.module.scss';
 
 export type PublicProps = {
+  _nextBaseVersionImprint?: string;
```
bobsilverberg: I gather this was added to support Storybook? I wonder if it would be better to set up the data so that this could just happen naturally with Storybook, rather than adding this "fake" property?

src/components/Navbar/index.spec.tsx
```diff
@@ -191,4 +212,213 @@ describe(__filename, () => {
...
+        versionsActions.setCurrentBaseVersionId({
+          versionId: currentBaseVersionId,
+        }),
+      );
+
+      const { _fetchVersion, dispatch, fakeThunk } = renderForFetching({
+        store,
+      });
+
+      expect(dispatch).toHaveBeenCalledWith(fakeThunk.thunk);
```
bobsilverberg: I find this a bit confusing because `fakeThunk` is not defined anywhere in this test, but rather in the `renderForFetching` helper. This means I have no idea what we're verifying here, without having to look it up in `renderForFetching`. I agree that the helper is quite useful, but I feel like this goes a bit against our approach of trying to set up data that will be asserted upon in the test case itself. I'm not sure if there is a better solution to this problem, but figured I'd raise it.

src/reducers/versions.tsx
```diff
@@ -339,7 +340,8 @@ export type VersionsState = {
   };
   currentBaseVersionId:
     | number // data successfully loaded
-    | undefined; // data has not yet loaded
+    | undefined // data has not yet loaded
+    | false; // data loaded but it is empty
```
bobsilverberg: I find this comment confusing. By "data loaded but it is empty", do you mean an ID has been loaded for baseVersion, but the baseVersion itself has not been loaded? (I think that's what `false` means). To me that's not what the comment suggests. It actually sounds like the data has been loaded and also has not been loaded (because it's empty), which is what I find confusing.

src/components/Navbar/index.tsx
kumar303: I had trouble with this but I figured out a way to do it ✨

src/components/Navbar/index.spec.tsx
kumar303: I see what you mean. Yeah, we typically make sure that anything used in an assertion is passed as a direct input to the helper.

The first option for fixing it is to make all inputs explicit, like:

<details>
<summary>Patch</summary>

```diff
diff --git a/src/components/Navbar/index.spec.tsx b/src/components/Navbar/index.spec.tsx
index 0228e6c..b44bf22 100644
--- a/src/components/Navbar/index.spec.tsx
+++ b/src/components/Navbar/index.spec.tsx
@@ -215,12 +215,15 @@ describe(__filename, () => {
 
   describe('loading data', () => {
     const renderForFetching = ({
+      fakeThunk = createFakeThunk(),
+      _fetchVersion = fakeThunk.createThunk,
       store = configureStore(),
       ...moreProps
-    }: RenderParams = {}) => {
+    }: {
+      fakeThunk?: ReturnType<typeof createFakeThunk>;
+      _fetchVersion?: ReturnType<typeof createFakeThunk>['createThunk'];
+    } & RenderParams = {}) => {
       const dispatch = spyOn(store, 'dispatch');
-      const fakeThunk = createFakeThunk();
-      const _fetchVersion = fakeThunk.createThunk;
 
       const props = { _fetchVersion, store, ...moreProps };
       const root = render(props);
@@ -258,7 +261,12 @@ describe(__filename, () => {
         }),
       );
 
-      const { _fetchVersion, dispatch, fakeThunk } = renderForFetching({
+      const fakeThunk = createFakeThunk();
+      const _fetchVersion = fakeThunk.createThunk;
+
+      const { dispatch } = renderForFetching({
+        _fetchVersion,
+        fakeThunk,
         store,
       });
 
```

</details>

However, this means that any assertion using `fakeThunk` also has to define `_fetchVersion` even if they don't use it in the assertion. That seems a little excessive because when `_fetchVersion` isn't used in an assertion, the reader has to ponder it unnecessarily.

I decided to return `_fetchVersionThunk` for use in assertions since that provides a much better hint to the reader about what it *is* which should alleviate the original problem you raised of having to look up the definition to figure it out.

<details>
<summary>Patch</summary>

```diff
diff --git a/src/components/Navbar/index.spec.tsx b/src/components/Navbar/index.spec.tsx
index 0228e6c..ea67a88 100644
--- a/src/components/Navbar/index.spec.tsx
+++ b/src/components/Navbar/index.spec.tsx
@@ -225,7 +225,12 @@ describe(__filename, () => {
       const props = { _fetchVersion, store, ...moreProps };
       const root = render(props);
 
-      return { _fetchVersion, dispatch, root, fakeThunk };
+      return {
+        _fetchVersion,
+        _fetchVersionThunk: fakeThunk.thunk,
+        dispatch,
+        root,
+      };
     };
 
     it('loads data on mount', () => {
@@ -258,11 +263,13 @@ describe(__filename, () => {
         }),
       );
 
-      const { _fetchVersion, dispatch, fakeThunk } = renderForFetching({
-        store,
-      });
+      const { _fetchVersion, _fetchVersionThunk, dispatch } = renderForFetching(
+        {
+          store,
+        },
+      );
 
-      expect(dispatch).toHaveBeenCalledWith(fakeThunk.thunk);
+      expect(dispatch).toHaveBeenCalledWith(_fetchVersionThunk);
       expect(_fetchVersion).toHaveBeenCalledWith({
         addonId,
         path: undefined,
@@ -273,16 +280,16 @@ describe(__filename, () => {
 
     it('does not fetch the base version without a current version', () => {
       const store = createStoreWithVersion({ makeCurrent: false });
-      const { dispatch, fakeThunk } = renderForFetching({ store });
+      const { _fetchVersionThunk, dispatch } = renderForFetching({ store });
 
-      expect(dispatch).not.toHaveBeenCalledWith(fakeThunk.thunk);
+      expect(dispatch).not.toHaveBeenCalledWith(_fetchVersionThunk);
     });
 
     it('does not fetch the base version without an ID', () => {
       const store = createStoreWithVersion({ makeCurrent: true });
-      const { dispatch, fakeThunk } = renderForFetching({ store });
+      const { _fetchVersionThunk, dispatch } = renderForFetching({ store });
 
-      expect(dispatch).not.toHaveBeenCalledWith(fakeThunk.thunk);
+      expect(dispatch).not.toHaveBeenCalledWith(_fetchVersionThunk);
     });
 
     it('does not fetch the base version when it already exists', () => {
@@ -299,9 +306,9 @@ describe(__filename, () => {
           versionId: currentBaseVersionId,
         }),
       );
-      const { dispatch, fakeThunk } = renderForFetching({ store });
+      const { _fetchVersionThunk, dispatch } = renderForFetching({ store });
 
-      expect(dispatch).not.toHaveBeenCalledWith(fakeThunk.thunk);
+      expect(dispatch).not.toHaveBeenCalledWith(_fetchVersionThunk);
     });
 
     it('does not fetch the base version when there was an error fetching it', () => {
@@ -317,9 +324,9 @@ describe(__filename, () => {
       store.dispatch(
         versionsActions.abortFetchVersion({ versionId: currentBaseVersionId }),
       );
-      const { dispatch, fakeThunk } = renderForFetching({ store });
+      const { _fetchVersionThunk, dispatch } = renderForFetching({ store });
 
-      expect(dispatch).not.toHaveBeenCalledWith(fakeThunk.thunk);
+      expect(dispatch).not.toHaveBeenCalledWith(_fetchVersionThunk);
     });
   });

 
```

</details>

Thanks for catching that. Let me know if you have further thoughts on improving it.

bobsilverberg: I think that's a good approach, thanks!

src/reducers/versions.tsx
kumar303: I updated the comments to make it more clear.

src/components/Navbar/index.tsx
kumar303: I agree. If you have a minute, I suggest filing a code quality issue to rename it to `versionString` or something.

src/pages/Index/index.tsx
kumar303: I'd say it's probably not worth it

Issue #1272, opened 2019-11-25
Adjust CommentSummaryButton config


kumar303:
The UI for https://github.com/mozilla/addons-code-manager/issues/757 will be sharing the new `PopoverButton` component. A few changes to `CommentSummaryButton` are needed for this.

**For QA**: please check for regressions in the *Comment Summary* button and its popover.

ioanarusiczki:
@kumar303 I verified this on dev on Browse and Compare views - looks good to me.
Verified with FF70(Win10).

PR #1273, opened 2019-11-25
Adjust CommentSummaryButton config

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/1272


src/components/CommentSummaryButton/index.spec.tsx
```diff
@@ -13,6 +11,7 @@ import {
   createStoreWithVersionComments,
   shallowUntilTarget,
   spyOn,
+  simulatePopover,
```
kumar303: I'll be using this helper in a forthcoming patch

Issue #1280, opened 2019-11-27
Slow loading indicator isn't accurate when browsing code


kumar303:
### Describe the problem and steps to reproduce it:

* Browse a file with 508 lines of code, such as https://code.addons.mozilla.org/en-US/browse/1479936/versions/3176697/?allowSlowPages=true&path=js%2Fbackground.js

### What happened?

The page warns about the file loading slowly

### What did you expect to happen?

The page should not warn about the file loading slowly because it loads quickly

ioanarusiczki:
@kumar303 
For this file with 1187 lines I can still see the warning at the top of the file "This file has been shortened to load faster. View the original file."

https://code.addons-dev.allizom.org/en-US/browse/502724/versions/1541510/?allowSlowPages=false&messageUid=c94a6ffe04da45bfac4a17ebf1dc3557&path=lib%2Fbrowser-polyfill.js

![load faster](https://user-images.githubusercontent.com/33448286/69876054-7b749b80-1273-11ea-94eb-8c74d4e4d717.png)

I could not see your [example](https://github.com/mozilla/addons-code-manager/issues/1280#issue-529431860) for the 508 lines of code. I tried to find it using the add-ons ID on prod but I get a 404 for https://addons.mozilla.org/en-US/firefox/addon/1479936/

ioanarusiczki:
@kumar303 I managed to upload using admin override an add-on with 683 lines of code 
https://code.addons-dev.allizom.org/en-US/browse/618459/versions/1689846/?path=vulcanized-polymer.js#I683
I don't see the message in Browse or Compare View. 
So I'm marking this verified on dev with FF70(Win10).

Issue #947, opened 2019-07-16
The browse page for very long files is loading slowly 


AlexandraMoga:
Extending issue #946 to the browse page as well

Try to open the browse page for the following files (~20k lines):

- https://code.addons-dev.allizom.org/en-US/browse/502868/versions/1688187/?path=src%2Fjs%2Feditor-popup.js
- https://code.addons-dev.allizom.org/en-US/browse/502868/versions/1688187/?path=src%2Fjs%2FGrammarly-bg.js
- https://code.addons-dev.allizom.org/en-US/browse/385621/versions/1655456?path=public%2Fjs%2Fbackground.js

**Actual result:**
It takes about 5 - 7 seconds for the code to load

**Expected result:**
The code should load faster

kumar303:
Thanks for filing. Even a file with 3,000 lines is slow: https://code.addons-dev.allizom.org/en-US/browse/385621/versions/1655456/?path=public%2Fjs%2Foptions.js

kumar303:
I did some profiling. The biggest win here would be trimming the content, like in https://github.com/mozilla/addons-code-manager/pull/951 . I played with disabling syntax highlighting but it doesn't speed it up by much.

AlexandraMoga:
Verified on -dev with Fx68, Win10x64

First of all, large/slow files are loaded faster now - still they aren't impressively fast (the API is also at fault here) but an improvement is indeed noticeable:

![2019-07-24_14-43-19](https://user-images.githubusercontent.com/31961530/61792455-74e99980-ae25-11e9-8483-135c79333c41.gif)
 

The reviewer has the option to view the original file by clicking on a link at the top of the code viewer. Additionally, they can switch back to trimmed code through the same option.

![image](https://user-images.githubusercontent.com/31961530/61792779-2e486f00-ae26-11e9-9178-d4fb0fd0c342.png)
![image](https://user-images.githubusercontent.com/31961530/61792780-30aac900-ae26-11e9-8fd2-2acfab3febd1.png)


@kumar303 I have the following scenario which might have been intended but I'd like to clarify it either way:
When I'm loading a large/slow file and then opt to "View the original file" the `allowSlowPages=true` parameter is set in the URL. Then, if I navigate to a different file path, `allowSlowPages=true` persists in the URL.  Would it be possible to reset this parameter as well, to avoid running into slow loading files again? Reviewers can opt to see the original file if needed.

kumar303:
[...]

PR #1281, opened 2019-11-27
Increase line limit for detecting slow files

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/1280

### DIFF
--- a/src/components/CodeView/index.tsx
+++ b/src/components/CodeView/index.tsx
@@ -25,7 +25,7 @@ import GlobalLinterMessages from '../GlobalLinterMessages';
 import SlowPageAlert from '../SlowPageAlert';
 
 // This is how many lines of code it takes to slow down the UI.
-const SLOW_LOADING_LINE_COUNT = 500;
+const SLOW_LOADING_LINE_COUNT = 1000;
 
 // This function mimics what https://github.com/rexxars/react-refractor does,
 // but we need a different layout to inline comments so we cannot use this


src/components/CodeView/index.tsx
```diff
@@ -25,7 +25,7 @@ import GlobalLinterMessages from '../GlobalLinterMessages';
 import SlowPageAlert from '../SlowPageAlert';
 
 // This is how many lines of code it takes to slow down the UI.
-const SLOW_LOADING_LINE_COUNT = 500;
+const SLOW_LOADING_LINE_COUNT = 1000;
```
bobsilverberg: Just wondering, are we just taking arbitrary guesses at this number? Might a file with 500 long lines be an issue, but one with 600 short lines not be an issue? I think this is fine to do, but I'm wondering where we are getting these numbers.

kumar303: Yeah, just a guess. I tested the files in https://github.com/mozilla/addons-code-manager/issues/947 which are legitimately slow and they are still flagged as such.

wagnerand: Yes, probably something for a follow-up, but just the line count might not be a good metric. Minified files only have a few lines, but they are often very slow to load.

kumar303: @wagnerand can you send me examples of this? When I profiled the browse view I could did not find a noticeable difference in minified files, only in diff view (which is mitigated somewhat). If necessary, we could change the detection to *waves math hands* some calculation of line width and line count.

wagnerand: Will do next time I come across it!

willdurand: We could maybe use the number of characters instead, line width/count wouldn't matter anymore.

bobsilverberg: But where's the fun in a simple solution like that? 😉

kumar303: I'm not sure if character count would be enough to detect how a single long line of minified code causes performance problems. It's worth a try, though.

wagnerand: A single metric, whatever it is, might not be enough.

PR #1294, opened 2019-12-02
Deal with delete-eofnl and insert-eofnl change types for diff c…

### LINKED ISSUE #1291: Deal with `insert-eofnl` and `delete-eofnl` change types
This is a follow-up from https://github.com/mozilla/addons-code-manager/issues/989#issuecomment-560448350.

Information that we receive about a diff from the API can include changes with change types of `insert-eofnl` and `delete-eofnl`. We should ensure we are dealing with those appropriately.

At a minimum, we need to update [ChangeType](https://github.com/mozilla/addons-code-manager/blob/2e1316294c5a855e7e8fdbe4f783504f3f58b84f/src/%40types/react-diff-view/index.d.ts#L4) to include these types. We should also disallow commenting on these two types for now, as we don't have a means of differentiating comments on these from comments on other changes which share the same line number.

I also think we should consider excluding these from the diff view entirely. To that end, we should consider the question:
- What value does including these in the diff add?

See https://code.addons-dev.allizom.org/en-US/compare/495710/versions/1541609...1688333/?path=manifest.json for an example of a diff that shows both types of changes.

@wagnerand @kumar303 What are your opinions on excluding these changes from the diff view?

For QA:

When this fix lands it should no longer be possible to comment on the lines that explain changes having to do with newlines at end of file. These appear in the diff view as a two-line comment with an empty line followed by the content "\ No newline at end of file". An example can be found at https://code.addons-dev.allizom.org/en-US/compare/495710/versions/1688329...1688333/?path=manifest.json

bobsilverberg:
Fixes #1291

### DIFF
--- a/src/@types/react-diff-view/index.d.ts
+++ b/src/@types/react-diff-view/index.d.ts
@@ -1,7 +1,13 @@
 /* eslint react/no-multi-comp: 0, max-classes-per-file: 0 */
 
 declare module 'react-diff-view' {
-  type ChangeType = 'delete' | 'insert' | 'normal';
+  type ChangeType =
+    | 'delete'
+    | 'delete-eofnl'
+    | 'insert'
+    | 'insert-eofnl'
+    | 'normal'
+    | 'normal-eofnl';
 
   export type ChangeInfo = {
     content: string;
--- a/src/@types/react-diff-view/index.d.ts
+++ b/src/@types/react-diff-view/index.d.ts
@@ -1,6 +1,7 @@
 /* eslint react/no-multi-comp: 0, max-classes-per-file: 0 */
 
 declare module 'react-diff-view' {
+  // The eofnl types are not coming from react-diff-view, but are returned by our API.
   type ChangeType =
     | 'delete'
     | 'delete-eofnl'
--- a/src/components/DiffView/index.spec.tsx
+++ b/src/components/DiffView/index.spec.tsx
@@ -1179,7 +1179,7 @@ describe(__filename, () => {
     );
 
     it.each(['delete', 'delete-eofnl', 'insert-eofnl', 'normal-eofnl'])(
-      'returns true for a(n) %s change',
+      'returns false for a(n) %s change',
       (changeType) => {
         const change = {
           ...fakeChangeInfo,
--- a/src/components/DiffView/index.spec.tsx
+++ b/src/components/DiffView/index.spec.tsx
@@ -3,6 +3,7 @@ import queryString from 'query-string';
 import React from 'react';
 import { Alert } from 'react-bootstrap';
 import {
+  ChangeType,
   Diff,
   HunkInfo,
   Hunks,
@@ -54,6 +55,7 @@ import DiffView, {
   DefaultProps,
   DiffViewBase,
   PublicProps,
+  changeCanBeCommentedUpon,
   diffCanBeHighlighted,
   getAllHunkChanges,
   trimHunkChanges,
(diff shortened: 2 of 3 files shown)


src/@types/react-diff-view/index.d.ts
```diff
@@ -1,7 +1,13 @@
 /* eslint react/no-multi-comp: 0, max-classes-per-file: 0 */
 
 declare module 'react-diff-view' {
-  type ChangeType = 'delete' | 'insert' | 'normal';
+  type ChangeType =
+    | 'delete'
+    | 'delete-eofnl'
```
kumar303: It might be good to add some comments about these since (I think?) they are not coming from `react-diff-view` but coming from our API instead.

src/components/DiffView/index.spec.tsx
```diff
@@ -719,21 +716,25 @@ describe(__filename, () => {
...
+      const _changeCanBeCommentedUpon = jest.fn().mockReturnValue(false);
+      const lineNumber = nextUniqueId();
       const version = createInternalVersion({
         ...fakeVersion,
         id: nextUniqueId(),
       });
       const diff = createDiffWithHunks([
-        createHunkWithChanges([{ type: 'delete', old_line_number: line }]),
+        createHunkWithChanges([
+          { new_line_number: lineNumber, old_line_number: lineNumber },
```
kumar303: I see `type` removed from a lot of these. Any reason? I think it's nice to see the `type` explicitly set because it makes the test easier to read.

src/components/DiffView/index.spec.tsx
```diff
@@ -1172,4 +1164,30 @@ describe(__filename, () => {
...
+          ...fakeChangeInfo,
+          type: changeType as ChangeType,
+        };
+
+        expect(changeCanBeCommentedUpon(change)).toEqual(true);
+      },
+    );
+
+    it.each(['delete', 'delete-eofnl', 'insert-eofnl', 'normal-eofnl'])(
+      'returns true for a(n) %s change',
```
kumar303: I think you meant `false`

src/components/DiffView/index.tsx
```diff
@@ -97,13 +97,18 @@ export const trimHunkChanges = (
   return trimmed;
 };
 
+export const changeCanBeCommentedUpon = (change: ChangeInfo) => {
+  return ['insert', 'normal'].includes(change.type);
```
kumar303: It would be a bit safer to use [`ChangeType`](https://github.com/mozilla/addons-code-manager/blob/55888e7084550467fcdfa0de14aca504bdd5d318/src/%40types/react-diff-view/index.d.ts#L4):

```tsx
['insert' as ChangeType, 'normal' as ChangeType].includes(change).type
```

kumar303: Sorry, what I meant was to try and make it more type safe (I don't think the code above will do that). Maybe something like this?

```tsx
const types: ChangeType[] = ['insert', 'normal'];
return types.includes(change.type);
```

src/components/DiffView/index.spec.tsx
bobsilverberg: I did this because whether or not comments are rendered in any of these tests depends on what's returned by `_changeCanBeCommentedUpon`, and not on the `type` at all, so the `type` is irrelevant. I didn't want to pass an explicit `type` because I was worried that someone reading the test would interpret that to mean that the `type` *was* affecting the test outcome.

Does that make sense?

src/components/DiffView/index.tsx
bobsilverberg: This is a good idea, thanks. I wonder, when looking at this, whether it would be better for `ChangeType` to be an `enum`? Perhaps that would be a problem for `react-diff-view`?

src/components/DiffView/index.spec.tsx
kumar303: ok, that makes sense

src/components/DiffView/index.tsx
kumar303: Yeah, making `ChangeType` an enum would probably help (maybe save that for another patch as it could produce a big diff).

bobsilverberg: Thanks. I filed #1322 as a follow-up.

PR #1312, opened 2019-12-05
Use correct line number for normal changes, and prevent widgets…

### LINKED ISSUE #877: Messages not shown in DiffView
~~Depends on https://github.com/mozilla/addons-server/issues/11853~~

I discovered this when investigating #876, but I don't know if it's at all related.

STR:

Start at http://localhost:3000/en-US/compare/502868/versions/1541685...1688183/?path=src%2Fjs%2FGrammarly-popup.js and click `a` twice to go to the previous linter message, which loads `Grammarly-bg.js`. This is a file that does contain changes. The CodeOverview shows a message at line 2, but all that is seen in the DiffView is a global linter message at the top and no messages are shown on line 2.

But here's the weirdest part: From `Grammarly-bg.js` press `z` to go back to `Grammarly-popup.js`, and then press `a` to go back up to `Grammarly-bg.js`. *Now* the messages appear!

Edit:
Following the above steps no longer seems to work to make the message appear, so now this issue seems to be about messages not appearing at all in DiffView.

bobsilverberg:
~~Depends on #1294~~
Fixes #877

### DIFF
--- a/src/components/DiffView/index.tsx
+++ b/src/components/DiffView/index.tsx
@@ -203,7 +203,11 @@ export class DiffViewBase extends React.Component<Props> {
 
     const allWidgets: WidgetMap = {};
 
-    for (const change of getAllHunkChanges(hunks)) {
+    // We only want widgets for delete, insert and normal changes, not eofnl changes.
+    const changesAllowingWidgets = getAllHunkChanges(hunks).filter(
+      (change) => change.isDelete || change.isInsert || change.isNormal,
+    );
+    for (const change of changesAllowingWidgets) {
       const changeKey = getChangeKey(change);
       const line = change.lineNumber;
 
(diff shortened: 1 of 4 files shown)


src/components/DiffView/index.tsx
```diff
@@ -201,7 +201,11 @@ export class DiffViewBase extends React.Component<Props> {
 
     const allWidgets: WidgetMap = {};
 
-    for (const change of getAllHunkChanges(hunks)) {
+    // We only want widgets for delete, insert and normal changes, not eofnl changes.
+    const changesAllowingWidgets = getAllHunkChanges(hunks).filter(
+      (change) => change.isDelete || change.isInsert || change.isNormal,
```
kumar303: Why deletes? It would be nice if you could reuse `changeCanBeCommentedUpon()` here.

kumar303: Ah, nevermind, it's because of linter messages.

Issue #677, opened 2019-05-03
Global linter messages are not visible in the overview


vcarciu:
Steps to reproduce:
1. Go to https://code.addons-dev.allizom.org/en-US/browse/494431/versions/1532144/?path=manifest.json
2.Observe the linter message at the beginning

Expected results:
The linter message should be signaled in the right side in order to be able to directly scroll at its position if you are in the middle of a long file

Actual result:
Linter message not highlighted at all 

Please see a screenshot for the issue :
![lintbegin](https://user-images.githubusercontent.com/1583842/57141499-fc91b000-6dc2-11e9-8cde-4715b89262ae.png)

bobsilverberg:
I think this is a good point. I recall that, when reviewing the code, it was intentional to not include them in the `CodeOverview`, which does make sense when you're at the top of the file. If you scroll down farther, however, it seems reasonable to expect an indication in the overview that there is a message at the top of the file.

willdurand:
Note that these are global messages, not inline messages.

bobsilverberg:
> Note that these are global messages, not inline messages.

Correct, but it still seems worthwhile to indicate that they exist in the overview.

kumar303:
I think makes sense to display global messages in the overview, at least for consistency.

ioanarusiczki:
Verified on dev with FF70(Win10). 

![global messages shown](https://user-images.githubusercontent.com/33448286/70446454-c9747500-1aa5-11ea-9a21-7e38cfb463de.png)

Issue #1315, opened 2019-12-06
CodeOverview story is broken: _fetchAndLoadComments is not a function


kumar303:
### Describe the problem and steps to reproduce it:

Go to https://mozilla.github.io/addons-code-manager/?path=/story/codeoverview--short-file

### What happened?

The story produces this error:

```
_fetchAndLoadComments is not a function
```

In the debugger, I can see that all imports from `../../reducers/comments` at the top of `CommentList/index.tsx` are undefined.

### What did you expect to happen?

It should load the story

### Anything else we should know?

git bisect says that https://github.com/mozilla/addons-code-manager/commit/76fdcaeb0d1ae06c87f996ba28e96ad425f7bb25 introduced the problem but it's not obvious why.

There is maybe some kind of storybook bug or webpack misconfiguration or who knows what. As a clue, I can make the error go away like this:

<details>
<summary>Patch</summary>

```diff
diff --git a/src/components/DiffView/index.tsx b/src/components/DiffView/index.tsx
index 49d9fb8..9cc1590 100644
--- a/src/components/DiffView/index.tsx
+++ b/src/components/DiffView/index.tsx
@@ -19,7 +19,6 @@ import { withRouter, RouteComponentProps } from 'react-router-dom';
 import makeClassName from 'classnames';
 
 import Commentable from '../Commentable';
-import CommentList from '../CommentList';
 import FadableContent from '../FadableContent';
 import LinterProvider, { LinterProviderInfo } from '../LinterProvider';
 import LinterMessage from '../LinterMessage';
@@ -40,6 +39,8 @@ import {
 import styles from './styles.module.scss';
 import 'react-diff-view/style/index.css';
 
+const CommentList = () => <div />;
+
 // This is the number of changes for which a diff starts loading slowly,
 // even when syntax highlighting is disabled.
 const SLOW_DIFF_CHANGE_COUNT = 1000;

```

</details>

kumar303:
This is yet another bug that https://github.com/mozilla/addons-code-manager/issues/425 probably would have caught.

mirefly:
It looks like that db4962c fixed this.

willdurand:
thanks @mirefly

ioanarusiczki:
@mirefly The page loads without errors. 

![without errors](https://user-images.githubusercontent.com/33448286/71409547-a7a0f380-264a-11ea-9709-292f6162c395.png)

PR #1316, opened 2019-12-06
Show global linter messages in overview

kumar303:
Fixes https://github.com/mozilla/addons-code-manager/issues/677

I wanted to add a story for this but was blocked by https://github.com/mozilla/addons-code-manager/issues/1315


kumar303:
> When you click on the line in the code overview, it jumps you to line 1 in the file...

Good catch, thanks. Fixed. We already had a mechanism for linking to the global linter container so I re-used that.

kumar303:
> Ideally it would jump you to line 0, which exists to support global messages, but I can see how that might be difficult, especially if there are also inline messages on line 1.

That's fine. I made it so that global messages take precedence (I added a test for that, too) since scrolling to the top will bring all messages into view.

PR #1330, opened 2019-12-10
Disable syntax highlighting whenever a file is truncated

### LINKED ISSUE #1328: Do not highlight code in a truncated file
This came up in discussions around #1086.

It seems like there is little value in highlighting code in a file that is also being truncated. This would lead to the conclusion that any time we are truncating a file we should also not highlight the code. That might give us a bit of a quick win in terms of performance in those cases where we are currently truncating and also highlighting.

@wagnerand Do you agree with that approach?

For QA:

After this change lands, any screen that truncates a file (both the Browse and Compare pages) should also disable syntax highlighting. Currently a file that has been truncated does not disable syntax highlighting on the Browse page, and only a file with 3000+ lines, or with a line with a length greater than 700, will disable syntax highlighting on the Compare page.

bobsilverberg:
Fixes #1328

### DIFF
--- a/src/components/CodeView/index.tsx
+++ b/src/components/CodeView/index.tsx
@@ -30,12 +30,16 @@ const SLOW_LOADING_LINE_COUNT = 1000;
 
 // This function mimics what https://github.com/rexxars/react-refractor does,
 // but we need a different layout to inline comments so we cannot use this
-// component.
-const renderCode = (
-  code: string,
-  language: string,
-  shouldHighlight: boolean,
-) => {
+// component. It also optionally does not highlight the code at all.
+const renderCode = ({
+  code,
+  language,
+  shouldHighlight,
+}: {
+  code: string;
+  language: string;
+  shouldHighlight: boolean;
+}) => {
   let value;
   if (shouldHighlight) {
     const ast = refractor.highlight(code, language);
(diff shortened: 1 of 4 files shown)


src/components/CodeView/index.tsx
```diff
@@ -182,7 +191,7 @@ export class CodeViewBase extends React.Component<Props> {
                             </td>
 
                             <td className={styles.code}>
-                              {renderHighlightedCode(code, language)}
+                              {renderCode(code, language, !codeWasTrimmed)}
```
willdurand: Could we switch to a single-object parameter for `renderCode()`? That would help understanding what the meaning of `!codeWasTrimmed` is.

src/components/CodeView/index.tsx
```diff
@@ -31,9 +31,16 @@ const SLOW_LOADING_LINE_COUNT = 1000;
 // This function mimics what https://github.com/rexxars/react-refractor does,
 // but we need a different layout to inline comments so we cannot use this
 // component.
-const renderHighlightedCode = (code: string, language: string) => {
-  const ast = refractor.highlight(code, language);
-  const value = ast.length === 0 ? code : ast.map(mapWithDepth(0));
+const renderCode = (
```
willdurand: Maybe we should update the comment above to mention that, in addition to highlighting code, it can also render non-highlighted code. The name of the function makes it clear but the comment above needs a new sentence I think (we should keep the rest of the comment because it is important).

PR #1339, opened 2019-12-13
Use same logic for not highlighting code for both Browse and Co…

### LINKED ISSUE #1337: Disable highlighting in both Compare and Browse mode
This is meant to address the issue described in #1086 

We currently disable highlighting both for long files, and also for files with long lines, in Compare mode. Files with long lines are also slow to load in Browse mode, and some quick and unscientific testing revealed that disabling highlighting helped greatly with this issue.

Therefore we should take the code we have for detecting files which should not be highlighted and share it between both Browse and Compare.

For QA:

When this change lands, files with long line lengths (greater than 500 chars) should not be highlighted either in Browse or Compare mode.

bobsilverberg:
Fixes #1337

### DIFF
--- a/src/components/CodeView/index.spec.tsx
+++ b/src/components/CodeView/index.spec.tsx
@@ -221,6 +221,19 @@ describe(__filename, () => {
     expect(line.find('.language-json')).toHaveProp('children', content);
   });
 
+  it('does not highlight code when code cannot be highlighted', () => {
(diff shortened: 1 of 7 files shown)


bobsilverberg:
Thanks for the review @willdurand. This logic may evolve over time as we also introduce truncating lines of wide files, but I think it makes sense to land little bits of it at a time like this.

PR #1348, opened 2019-12-16
Truncate wide lines, in the CodeView

### LINKED ISSUE #1347: Truncate file content based on number of characters in the CodeView
As discussed in the comments below, it is believed that all we need to do to address performance issues with large files is to truncate the contents of any file that is above a particular threshold. We do not think that we need to disable highlighting for this truncated file.

The contents should include an indication that the file was truncated (e.g., a comment added by code-manager), and a button should be available to view the whole file. The latter part is already implemented via `SlowPageAlert`.

The suggested limit for number of characters is 3000. If any file is larger than that then only the first 2000 characters of the file will be shown.

Note that follow-ups should be filed to do the same for the Compare page, as well as to remove the code that disables highlighting when a file is truncated.

bobsilverberg:
Fixes #1347

### DIFF
--- a/src/components/CodeView/index.tsx
+++ b/src/components/CodeView/index.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { withRouter, Link, RouteComponentProps } from 'react-router-dom';
 import makeClassName from 'classnames';
+import { Alert } from 'react-bootstrap';
 
 import styles from './styles.module.scss';
 import Commentable from '../Commentable';
@@ -16,6 +17,7 @@ import {
 } from './utils';
 import refractor from '../../refractor';
 import {
+  HIGHLIGHT_WIDE_LINE_LENGTH,
   codeCanBeHighlighted,
   gettext,
   getLanguageFromMimeType,
@@ -45,6 +47,9 @@ const renderCode = ({
   if (shouldHighlight) {
     const ast = refractor.highlight(code, language);
     value = ast.length === 0 ? code : ast.map(mapWithDepth(0));
+  } else {
+    // If the code cannot be highlighted, truncate the line as well.
+    value = code.substring(0, HIGHLIGHT_WIDE_LINE_LENGTH);
   }
 
   return (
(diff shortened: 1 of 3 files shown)


src/components/CodeView/index.tsx
```diff
@@ -45,6 +47,9 @@ const renderCode = ({
   if (shouldHighlight) {
     const ast = refractor.highlight(code, language);
     value = ast.length === 0 ? code : ast.map(mapWithDepth(0));
+  } else {
+    // If the code cannot be highlighted, truncate the line as well.
+    value = code.substring(0, HIGHLIGHT_WIDE_LINE_LENGTH);
```
bobsilverberg: I was thinking about this last night, and I wonder if this should be:

```js
value = code.length > HIGHLIGHT_WIDE_LINE_LENGTH ? code.substring(0, HIGHLIGHT_WIDE_LINE_LENGTH) : code;
```

I'm not sure how efficient js is at doing a `substring`, so maybe we shouldn't be calling `substring` for every line, but only for those that are in fact too wide.

WDYT?

willdurand: I think that's a micro-optimization. Let's keep things as simple as possible, I don't think this particular change will lead to a significant improvement.

(From what I am seeing, even syntax highlighting is not always the problem)

src/components/CodeView/index.tsx
```diff
@@ -140,6 +145,13 @@ export class CodeViewBase extends React.Component<Props> {
 
     const shouldHighlight =
       !codeWasTrimmed && _codeCanBeHighlighted({ code: codeLines });
+    const highlightAlert = !shouldHighlight ? (
+      <Alert className={styles.highlightAlert} variant="warning">
+        {gettext(
+          'Syntax highlighting was disabled, and some lines may have been truncated, for performance.',
```
willdurand: This makes me think that we should maybe mark the lines that are truncated. Can we append a string to each line maybe? Something like ` /* truncated */` maybe?

bobsilverberg: That's a good idea, thanks!

bobsilverberg:
We decided not to proceed with this patch.

Issue #1334, opened 2019-12-12
Wait until we load the linter messages to render the code in `CodeView`


willdurand:
I know this sounds counter-productive but when profiling [a browse page with a large file](https://code.addons-dev.allizom.org/en-US/browse/494431/versions/1532144/?allowSlowPages=true&path=lib%2Fadblockplus.js), we can clearly see that we render `CodeView` twice:

1. when we get the version data
2. when we get the linter messages

The problem is that rendering `CodeView` takes time (~8s per render in my case).

If we render the actual code only when we have all the data we need, then we will render the `CodeView` only once and that speeds up the page load.

Sketch of patch:

```diff
diff --git a/src/components/CodeView/index.tsx b/src/components/CodeView/index.tsx
index b39d3a8..6ccb11a 100644
--- a/src/components/CodeView/index.tsx
+++ b/src/components/CodeView/index.tsx
@@ -83,7 +83,7 @@ export class CodeViewBase extends React.Component<Props> {
     enableCommenting: process.env.REACT_APP_ENABLE_COMMENTING === 'true',
   };
 
-  renderWithLinterInfo = ({ selectedMessageMap }: LinterProviderInfo) => {
+  renderWithLinterInfo = ({ messageMap, selectedMessageMap }: LinterProviderInfo) => {
     const {
       _scrollToSelectedLine,
       _slowLoadingLineCount,
@@ -94,6 +94,10 @@ export class CodeViewBase extends React.Component<Props> {
       version,
     } = this.props;
 
+    if (!messageMap) {
+      return (<p>still loading</p>);
+    }
+
     const language = getLanguageFromMimeType(mimeType);
     let codeLines = getLines(content);
     let codeWasTrimmed = false;
```

bobsilverberg:
This seems like a good idea. I guess one question is whether, with a smaller file, it seems like performance is worse because it takes longer to see the code (because we are waiting for the messages too). Do you think that is likely to happen, @willdurand?

willdurand:
> This seems like a good idea. I guess one question is whether, with a smaller file, it seems like performance is worse because it takes longer to see the code (because we are waiting for the messages too). Do you think that is likely to happen, @willdurand?

Indeed, I have the same concern. Given that we would have the version content, we could likely decide whether to render the code or not depending on how big the content is, maybe? (assuming that not rendering the code without linter messages feels slow for small files).

willdurand:
[...]

PR #1350, opened 2019-12-17
Only render the `VersionFileViewer` children when linter messages are loaded

willdurand:
Fixes https://github.com/mozilla/addons-code-manager/issues/1334

---

This patch optimizes "cold starts" only because we only fetch the validation information once for a given version (if I am not mistaken). That being said, I can see huge performance boosts when cold loading a large file.

### DIFF
--- a/src/components/VersionFileViewer/index.spec.tsx
+++ b/src/components/VersionFileViewer/index.spec.tsx
@@ -122,11 +122,38 @@ describe(__filename, () => {
       .filterWhere((c) => c.prop('title') === title);
   };
 
-  it('renders children', () => {
+  it('renders a loading message when linter messages are not loaded yet', () => {
     const childClass = 'ExampleClass';
-    const root = renderWithLinterProvider({
-      children: <div className={childClass} />,
-    });
+    const root = renderWithLinterProvider(
+      {
+        children: <div className={childClass} />,
+      },
+      {
+        messageMap: undefined,
+      },
+    );
+
+    expect(root.find(`.${childClass}`)).toHaveLength(0);
+    expect(root.find(Loading)).toHaveLength(1);
+    expect(root.find(Loading)).toHaveProp(
+      'message',
+      'Loading linter information...',
+    );
+  });
+
+  it('renders children when linter messages are loaded', () => {
+    const childClass = 'ExampleClass';
+    const messageMap = getMessageMap(
+      createFakeExternalLinterResult({ messages: [] }),
+    );
+    const root = renderWithLinterProvider(
+      {
+        children: <div className={childClass} />,
+      },
+      {
+        messageMap,
+      },
+    );
 
     expect(root.find(`.${childClass}`)).toHaveLength(1);
   });
--- a/src/components/VersionFileViewer/index.tsx
+++ b/src/components/VersionFileViewer/index.tsx
@@ -103,7 +103,11 @@ const VersionFileViewer = ({
           ) : null
         }
       >
-        {children}
+        {messageMap ? (
+          children
+        ) : (
+          <Loading message={gettext('Loading linter information...')} />
+        )}
       </ContentShell>
     );
   };


willdurand:
I am confused because I retried the experiments and there are differences.

### STR:

URL for the test: http://localhost:3000/en-US/browse/494431/versions/1532144/?allowSlowPages=true&path=lib%2Fadblockplus.js

- checkout branch
- run `yarn start-local-dev`
- In Chrome, open the devtools
- Select Network panel
- Make sure no cache is selected
- Use hard refresh (<kbd>cmd</kbd> + <kbd>shift</kbd> + <kbd>r</kbd>) every time
- Record the "Finish" time once it is stable


### Results:

- [`master`](https://github.com/mozilla/addons-code-manager/commit/410cd7d495ca4b200c3bc4b8dc3982345fc6c694) branch:

7.23s
7.36s
7.22s
7.17s
6.99s

- [`wait-for-linter`](https://github.com/mozilla/addons-code-manager/pull/1350/commits/11100669ebe01c7cf8599ffdb0c32026bae4afd6) branch:

4.51s
4.52s
4.52s
4.39s
4.93s

bobsilverberg:
> I am confused because I retried the experiments and there are differences.
> ### STR:
> 
> URL for the test: http://localhost:3000/en-US/browse/494431/versions/1532144/?allowSlowPages=true&path=lib%2Fadblockplus.js
> 
>     * checkout branch
> 
>     * run `yarn start-local-dev`
> 
>     * In Chrome, open the devtools
> 
>     * Select Network panel
> 
>     * Make sure no cache is selected
> 
>     * Use hard refresh (cmd + shift + r) every time
> 
>     * Record the "Finish" time once it is stable
> 
> 
> ### Results:
> 
>     * [`master`](https://github.com/mozilla/addons-code-manager/commit/410cd7d495ca4b200c3bc4b8dc3982345fc6c694) branch:
> 
> 
> 7.23s
> 7.36s
> 7.22s
> 7.17s
> 6.99s
> 
>     * [`wait-for-linter`](https://github.com/mozilla/addons-code-manager/pull/1350/commits/11100669ebe01c7cf8599ffdb0c32026bae4afd6) branch:
> 
> 
> 4.51s
> 4.52s
> 4.52s
> 4.39s
> 4.93s

I am seeing the same thing as you when I follow those steps and look at timings from Chrome's devtools, but... if I use Firefox, and just observe the timing (not using any tools, but simply counting), I still find that there is virtually no difference between `master` and `wait-for-linter`, in terms of when I finally arrive at a working screen. In fact, it sometimes seems that `wait-for-linter` is slower.

When I try a similar STR using Firefox, using the Network panel and the "Finish" time, I observe two things:
1. The "Finish" time does not reflect when the screen is actually ready. It stops counting while the "Loading Linter information..." message is still displayed, so it doesn't see a totally accurate way of timing.
2. Timing with a stopwatch from the time I hard refresh to the time the screen is ready, I'm still seeing virtually no difference, and in fact sometimes `wait-for-linter` is slower.

As Firefox is our target platform (I assume), we should probably make sure that performance improvements affect that platform. Perhaps you could do some testing on Firefox and see if you're seeing the same things as me?

bobsilverberg:
More about this:

When I do the manual stopwatch method, where I start a stopwatch when I reload, and stop it when I observe that the screen is ready, even in Chrome I am not seeing a significant difference between the branches. The loading time seems to average about 10 seconds for both branches, with each branch varying by 1-2 seconds, but not uniformly. This is making me question whether using the "Finish" time in the Network panel is an accurate way of measuring this. @willdurand perhaps you could try a similar test on Chrome and see if you observe something similar?

willdurand:
> As Firefox is our target platform (I assume), we should probably make sure that performance improvements affect that platform. Perhaps you could do some testing on Firefox and see if you're seeing the same things as me?

I only use Chrome because the finish time is reliable, not because I want to "optimize" for it. I should have mentioned earlier obviously but Firefox time is flawed as you discovered.

In Chrome, the finish time stops when the page does not change, which is very accurate. You can try to click a button in the page a few minutes and you'll see the finish time being updated.

I don't think we can reliably use Firefox to measure time and I don't think we can use a stopwatch manually because the human reaction time is 0.25s.

I found this performance trick by profiling our app and it seems logical because rendering code is slow as we have to render lots of components with HL, linter messages, etc.

bobsilverberg:
> > As Firefox is our target platform (I assume), we should probably make sure that performance improvements affect that platform. Perhaps you could do some testing on Firefox and see if you're seeing the same things as me?
> 
> I only use Chrome because the finish time is reliable, not because I want to "optimize" for it. I should have mentioned it earlier obviously but Firefox time is flawed (as you discovered).
> 
> In Chrome, the finish time stops when the page does not change, which is very accurate. You can try to click a button in the page a few minutes later and you'll see the finish time being updated.
> 
> I don't think we can reliably use Firefox to measure time and I don't think we can use a stopwatch manually because the human reaction time is 0.25s.
> 

Well, if the differences we are seeing are in the number of seconds, then 0.25s seems like an acceptable margin of error.

> I found this performance trick by profiling our app and it seems logical because rendering code is slow as we have to render lots of components with HL, linter messages, etc.

I'm just concerned because I don't feel like I'm "seeing" any improvement in Firefox, but if you are confident that this will help, even in Firefox, then I guess it's fine. I'll take a final look at the code.

bobsilverberg:
> 
> Well, if the differences we are seeing are in the number of seconds, then 0.25s seems like an acceptable margin of error.
>

I'm not sure I worded that very well. What I mean is, if the difference is 3 seconds, then 0.25 seems like a small enough fraction of that to make a manual stopwatch-style test still valid.

willdurand:
> When I do the manual stopwatch method, where I start a stopwatch when I reload, and stop it when I observe that the screen is ready, even in Chrome I am not seeing a significant difference between the branches. The loading time seems to average about 10 seconds for both branches, with each branch varying by 1-2 seconds, but not uniformly. This is making me question whether using the "Finish" time in the Network panel is an accurate way of measuring this. @willdurand perhaps you could try a similar test on Chrome and see if you observe something similar?

I see. For me the Finish time in Chrome is very reliable, it changes when the code gets loaded and then it does not anymore (in FF, it does not update after a few seconds without any change).

I get that we should target FF though, but it is actually hard :( While the hosted envs in FF work well (and that's quite fast), local dev is slow and I often have to kill some tabs.

willdurand:
> I'm just concerned because I don't feel like I'm "seeing" any improvement in Firefox, but if you are confident that this will help, even in Firefox, then I guess it's fine. I'll take a final look at the code.

I see. That seems to be a good idea to me, but I am not 100% sure, mainly because -dev and local dev (production) behave very differently for me (in FF). That is why I focused on numbers that I did not measure myself.

> I'm not sure I worded that very well. What I mean is, if the difference is 3 seconds, then 0.25 seems like a small enough fraction of that to make a manual stopwatch-style test still valid.

That is true but every second matters I think. I have the feeling that this would be helpful but maybe it isn't in the end.

bobsilverberg:
> > I'm just concerned because I don't feel like I'm "seeing" any improvement in Firefox, but if you are confident that this will help, even in Firefox, then I guess it's fine. I'll take a final look at the code.
> 
> I see. That seems to be a good idea to me, but I am not 100% sure, mainly because -dev and local dev (production) behave very differently for me (in FF). That is why I focused on numbers that I did not measure myself.
> 
> > I'm not sure I worded that very well. What I mean is, if the difference is 3 seconds, then 0.25 seems like a small enough fraction of that to make a manual stopwatch-style test still valid.
> 
> That is true but every second matters I think. I have the feeling that this would be helpful but maybe it isn't in the end.

I wonder if we should try to land some sort of way to measure this, so that once it is deployed to other environments we can verify whether there is in fact an improvement? Are you still working on trying to get `window.performance` working reliably?

willdurand:
> I wonder if we should try to land some sort of way to measure this, so that once it is deployed to other environments we can verify whether there is in fact an improvement?

That would be nice indeed.

> Are you still working on trying to get `window.performance` working reliably?

I am. I am wondering where we could send such data because that would likely give more useful metrics if we were able to collect metrics for all users.

bobsilverberg:
> > I wonder if we should try to land some sort of way to measure this, so that once it is deployed to other environments we can verify whether there is in fact an improvement?
> 
> That would be nice indeed.
> 
> > Are you still working on trying to get `window.performance` working reliably?
> 
> I am. I am wondering where we could send such data because that would likely give more useful metrics if we were able to collect metrics for all users.

Agreed. I'm not sure what the answer is. TBH, I'm not familiar with all of the tools we have at our disposal for logging and reporting. Maybe @eviljeff or @EnTeQuAk has some ideas?

willdurand:
> r+wc (The comments being all that stuff about performance, but if you are confident this will be an improvement then I'm fine with the code landing.)

I'd like to land this patch and see if it changes anything in a "production-like" setup. That will tell us whether it was a good idea and also if using Chrome for perf-related work was wise.

Issue #1351, opened 2019-12-17
TypeError: _codeCanBeHighlighted is not a function


willdurand:
### Describe the problem and steps to reproduce it:

1. Open https://code.addons-dev.allizom.org/en-US/compare/502955/versions/1541798...1541798/?path=popup%2Fpopup.js

### What happened?

```
Oops, something went wrong.
```

In local dev, the error is:

```
TypeError: _codeCanBeHighlighted is not a function
```

### What did you expect to happen?

No error.

### Anything else we should know?

It looks like a compilation error (webpack related) because the source code seems correct. The patch that introduced this "issue" is: https://github.com/mozilla/addons-code-manager/pull/1339

ioanarusiczki:
@willdurand I could not reproduce this issue on dev with FF71(Win10). 

![no error](https://user-images.githubusercontent.com/33448286/71360034-1d4b8780-2597-11ea-8a12-b04d403f553a.gif)

PR #1352, opened 2019-12-17
Move shared utils to src/utils.tsx

willdurand:
Fixes https://github.com/mozilla/addons-code-manager/issues/1351


willdurand:
> Looks good, @willdurand, but why does it fix #1351?

ah sorry, I should have mentioned this. There was a circular dependency and that caused the function to not be injected correctly in the bundle generated by webpack :/

bobsilverberg:
> > Looks good, @willdurand, but why does it fix #1351?
> 
> ah sorry, I should have mentioned this. There was a circular dependency and that caused the function to not be injected correctly in the bundle generated by webpack :/

Ah, that makes sense! It seems odd that none of our tests, nor even `tsc`, caught this. Does that surprise you, or is that expected?

willdurand:
> Ah, that makes sense! It seems odd that none of our tests, nor even `tsc`, caught this. Does that surprise you, or is that expected?

The fact that the CI did not catch this problem does not surprise me because we are not testing the bundle generated by webpack. What surprises me is that I did not catch it locally before, though.

Circular dependencies are hard to detect. There are webpack plugins for that and I thought webpack was smart enough to detect them too...

bobsilverberg:
> > Ah, that makes sense! It seems odd that none of our tests, nor even `tsc`, caught this. Does that surprise you, or is that expected?
> 
> The fact that the CI did not catch this problem does not surprise me because we are not testing the bundle generated by webpack. What surprises me is that I did not catch it locally before, though.
> 
> Circular dependencies are hard to detect. There are webpack plugins for that and I thought webpack was smart enough to detect them too...

Yes, I should have caught this even before creating the PR. I guess I only manually tested the CodeView and not the DiffView.

PR #1361, opened 2019-12-18
Extract and update createHunkWithChanges

### LINKED ISSUE #1357: Extract and update createHunkWithChanges
While working on #920, I discovered that I need a `createHunkWithChanges` helper in some new tests as well, so this issue is about moving it from `src/components/DiffView/index.spec.tsx` into `src/test-helpers.tsx`, and also updating it so that it can be passed changes that include a simple `lineNumber`, instead of requiring `old_line_number` and `new_line_number`.

bobsilverberg:
Fixes #1357 

Note that I didn't add tests for the new `createHunkWithChanges` helper as we don't generally add tests for test helpers that generate data. The tests that use the helpers provide coverage, kind of. I am happy to add tests for it if you think that would be good, as it does include some logic.

### DIFF
--- a/src/components/DiffView/index.spec.tsx
+++ b/src/components/DiffView/index.spec.tsx
@@ -32,7 +32,6 @@ import {
   getAllHunkChanges,
 } from '../../utils';
 import {
-  ExternalChange,
   ExternalHunk,
   ScrollTarget,
   createInternalDiff,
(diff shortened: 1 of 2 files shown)


bobsilverberg:
@willdurand Hold off on reviewing this. I think I might not need it at all. I might be able to use a mock in the test in which I thought I'd need this.

bobsilverberg:
This is a nice enhancement to the helper, but we don't actually need it right now, so I'm going to close the PR.

PR #1363, opened 2019-12-19
Truncate file content based on number of characters in the Code…

### LINKED ISSUE #1347: Truncate file content based on number of characters in the CodeView

bobsilverberg:
Fixes #1347

### DIFF
--- a/src/components/CodeView/index.spec.tsx
+++ b/src/components/CodeView/index.spec.tsx
@@ -152,21 +152,19 @@ describe(__filename, () => {
   };
 
   const renderSlowLoadingCode = ({
-    _slowLoadingLineCount = 3,
-    contentLineCount,
-    content = new Array(
-      contentLineCount !== undefined
-        ? contentLineCount
+    _slowLoadingCharCount = 3,
+    contentCharCount,
+    content = 'a'.repeat(
+      contentCharCount !== undefined
+        ? contentCharCount
         : // Simulate a long file (which will load slowly) by exceeding the
-          // line limit.
-          _slowLoadingLineCount + 1,
-    )
-      .fill('// example code')
-      .join('\n'),
+          // char limit.
+          _slowLoadingCharCount + 1,
+    ),
     ...moreProps
-  }: RenderParams & { contentLineCount?: number } = {}) => {
+  }: RenderParams & { contentCharCount?: number } = {}) => {
     return renderWithLinterProvider({
-      _slowLoadingLineCount,
+      _slowLoadingCharCount,
       content,
       ...moreProps,
     });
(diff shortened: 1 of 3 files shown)


bobsilverberg:
@willdurand I am unsure about a couple of things, but figured it was worth proposing a PR for discussion:

1. Why, in your example, did you make the number of chars used to determine whether a file was too large (3000) different from the number of chars used when trimming the file (2000)? I implemented it as you suggested, but I'm not sure why we need two different numbers.

2. I'm not convinced that 3000 is a good number. When I was casually paging through some files in an add-on, I came across a number that were shortened, but weren't (in my mind) ridiculously long, nor were they minified. I guess my concern is that users are going to be frustrated if they encounter a large number of files that they then have to manually choose to un-trim. I fear we might have trouble finding the balance between minified files which slow down the UI, and non-minified files that would be perfectly fine without truncation, but yet have a lot of characters. I think we discussed this already, but I guess we can see what reviewers say and try to tweak the settings in a way that works for both cases.

willdurand:
>     1. Why, in your example, did you make the number of chars used to determine whether a file was too large (3000) different from the number of chars used when trimming the file (2000)? I implemented it as you suggested, but I'm not sure why we need two different numbers.

That's because I tried different values and the final gist contained different values, sorry about that :/

>     2\. I'm not convinced that 3000 is a good number. When I was casually paging through some files in an add-on, I came across a number that were shortened, but weren't (in my mind) ridiculously long, nor were they minified. I guess my concern is that users are going to be frustrated if they encounter a large number of files that they then have to manually choose to un-trim. I fear we might have trouble finding the balance between minified files which slow down the UI, and non-minified files that would be perfectly fine without truncation, but yet have a lot of characters. I think we discussed this already, but I guess we can see what reviewers say and try to tweak the settings in a way that works for both cases.

We should have a value that makes page loading super fast so that reviewers can move fast between files.  Loading the content entirely of a "not ridiculously long" file should also be quite fast anyway.

That makes me think that the two different values above could help because we could consider a "slow file" at X characters and only render Y characters so that it is still blazing fast.

Between Y and X characters, loading a file (entirely) will be a bit slower but maybe acceptable to address your 2nd point.

bobsilverberg:
> 
> We should have a value that makes page loading super fast so that reviewers can move fast between files. Loading the content entirely of a "not ridiculously long" file should also be quite fast anyway.
>

Yes, if their priority is to be able to page through files quickly then this seems like it could be okay. My concern is that they're not going to be happy having to manually choose to view the entire file often, if the number we choose is too low, and if when they choose to view the whole file it's still quite fast, they may question why we shortened it at all in the first place.

willdurand:
> Yes, if their priority is to be able to page through files quickly then this seems like it could be okay. My concern is that they're not going to be happy having to manually choose to view the entire file often, if the number we choose is too low, and if when they choose to view the whole file it's still quite fast, they may question why we shortened it at all in the first place.

Yep, we don't have any of that right now, though :/

Maybe adjust the numbers so that we truncate at 4k or 5k but only render 2 or 3k.

bobsilverberg:
> > Yes, if their priority is to be able to page through files quickly then this seems like it could be okay. My concern is that they're not going to be happy having to manually choose to view the entire file often, if the number we choose is too low, and if when they choose to view the whole file it's still quite fast, they may question why we shortened it at all in the first place.
> 
> Yep, we don't have any of that right now, though :/
> 
> Maybe adjust the numbers so that we truncate at 4k or 5k but only render 2 or 3k.

What do you think of the idea of making this user-configurable? Not as part of this patch, but as a follow-up? Then users could potentially tweak it to suit them, and their machine (as I imagine the power of the machine might have a pretty big impact on this).

willdurand:
> What do you think of the idea of making this user-configurable? Not as part of this patch, but as a follow-up? Then users could potentially tweak it to suit them, and their machine (as I imagine the power of the machine might have a pretty big impact on this).

I don't think we should provide that.

We should rather speed up the rendering of a full content (client-side if possible). This trick to load partial files is a trick to buy us some more time. GitHub does it too, maybe less frequently but still, which means two things: (1) it is hard and (2) we'll have to keep this trick in place anyway.

PR #1401, opened 2020-01-20
Fix travis config

willdurand:


muffinresearch:
@willdurand was there some context for the change?

willdurand:
@muffinresearch sorry, I should have filed an issue first. Here it is: https://github.com/mozilla/addons-code-manager/issues/1403

.travis.yml
```diff
@@ -13,9 +13,6 @@ cache:
 # See: https://github.com/mozilla/addons-frontend/issues/3034
 install: yarn install --pure-lockfile
 
-matrix:
-  fast_finish: true
```
willdurand: When I looked at the travis manual, the setting was only shown for `matrix` so I removed it to make sure subsequent builds would pass. I now see that it works for `jobs` too, so we can likely add it back..

PR #1404, opened 2020-01-21
Re add `fast_finish` to travis config

willdurand:
See also: https://github.com/mozilla/addons-code-manager/pull/1401

---

I guess I should have requested a review instead of quick-fixing the build.. Sorry about that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ cache:
 install: yarn install --pure-lockfile
 
 jobs:
+  fast_finish: true
   include:
     - name: unit tests (node@production)
       <<: *node-prod


willdurand:
> Can you explain this a bit? What exactly does `fast_finish` do?

https://blog.travis-ci.com/2013-11-27-fast-finishing-builds

willdurand:
I am wondering why this isn't a default setting though. Who wants to wait when a job has failed? 🤔

bobsilverberg:
> I am wondering why this isn't a default setting though. Who wants to wait when a job has failed? 🤔

After reading the post you linked to, it sounded to me like it was more about waiting for a step to finish that's allowed to fail. That's a bit different than waiting for a job after a step has failed. But maybe that's what you meant.

willdurand:
mmh according to this paragraph, it can also be when a job has failed, no?

> With fast finishing enabled, Travis CI will mark your build as finished as soon as one of two conditions are met: The only remaining jobs are allowed to fail, **or a job has already failed**. In these cases, the status of the build can already be determined, so there’s no need to wait around until the other jobs finish.

bobsilverberg:
> mmh according to this paragraph, it can also be when a job has failed, no?
> 
> > With fast finishing enabled, Travis CI will mark your build as finished as soon as one of two conditions are met: The only remaining jobs are allowed to fail, **or a job has already failed**. In these cases, the status of the build can already be determined, so there’s no need to wait around until the other jobs finish.

You're right. I must have stopped reading after the first case. :-P

PR #1411, opened 2020-01-23
Add a tokenize worker to speed up code highlighting - WIP

bobsilverberg:
This is a patch that does most of what we need to do, but it's not quite working. When the value of `state.tokens` is updated via `this.setState()`, the `Diff` component does not rerender. I believe this is because of how `LinterProvider` works and the fact that the code that needs to react to this change of state is buried inside `renderWithMessages`.

Note also that this patch contains a bunch of debugging statements, and also some hacks, so it's not at all ready for prime time, but more of a proof of concept.

Any ideas about what we can do to get `Diff` to rerender when `state.tokens` changes would be appreciated.


bobsilverberg:
@willdurand Here's a patch with most of the work I've done thus far to get the web worker thing working.

bobsilverberg:
@willdurand Here is a working version if you want to try it for yourself.

bobsilverberg:
I'm going to close this as we've decided that for now this doesn't bring enough of a performance boost to accept the UI issues with the flash of content.

PR #1426, opened 2020-02-07
Add Google Analytics support to the app

### LINKED ISSUE #1425: Add code to support Google Analytics
This is a spin-off from #1424 and is about implementing a `tracking` utility, similar to the one we have in addons-frontend, so that we can use Google Analytics for Code Manager.

bobsilverberg:
Fixes #1425 

Note that this code is based on the code we have in addons-frontend for GA. I assumed we want to keep the same logic to continue to support DNT.


.env.common-local
```diff
@@ -22,3 +22,8 @@ REACT_APP_UNSAFE_ERROR_SIMULATION=true
 #
 # Note: this variable is ignored when NODE_ENV is not `production`.
 PUBLIC_URL=/
+
+# Variables for Google Analytics
+REACT_APP_ANALYTICS_HOST=https://www.google-analytics.com
```
eviljeff: this doesn't need to be specified again if it's already in the root `.env` I presume.

.env
```diff
@@ -27,3 +27,8 @@ REACT_APP_UNSAFE_ERROR_SIMULATION=false
 
 # When true, commenting features will be enabled.
 REACT_APP_ENABLE_COMMENTING=true
+
+# Variables for Google Analytics
+REACT_APP_ANALYTICS_HOST=https://www.google-analytics.com
+REACT_APP_TRACKING_ENABLED=true
+REACT_APP_TRACKING_ID=GA-ID-would-go-here
```
eviljeff: Where are the GA ids?

src/server/index.spec.tsx
```diff
@@ -9,6 +9,8 @@ import { RequestWithCookies } from 'universal-cookie-express';
 import Cookies from 'universal-cookie';
```
eviljeff: Is this a test file?  How it is we're expected to differentiate between tests and live code?

.env
bobsilverberg: We have requested that this site be set up for GA, but that hasn't been completed yet. See https://bugzilla.mozilla.org/show_bug.cgi?id=1613724.

src/server/index.spec.tsx
bobsilverberg: In code-manager, files with `.spec` in their filename are test files. We generally name the tests the same as the source files, with `.spec`, so in this case `src/server/index.tsx` and `src/server/index.spec.tsx`.

.env
bobsilverberg: We've received the ids now, so I've updated the PR to include them.

PR #1440, opened 2020-02-17
Try using notifications.slack.room

muffinresearch:


muffinresearch:
The other places have started working by themselves after a time, so I'm inclined to close this and see if it starts up by itself.

PR #1454, opened 2020-02-21
Send fetch timings to GA

### LINKED ISSUE #1450: Record fetch performance in Google Analytics
This is a spin-off from #1423.

This is about capturing `fetch` performance via `window.Performance` and recording it in Google Analytics so that we can report on it and identify performance bottlenecks.

For QA:

After this lands, "beacons" for Google Analytics pings should appear in the network panel to indicate that a timing was sent for each `fetch` operation. There will also be a console message for each one. The console messages will look like:

`[GA: ON] Sending timing for {"category":"resource","label":"http://localhost:3000/api/v5/accounts/profile/?lang=en-US","value":298,"variable":"fetch"}`

The network panel entries will look like:

Screenshot 2020-02-24 14 42 40

and if you click on the network panel entry for the "beacon", and check the "Params" tab, you'll see the content of the beacon, which, for the example quoted above, would look like `v=1&_v=j81d&a=132045738&t=timing&_s=1&dl=http%3A%2F%2Flocalhost%2Fen-US%2Fbrowse%2F502955%2Fversions%2F1541786%2F&ul=en-us&de=UTF-8&dt=Addons%20Code%20Manager&sd=24-bit&sr=960x600&vp=1455x419&je=0&fl=32.0%20r0&utc=resource&utv=fetch&utl=http%3A%2F%2Flocalhost%3A3000%2Fapi%2Fv5%2Faccounts%2Fprofile%2F%3Flang%3Den-US&utt=298&_u=SACAAAAB~&jid=1792451120&gjid=1478828136&cid=474789440.1575918409&tid=UA-77033033-31&_gid=1676997539.1582562163&_r=1&z=1512349024`

Note that the value of `utl` above is the same as the `label` in the console message, and that the value of `utt` is the same as the `value` in the console message.

bobsilverberg:
Fixes #1450 
~~Depends on #1452~~

### DIFF
--- a/src/components/App/index.tsx
+++ b/src/components/App/index.tsx
@@ -32,14 +32,33 @@ import Compare from '../../pages/Compare';
 import Index from '../../pages/Index';
 import NotFound from '../../pages/NotFound';
 import { gettext } from '../../utils';
+import tracking from '../../tracking';
+
+export const resourceObserverCallback = (
+  _tracking: typeof tracking,
+  resource: Partial<PerformanceResourceTiming>,
+) => {
+  if (resource.initiatorType === 'fetch' && resource.duration) {
+    _tracking.timing({
+      category: 'resource',
+      label: resource.name,
+      value: resource.duration,
+      variable: 'fetch',
+    });
+  }
+};
+
+export type MockObserver = { disconnect: () => void; observe: () => void };
 
 export type PublicProps = {
+  _mockObserver?: MockObserver;
   authToken: string | null;
 };
 
 export type DefaultProps = {
   _fetchCurrentUser: typeof fetchCurrentUser;
   _log: typeof log;
+  _tracking: typeof tracking;
 };
 
 type PropsFromState = {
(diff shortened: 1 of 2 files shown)


bobsilverberg:
For some reason a coverage report was not run for this PR. Is it because it includes two commits? That would be strange. This depends on #1452, so maybe after that lands, and I rebase this, the coverage report will run?

bobsilverberg:
Note that this patch does not have complete test coverage as I was unable to figure out how to write a test to cover the firing of the PerformanceObserver. I asked a [question over at SO](https://stackoverflow.com/questions/60338965/how-to-mock-performanceobserver-with-jest) about it, but nobody has made any suggestions yet. I managed to create partial coverage for the observer, but not as much as I'd like. I suppose it's not particularly critical as this code doesn't support a feature that will affect users, and is only there to generate data for ourselves, so if it were to break that would only be an issue for us, and we would be in a position to notice it.

bobsilverberg:
@diox FYI, I am hoping to land this for tomorrow's tag, if possible.

src/components/App/index.tsx
```diff
@@ -61,8 +80,38 @@ export class AppBase extends React.Component<Props> {
...
+  resourceObserver: PerformanceObserver | MockObserver;
+
+  createResourceObserver() {
+    const { _mockObserver, _tracking } = this.props;
+
+    if (_mockObserver) {
+      return _mockObserver;
+    }
+
+    const observer = new PerformanceObserver((list) => {
```
diox: So, I know you mentioned this in a PR comment and I am not super familiar with how mocking such web APIs is done elsewhere already, but I'm not a fan of the mocking approach here. It feels like a weird compromise: you have some test/mocking specific code in `createResourceObserver` (creating a huge difference in how the whole thing behaves in tests and in "production") but at the same time don't go all the way and just return the mock directly, bypassing a bunch of code.

What's preventing you from instantiating `resourceObserver` here ?

bobsilverberg: I too am not very happy with this approach, but I've been trying to figure out how to get the whole thing mocked and thus far have had no luck. I suppose if I can create a true mocked version of `PerformanceObserver` I should be able to instantiate it here. I'll give that a try. It's fine if this doesn't make it into the tag. I'd rather get this mocking working fully.

bobsilverberg: I have made an improvement to the mocking, but it's still unable to fully test the observer. I injected a mock class so that we can better test the constructor, and we do not have separate paths for test vs. production, which was poor.

I am still unable to figure out how to test the body of the observer. The callback is called whenever a performance entry for a `resource` is recorded, but I can't figure out how to emulate that via a mock and a test. I will continue searching, and will see if I can ask around Mozilla, but at this point I'm just not sure how to do it.

I think what is here is an improvement over the previous version, and it should give us a bit more coverage, but coverage for the actual observer callback is still missing.

bobsilverberg:
I'm going to land this now so we can verify that it works on dev. I will file a follow-up to attempt to get coverage for the small amount of code in the observer not yet covered by tests.

PR #1464, opened 2020-02-28
More mocking for PerformanceObserver

### LINKED ISSUE #1463: Try to get more code coverage for PerformanceObserver
This is a follow-up to #1454, which introduced a `PerformanceObserver` into the `App` component. I was unable to figure out how to get code coverage for the body of the callback that is defined when we create the new `PerformanceObserver`, but I intend to continue looking into this and hopefully will be able to figure it out and fix the coverage that is missing according to https://codecov.io/gh/mozilla/addons-code-manager/pull/1454/diff.

bobsilverberg:
Fixes #1463

### DIFF
--- a/src/components/App/index.spec.tsx
+++ b/src/components/App/index.spec.tsx
@@ -9,6 +9,7 @@ import ContentShell from '../FullscreenGrid/ContentShell';
 import { actions as apiActions } from '../../reducers/api';
 import { actions as errorsActions } from '../../reducers/errors';
 import { actions as userActions } from '../../reducers/users';
+import tracking from '../../tracking';
 import {
   createContextWithFakeRouter,
   createFakeThunk,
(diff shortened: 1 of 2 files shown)


bobsilverberg:
> How can we see if the pr fixes/increases the codecoverage (given that was the stated aim of the issue)?

This is a good question. I have no idea why the code coverage report for this was not generated, not how to try to force one. It usually "just happens". 

I am going to play around with a different approach as well, so I'll remove your review request for now and maybe the next version will properly generate a coverage report?

bobsilverberg:
Hmm, I cannot see how to remove a request. Maybe the fact that you replied makes it think you're already removed, as there is a "re-request" icon there now.

bobsilverberg:
I'm just going to close this for now. The code in question is (likely) temporary and is only used for logging performance data to GA, so I think what we have in place for testing thus far is fine. I'm also not sure when I'll have time to get back to this with everything else on my plate.

PR #1477, opened 2020-03-06
Update PartialExternalVersion type after API changes land

### LINKED ISSUE #1476: Update PartialExternalVersion type after API changes land
https://github.com/mozilla/addons-server/issues/13651 is removing some fields from the API response for reviewers APIs. We should update our types accordingly, after the changes land.

It looks like the only type affected is [PartialExternalVersion](https://github.com/mozilla/addons-code-manager/blob/master/src/reducers/versions.tsx#L93), but whoever works on this should verify that.

bobsilverberg:
Fixes #1476 

Note that I wasn't able to test the locally because of https://github.com/mozilla/addons-server/issues/13665, which although it's been fixed, hasn't deployed to dev yet.

I suggest someone wait to test this locally before merging.

### DIFF
--- a/src/reducers/versions.tsx
+++ b/src/reducers/versions.tsx
@@ -34,21 +34,6 @@ export enum ScrollTarget {
   lastDiff = 'lastDiff',
 }
 
-type VersionCompatibility = {
-  [appName: string]: {
-    min: string;
-    max: string;
-  };
-};
-
-type VersionLicense = {
-  id: number;
-  isCustom: boolean;
-  name: LocalizedStringMap;
-  text: LocalizedStringMap;
-  url: string;
-};
-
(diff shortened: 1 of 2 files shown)


willdurand:
> I've not done that.

@eviljeff any particular reason for not doing it? I can double-check very quickly

willdurand:
> I believe it may take a considerable amount of time for me to get a copy of addons-code-manager running so I didn't attempt it.

I see. So it does not take much time actually. Assuming you have a working node env (also required for frontend and the sectools), it is a matter of `yarn && yarn dev` after cloning the repo :wink:

I checked this patch and it looks ok.

willdurand:
Merging it now so that it won't conflict if we ever contribute to this project in the next two weeks.

PR #1496, opened 2020-03-24
Adjust reducer to store and return isMinified

### LINKED ISSUE #1495: Add uses_unknown_minified_code to the versions reducer
This is a spin off from #1462, as the patch for that looks like it's going to be fairly large.

This is just about taking the data provided by the API for `uses_unknown_minified_code` and adding that data to the file data that is stored by the `versions` reducer.

bobsilverberg:
Fixes #1495

### DIFF
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -933,6 +933,7 @@ describe(__filename, () => {
     /* eslint-disable @typescript-eslint/camelcase */
     it('returns a version file', () => {
       const downloadURL = 'http://example.org/download/file';
+      const isMinified = true;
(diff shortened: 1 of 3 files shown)


willdurand:
Do we have a code quality issue to pass a `file` instead of all the props already?

bobsilverberg:
> Do we have a code quality issue to pass a `file` instead of all the props already?

Nope. That was actually related to the next patch I'm working on, not this one, but thanks for the reminder. I'll open one.

bobsilverberg:
@willdurand I opened https://github.com/mozilla/addons-code-manager/issues/1499.

PR #1498, opened 2020-03-25
Truncate all minified files by default

### LINKED ISSUE #1462: Truncate all minified files by default
~~Depends on #1495~~

This was discussed in a meeting on Feb. 26, 2020. There is no need for a reviewer to see the full contents of a minified file, and it is known that minified files cause performance issues for the code manager, both in Browse and in Compare mode. While this will not solve the performance issue for all large files, it is seen as a quick win which is fairly easy to implement and will produce an immediate improvement.

We will use the info generated by the linter to identify a file as minified. It is possible that this could be a false positive, but in these cases a reviewer can simply click the link that allows them to view the whole file.

This will require a change to the API to inform the front end that a file is minified.

For QA:

After this lands, all minified files should be trimmed by default in the Browse and Compare views, regardless of their size. Note that, prior to this patch, in the Compare view, files *were not* trimmed by default, even if they are considered too large. After this patch lands all minified files *will be* trimmed by default in the Compare view.

bobsilverberg:
Fixes #1462

### DIFF
--- a/src/components/CodeView/index.spec.tsx
+++ b/src/components/CodeView/index.spec.tsx
@@ -64,6 +64,7 @@ describe(__filename, () => {
   }: RenderParams = {}) => {
     const props = {
       content: 'some content',
+      isMinified: false,
       linterMessagesByLine: undefined,
       mimeType: 'mime/type',
       version: createInternalVersion(fakeVersion),
@@ -483,6 +484,43 @@ describe(__filename, () => {
     );
   });
 
+  it('trims the code when identified as minified', () => {
+    const content = 'some-content';
+    const _slowLoadingCharCount = content.length + 1;
+    const _trimmedCharCount = content.length - 2;
+    const isMinified = true;
+    const mimeType = 'mime/type';
+
+    const root = renderWithLinterProvider({
+      _slowLoadingCharCount,
+      _trimmedCharCount,
+      content,
+      isMinified,
+    });
+
+    const fadable = root.find(FadableContent);
+    expect(fadable).toHaveProp('fade', true);
+
+    // Show the warning twice: top and bottom.
+    expect(root.find(SlowPageAlert)).toHaveLength(2);
+
+    const { renderContent } = simulateCommentableLine({
+      _slowLoadingCharCount,
+      _trimmedCharCount,
+      content,
+      isMinified,
+      mimeType,
+    });
+    const line = renderContent();
+
+    expect(line.find('.innerHighlightedCode')).toHaveText(
+      `${content.substring(
+        0,
+        _trimmedCharCount,
+      )} /* truncated by code-manager */`,
+    );
+  });
+
   it('does not trim code when slow pages are allowed', () => {
     const content = 'some-content';
     const _slowLoadingCharCount = content.length - 2;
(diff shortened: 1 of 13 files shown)


src/components/CodeView/index.spec.tsx
```diff
@@ -634,7 +672,7 @@ describe(__filename, () => {
     it('does not break when the element is null', () => {
       const element = null;
 
-      scrollToSelectedLine(element);
+      expect(() => scrollToSelectedLine(element)).not.toThrow();
```
bobsilverberg: Note that this was a drive-by fix for a linter warning that I figured I'd make while I was in the file.

src/components/DiffView/index.tsx
```diff
@@ -365,8 +367,11 @@ export class DiffViewBase extends React.Component<Props> {
     if (diff) {
       hunksToDisplay = diff.hunks;
 
-      if (getChangeCharCount(hunksToDisplay) >= _slowLoadingCharCount) {
-        if (!shouldAllowSlowPages({ allowByDefault: true, location })) {
+      if (
+        getChangeCharCount(hunksToDisplay) >= _slowLoadingCharCount ||
+        isMinified
```
eviljeff: can we unify this check somewhere (either with a function or with a variable) so it's not defined in both `DiffView` and `CodeView`?

src/components/CodeView/index.spec.tsx
bobsilverberg: We have a bunch of tests where we just want to make sure that something doesn't throw (because a bug was found where it did throw, and then we fixed that bug). At some point recently our linter started complaining about any tests that do not include any explicit assertions, so these tests now generate warnings. That's why we need to wrap them in an `expect` with `not.toThrow()`.

src/components/CodeView/index.tsx
```diff
@@ -101,7 +103,7 @@ export class CodeViewBase extends React.Component<Props> {
     let codeWasTrimmed = false;
     let slowAlert;
 
-    if (content.length >= _slowLoadingCharCount) {
+    if (content.length >= _slowLoadingCharCount || isMinified) {
```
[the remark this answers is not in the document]
bobsilverberg: Yeah, it's a bit unclear. Lots of similar names, probably because this evolved over time. `shouldAllowSlowPages` takes into account a URL parameter that we use for the user to be able to control whether files are truncated or not. This allows them to show the full file, even if the app thinks it should be truncated. `shouldAllowSlowPages` also allows us to configure what the default is, if no URL parameter has been provided. This is to allow the Compare view to default to no truncation.

`allowSlowPages`, farther down, is the actual value of the URL parameter in question.

These don't directly relate to the conditions `content.length >= _slowLoadingCharCount` and `isMinified`. Those are used as an initial gate to let the app know whether the content should be truncated at all or not. So if neither of those conditions is `true`, then we don't care about `shouldAllowSlowPages` or `allowSlowPages` because we are never going to truncate the file.

I hope that sheds some light on all of this.

src/components/DiffView/index.tsx
bobsilverberg: Yes, I could do that. Thanks for the suggestion.

PR #1502, opened 2020-03-27
Use React.Profiler to send render timings to GA

### LINKED ISSUE #1423: Add performance tracking
Performance is an issue at times in the app. In order to get a better handle on exactly how often performance concerns occur, we intend to add performance tracking which will contribute data to InfluxDB.

This issue is about investigating how to do that, and adding the code to record the metrics that concern us.

bobsilverberg:
Fixes #1423 

This will send data on the performance of the render function for both `CodeView` and `DiffView` to GA for future analysis. There has been some discussion about how much highlighting affects the performance, so I built in a separate measure of just the highlighting piece for the `CodeView` component.

### DIFF
--- a/src/components/CodeView/index.tsx
+++ b/src/components/CodeView/index.tsx
@@ -29,8 +29,6 @@ import LinterProvider, { LinterProviderInfo } from '../LinterProvider';
 import GlobalLinterMessages from '../GlobalLinterMessages';
 import SlowPageAlert from '../SlowPageAlert';
 
-const { Profiler } = React;
-
 // This function mimics what https://github.com/rexxars/react-refractor does,
 // but we need a different layout to inline comments so we cannot use this
 // component.
@@ -91,7 +89,8 @@ export class CodeViewBase extends React.Component<Props> {
     enableCommenting: process.env.REACT_APP_ENABLE_COMMENTING === 'true',
   };
 
-  onRenderCallback = (id: string, phase: string, actualDuration: number) => {
+  // See https://github.com/reactjs/rfcs/blob/master/text/0051-profiler.md
+  onRenderProfiler = (id: string, phase: string, actualDuration: number) => {
     this.props._sendPerfTiming({ actualDuration, id, phase });
   };
 
@@ -205,18 +204,12 @@ export class CodeViewBase extends React.Component<Props> {
                               >{`${line}`}</Link>
                               {enableCommenting && addCommentButton}
                             </td>
-
-                            <Profiler
-                              id="CodeView-Highlighting"
-                              onRender={this.onRenderCallback}
-                            >
-                              <td className={styles.code}>
-                                {renderCode({
-                                  code,
-                                  language,
-                                })}
-                              </td>
-                            </Profiler>
+                            <td className={styles.code}>
+                              {renderCode({
+                                code,
+                                language,
+                              })}
+                            </td>
                           </>
                         )}
                       </Commentable>
@@ -270,7 +263,7 @@ export class CodeViewBase extends React.Component<Props> {
     const { version } = this.props;
 
     return (
-      <Profiler id="CodeView-Render" onRender={this.onRenderCallback}>
+      <React.Profiler id="CodeView-Render" onRender={this.onRenderProfiler}>
         <LinterProvider
           versionId={version.id}
           validationURL={version.validationURL}
@@ -281,7 +274,7 @@ export class CodeViewBase extends React.Component<Props> {
           // comments per line.
           (info: LinterProviderInfo) => this.renderWithLinterInfo(info)}
         </LinterProvider>
-      </Profiler>
+      </React.Profiler>
     );
   }
 }
--- a/src/components/CodeView/index.tsx
+++ b/src/components/CodeView/index.tsx
@@ -21,12 +21,15 @@ import {
   codeShouldBeTrimmed,
   gettext,
   getLanguageFromMimeType,
+  sendPerfTiming,
   shouldAllowSlowPages,
 } from '../../utils';
 import { Version } from '../../reducers/versions';
 import LinterProvider, { LinterProviderInfo } from '../LinterProvider';
 import GlobalLinterMessages from '../GlobalLinterMessages';
 import SlowPageAlert from '../SlowPageAlert';
+
+const { Profiler } = React;
 
 // This function mimics what https://github.com/rexxars/react-refractor does,
 // but we need a different layout to inline comments so we cannot use this
@@ -71,6 +74,7 @@ export type PublicProps = {
 
 export type DefaultProps = {
   _scrollToSelectedLine: typeof scrollToSelectedLine;
+  _sendPerfTiming: typeof sendPerfTiming;
   _slowLoadingCharCount: number;
   _trimmedCharCount: number;
   enableCommenting: boolean;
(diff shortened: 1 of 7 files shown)


bobsilverberg:
@willdurand I assigned this to the team, and Mat won the lottery, but I would like to have your feedback on the patch as well, if you don't mind, as it's something different from what we've ever done before. Thanks!

bobsilverberg:
> This seems to work (*) and make sense, but I'm not familiar with any of this.
> 
> (*) Weirdly, even if I had deactivated tracking protection for this site (and had deactivated my adblocker as well) Firefox kept answering `"1"` for `navigator.noNotTrack` (and the pref was set to follow tracking protection, it wasn't set to `always`) so I had to comment the code that deactivates tracking when DNT is enabled to see this patch in action. Not sure what's going on there...

There is a config setting called `trackingEnabled` which needs to be set to `true` in order for tracking to work. This is legacy and I'm not sure why it's set up that way, but it is. I have a file called `config/local-development-amo.js`, in which I have the following code:

```
module.exports = {
    enableStrictMode: false,
    trackingEnabled: true,
    allowErrorSimulation: true,
};
```

which configures my dev environment. Perhaps if you turn on `trackingEnabled` you'll find it works for you too?

bobsilverberg:
> > This seems to work (*) and make sense, but I'm not familiar with any of this.
> > (*) Weirdly, even if I had deactivated tracking protection for this site (and had deactivated my adblocker as well) Firefox kept answering `"1"` for `navigator.noNotTrack` (and the pref was set to follow tracking protection, it wasn't set to `always`) so I had to comment the code that deactivates tracking when DNT is enabled to see this patch in action. Not sure what's going on there...
> 
> There is a config setting called `trackingEnabled` which needs to be set to `true` in order for tracking to work. This is legacy and I'm not sure why it's set up that way, but it is. I have a file called `config/local-development-amo.js`, in which I have the following code:
> 
> ```
> module.exports = {
>     enableStrictMode: false,
>     trackingEnabled: true,
>     allowErrorSimulation: true,
> };
> ```
> 
> which configures my dev environment. Perhaps if you turn on `trackingEnabled` you'll find it works for you too?

As @willdurand pointed out, that code, above is actually from addons-frontend, and not addons-code-manager, so everything I said was irrelevant. I'll look into @diox's issue a bit more.

willdurand:
> This will send data on the performance of the render function for both `CodeView` and `DiffView` to GA for future analysis. There has been some discussion about how much highlighting affects the performance, so I built in a separate measure of just the highlighting piece for the `CodeView` component.

The issue mentions InfluxDB and not GA, so I guess we won't be using InfluxDB/Grafana?

src/components/CodeView/index.tsx
```diff
@@ -21,13 +21,16 @@ import {
...
   getLanguageFromMimeType,
+  sendPerfTiming,
   shouldAllowSlowPages,
 } from '../../utils';
 import { Version } from '../../reducers/versions';
 import LinterProvider, { LinterProviderInfo } from '../LinterProvider';
 import GlobalLinterMessages from '../GlobalLinterMessages';
 import SlowPageAlert from '../SlowPageAlert';
 
+const { Profiler } = React;
```
willdurand: That's weird but why not. Can't you import it as usual? (or use `React.Profiler`)

src/components/CodeView/index.tsx
```diff
@@ -81,11 +85,16 @@ type Props = PublicProps & DefaultProps & RouteComponentProps;
 export class CodeViewBase extends React.Component<Props> {
   static defaultProps: DefaultProps = {
     _scrollToSelectedLine: scrollToSelectedLine,
+    _sendPerfTiming: sendPerfTiming,
     _slowLoadingCharCount: SLOW_LOADING_CHAR_COUNT,
     _trimmedCharCount: TRIMMED_CHAR_COUNT,
     enableCommenting: process.env.REACT_APP_ENABLE_COMMENTING === 'true',
   };
 
+  onRenderCallback = (id: string, phase: string, actualDuration: number) => {
```
willdurand:
```suggestion
  onRenderProfiler = (id: string, phase: string, actualDuration: number) => {
```

src/components/CodeView/index.tsx
```diff
@@ -197,12 +206,17 @@ export class CodeViewBase extends React.Component<Props> {
...
-                            <td className={styles.code}>
-                              {renderCode({
-                                code,
-                                language,
-                              })}
-                            </td>
+                            <Profiler
+                              id="CodeView-Highlighting"
+                              onRender={this.onRenderCallback}
+                            >
```
willdurand: So we are going to profile every single line of code?

How are we going to determine that the `Profiler` isn't slowing down the page more than without `Profiler`? I read that impact should be minimal here: https://gist.github.com/bvaughn/25e6233aeb1b4f0cdb8d8366e54a3977#gistcomment-2744384, but this part of our code is going to render a `Profiler` for every single LOC.

src/components/DiffView/index.tsx
```diff
@@ -41,11 +41,14 @@ import {
   getAllHunkChanges,
   getLanguageFromMimeType,
   gettext,
+  sendPerfTiming,
   shouldAllowSlowPages,
 } from '../../utils';
 import styles from './styles.module.scss';
 import 'react-diff-view/style/index.css';
 
+const { Profiler } = React;
```
willdurand: Same question.

src/components/DiffView/index.tsx
```diff
@@ -131,13 +135,18 @@ export class DiffViewBase extends React.Component<Props> {
...
     _getRelativeDiffAnchor: getRelativeDiffAnchor,
+    _sendPerfTiming: sendPerfTiming,
     _slowLoadingCharCount: SLOW_LOADING_CHAR_COUNT,
     _tokenize: tokenize,
     _trimmedCharCount: TRIMMED_CHAR_COUNT,
     enableCommenting: process.env.REACT_APP_ENABLE_COMMENTING === 'true',
     viewType: 'unified',
   };
 
+  onRenderCallback = (id: string, phase: string, actualDuration: number) => {
```
willdurand:
```suggestion
  onRenderProfiler = (id: string, phase: string, actualDuration: number) => {
```

src/utils.tsx
```diff
@@ -270,3 +271,22 @@ export const codeShouldBeTrimmed = (
...
+  actualDuration: number;
+  id: string;
+  phase: string;
+}) => {
+  _tracking.timing({
+    category: 'renderPerf',
+    label: phase,
+    value: actualDuration,
+    variable: id,
+  });
```
willdurand: How is it looking in GA? Can we graph these metrics? Or can we use aggregate functions at least (to compute percentiles, etc.)?

src/components/CodeView/index.tsx
willdurand: We should likely add a link to the docs: https://github.com/reactjs/rfcs/blob/master/text/0051-profiler.md above this method.

src/components/DiffView/index.tsx
willdurand: We should likely add a link to the docs: https://github.com/reactjs/rfcs/blob/master/text/0051-profiler.md above this method.

bobsilverberg:
> > This will send data on the performance of the render function for both `CodeView` and `DiffView` to GA for future analysis. There has been some discussion about how much highlighting affects the performance, so I built in a separate measure of just the highlighting piece for the `CodeView` component.
> 
> The issue mentions InfluxDB and not GA, so I guess we won't be using InfluxDB/Grafana?

Correct. I had a discussion on Slack in #infraeng-influxdb and the result was a decision not to go that route and use GA instead. I will update the issue.

src/components/CodeView/index.tsx
bobsilverberg: Ah, I didn't realize we were highlighting the code one line at a time - I thought we were doing it to all the code at once. It's odd, because when I tested this locally, I only saw one call to send the data to GA for highlighting per render, not a call for every line of code, so I wonder what exactly is happening here? Maybe `onRender` is only getting called once? If so, what exactly is being measured?

It's fine to just remove this for now, as I think we're more concerned at the moment with measuring overall performance of the component and not trying to isolate highlighting performance. That will resolve this particular issue.

src/components/CodeView/index.tsx
bobsilverberg: Sure, I can try the former and if that doesn't work use the latter. I cannot remember exactly why I needed to do this.

src/utils.tsx
bobsilverberg: We have data in GA now for all the fetch timings and I've taken a look at that and it does look like we can manipulate it in useful ways. You can take a look too, but you might need to ask for access to the code manager sites in GA.

bobsilverberg:
Thanks for the review @willdurand. I have pushed a new commit that addresses your issues. I am going to run an errand this morning, but hope to be back before noon. I'd love to get this into today's tag if possible.

PR #1517, opened 2020-04-03
Change logic for legacy file viewer links

### LINKED ISSUE #1509: Change logic for legacy file viewer links
Depends on https://github.com/mozilla/addons-server/issues/13634.

https://github.com/mozilla/addons-server/issues/13634 describes a change that will be made to the way code-manager will build links to the legacy file viewer. The exact method is TBD, but in any case will require changes to code-manager to the logic that generates the links.

This issue can be used to track the changes needed in code-manager once https://github.com/mozilla/addons-server/issues/13634 lands.

bobsilverberg:
Fixes #1509

### DIFF
--- a/src/components/Navbar/index.spec.tsx
+++ b/src/components/Navbar/index.spec.tsx
@@ -17,12 +17,9 @@ import {
   createStoreWithVersion,
   createStoreWithVersionComments,
   dispatchLoadVersionInfo,
-  fakeExternalDiff,
   fakeUser,
   fakeVersion,
   fakeVersionAddon,
-  fakeVersionEntry,
-  fakeVersionFile,
   nextUniqueId,
   shallowUntilTarget,
   spyOn,
@@ -92,18 +89,6 @@ describe(__filename, () => {
     });
   };
 
-  const _loadVersionFile = ({
-    store = configureStore(),
-    version = { ...fakeVersion, id: nextUniqueId() },
-  }) => {
-    store.dispatch(
-      versionsActions.loadVersionFile({
-        path: version.file.selected_file,
-        version,
-      }),
-    );
-  };
-
   describe('when a version is loaded', () => {
     it('renders addon name', () => {
       const addonName = 'addon name example';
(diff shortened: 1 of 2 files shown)


willdurand:
Sorry, Chris is out this week so I am going to look at this patch instead :)

src/components/Navbar/index.tsx
```diff
@@ -220,16 +216,16 @@ export class NavbarBase extends React.Component<Props, State> {
...
+          {currentVersionId ? (
             <a
               className={styles.legacyLink}
               href={
-                baseFileId
-                  ? `${baseUrlToLegacy}/compare/${file.id}...${baseFileId}/${
-                      path ? `file/${path}` : ''
+                currentBaseVersionId
+                  ? `${baseUrlToLegacy}/compare-redirect/${currentVersionId}...${currentBaseVersionId}/${
+                      path ? `?file=${path}` : ''
```
willdurand, line 225: Would having a helper function to create this query string (taking `path` as unique argument) make sense?

src/components/Navbar/index.tsx
```diff
@@ -220,16 +216,16 @@ export class NavbarBase extends React.Component<Props, State> {
...
-          {currentVersion && file ? (
+          {currentVersionId ? (
             <a
               className={styles.legacyLink}
               href={
-                baseFileId
-                  ? `${baseUrlToLegacy}/compare/${file.id}...${baseFileId}/${
-                      path ? `file/${path}` : ''
+                currentBaseVersionId
+                  ? `${baseUrlToLegacy}/compare-redirect/${currentVersionId}...${currentBaseVersionId}/${
```
willdurand, line 224: So why would we use a (default) prop for `reviewersHost` and not for the api lang? Both are env vars and behave the same.

src/components/Navbar/index.spec.tsx
```diff
@@ -264,10 +152,40 @@ describe(__filename, () => {
       expect(root.find(VersionChooser)).toHaveLength(0);
     });
 
-    it('does not render a link to the legacy file viewer', () => {
-      const root = render();
+    it('renders a link to the legacy file viewer in the browse page', () => {
```
willdurand, line 155: Nothing in this test case makes me think that we are in the browse page. Shouldn't it be something like: `...file viewer when only the currentVersionId is set`?

src/components/Navbar/index.spec.tsx
```diff
@@ -264,10 +152,40 @@ describe(__filename, () => {
...
+
+      const root = render({ reviewersHost, store, path });
+
+      expect(root.find(`.${styles.legacyLink}`)).toHaveProp(
+        'href',
+        `${reviewersHost}/${process.env.REACT_APP_DEFAULT_API_LANG}/firefox/files/browse-redirect/${versionId}/?file=${path}`,
+      );
+    });
+
+    it('renders a link to the legacy diff viewer in the compare page', () => {
```
willdurand, line 170: Nothing in this test case makes me think that we are in the compare page. Shouldn't it be something like: `...file viewer when both the currentVersionId and baseVersionId are set`?

src/components/Navbar/index.spec.tsx
```diff
@@ -513,16 +431,7 @@ describe(__filename, () => {
...
-          history: createFakeHistory(),
-          location: createFakeLocation(),
-          match: {
-            params: {},
-            isExact: true,
-            path: '/some-path',
-            url: 'some-url',
-          },
-        }),
+        ...mapStateToProps(store.getState()),
```
willdurand, line 434: Why is this change needed?

src/components/Navbar/index.tsx
bobsilverberg: There doesn't seem to be a good reason for this. I guess it evolved over time, as `reviewersHost` was added first, before there was a need for api lang, and then when the latter was added it got added in a different manner. I can change it so they are both props.

src/components/Navbar/index.tsx
bobsilverberg: It seems like we'd be adding more lines of code to define (and call) a function that does this simple calculation, than we currently have, but you're right, it is duplicated logic, so I guess it makes sense. I'll add one. I don't think we want to go the next step and use a mock for that function in tests, and have a separate test for the function though, would we? That seems like creating a lot of extra code and work for very little benefit.

willdurand: No, no mock. We can reuse the helper function in the test, though.

src/components/Navbar/index.spec.tsx
bobsilverberg: Because `ownProps` was removed as an argument to `mapStateToProps`.

PR #1542, opened 2020-04-14
Use a client-side redirect when switching between allowing slow code and not

### LINKED ISSUE #1540: Do not reload full page when switching `allowSlow`
I noticed that the page is fully reloaded when we switch between the truncated and non-truncated view.

If there is no particular reason to do this, let's see if we can avoid that and instead just reload the file contents.

bobsilverberg:
Fixes #1540

### DIFF
--- a/src/components/SlowPageAlert/index.tsx
+++ b/src/components/SlowPageAlert/index.tsx
@@ -48,6 +48,4 @@ export const SlowPageAlertBase = ({
   );
 };
 
-export default withRouter(SlowPageAlertBase) as React.ComponentType<
-  PublicProps
->;
+export default withRouter(SlowPageAlertBase);
--- a/src/components/SlowPageAlert/index.tsx
+++ b/src/components/SlowPageAlert/index.tsx
@@ -1,12 +1,9 @@
-import { Location } from 'history';
+import { History, Location } from 'history';
+import queryString from 'query-string';
 import * as React from 'react';
 import { Alert } from 'react-bootstrap';
 
-import {
-  allowSlowPagesParam,
-  createAdjustedQueryString,
-  shouldAllowSlowPages,
-} from '../../utils';
+import { allowSlowPagesParam, shouldAllowSlowPages } from '../../utils';
 
 type GetTextForSlowState = (allowSlowPages: boolean) => string;
 
@@ -15,6 +12,7 @@ export type PublicProps = {
   allowSlowPagesByDefault?: boolean;
   getLinkText: GetTextForSlowState;
   getMessage: GetTextForSlowState;
+  history: History;
   location: Location;
 };
 
@@ -23,6 +21,7 @@ const SlowPageAlertBase = ({
   allowSlowPagesByDefault,
   getLinkText,
   getMessage,
+  history,
   location,
 }: PublicProps) => {
   const allowSlowPages = _shouldAllowSlowPages({
@@ -30,17 +29,21 @@ const SlowPageAlertBase = ({
     location,
   });
 
-  const newLocation = `${location.pathname}${createAdjustedQueryString(
-    location,
-    {
-      [allowSlowPagesParam]: !allowSlowPages,
-    },
-  )}`;
+  const onClick = () => {
+    const newLocation = {
+      ...location,
+      search: queryString.stringify({
+        ...queryString.parse(location.search),
+        [allowSlowPagesParam]: !allowSlowPages,
+      }),
+    };
+    history.push(newLocation);
+  };
 
   return (
     <Alert variant="danger">
       {getMessage(allowSlowPages)}{' '}
-      <Alert.Link href={newLocation}>{getLinkText(allowSlowPages)}</Alert.Link>
+      <Alert.Link onClick={onClick}>{getLinkText(allowSlowPages)}</Alert.Link>
     </Alert>
   );
 };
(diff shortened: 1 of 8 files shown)


src/test-helpers.tsx
```diff
@@ -480,21 +480,25 @@ export const createFakeLocation = ({
   };
 };
 
+// The funny business with jestExists below is so we can use this helper with
+// Storybook, where jest does not exist.
 export const createFakeHistory = ({
   location = createFakeLocation(),
 } = {}): History => {
+  const jestExists = typeof jest !== 'undefined';
+  const fakeCallback = jestExists ? null : () => () => null;
```
willdurand, line 489: I'd rather create a fake `jest` global object in Storybook. This does not look pretty :)

src/components/SlowPageAlert/index.tsx
```diff
@@ -23,24 +21,29 @@ const SlowPageAlertBase = ({
   allowSlowPagesByDefault,
   getLinkText,
   getMessage,
+  history,
```
willdurand, line 24: Do we have the `withRouter` available in this project? That might be time to use it since we need both `history` and `location` in this component.

src/components/SlowPageAlert/index.tsx
```diff
@@ -23,24 +21,29 @@ const SlowPageAlertBase = ({
...
-  const newLocation = `${location.pathname}${createAdjustedQueryString(
-    location,
-    {
-      [allowSlowPagesParam]: !allowSlowPages,
-    },
-  )}`;
+  const onClick = () => {
+    const newLocation = {
+      ...location,
+      search: queryString.stringify({
```
willdurand, line 35: I forgot how it works but can't we use `location.query` here?

src/test-helpers.tsx
bobsilverberg: I looked into trying to get a real instance of `jest` in Storybook and didn't have much luck, but you're right, we could just have a fake `jest`. Where would you put that?

bobsilverberg:
@willdurand As long as you're looking at this... the build is failing on the `build` and `yarn serve` steps, and all I see in the log is `null`. It's already been re-run once, but I'll try yet again. If that doesn't work, do you have any advice for debugging this? `null` isn't much help.

willdurand:
> @willdurand As long as you're looking at this... the build is failing on the `build` and `yarn serve` steps, and all I see in the log is `null`. It's already been re-run once, but I'll try yet again. If that doesn't work, do you have any advice for debugging this? `null` isn't much help.

I cleared the travis-ci cache :)

bobsilverberg:
> > @willdurand As long as you're looking at this... the build is failing on the `build` and `yarn serve` steps, and all I see in the log is `null`. It's already been re-run once, but I'll try yet again. If that doesn't work, do you have any advice for debugging this? `null` isn't much help.
> 
> I cleared the travis-ci cache :)

I guess I should have thought of that. Is that a new thing we're going to have to do all the time now?

src/test-helpers.tsx
willdurand: We could put that in the storybook config:

```diff
diff --git a/stories/setup/config.tsx b/stories/setup/config.tsx
index b25ea72..6877cf2 100644
--- a/stories/setup/config.tsx
+++ b/stories/setup/config.tsx
@@ -19,6 +19,14 @@ import '../../src/styles.scss';
 // Apply some custom styles to storybook.
 import './styles.scss';
 
+// Fake jest so that we can use test helpers in stories.
+window.jest = {
+  // @ts-ignore
+  fn() {
+    return null;
+  },
+};
+
 configureApplication();
 
 // Automatically import all files ending in *.stories.tsx
```

willdurand:
> I guess I should have thought of that. Is that a new thing we're going to have to do all the time now?

I probably cleared the Travis CI cache 3 times in my life.

I believe travis updated the node version or something and it put the cache in a weird state.

src/components/SlowPageAlert/index.tsx
bobsilverberg: There is no such thing as `location.query`, just `location.search`. https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/history/index.d.ts#L29-L35

src/components/SlowPageAlert/index.tsx
bobsilverberg: Good idea. This also eliminates the issue I had with needing a fake `history` object in the story.

src/components/SlowPageAlert/index.tsx
willdurand: oh right, in addons-frontend we patch the `location`: https://github.com/mozilla/addons-frontend/blob/43c481afa8e0ab37494f250639efc2511de93fc2/src/core/server/base.js#L56

src/components/SlowPageAlert/index.tsx
```diff
@@ -15,34 +12,42 @@ export type PublicProps = {
...
       {getMessage(allowSlowPages)}{' '}
-      <Alert.Link href={newLocation}>{getLinkText(allowSlowPages)}</Alert.Link>
+      <Alert.Link onClick={onClick}>{getLinkText(allowSlowPages)}</Alert.Link>
     </Alert>
   );
 };
 
-export default SlowPageAlertBase;
+export default withRouter(SlowPageAlertBase) as React.ComponentType<
+  PublicProps
```
willdurand, line 52: Do we still have to do that? (specifying a type alias with `as`)

bobsilverberg: I thought we needed it to tell TS only to accept `PublicProps` for the component, because we declare `Props` above to also include `RouteComponentProps`, but I see now that there's some magic in `withRouter` that takes care of that for us, so you are correct, we don't need the type alias anymore.

PR #1544, opened 2020-04-16
Send the URL along with the performance timings

### LINKED ISSUE #1543: Send the URL along with the performance timings
I had originally assumed that the URL of the page that the user was on when a performance timing was sent would have somehow been recorded by GA, but that seems to not be the case. It would be good, for investigative purposes, if we could see the URL for instances in which there was a performance issue, so this issue is about making sure we send the URL along with the performance measures.

bobsilverberg:
Fixes #1543

### DIFF
--- a/src/utils.tsx
+++ b/src/utils.tsx
@@ -288,17 +288,21 @@ export const sendPerfTiming = ({
   _tracking = tracking,
   actualDuration,
   id,
-  phase,
+  url,
 }: {
   _tracking?: typeof tracking;
   actualDuration: number;
   id: string;
-  phase: string;
+  url: string;
 }) => {
   _tracking.timing({
     category: 'renderPerf',
-    label: phase,
+    label: url,
     value: actualDuration,
     variable: id,
   });
+};
+
+export const getUrlFromLocation = (location: Location) => {
+  return `${location.pathname}${location.search}`;
 };
--- a/src/utils.tsx
+++ b/src/utils.tsx
@@ -286,18 +286,18 @@ export const codeShouldBeTrimmed = ({
 
 export const sendPerfTiming = ({
   _tracking = tracking,
+  _window = window,
   actualDuration,
   id,
-  url,
 }: {
   _tracking?: typeof tracking;
+  _window?: typeof window;
   actualDuration: number;
   id: string;
-  url: string;
 }) => {
   _tracking.timing({
     category: 'renderPerf',
-    label: url,
+    label: _window.location.href,
     value: actualDuration,
     variable: id,
   });
--- a/src/utils.tsx
+++ b/src/utils.tsx
@@ -302,7 +302,3 @@ export const sendPerfTiming = ({
     variable: id,
   });
 };
-
-export const getUrlFromLocation = (location: Location) => {
-  return `${location.pathname}${location.search}`;
-};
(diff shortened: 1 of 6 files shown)


src/utils.tsx
```diff
@@ -288,17 +288,21 @@ export const sendPerfTiming = ({
...
   _tracking?: typeof tracking;
   actualDuration: number;
   id: string;
-  phase: string;
+  url: string;
 }) => {
   _tracking.timing({
     category: 'renderPerf',
-    label: phase,
+    label: url,
```
willdurand, line 300: Couldn't we set `url` to `location.href` directly here so that it takes the full url from the browser? That would avoid a lot of code.

bobsilverberg: There is no such thing as `location.href`. Is that another thing we have in addons-frontend, but not in code-manager?

willdurand: ah well, `window.location.href` that is

src/utils.tsx
```diff
@@ -288,17 +288,21 @@ export const sendPerfTiming = ({
   _tracking = tracking,
   actualDuration,
   id,
-  phase,
+  url,
 }: {
   _tracking?: typeof tracking;
   actualDuration: number;
   id: string;
-  phase: string;
```
eviljeff, line 296: what was `phase` and why don't we need it any longer?

src/utils.tsx
bobsilverberg: Ah, I was just checking what we had available in the `location` object. Yes, we could use `window.location.href`. I'll change that.

src/utils.tsx
bobsilverberg: `phase` is passed into the `onRender` callback, and it represents the render phase, which is either `mount` or `unmount`. I was passing it to GA as an extra piece of data, but we really don't care about the difference between `mount` and `unmount` for these timings, and we only have room to send so much data to GA, so I have opted to send the `url` instead of the `phase`.

src/utils.tsx
```diff
@@ -288,16 +288,16 @@ export const sendPerfTiming = ({
...
+  url: string;
 }) => {
   _tracking.timing({
     category: 'renderPerf',
-    label: phase,
+    label: url,
```
willdurand, line 300: We could probably set the value of `label: window.location.href`.

bobsilverberg: Make that the default, or just always set it to `window.location.href`? I guess the latter is okay for now, as I see no reason we'd need to specify a different url for any of the current use cases.

bobsilverberg: Just thinking if that will make the test cases more complicated. I guess we can just test the value of `url` in the test for `sendPerfTiming` and not even bother checking it for the tests for `CodeView` and `DiffView`.

willdurand: We don't necessarily need to test the value of the `url`. I see this value as an implementation detail of the `sendPerfTiming` function.

bobsilverberg: Interesting. So even though there will be logic that sets the value of the `url` based on `window.location`, you don't think we should have a test anywhere that verifies that logic? Or are you just agreeing that we only need to test it in the test for `sendPerfTimimg`?

willdurand: > Or are you just agreeing that we only need to test it in the test for `sendPerfTimimg`?

That, indeed. sorry if that wasn't clear, we should have only one test case.

bobsilverberg:
> seems fine to me, but I'm not clear from the conversation if you're still planning on changing something 🤷

Thanks @eviljeff. I did make some more changes, but it has only resulted in an even smaller patch, so I don't think it needs another review, but feel free to take a peek if you like.

PR #1556, opened 2020-04-22
Change logic for truncation of files

### LINKED ISSUE #1552: Change logic for truncation of files
I am repurposing this issue to be about implementing the changes we discussed today via Zoom.  Here is what we decided:

1. *All* minified files will be truncated, regardless of their length.
2. Non-minified files will be truncated if they contain more than 5000 lines.
3. All truncation, for both file types, will be done based on number of characters. We will display only the first 300 characters of any files that have been truncated.

@willdurand @wagnerand Please confirm if I accurately captured the requirements we discussed.

bobsilverberg:
Fixes #1552

### DIFF
--- a/src/components/CodeView/index.tsx
+++ b/src/components/CodeView/index.tsx
@@ -16,7 +16,7 @@ import {
 } from './utils';
 import refractor from '../../refractor';
 import {
-  MINIFIED_FILE_TRIMMED_CHAR_COUNT,
+  TRIMMED_CHAR_COUNT,
   SLOW_LOADING_LINE_COUNT,
   codeShouldBeTrimmed,
   contentAddedByTrimmer,
@@ -64,6 +64,10 @@ export const scrollToSelectedLine = (element: HTMLElement | null) => {
   }
 };
 
+export const trimCode = (content: string) => {
+  return `${content.substring(0, TRIMMED_CHAR_COUNT)} ${contentAddedByTrimmer}`;
+};
+
 export type PublicProps = {
   content: string;
   isMinified: boolean;
@@ -72,7 +76,9 @@ export type PublicProps = {
 };
 
 export type DefaultProps = {
-  _minifiedFileTrimmedCharCount: number;
+  _codeShouldBeTrimmed: typeof codeShouldBeTrimmed;
+  _trimCode: typeof trimCode;
+  _trimmedCharCount: number;
   _scrollToSelectedLine: typeof scrollToSelectedLine;
   _sendPerfTiming: typeof sendPerfTiming;
   _slowLoadingLineCount: number;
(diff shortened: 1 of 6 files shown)


src/components/CodeView/index.tsx
```diff
@@ -64,6 +64,10 @@ export const scrollToSelectedLine = (element: HTMLElement | null) => {
   }
 };
 
+export const trimCode = (content: string) => {
+  return `${content.substring(0, TRIMMED_CHAR_COUNT)} ${contentAddedByTrimmer}`;
+};
```
eviljeff, line 69: if `codeShouldBeTrimmed` is in utils.py, should this function be too?

src/components/CodeView/index.tsx
```diff
@@ -114,26 +124,16 @@ export class CodeViewBase extends React.Component<Props> {
     let slowAlert;
 
     if (
-      codeShouldBeTrimmed({
+      _codeShouldBeTrimmed({
```
eviljeff, line 127: tbh I don't really understand why our(? React?) code is written like this - `codeShouldBeTrimmed` is a function we import... so why are we declaring it in `DefaultProps`, which then has to be set in `defaultProps` and again in `renderWithLinterInfo` - it's like 3 extra lines of code cruft for this function alone.

src/utils.tsx
```diff
@@ -12,9 +12,9 @@ import { getCodeLineAnchor } from './components/CodeView/utils';
 import tracking from './tracking';
 
 // This is how many lines of code it takes to slow down the UI.
-export const SLOW_LOADING_LINE_COUNT = 2000;
-// This is how many characters we will include in trimmed minfied file content.
-export const MINIFIED_FILE_TRIMMED_CHAR_COUNT = 2000;
+export const SLOW_LOADING_LINE_COUNT = 5000;
+// This is how many characters we will include in trimmed content.
+export const TRIMMED_CHAR_COUNT = 300;
```
eviljeff, line 17: I'm sure you discussed this in your meeting, but why the big discrepancy between the `SLOW_LOADING_LINE_COUNT` of 5000 and `TRIMMED_CHAR_COUNT` - you have a file that's 5000 lines long, with, say 100,000  characters that's displayed, but if there's an extra line return so it's 5001 lines then we only display 300 characters?

willdurand: Truncated content does not give any useful information to reviewers and it is actually dangerous because if we give too much to reviewers, they might not open the entire file. That's why we have this big discrepancy. By having a small number of characters, we can display the truncated view very fast.

5k lines is what we consider large files that are slow to render in code-manager by default.

src/components/CodeView/index.tsx
willdurand: That's dependency injection, mainly for testing purposes.

src/components/CodeView/index.tsx
bobsilverberg: We have different logic for trimming code in the browse view and the diff view, so each component (`CodeView` and `DiffView`) contains its own method for doing this. I think that makes it easier to understand than moving them both to `utils` (`.tsx`, btw, not `.py`, 😉 ) and giving them unique names there. It also means that this is only ever called once, from here, so moving it to `utils` doesn't really yield a benefit.

PR #1565, opened 2020-04-28
Adjust logic for the versions reducer to reflect API changes

### LINKED ISSUE #1559: Adjust logic for the versions reducer to reflect API changes
Depends on https://github.com/mozilla/addons-server/pull/14043

Changes are landing in https://github.com/mozilla/addons-server/pull/14043 which will move where some data can be found in the responses for the browse and compare APIs. Specifically, code manager currently looks for `sha256` and `mimetype` in the `entries` for a file, but after the above mentioned patch lands, those properties will be found with the `file` itself, not in the `entity` for the file.

As well, `size` will no longer exist for an `entity`, nor will `modified`. Code manager does not currently use either of those, but we should still update our types to reflect this new reality.

bobsilverberg:
Fixes #1559

### DIFF
--- a/src/pages/Browse/index.tsx
+++ b/src/pages/Browse/index.tsx
@@ -191,7 +191,7 @@ export class BrowseBase extends React.Component<Props> {
     if (!file || !version) {
       return <Loading message={gettext('Loading content...')} />;
     }
-    if (file.type === 'image' && file.downloadURL) {
+    if (file.fileType === 'image' && file.downloadURL) {
       return (
         <div className={styles.Image}>
           <img
--- a/src/reducers/versions.tsx
+++ b/src/reducers/versions.tsx
@@ -159,7 +159,7 @@ type InternalVersionFile = {
   mimeType: string;
   sha256: string;
   size: number;
-  type: VersionEntryType;
+  fileType: VersionEntryType;
 };
 
 export type VersionFile = {
@@ -175,7 +175,7 @@ export type VersionFile = {
   path: string;
   sha256: string;
   size: number;
-  type: VersionEntryType;
+  fileType: VersionEntryType;
   version: string;
 };
 
@@ -439,7 +439,7 @@ export const createInternalVersionFile = (
     mimeType: file.mimetype,
     sha256: file.sha256,
     size: file.size,
-    type: file.mime_category,
+    fileType: file.mime_category,
   };
 };
 
--- a/src/reducers/versions.tsx
+++ b/src/reducers/versions.tsx
@@ -41,11 +41,7 @@ export type ExternalVersionEntry = {
   depth: number;
   filename: string;
   mime_category: VersionEntryType;
-  mimetype: string;
-  modified: string;
   path: string;
-  sha256: string | null;
-  size: number | null;
   status?: VersionEntryStatus;
 };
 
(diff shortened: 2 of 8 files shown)


bobsilverberg:
This depends on, and also blocks https://github.com/mozilla/addons-server/pull/14043, so they both need to land for the same push. It's not critical to get them into this week's push, but if this can be reviewed in time, then we can land both. Otherwise I'll hold off merging https://github.com/mozilla/addons-server/pull/14043 until after the tag.

src/reducers/versions.tsx
```diff
@@ -153,10 +153,13 @@ type InternalVersionFile = {
   content: string;
   created: string;
   downloadURL: string | null;
+  filename: string;
   id: number;
   isMinified: boolean;
-  sha256: string | null;
+  mimeType: string;
+  sha256: string;
```
eviljeff, line 160: it can't be `null` any longer?

src/reducers/versions.tsx
```diff
@@ -425,21 +426,20 @@ export const getParentFolders = (path: string): string[] => {
...
     created: file.created,
     downloadURL: file.download_url,
+    filename: file.filename,
     id: file.id,
     isMinified: file.uses_unknown_minified_code,
-    sha256: file.entries[path] ? file.entries[path].sha256 : null,
+    mimeType: file.mimetype,
+    sha256: file.sha256,
     size: file.size,
+    type: file.mime_category,
```
eviljeff, line 442: mmm, is `type` the same thing as `mime_category`?  - at a glance it'd be confusing that they were the same thing (to me, anyway).  But I'm not familiar with the possible values.

src/reducers/versions.tsx
bobsilverberg: Correct. When `sha256` was in `entries` it could be `null`, but in `file` it should never be `null`.

src/reducers/versions.tsx
bobsilverberg: It's true that `type` isn't such a great name, but it's been there for quite some time. It's like the "type" of file, either "text", "binary" or "image". Maybe I should change it to `fileType`?

src/pages/Browse/index.tsx
```diff
@@ -191,7 +191,7 @@ export class BrowseBase extends React.Component<Props> {
     if (!file || !version) {
       return <Loading message={gettext('Loading content...')} />;
     }
-    if (file.type === 'image' && file.downloadURL) {
+    if (file.fileType === 'image' && file.downloadURL) {
```
willdurand, line 194: That seems redundant, no?

willdurand: I suggest we revert that.

bobsilverberg: I did it because I agreed with @eviljeff's observation that `type` was a bit ambiguous. Perhaps that's not even what he was saying, but we do have a few different `type` fields on different objects, and I like this change. I know it seems a bit redundant, but `type` for file could imply `mimeType`, or type based on extension, or maybe even something else that a reader might infer. To me `fileType` says "what is the type of this file?"

If you feel really strongly about it I can revert it, but I actually like it and don't think it's worth the time and effort to revert it.

PR #1568, opened 2020-04-30
Do not highlight code that has been trimmed in the DiffView

### LINKED ISSUE #1567: Make the first 300 chars visible in compare view for minified files
Followup for https://github.com/mozilla/addons-code-manager/issues/1552

For some reason there are some minified trimmed files which don't show the first 300 chars in compare view. They will only show /* truncated by code-manager */
 The issue is not reproducible for the Browse files.

Examples:
https://code.addons-dev.allizom.org/en-US/compare/618937/versions/1690424...1690425/?path=src%2Fmanifest.js

https://code.addons-dev.allizom.org/en-US/compare/618937/versions/1690424...1690425/?allowSlowPages=false&path=src%2Fcontent%2Fengines%2FauDesign.js

https://code.addons-dev.allizom.org/en-US/compare/618937/versions/1690424...1690425/?allowSlowPages=false&path=src%2Fcontent%2Fengines%2Fbaidu.js

https://code.addons-dev.allizom.org/en-US/compare/618937/versions/1690424...1690425/?allowSlowPages=false&path=src%2Fcontent%2Fengines%2Fbing.js

https://code.addons-dev.allizom.org/en-US/compare/618937/versions/1690424...1690425/?allowSlowPages=false&path=src%2Fcontent%2Fengines%2Fdepositphotos.js

https://code.addons-dev.allizom.org/en-US/compare/618937/versions/1690424...1690425/?allowSlowPages=false&path=src%2Fmanifest.js

bobsilverberg:
Fixes #1567

### DIFF
--- a/src/@types/react-diff-view/index.d.ts
+++ b/src/@types/react-diff-view/index.d.ts
@@ -105,6 +105,9 @@ declare module 'react-diff-view' {
         type: string;
         value: string;
       }
+    // When tokenize fails to properly parse some content, it returns an
+    // empty array as the Token, which is what the following is meant to
+    // illustrate.
     | string[];
 
   type Tokens = {
--- a/src/@types/react-diff-view/index.d.ts
+++ b/src/@types/react-diff-view/index.d.ts
@@ -100,10 +100,12 @@ declare module 'react-diff-view' {
     refractor: typeof import('refractor');
   };
 
-  type Token = {
-    type: string;
-    value: string;
-  };
+  type Token =
+    | {
+        type: string;
+        value: string;
+      }
+    | string[];
 
   type Tokens = {
     old: Token[];
--- a/src/components/DiffView/index.spec.tsx
+++ b/src/components/DiffView/index.spec.tsx
@@ -158,7 +158,7 @@ describe(__filename, () => {
 
   const createInternalHunkWithChanges = (
     changes: Partial<ExternalChange>[],
-  ): HunkInfo => {
+  ): ReactDiffView.HunkInfo => {
     return createInternalHunk(createHunkWithChanges(changes));
   };
 
@@ -1129,7 +1129,7 @@ describe(__filename, () => {
       side = 'new',
       version = createInternalVersion(fakeVersion),
       wrapInAnchor = jest.fn().mockReturnValue(defaultGutter),
-    }: Partial<RenderGutterParams> &
+    }: Partial<ReactDiffView.RenderGutterParams> &
       RenderParams & {
         defaultGutter?: React.ReactElement;
       } = {}) => {
@@ -1145,7 +1145,7 @@ describe(__filename, () => {
       });
       const { renderGutter } = root.find(ReactDiffView.Diff).props();
       if (renderGutter) {
-        const params: RenderGutterParams = {
+        const params: ReactDiffView.RenderGutterParams = {
           change,
           renderDefault,
           inHoverState: false,
--- a/src/components/DiffView/index.spec.tsx
+++ b/src/components/DiffView/index.spec.tsx
@@ -276,6 +276,22 @@ describe(__filename, () => {
     });
   });
 
+  it('does not tokenize the hunks to add syntax highlighting if the diff has been trimmed', () => {
+    const _codeCanBeHighlighted = jest.fn().mockReturnValue(true);
+    const _codeShouldBeTrimmed = jest.fn().mockReturnValue(true);
+    const _tokenize = jest.fn();
+
+    renderWithLinterProvider({
+      _codeCanBeHighlighted,
+      _codeShouldBeTrimmed,
+      _tokenize,
+      // A minified file will be trimmed by default.
+      isMinified: true,
+    });
+
+    expect(_tokenize).not.toHaveBeenCalled();
+  });
+
   it('configures anchors/links on each line number', () => {
     const root = renderWithLinterProvider();
 
--- a/src/components/DiffView/index.spec.tsx
+++ b/src/components/DiffView/index.spec.tsx
@@ -2,16 +2,7 @@
 import queryString from 'query-string';
 import React from 'react';
 import { Alert } from 'react-bootstrap';
-import {
-  Diff,
-  HunkInfo,
-  Hunks,
-  RenderGutterParams,
-  WidgetMap,
-  parseDiff,
-  getChangeKey,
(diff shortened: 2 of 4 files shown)


bobsilverberg:
@willdurand based on our conversation this morning, I added a check to catch when invalid tokens are generated from `tokenize`. Unfortunately, the format of the invalid tokens is a bit obscure, so I had to write some nasty looking code, but it still seems like a worthwhile safety hatch.

src/components/DiffView/index.tsx
```diff
@@ -426,12 +426,31 @@ export class DiffViewBase extends React.Component<Props> {
...
+    ) {
       // TODO: always highlight when we can use a Web Worker.
       // https://github.com/mozilla/addons-code-manager/issues/928
       tokens = _tokenize(hunks, options);
     }
 
+    // If tokenize fails to create valid tokens, we can end up with an object
+    // that looks like:
+    // { new: [[]], old: [[]] }
+    // which is what we're trying to catch here so that invlaid tokens do not
```
willdurand, line 442:
```suggestion
    // which is what we're trying to catch here so that invalid tokens do not
```

src/components/DiffView/index.spec.tsx
```diff
@@ -276,6 +277,39 @@ describe(__filename, () => {
...
+
+    expect(_tokenize).not.toHaveBeenCalled();
+  });
+
+  it('resets tokens to undefined if they seem invalid', () => {
+    const _codeCanBeHighlighted = jest.fn().mockReturnValue(true);
+    const _codeShouldBeTrimmed = jest.fn().mockReturnValue(false);
+    const _tokenize = jest
+      .fn()
+      .mockReturnValue(createFakeTokens({ invalid: true }));
```
willdurand, line 301: Could we use the `tokenize` function here, instead of a fake object? The code in the component is not very future-proof, could easily break and we don't have coverage around the returned object of `tokenize` when it has failed.

src/@types/react-diff-view/index.d.ts
```diff
@@ -100,10 +100,12 @@ declare module 'react-diff-view' {
...
-  type Token = {
-    type: string;
-    value: string;
-  };
+  type Token =
+    | {
+        type: string;
+        value: string;
+      }
+    | string[];
```
willdurand, line 108: Maybe we should have a comment to indicate that `string[]` is actually used for empty array?

src/components/DiffView/index.spec.tsx
```diff
@@ -276,12 +272,55 @@ describe(__filename, () => {
...
+  });
+
+  it('resets tokens to undefined if they seem invalid', () => {
+    const _codeCanBeHighlighted = jest.fn().mockReturnValue(true);
+    const _codeShouldBeTrimmed = jest.fn().mockReturnValue(false);
+    const _tokenize = (jest.spyOn(
+      ReactDiffView,
+      'tokenize',
+    ) as unknown) as typeof ReactDiffView.tokenize;
+    const badCode =
```
willdurand, line 298: Is it really bad code or partial code that will break the tokenize function? Maybe we should have a comment here too. This comment could also say that if this test case starts to fail, one should look into the return value of the `tokenize` function.

src/components/DiffView/index.spec.tsx
```diff
@@ -276,12 +272,55 @@ describe(__filename, () => {
...
+    expect(_tokenize).not.toHaveBeenCalled();
+  });
+
+  it('resets tokens to undefined if they seem invalid', () => {
+    const _codeCanBeHighlighted = jest.fn().mockReturnValue(true);
+    const _codeShouldBeTrimmed = jest.fn().mockReturnValue(false);
+    const _tokenize = (jest.spyOn(
+      ReactDiffView,
+      'tokenize',
+    ) as unknown) as typeof ReactDiffView.tokenize;
```
willdurand, line 297: We probably don't even need to spy on this function. What matters is the `tokens` prop.

src/components/DiffView/index.spec.tsx
```diff
@@ -2,16 +2,7 @@
...
-  Diff,
-  HunkInfo,
-  Hunks,
-  RenderGutterParams,
-  WidgetMap,
-  parseDiff,
-  getChangeKey,
-  tokenize,
-} from 'react-diff-view';
+import * as ReactDiffView from 'react-diff-view';
```
willdurand, line 5: I am fine with this change but I am wondering why it is needed.

src/@types/react-diff-view/index.d.ts
bobsilverberg: Good idea. I believe I can actually remove this change because there is nowhere in our code that uses the empty array. I needed it before when I had the `createFakeTokens` helper that created a set of invalid tokens, but now that we're using `tokenize` in that test we no longer need this. On the other hand, we do know that it is possible for this to be an empty array and are checking for that, so it probably makes sense to keep this change. Do you agree?

willdurand: I agree, having accurate types is always good.

src/components/DiffView/index.spec.tsx
bobsilverberg: Good idea about the variable name change, and I also added a detailed comment above the assertion.

src/components/DiffView/index.spec.tsx
bobsilverberg: I believe we do need this. I added it because it is possible that `tokens` can be `undefined` for other reasons. We try to ensure that the test setup covers that (by having `_codeCanBeHighlighted` return `true` and `_codeShouldBeTrimmed` return `false`), but there are just enough moving parts in here that I don't think we should rely on test setup to ensure that the case does what we expect it to. If any of that logic were to change, or new code introduced, it's possible that `tokenize` might not be called at all in this test case, and that would cause `tokens` to be `undefined` and hence a false positive. If we spy on `tokenize` then we can be sure that it is actually called, and therefore this test case is valid.

src/components/DiffView/index.spec.tsx
bobsilverberg: I needed to do this to get `jest.spyOn(ReactDiffView, 'tokenize')` to work. I checked docs and Googled quite a bit and couldn't find a way to use `spyOn` with just an imported function, as opposed to having to specify the imported module and function.

PR #1571, opened 2020-05-01
Turn off debug mode for Google Analytics

### LINKED ISSUE #1569: Turn off debug mode for Google Analytics
When we first implemented GA for Code Manager, debug mode was enabled for local development by default. This provides a lot of information that can be useful when working directly with GA, but the amount of info it adds is actually quite annoying when you don't really care about GA.

This issue is about changing the default to be `debug: false`, but allowing it to be changed locally via `.env.common-local`.

For QA:
There will not be any observable differences from this change on any of the hosted environments, but please check that the existing code that sends pings to Google Analytics remains unaffected.

bobsilverberg:
Fixes #1569 

I also found a bug in the way we were handling boolean values from `env` files in `tracking.tsx` which I fixed.

### DIFF
--- a/.env.common-local
+++ b/.env.common-local
@@ -26,3 +26,6 @@ PUBLIC_URL=/
 # Variables for Google Analytics
 REACT_APP_TRACKING_ENABLED=false
 REACT_APP_TRACKING_ID=UA-77033033-31
+
+# This is used for local development to toggle debug mode for Google Analytics
+REACT_APP_GA_DEBUG_MODE=false
--- a/src/tracking.tsx
+++ b/src/tracking.tsx
@@ -46,7 +46,7 @@ export class Tracking {
   constructor({
     _isDoNotTrackEnabled = isDoNotTrackEnabled,
     _reactGA = ReactGA,
-    trackingEnabled = Boolean(process.env.REACT_APP_TRACKING_ENABLED),
+    trackingEnabled = process.env.REACT_APP_TRACKING_ENABLED === 'true',
     trackingId = process.env.REACT_APP_TRACKING_ID,
     // We need to set useMock to true in the tests in order to use a provided
     // mock, because of the check below for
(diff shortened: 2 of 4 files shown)


.env.common-local
```diff
@@ -26,3 +26,6 @@ PUBLIC_URL=/
 # Variables for Google Analytics
 REACT_APP_TRACKING_ENABLED=false
 REACT_APP_TRACKING_ID=UA-77033033-31
+
+# This is used for local development to toggle debug mode for Google Analytics
+REACT_APP_GA_DEBUG_MODE=false
```
eviljeff, line 31: Does it need to be redefined in this file? (`.env.common-local` inherits from `.env`?  Or is there an intermediate file where it's `true`?)

src/tracking.tsx
```diff
@@ -46,7 +46,7 @@ export class Tracking {
   constructor({
     _isDoNotTrackEnabled = isDoNotTrackEnabled,
     _reactGA = ReactGA,
-    trackingEnabled = Boolean(process.env.REACT_APP_TRACKING_ENABLED),
+    trackingEnabled = process.env.REACT_APP_TRACKING_ENABLED === 'true',
```
eviljeff, line 49: curiosity: why does `REACT_APP_TRACKING_ENABLED` change to `'true'` from `true` along the way from the `.env` file to here?

.env.common-local
bobsilverberg: I added it here as a clue to a developer that this is where you'd want to change it if you want to affect your local environment. I suppose someone could just edit `.env` instead (I think), but this is where we encourage people to adjust their local environment variables.

src/tracking.tsx
bobsilverberg: I think it's just the nature of how `process.env` works. It always returns strings to us.

PR #1591, opened 2020-05-19
Only retrieve file info when that's all we need, in the Browse view

### LINKED ISSUE #1590: Only retrieve file info when that's all we need, in the Browse view
This is a spin-off from #1574 

While working on that issue I found a number of things that can/should be fixed in terms of how we're calling APIs and returning data for both the Browse and Compare views. Because this is a fairly large task, I'm breaking it down into smaller pieces.

This issue is simply about changing the logic for the Browse view so that when we only need to retrieve file information (because we already have version information), we use the new `file_only` param of the browse API to avoid returning unnecessary data.

For QA:

When looking at the API requests generated by code manager to `/api/v5/reviewers/addon/{addonId}/versions/{versionId}/`, the first API request for a version will *not* have the `file_only` param specified, and the response will be a complete `version` object, including a `file_entries` property and an embedded complete `file` object. For each subsequent request to the API for a different file (there will be a `file=` param on the URL), bur for the same version, the `file_only=true` param should be specified, and the response will contain just an `id` and a `file` object.

bobsilverberg:
Fixes #1590

### DIFF
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -7,6 +7,8 @@ import { ApiState } from '../reducers/api';
 import { ExternalComment } from '../reducers/comments';
 import { ExternalUser } from '../reducers/users';
 import {
+  ExternalVersionWithContentFileOnly,
+  ExternalVersionWithDiffFileOnly,
   ExternalVersionWithContent,
   ExternalVersionWithDiff,
   ExternalVersionsList,
@@ -104,6 +106,8 @@ export enum HttpMethod {
   PUT = 'PUT',
 }
 
+type CallApiQuery = { [key: string]: string };
+
 type CallApiParams<BodyDataType extends undefined | {}> = {
   _makeQueryString?: typeof makeQueryString;
   apiState: ApiState;
@@ -113,7 +117,7 @@ type CallApiParams<BodyDataType extends undefined | {}> = {
   includeCredentials?: boolean;
   lang?: string;
   method?: HttpMethod;
-  query?: { [key: string]: string };
+  query?: CallApiQuery;
   version?: string;
 };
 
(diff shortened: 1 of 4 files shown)


src/api/index.tsx
```diff
@@ -302,6 +306,17 @@ export const fetchAllPages = async <ResultsType,>(
   );
 };
 
+const makeApiQuery = (path?: string, fileOnly?: boolean) => {
+  const query: CallApiQuery = {};
+  if (path) {
```
eviljeff, line 311: when would `path` ever be missing/empty?

src/reducers/versions.tsx
```diff
@@ -1333,6 +1361,27 @@ export const createReducer = ({
...
         };
       }
+      case getType(actions.loadVersionFileFromVersion): {
+        const { path, version } = action.payload;
+
+        return {
+          ...state,
+          versionFiles: {
+            ...state.versionFiles,
+            [version.id]: {
```
eviljeff, line 1371: I'm reading this (perhaps through my python sunglasses) as wrapping `version.id` in a list and making the list an object key/property ... which seems a strange thing to do.

src/api/index.tsx
bobsilverberg: When you go from Reviewer Tools to Code Manager it takes you to the default file, which is generally `manifest.json`, but at that point no file is actually selected so no `path` is sent to the API.

src/reducers/versions.tsx
bobsilverberg: It makes sense that your python sunglasses would show you that. If you switch those for these JS sunglasses 🕶️ , then you'll see that what's being said here is that the name of the property in question is the value of `version.id`, so if `version.id == 1` then the property would be called `1`.

eviljeff: interesting.  I guess the obvious ` version.id: {` wouldn't work because Javascript's syntax is so overloaded.

bobsilverberg: Yeah, I'm pretty sure that's not legal.

PR #1602, opened 2020-05-27
Remove version info from draft_comments API

### LINKED ISSUE #1594: Update Code Manager to address changes in the draft_comments API
Depends on https://github.com/mozilla/addons-server/pull/14361

When https://github.com/mozilla/addons-server/pull/14361 lands, the `draft_comments` API will no longer return a full `version` object, but rather will return just a `version_id`. Code manager is not currently using returned the `version` object for anything, but code does exist to load it into the redux state, so that code is going to need to change when https://github.com/mozilla/addons-server/pull/14361 lands.

For QA:

No changes in behaviour should be observed. Please check the comments features for regressions.

bobsilverberg:
Fixes #1594

### DIFF
--- a/src/reducers/comments.spec.tsx
+++ b/src/reducers/comments.spec.tsx
@@ -509,7 +509,7 @@ describe(__filename, () => {
             url: userUrl,
             username: userUsername,
           },
-          version: { id: versionId },
+          version_id: versionId,
         }),
       ).toEqual({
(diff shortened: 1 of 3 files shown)


bobsilverberg:
> thought I approved this yesterday, r+

Maybe it's the server part of this that you're remembering. In any case, thanks!

PR #1628, opened 2020-06-24
Call validation.json endpoint as an API

### LINKED ISSUE #1627: Fetch validation results as an API call
When https://github.com/mozilla/addons-server/pull/14717 lands, which should fix https://github.com/mozilla/addons-server/issues/14655, which currently breaks code manager, we will need to change the logic we use to fetch validation results. 

Currently we simply fetch a URL, bit now we need to make an API call, which will require some logic changes in code-manager code.

bobsilverberg:
Fixes #1627

### DIFF
--- a/src/api/index.tsx
+++ b/src/api/index.tsx
@@ -5,6 +5,7 @@ import log from 'loglevel';
 
 import { ApiState } from '../reducers/api';
 import { ExternalComment } from '../reducers/comments';
+import { ExternalLinterResult } from '../reducers/linter';
 import { ExternalUser } from '../reducers/users';
 import {
   ExternalVersionWithContentFileOnly,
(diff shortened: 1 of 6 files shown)


bobsilverberg:
@willdurand This relates to a P1, so needs to be looked at ASAP. Sorry for this. I'm going to land the related change on server soon, so you should be able to test this against the dev server once that lands.

src/api/index.tsx
```diff
@@ -542,3 +543,15 @@ export const deleteComment = async ({
...
+
+type GetValidationParams = {
+  apiState: ApiState;
+  url: string;
+};
+
+export const getValidation = async ({ apiState, url }: GetValidationParams) => {
+  return callApi<ResponseOnly<ExternalLinterResult>>({
+    apiState,
+    endpointUrl: url,
```
willdurand, line 555: Can we construct the endpoint here instead? We don't need to pass the `url` anymore I think. We can probably pass the addon ID.

willdurand:
I haven't tried the patch but it looks good :) If you land the api patch, I can give it a try

src/api/index.tsx
bobsilverberg: We need the file ID as well, but yes, we should be able to do this. I'll file a follow-up.

bobsilverberg:
> I haven't tried the patch but it looks good :) If you land the api patch, I can give it a try

I've landed the API patch, but I'm not sure if it's been deployed yet.

PR #1634, opened 2020-06-25
Remove references to `oldRevision` and `newRevision` from the code

### LINKED ISSUE #1633: Remove references to `oldRevision` and `newRevision` from the code

Neither `oldRevision` nor `newRevision` are used anywhere in our code, so, for cleanup purposes, we should remove any references to them.

For QA:

There should be no observable changes in the app. Please check for regressions on the Compare page.

bobsilverberg:
Fixes #1633

### DIFF
--- a/src/components/DiffView/index.tsx
+++ b/src/components/DiffView/index.tsx
@@ -363,9 +363,10 @@ export class DiffViewBase extends React.Component<Props> {
       _trimHunkChars,
       diff,
       isMinified,
-      mimeType,
-      viewType,
       location,
+      mimeType,
+      version,
+      viewType,
     } = this.props;
 
     const options = {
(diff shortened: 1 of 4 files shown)


src/components/DiffView/index.tsx
```diff
@@ -480,7 +481,7 @@ export class DiffViewBase extends React.Component<Props> {
         {extraMessages.length ? this.renderExtraMessages(extraMessages) : null}
 
         {diff && hunks && (
-          <React.Fragment key={`${diff.oldRevision}-${diff.newRevision}`}>
+          <React.Fragment key={version.id}>
```
willdurand, line 484: Is this going to work when, on a compare page, we select a different base or head version? Does it update the page correctly?

bobsilverberg: Hmm, let me check that...

bobsilverberg: It seems to work fine on dev. Changing either pushes a new route, which reloads most of the page, so yes, it works. I don't think there's a use case for just updating the `DiffView` in place for a different set of versions. The `diff` changes, which means that the `FileTree` and the `CodeOverview` also change.

PR #1636, opened 2020-06-25
Introduce different types for external files with content vs diffs

### LINKED ISSUE #1635: Introduce different types for external files with content vs diffs

We're getting rid of `CompareInfo` and replacing it with version files which contain diffs, which are similar to the existing version files which contain content. This results in a lot of type churn in existing code that uses the existing version files, so it makes sense to extract these changes out into a separate patch.
For QA:

There should be no observable changes in the app. Please check for regressions on the Browse page.

bobsilverberg:
Fixes #1635 

Most of these changes are just replacing the `Version` type with `VersionWithContent` and the `VersionFile` type with `VersionFileWithConent`. The new `VersionWithContent` and `VersionFileWithConent` types are also defined, and a change was made to remove `versionString` from the `VersionFileWithConent` type because it should be able to be created from just a `file`, which doesn't have the `versionString`.

A future patch will introduce the `VersionWithDiff` and `VersionFileWithDiff` types, including functions to allow differentiating between them. This patch is mostly just groundwork that needs to be done and will reduce the size of future patches.

### DIFF
--- a/src/components/CodeOverview/index.spec.tsx
+++ b/src/components/CodeOverview/index.spec.tsx
@@ -15,7 +15,7 @@ import {
   createFakeExternalLinterResult,
   createFakeLocation,
   createFakeRef,
-  fakeVersion,
+  fakeVersionWithContent,
   getInstance,
   shallowUntilTarget,
   simulateLinterProvider,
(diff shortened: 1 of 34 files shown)


bobsilverberg:
> Probably don't want the `yarn.lock` changes in this commit though

Correct, thanks! I'll fix that.

PR #1639, opened 2020-06-26
Add types to support VersionDiffs

### LINKED ISSUE #1638: Add types to support VersionDiffs

We will be storing diffs in much the same way we store files currently, so in addition to `VersionFile`s we'll have `VersionDiff`s. This issue is about adding all of the types required for those, as well as the reducer functions required to create them.

For QA:

There should be no observable changes in the app. Please check for regressions on the Browse and Compare pages.

bobsilverberg:
Fixes #1638 
~~Depends on #1636~~

It probably makes sense to wait for #1636 to land before reviewing this, as it will make the diff much smaller.

### DIFF
--- a/src/components/DiffView/index.spec.tsx
+++ b/src/components/DiffView/index.spec.tsx
@@ -163,26 +163,9 @@ describe(__filename, () => {
   };
 
   const createDiffWithHunks = (hunks: ExternalHunk[]) => {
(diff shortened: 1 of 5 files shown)


bobsilverberg:
@diox I have rebased this on top of https://github.com/mozilla/addons-code-manager/pull/1636.

PR #1641, opened 2020-06-26
Fix logic for analyticsPath for local dev

### LINKED ISSUE #1640: Fix logic for analyticsPath for local dev
When I landed https://github.com/mozilla/addons-code-manager/pull/1571 to disable debug mode for GA by default, I missed one thing, which was causing the debug library to still be loaded for local development. This issue is about changing that.

For QA:

There's not need to verify this specific change, as it only relates to running the site locally, but please do check for regressions in the calls to GA.

bobsilverberg:
Fixes #1640

### DIFF
--- a/src/server/index.spec.tsx
+++ b/src/server/index.spec.tsx
@@ -701,6 +701,32 @@ describe(__filename, () => {
         expect(policy['script-src']).toEqual([
           "'self'",
           "'unsafe-inline'",
+          `${devEnv.REACT_APP_ANALYTICS_HOST}${ANALYTICS_PATH}`,
+        ]);
+      });
+
+      it('uses debug analytics path when REACT_APP_GA_DEBUG_MODE is true', async () => {
+        const content = '<h1>It works!</h1>';
+        fakeCreateReactAppServerApp.get('/*', (req, res) => res.send(content));
+
+        const server = request(
+          createServer({
+            env: {
+              ...devEnv,
+              REACT_APP_GA_DEBUG_MODE: 'true',
+            } as ServerEnvVars,
+            rootPath: fixturesPath,
+          }),
+        );
+
+        const response = await server.get('/');
+        expect(response.status).toEqual(200);
+        expect(response.header).toHaveProperty('content-security-policy');
+
+        const policy = cspParser(response.header['content-security-policy']);
+        expect(policy['script-src']).toEqual([
+          "'self'",
+          "'unsafe-inline'",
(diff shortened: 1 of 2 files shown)


src/server/index.spec.tsx
```diff
@@ -697,6 +697,32 @@ describe(__filename, () => {
...
+              ...devEnv,
+              REACT_APP_GA_DEBUG_MODE: 'true',
+            } as ServerEnvVars,
+            rootPath: fixturesPath,
+          }),
+        );
+
+        const response = await server.get('/');
+        expect(response.status).toEqual(200);
+        expect(response.header).toHaveProperty('content-security-policy');
```
eviljeff, line 724: I'm not sure what's being tested here exactly - it seems the same to the other surrounding tests.

src/server/index.spec.tsx
```diff
@@ -697,6 +697,32 @@ describe(__filename, () => {
...
+
+      it('uses debug analytics path when REACT_APP_GA_DEBUG_MODE is true', async () => {
+        const content = '<h1>It works!</h1>';
+        fakeCreateReactAppServerApp.get('/*', (req, res) => res.send(content));
+
+        const server = request(
+          createServer({
+            env: {
+              ...devEnv,
+              REACT_APP_GA_DEBUG_MODE: 'true',
```
eviljeff, line 716: Are there already existing tests that check the opposite case, when it's 'false?

src/server/index.spec.tsx
bobsilverberg: It's really just the assertion below about a specific `content-security-policy` that we care about, but these guard assertions will make a test failure easier to debug if one happens, so I can see value in them. Do you feel strongly that they should be removed?

src/server/index.spec.tsx
bobsilverberg: No, there's isn't one. I'll add one. Thanks!

bobsilverberg: [`3e80765` (#1641)](https://github.com/mozilla/addons-code-manager/pull/1641/commits/3e80765850bf2157c0f46df76a9404d1cb86ffeb)

src/server/index.spec.tsx
eviljeff: guard assumptions are fine, but I couldn't identify what was uniquely being tested in this test versus the other tests that seemed to share the same code.

PR #1644, opened 2020-06-30
Remove code for base_file

### LINKED ISSUE #1643: Remove code for base_file
We used to need the `base_file` from the API response to construct the legacy file viewer links, but we do not need that info anymore, so it can be removed. This is part of the big refactor I did for https://github.com/mozilla/addons-code-manager/issues/1599, so I figured I'd split it out into a separate issue to make the patch for that a bit smaller.

For QA:

Check for regressions on the Compare page and verify that legacy file viewer links still work.

bobsilverberg:
Fixes #1643

### DIFF
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -698,7 +698,6 @@ describe(__filename, () => {
 
     it('loads compare info', () => {
       const addonId = nextUniqueId();
-      const baseFileId = nextUniqueId();
       const baseVersionId = nextUniqueId();
       const path = 'manifest.json';
       const mimeType = 'mime/type';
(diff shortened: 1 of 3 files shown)


src/reducers/versions.spec.tsx
```diff
@@ -1589,12 +1584,9 @@ describe(__filename, () => {
...
     ): ExternalVersionWithDiff => {
       return {
-        ...fakeVersionWithContent,
+        ...fakeVersionWithDiff,
         file: {
-          ...fakeVersionWithContent.file,
-          base_file: {
-            id: nextUniqueId(),
-          },
+          ...fakeVersionWithDiff.file,
```
eviljeff, line 1589: these two changes from `fakeVersionWithContent` to `fakeVersionWithDiff` - are they related to removing `base_file` too?  I'm not overly familiar with this code but it seems unrelated.

bobsilverberg: No, that was a drive-by fix that I noticed should be done while I was updating this test code.

PR #1646, opened 2020-07-02
Add reducer support for retrieving and storing diff files

### LINKED ISSUE #1645: Add reducer support for retrieving and storing diff files

This will add support to the `versions` reducer to retrieve and store individual diff files. This is likely the last part that can be split out of the work needed for #1599. The next patch will make use of all of the bits that have been landing for this.

For QA:

There should be no observable changes in the app. Please check for regressions on the Browse and Compare pages.

bobsilverberg:
Fixes #1645

### DIFF
--- a/src/reducers/versions.spec.tsx
+++ b/src/reducers/versions.spec.tsx
@@ -1148,7 +1148,6 @@ describe(__filename, () => {
     const headVersionId = baseVersionId + 1;
     const path = 'some-file.js';
 
-    /* eslint-disable @typescript-eslint/camelcase */
     it('returns a version diff', () => {
       const downloadURL = 'http://example.org/download/file';
       const isMinified = true;
@@ -1205,7 +1204,6 @@ describe(__filename, () => {
         fileType: type,
       });
     });
-    /* eslint-enable @typescript-eslint/camelcase */
 
     it('returns `null` when the file was not retrieved from the API', () => {
       const versions = reducer(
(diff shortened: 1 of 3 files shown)


src/reducers/versions.tsx
```diff
@@ -974,6 +1087,71 @@ export const fetchVersionFile = ({
...
+type FetchDiffFileParams = {
+  _getDiffFileOnly?: typeof getDiffFileOnly;
+  addonId: number;
+  baseVersionId: number;
+  headVersionId: number;
+  path: string;
+};
+
+export const fetchDiffFile = ({
+  _getDiffFileOnly = getDiffFileOnly,
```
bobsilverberg, line 1099: This line is missing coverage because the only tests we have right now are for the reducer itself, and we always pass a mock in for this. I believe when the next patch comes out, that integrates all the pieces together, we should have a test for the `Compare` component that provides coverage for this line.

src/reducers/versions.spec.tsx
```diff
@@ -2788,6 +3520,80 @@ describe(__filename, () => {
...
+    const headVersionId = nextUniqueId();
+
+    it('computes a key given an addonId, baseVersionId and headVersionId', () => {
+      expect(getDiffKey({ addonId, baseVersionId, headVersionId })).toEqual(
+        `${addonId}/${baseVersionId}/${headVersionId}`,
+      );
+    });
+  });
+
+  describe('isDiffFileLoading', () => {
```
willdurand, line 3535: I dont see coverage for a `path` that is falsey.

src/reducers/versions.spec.tsx
```diff
@@ -975,6 +1142,127 @@ describe(__filename, () => {
     });
   });
 
+  describe('getVersionDiff', () => {
+    const addonId = nextUniqueId();
+    const baseVersionId = nextUniqueId();
+    const headVersionId = baseVersionId + 1;
+    const path = 'some-file.js';
+
+    /* eslint-disable @typescript-eslint/camelcase */
```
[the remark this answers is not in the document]
bobsilverberg: You're right, we did! I'll open a separate issue to remove all of these comments, but I'll remove any that I added in this PR.

PR #1647, opened 2020-07-02
Only retrieve file info when that's all we need, in the Compare view

### LINKED ISSUE #1327: Diff is not loading for removed files
STR:
1. Load https://code.addons-dev.allizom.org/en-US/compare/7560/versions/1252898...1510628/ for
 old version `3.3.6` and new version is `5.1.6`
2. Observe the page 

Expected result:
The Diff loads without errors.

Actual result:
An error is displayed. 

69337214-0698d480-0c69-11ea-9efa-7339a8c81c95

It's reproducible on dev with FF70(Win10).

### LINKED ISSUE #1574: Retrieve only file information when that's all we need
~~Depends on https://github.com/mozilla/addons-code-manager/issues/1587~~

Every time code manager retrieves information for a specific file, all of the version information, including all of the entries for the version, is returned. This represents a lot of data being returned to the client that is unneccesary, and likely contributes to some of the API slowness we are seeing.

With the landing of https://github.com/mozilla/addons-server/pull/14163, the Browse and Compare APIs now support a `file_only` param. We should update code manager to use those params when we already have version information.

### LINKED ISSUE #1599: Only retrieve file info when that's all we need, in the Compare view
Similar to https://github.com/mozilla/addons-code-manager/issues/1590, this issue is about adjusting the code for the Compare view to use the new `file_only` param to make sure that we only request file information if we already have version information for the current version.

This ended up requiring a huge refactoring (with 1,344 additions and 1,390 deletions) so I'm going to try to break that up into smaller patches, where possible, and I'll create issues for them referencing this issue.

bobsilverberg:
Fixes #1599 
Fixes #1574 
Fixes #1327 

~~Depends on #1646~~

I realize that this is still a huge patch, but I have already spent a lot of time breaking a larger patch into smaller pieces, and I'm not sure I can reduce this one much more, as making the change impacts a lot of other areas. Probably the majority of the changes in this patch are to tests, but I know that doesn't make it any easier to review.

This patch finishes implementing everything we need to do the following:
1. Once a version is loaded, we will only retrieve the diff info for each subsequent file, rather than retrieving the entire version each time.
2. As a side effect of this change, we no longer need to also retrieve the contents of the file for the current version - we can get away with only retrieving the diff.

The site should continue to function exactly as it did before. There will just be fewer requests generated, and less data returned to the client.

### DIFF
--- a/src/components/KeyboardShortcuts/index.spec.tsx
+++ b/src/components/KeyboardShortcuts/index.spec.tsx
@@ -12,19 +12,23 @@ import { getMessageMap } from '../../reducers/linter';
 import {
   actions as versionsActions,
   createInternalVersion,
+  createInternalVersionFile,
+  VersionFileWithContent,
+  VersionFileWithDiff,
 } from '../../reducers/versions';
 import { actions as fullscreenGridActions } from '../../reducers/fullscreenGrid';
 import {
   CreateKeydownEventParams,
   createContextWithFakeRouter,
   createKeydownEvent,
-  createFakeCompareInfo,
   createFakeExternalLinterResult,
   createFakeLocation,
   createFakeThunk,
   createStoreWithVersion,
   fakeAction,
   fakeExternalLinterMessage,
+  fakeVersionFileWithContent,
+  fakeVersionFileWithDiff,
   fakeVersionWithContent,
   fakeVersionEntry,
   fakeVersionWithDiff,
@@ -92,8 +96,8 @@ describe(__filename, () => {
     _goToRelativeFile = jest.fn().mockReturnValue(fakeAction),
     _goToRelativeMessage = jest.fn().mockReturnValue(fakeAction),
     comparedToVersionId = 1,
-    compareInfo = null,
     currentPath = 'file1.js',
+    file = null,
     location = createFakeLocation(),
     messageMap = getMessageMap(
       createFakeExternalLinterResult({ messages: [fakeExternalLinterMessage] }),
@@ -107,9 +111,9 @@ describe(__filename, () => {
       _goToRelativeDiff,
       _goToRelativeFile,
       _goToRelativeMessage,
-      compareInfo,
       comparedToVersionId,
       currentPath,
+      file,
       location,
       messageMap,
       versionId,
@@ -167,7 +171,11 @@ describe(__filename, () => {
   });
 
   it('renders one line descriptions for keys "p"/"n" and "k"/"j" in Browse', () => {
-    const root = render({ compareInfo: null });
+    const root = render({
+      file: createInternalVersionFile(
+        fakeVersionFileWithContent,
+      ) as VersionFileWithContent,
+    });
 
     expect(root.find(`dt.${styles.hasAlias}`)).toHaveLength(2);
     expect(root.find(`dd.${styles.hasAlias}`)).toHaveLength(2);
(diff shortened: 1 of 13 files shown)


bobsilverberg:
Hmm, I rebased this and now all the tasks, including `tsc` and linter are failing, which is odd. I'll take a look.

bobsilverberg:
It was a simple fix. This is fine now and ready to be reviewed.

src/components/KeyboardShortcuts/index.spec.tsx
```diff
@@ -167,7 +171,11 @@ describe(__filename, () => {
   });
 
   it('renders one line descriptions for keys "p"/"n" and "k"/"j" in Browse', () => {
-    const root = render({ compareInfo: null });
+    const root = render({
+      file: createInternalVersionFile(
+        fakeVersionFileWithContent,
+      ) as VersionFileWithContent,
```
willdurand, line 177: So, does that mean TS isn't able to infer the type of the returned value here?

src/components/VersionFileViewer/index.spec.tsx
```diff
@@ -41,61 +42,48 @@ import VersionFileViewer, { ItemTitles, PublicProps } from '.';
 
 describe(__filename, () => {
   const getInternalVersionAndFile = ({
-    store = configureStore(),
-    path = 'background.js',
+    contentOrDiff = 'content',
```
willdurand, line 45: Would it make sense to use constants (or an enum) here? This seems to only be used in the test file, though.

src/reducers/versions.tsx
```diff
@@ -326,14 +335,17 @@ export const actions = {
...
-  abortFetchDiff: createAction('ABORT_FETCH_DIFF', (resolve) => {
-    return (payload: {
-      addonId: number;
-      baseVersionId: number;
-      headVersionId: number;
-      path?: string;
-    }) => resolve(payload);
-  }),
+  abortFetchVersionWithDiff: createAction(
+    'ABORT_FETCH__VERSION_WITH_DIFF',
```
willdurand, line 339: Why does this one have two underscores?

src/components/KeyboardShortcuts/index.spec.tsx
bobsilverberg: I'm pretty sure that is the case, or I wouldn't have done it, but I'll double check.

src/components/VersionFileViewer/index.spec.tsx
bobsilverberg: Yeah, this is just a helper function for the tests, so I'm not sure it's worth doing that. WDYT?

src/reducers/versions.tsx
bobsilverberg: Because it's super special! 😉

bobsilverberg: I'll fix it.

src/components/VersionFileViewer/index.spec.tsx
willdurand: it's fine, I guess

src/components/KeyboardShortcuts/index.spec.tsx
bobsilverberg: It turns out it's not actually needed here, but that's not because TS is able to infer the type, it's just because there is no code below this that relies on it being a particular type. There are plenty of other places where we do need to assign an explicit type to the result of `createInternalVersionFile` even though one might expect TS to be able to infer it.

I have removed these extra `as`'s where we don't actually need them.

src/components/VersionFileViewer/index.spec.tsx
bobsilverberg: You sounded so sad about this, I decided to change it anyway. It wasn't much effort.

PR #1651, opened 2020-07-07
Call validation endpoint the same way we call other APIs

### LINKED ISSUE #1631: Call validation endpoint the same way we call other APIs
This is a follow up from https://github.com/mozilla/addons-code-manager/pull/1628.

As per the suggestion at https://github.com/mozilla/addons-code-manager/pull/1628#pullrequestreview-436800937, in code-manager we should construct the API endpoint similar to how we call other APIs, rather than using a URL provided in another API response. Once that lands we could remove the `validation_url_json` from the response of the Browse and Compare APIs on addons-server, but maybe we want to keep it there anyway, for other potential consumers?

For QA:

Nothing should change in code manager. Please check for regressions on the Browse and Compare pages.

bobsilverberg:
Fixes #1631 

Note that in order to make the changes desired for #1631, the following additional changes were made, which make up the bulk of this patch:
- Change `LinterProvider` so that it accepts a `version` property, rather than `versionId` and `validationURL`.
- Change the `fetchLinterMessagesIfNeeded` reducer function so that it accepts a `version` argument, rather than `versionId` and `url`.
- Store a `fileId` as part of the internal `Version` type.

Note that this will need to be rebased after #1646 and #1647 land, and that will be much easier to do that the opposite (land this first and rebase those), so it probably makes sense to hold off on this one until those land.

### DIFF
--- a/src/components/DiffView/index.spec.tsx
+++ b/src/components/DiffView/index.spec.tsx
@@ -597,8 +597,7 @@ describe(__filename, () => {
     const root = render({ version });
 
     const provider = root.find(LinterProvider);
-    expect(provider).toHaveProp('versionId', version.id);
-    expect(provider).toHaveProp('validationURL', version.validationURL);
+    expect(provider).toHaveProp('version', version);
     expect(provider).toHaveProp('selectedPath', version.selectedPath);
   });
 
--- a/src/api/index.spec.tsx
+++ b/src/api/index.spec.tsx
@@ -1145,11 +1145,21 @@ describe(__filename, () => {
 
   describe('getValidation', () => {
     it('calls the API to retrieve validation information', async () => {
-      const url = 'some-url';
+      const addonId = nextUniqueId();
+      const fileId = nextUniqueId();
 
-      await getValidation({ apiState: defaultApiState, url });
+      await getValidation({
+        apiState: defaultApiState,
+        addonId,
+        fileId,
+      });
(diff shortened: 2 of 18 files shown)


src/components/DiffView/index.spec.tsx
```diff
@@ -597,8 +597,7 @@ describe(__filename, () => {
     const root = render({ version });
 
     const provider = root.find(LinterProvider);
-    expect(provider).toHaveProp('versionId', version.id);
-    expect(provider).toHaveProp('validationURL', version.validationURL);
```
willdurand, line 601: Do we have an issue in addons-server to remove this property from the API response?

bobsilverberg:
Wow, I expected this to need a rebase after #1647 landed. I'm shocked!  I'll just double check that this is okay.

src/components/DiffView/index.spec.tsx
bobsilverberg: We don't. I wasn't sure if we should do that, but I suppose we're the only consumers right now, so it shouldn't affect anyone. @diox Do you think we're fine to remove this property from the API response as Code Manager no longer needs it?

bobsilverberg:
Nope, everything is fine. I guess it's because the code here only deals with `Version`s and not `VersionFile`s.  Thanks for the review, @willdurand!

PR #1658, opened 2020-07-15
Make reviewer tools link work for deleted add-ons

### LINKED ISSUE #1656: Link to reviewer tools is .../null for a deleted add-on

If the add-on is deleted, the link to the reviewer tools in the top left corner is: https://reviewers.addons.mozilla.org/reviewers/review/null

### What happened?

Link is https://reviewers.addons.mozilla.org/reviewers/review/null

Link should be `https://reviewers.addons.mozilla.org/reviewers/review/<addon-id>`
(Note that we can't use slug here since deleted add-ons don't have a slug (that's why we link to .../null)

wagnerand:
Fix #1656.

### DIFF
--- a/src/reducers/versions.tsx
+++ b/src/reducers/versions.tsx
@@ -68,7 +68,7 @@ export type ExternalVersionAddon = {
   icon_url: string;
   id: number;
   name: LocalizedStringMap;
-  slug: string;
+  slug: string | null;
 };
 
 export type PartialExternalVersion = {
@@ -195,7 +195,7 @@ type VersionAddon = {
   iconUrl: string;
   id: number;
   name: LocalizedStringMap;
-  slug: string;
+  slug: string | null;
 };
 
 export type Version = {
--- a/src/components/Navbar/index.spec.tsx
+++ b/src/components/Navbar/index.spec.tsx
@@ -127,6 +127,24 @@ describe(__filename, () => {
       );
     });
 
+    it('renders a link to reviewer tools for a deleted add-on', () => {
+      const reviewersHost = 'https://example.com';
+      const id = 456;
+      const store = createStoreWithVersion({
+        version: {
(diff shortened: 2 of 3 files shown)


src/reducers/versions.tsx
```diff
@@ -195,7 +195,7 @@ type VersionAddon = {
   iconUrl: string;
   id: number;
   name: LocalizedStringMap;
-  slug: string;
+  slug: string | null;
```
willdurand, line 198: It would be nice to update the addons-server docs too: https://addons-server.readthedocs.io/en/latest/topics/api/reviewers.html#get--api-v4-reviewers-addon-(int-addon_id)-versions-(int-version_id)-

wagnerand: https://github.com/mozilla/addons-server/pull/14963

bobsilverberg: I had a further thought about this. If we didn't want to allow `null` in `VersionAddon`, we could change `createInternalVersion` to that it stores an empty string in `slug` if it finds a `null` in the response. Doing that would eliminate the need for the ternary added to `src/components/Navbar/index.tsx` above. The downside is that we would no longer be able to tell the difference between a slug with a value of `''` and a slug with a value of `null`, so it might be better to just leave it like this.

@willdurand WDYT?

willdurand: I'd prefer to have the `slug` declared as nullable like it's done here. @eviljeff confirmed that we have `null` slugs for deleted add-ons, which is something we definitely need to support in a tool designed for reviewers.

As you said, using an empty string would be confusing because this isn't allowed.

bobsilverberg: Thanks @willdurand. In case it isn't entirely clear, @wagnerand, this means just leave it as it is now.

PR #1712, opened 2020-09-24
Upgrade typescript-eslint

willdurand:


src/api/index.tsx
```diff
@@ -155,8 +154,8 @@ type EmptyRequestAndResponse = {
 
 export const callApi = async <
   T extends {
-    requestData: undefined | {};
-    successfulResponse: {};
+    requestData: undefined | Record<string, unknown>;
+    successfulResponse: unknown;
```
bobsilverberg, line 158: I know you don't really want to spend time fighting with TS (as discussed in our last engineering meeting), but I'm just wondering, how is `unknown` any better than `any`? This isn't code interacting with a third party, it's our own code, so, theoretically we should be able to have a proper type for this.

src/components/App/index.spec.tsx
```diff
@@ -237,6 +237,7 @@ describe(__filename, () => {
     const root = render({ store });
 
     // User clicks the "close" button of the first error shown.
+    // eslint-disable-next-line @typescript-eslint/ban-types
```
bobsilverberg, line 240: Can we not fix this with a specific function type?

src/components/CommentSummaryButton/index.tsx
```diff
@@ -12,9 +12,9 @@ import CommentSummary from '../CommentSummary';
 import PopoverButton from '../PopoverButton';
 import styles from './styles.module.scss';
 
-export type PublicProps = {};
+export type PublicProps = Record<string, unknown>;
 
-export type DefaultProps = {};
+export type DefaultProps = Record<string, unknown>;
```
bobsilverberg, line 17: Why do we even need `PublicProps` and `DefaultProps`? It seems odd that we have them and set them to `{}	`.

src/components/FileTree/index.spec.tsx
```diff
@@ -202,6 +201,7 @@ describe(__filename, () => {
       const treeFold = root.find(Treefold);
       expect(treeFold).toHaveProp('onToggleExpand');
 
+      // eslint-disable-next-line @typescript-eslint/ban-types
       const onToggleExpand = treeFold.prop('onToggleExpand') as Function;
```
bobsilverberg, line 205: As discussed above, can we not have a specific function type?

src/components/FileTree/index.spec.tsx
```diff
@@ -241,6 +241,7 @@ describe(__filename, () => {
       const treeFold = root.find(Treefold);
       expect(treeFold).toHaveProp('isNodeExpanded');
 
+      // eslint-disable-next-line @typescript-eslint/ban-types
       const isNodeExpanded = treeFold.prop('isNodeExpanded') as Function;
```
bobsilverberg, line 245: As discussed above, can we not have a specific function type?

src/components/FileTree/index.spec.tsx
```diff
@@ -261,6 +262,7 @@ describe(__filename, () => {
...
```
bobsilverberg, line 266: As discussed above, can we not have a specific function type?

src/components/LoginButton/index.tsx
```diff
@@ -9,7 +9,7 @@ import { makeApiURL } from '../../api';
 import { gettext } from '../../utils';
 import styles from './styles.module.scss';
 
-type PublicProps = {};
+type PublicProps = Record<string, unknown>;
```
bobsilverberg, line 12: As mentioned above, I'm not sure I understand why we even need `PublicProps` here at all.

src/pages/Browse/index.tsx
```diff
@@ -34,7 +34,7 @@ import Loading from '../../components/Loading';
 import CodeView from '../../components/CodeView';
 import styles from './styles.module.scss';
 
-export type PublicProps = {};
+export type PublicProps = Record<string, unknown>;
```
bobsilverberg, line 37: Is this another case where we don't need `PublicProps` at all?

src/pages/Index/index.tsx
```diff
@@ -11,7 +11,7 @@ import styles from './styles.module.scss';
 import { ConnectedReduxProps } from '../../configureStore';
 import { actions as versionsActions } from '../../reducers/versions';
 
-export type PublicProps = {};
+export type PublicProps = Record<string, unknown>;
```
bobsilverberg, line 14: As above, do we need `PublicProps`?

src/pages/NotFound/index.tsx
```diff
@@ -4,7 +4,7 @@ import { Col } from 'react-bootstrap';
 
 import { gettext } from '../../utils';
 
-type PublicProps = {};
+type PublicProps = Record<string, unknown>;
```
bobsilverberg, line 7: Same as above.

bobsilverberg (commented):
Thanks for taking this on @willdurand. I know you probably don't want to spend a lot of time on this, being of the opinion that these types are not particularly helping us, and take time to deal with, but I did have a few observations about some of these changes. Let me know what you think.

.eslintrc
```diff
@@ -38,7 +38,15 @@
     // The beauty of TS is that it infers types quite well, so let's not write
     // too much code.
     "@typescript-eslint/explicit-function-return-type": "off",
-    "@typescript-eslint/no-explicit-any": "error"
+    "@typescript-eslint/no-explicit-any": "error",
+    // See: https://github.com/typescript-eslint/typescript-eslint/issues/2502
+    "no-use-before-define": "off",
```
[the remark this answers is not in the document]
willdurand: See this comment: https://github.com/typescript-eslint/typescript-eslint/issues/2502#issuecomment-689595020. We cannot use this rule with TS.

src/api/index.tsx
willdurand: I don't think `unknown` is better than `any` but the compiler suggested `unknown` to me so 🤷

src/components/CommentSummaryButton/index.tsx
willdurand: That's what we've done in the past. It means "there is no default props" or "no public props". I don't think we necessarily need the `DefaultProps` all the time but having `PublicProps` and `InternalProps` makes sense to me–one can quickly scan a component file and see the different props.

willdurand: This is valid in TS, it's just the linter that does not like `{}`

willdurand: I removed this `DefaultProps` type.

src/pages/Browse/index.tsx
willdurand: I don't think we should change that right now. That was working before and I don't think we complained about it.

src/components/FileTree/index.spec.tsx
willdurand: Yep, we could and I added them.

src/pages/Browse/index.tsx
bobsilverberg: Fair enough. I am a bit concerned about the fact that I don't think that `Record<string, unknown>` is actually equivalent to `{}`. When we used `{}` what we were really saying is "There are no public props", which I don't think is the same thing as `Record<string, unknown>`. I think we're just doing a replacement to please the linter, but we're replacing it with something inaccurate.

I wonder if there's a different, valid, notation we can use in TS to say, "this is an empty type with no properties"?

willdurand: Yeah, that's part of the reasons why I don't like TS anymore.. I tried to use `{} as Record<...>` but because it's a type, we cannot do that.

bobsilverberg: Ok, so I guess our strategy for now is to just keep TS happy with a minimum of effort and possibly migrate away from it some day?

willdurand: I don't really have a better alternative for now. Given that it compiles and the linter is happy, I guess we're good?

bobsilverberg: One more thing about this `PublicProps` thing. I suggested we remove it (here and elsewhere) as it's not actually used anywhere in the file. I.e., TS doesn't enforce it. So all it's really doing is serving as some documentation to show that there are no public props. Documentation that could easily be missed or ignored by a developer, and not necessarily any more useful than recognizing that no `PublicProps` type means that no public props are allowed.

bobsilverberg: > I don't really have a better alternative for now. Given that it compiles and the linter is happy, I guess we're good?

That's what my second-to-last comment was about. I guess we can just accept it for now. I agree with not spending too much time on nitty things for TS.

bobsilverberg (approved):

PR #1779, opened 2020-11-04
Permanently show current file info panel

### LINKED ISSUE #1765: Consider permanently showing current file info panel
The "Information" info panel for the current file is collapsed by default and for many monitors there is a lot of blank space in the overall left panel because lots of extensions don't have too many files or not all folders are expanded. We could use some of that space to show the "Information" panel permanently, like so:

code-manager

We can also drop the "Version" section in that panel as the version is already shown in the top bar.

This is especially handy when coming across a file that has been truncated, but you still want to take a look at it. Viewing the full file in code-manager can make the entire browser hang (and you won't know in advance whether it would), so I have been using the "Download" link regularly to open the file in a local code editor.

@kewisch Do you have an opinion on this?

bobsilverberg:
Fixes #1765

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -39,6 +39,7 @@
     // too much code.
     "@typescript-eslint/explicit-function-return-type": "off",
     "@typescript-eslint/no-explicit-any": "error",
+    "@typescript-eslint/explicit-module-boundary-types": "off",
     // See: https://github.com/typescript-eslint/typescript-eslint/issues/2502
     "no-use-before-define": "off",
     "@typescript-eslint/ban-ts-comment": ["error", {
--- a/src/components/AccordionMenu/index.tsx
+++ b/src/components/AccordionMenu/index.tsx
@@ -3,7 +3,6 @@ import { connect } from 'react-redux';
 import makeClassName from 'classnames';
 
 import Button from '../Button';
-import { infoPanelTitle } from '../VersionFileViewer';
 import { ApplicationState } from '../../reducers';
 import { ConnectedReduxProps } from '../../configureStore';
 import { actions, isExpanded } from '../../reducers/accordionMenu';
@@ -24,6 +23,7 @@ export type PublicItemProps = {
 };
 
 export type DefaultItemProps = {
+  alwaysExpanded: boolean;
   expandedByDefault: boolean;
 };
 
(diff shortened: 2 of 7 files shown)


.eslintrc
```diff
@@ -39,6 +39,7 @@
     // too much code.
     "@typescript-eslint/explicit-function-return-type": "off",
     "@typescript-eslint/no-explicit-any": "error",
+    "@typescript-eslint/explicit-module-boundary-types": "off",
```
willdurand, line 42: what does this rule do?

bobsilverberg: It requires [explicit types for exported functions](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md), which generates a ton of warnings. As mentioned in the comment above these rules, we rely on Typescript's ability to infer types, and if we allow this rule to be in place it would require us to add a lot of code to our files.  When also considering the fact that we are questioning the value of TS at all, I think that any rule that requires us to add a bunch of type code to our files should be avoided.

src/components/AccordionMenu/index.tsx
```diff
@@ -56,6 +57,7 @@ export class AccordionItemBase extends React.Component<ItemProps> {
   render() {
     const { children, expanded, title } = this.props;
     const contentId = makeItemContentId(title);
+    const expandPanel = expanded || title === infoPanelTitle;
```
eviljeff, line 60: this feels a bit like magic functionality - that if the title of a panel happens to be the same string as the information panel it's expanded.  Can't we explicitly pass down an arg from where the `AccordionItem` is used in `renderFileInfo` to specify that this panel should be expanded instead?

eviljeff (approved):
seems reasonable otherwise

src/components/AccordionMenu/index.tsx
bobsilverberg: Good idea!

bobsilverberg: [`6e875ac` (#1779)](https://github.com/mozilla/addons-code-manager/pull/1779/commits/6e875ac8df443c73a0054b6c00f028bec327ff52)

PR #1804, opened 2020-12-16
Fully migrate to Circle CI

willdurand:


willdurand:
I can reproduce the failure on Node 14 locally so I guess this patch works as intended :D

src/server/index.tsx
```diff
@@ -294,28 +291,56 @@ export const createServer = ({
...
-            // eslint-disable-next-line no-param-reassign
-            proxyResponse.headers['cache-control'] = 'private, no-store';
-
-            modifyResponse(
-              res,
-              proxyResponse.headers['content-encoding'],
-              (body: string) => _injectAuthenticationToken(req, body, env),
-            );
-          }
+          const bodyChunks: Buffer[] = [];
```
willdurand, line 309: I rewrote that part without `modifyResponse` which was not working anymore on Node 14. This library hasn't been updated for a while and it's a kinda obscure one so I preferred to add the rewrite code here.

willdurand: Note: that rewrite part is needed because tests fail on node 14, which wasn't covered before but I do reproduce locally.

willdurand:
We save more than a minute thanks to the last commit.

willdurand:
This patch is now ready for a review.

src/server/index.tsx
```diff
@@ -294,28 +291,56 @@ export const createServer = ({
...
+              res.setHeader(key, proxyResponse.headers[key] || '');
+            });
+
+            const contentType = proxyResponse.headers['content-type'] || '';
+            const contentEncoding =
+              proxyResponse.headers['content-encoding'] || '';
+
+            let body;
+            if (contentEncoding.includes('gzip')) {
+              body = zlib
```
bobsilverberg, line 327: I don't see any test coverage for this block. Should we have some?

src/server/index.tsx
```diff
@@ -294,28 +291,56 @@ export const createServer = ({
...
+            } else {
+              body = Buffer.concat(bodyChunks).toString('utf8');
+            }
+
+            if (contentType.includes('text/html')) {
+              res.setHeader('cache-control', 'private, no-store');
+              body = _injectAuthenticationToken(req, body, env);
+            }
+
+            res.setHeader('content-length', body.length);
```
bobsilverberg, line 341: I don't see any test coverage for this. Should we have some?

bobsilverberg (approved):
Looks good, thanks for taking this on @willdurand. I checked for test coverage for some of the changes you introduced into `src/server/index.tsx` and it seems there is a bit of coverage missing. Other than that it looks good!

r+wc

.circleci/config.yml
```diff
@@ -3,12 +3,115 @@
...
+      - checkout
+      - *restore_build_cache
+      - *run_yarn_install
+      - *save_build_cache
+      - run: yarn test-ci
+      - run:
+          name: yarn serve
+          command: |
+            yarn build
+            yarn serve &
```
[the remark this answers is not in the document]
willdurand: yeah, I can't remember exactly why we have that but we put it for a good reason :|

.circleci/config.yml
```diff
@@ -3,12 +3,107 @@
...
+      - *run_yarn_install
+      - *save_next_build_cache
+      # See: https://support.circleci.com/hc/en-us/articles/360038192673-NodeJS-Builds-or-Test-Suites-Fail-With-ENOMEM-or-a-Timeout
+      - run: yarn test --maxWorkers=2
+
+  deploy-storybook:
+    <<: *defaults
+    steps:
+      - checkout
+      - *restore_next_build_cache
```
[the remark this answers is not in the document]
willdurand: doh

src/server/index.tsx
willdurand: yes

PR #1878, opened 2021-03-08
upgrade react-scripts to v4

willdurand:


src/components/SidePanel/index.tsx
```diff
@@ -27,7 +27,6 @@ export const SidePanelBase = ({
 
   return (
     <aside
-      aria-expanded={isExpanded ? 'true' : 'false'}
```
willdurand, line 30: this isn't valid.

diox (approved):

PR #1882, opened 2021-03-10
Upgrade Storybook to v6

willdurand:
---


willdurand:
@bobsilverberg before I go ahead and move all existing stories, what do you think of this approach? I removed the `chapters` plugin (not maintained, outdated, kinda broken with v6) and I rebuilt the config from scratch.

bobsilverberg (commented):
I like it! I think it's better than the current approach. Nice work @willdurand!

The only odd thing I noticed was when I run `yarn storybook`, it opens Google Chrome (even though Firefox is my default browser and already open), but it doesn't load the URL into Chrome. It would be nice if there was a switch to turn that behaviour off.

willdurand:
> The only odd thing I noticed was when I run `yarn storybook`, it opens Google Chrome (even though Firefox is my default browser and already open), but it doesn't load the URL into Chrome. It would be nice if there was a switch to turn that behaviour off.

mmm for me, it uses Firefox. I'll take a look, thanks.

bobsilverberg:
@willdurand Some of the security updates that cannot be applied to this repo seem to be blocked by Storybook < 6, so it might be a good time to finish this off. Do you think you can get this in shape to merge?

willdurand:
thank you for finishing this patch @bobsilverberg!

bobsilverberg:
> thank you for finishing this patch @bobsilverberg!

You can thank me when it's actually finished. 😉

bobsilverberg:
@willdurand I finished migrating the stories, and then I rebased everything on top of the current `master`, and now I am getting a lot of `eslint` errors and even an error from `tsc`. 😢 

The `eslint` errors shouldn't be too hard to fix (we might want to disable some rules), but I have no idea where the `tsc` error comes from as it doesn't seem to be related to anything changed in this PR. I'll dig in some more.

bobsilverberg:
> @willdurand I finished migrating the stories, and then I rebased everything on top of the current `master`, and now I am getting a lot of `eslint` errors and even an error from `tsc`. 😢
> 
> The `eslint` errors shouldn't be too hard to fix (we might want to disable some rules), but I have no idea where the `tsc` error comes from as it doesn't seem to be related to anything changed in this PR. I'll dig in some more.

These had something to do with a bad `yarn.lock` file. I restored the one from `master` and that seemed to make these problems go away. 😄

bobsilverberg:
This is done and ready to land. There is one issue, but I'd like to treat it as a follow-up, as this has been sitting around for so long and I'd prefer it not bit rot. I opened https://github.com/mozilla/addons-code-manager/issues/2144, which describes the issue, but I don't think it needs to block landing this.

I am going to request a review from `addons-amo-engineering` as both @willdurand and I have worked on this patch.

willdurand:
> These had something to do with a bad `yarn.lock` file. I restored the one from `master` and that seemed to make these problems go away. 😄

Isn't the `yarn.lock` file supposed to be updated when there are changes in `package.json`?

bobsilverberg:
> > These had something to do with a bad `yarn.lock` file. I restored the one from `master` and that seemed to make these problems go away. 😄
> 
> Isn't the `yarn.lock` file supposed to be updated when there are changes in `package.json`?

True, let me see if I can get that to work.

bobsilverberg:
> > > These had something to do with a bad `yarn.lock` file. I restored the one from `master` and that seemed to make these problems go away. 😄
> > 
> > 
> > Isn't the `yarn.lock` file supposed to be updated when there are changes in `package.json`?
> 
> True, let me see if I can get that to work.

@willdurand Ugh, for some reason, after I run `yarn` locally, allowing it to update the `yarn.lock` file, this is now generating `tsc` errors again. I'm not sure why that's happening locally but not with this PR. I guess some dependency that is getting updated as a result of changes to `package.json` in this PR is creating some sort of problem.

I will dig into it, but I am concerned about the fact that this is going to eat up a huge chunk of time. 😢

bobsilverberg:
> > > > These had something to do with a bad `yarn.lock` file. I restored the one from `master` and that seemed to make these problems go away. 😄
> > > 
> > > 
> > > Isn't the `yarn.lock` file supposed to be updated when there are changes in `package.json`?
> > 
> > 
> > True, let me see if I can get that to work.
> 
> @willdurand Ugh, for some reason, after I run `yarn` locally, allowing it to update the `yarn.lock` file, this is now generating `tsc` errors again. I'm not sure why that's happening locally but not with this PR. I guess some dependency that is getting updated as a result of changes to `package.json` in this PR is creating some sort of problem.
> 
> I will dig into it, but I am concerned about the fact that this is going to eat up a huge chunk of time. 😢

And now it's failing on CI too, even with no changes included in `yarn.lock`. This is essentially the same patch that we just recently passing. I think this might have to do with the yarn cache on CI, but I can't find an option anywhere in Circle to delete/clear the yarn cache.

bobsilverberg:
> I think we should update the yarn lockfile based on what `yarn install` does, then fix the error. It is probably related to some dep changes (as you said) but that does not mean they are invalid.

Yes, that is what I am attempting to do, but I think I found a serious problem. A bunch of `tsc` errors are related to types to the `Route` component. Looking at changes in the lockfile, I see this:
```
"@storybook/router@6.4.0":
  version "6.4.0"
  resolved "https://registry.yarnpkg.com/@storybook/router/-/router-6.4.0.tgz#becdd52975d221695dcbddc99bced22ef42c81a1"
  integrity sha512-qd6GisJUIf/8fjrTlzIg5MfXLm2b2u33PiHYknm9Pu2wu4FOj8A4jfwKccy9tQ1w8GcEysjF4YQseIG1RK0VTw==
  dependencies:
    "@storybook/client-logger" "6.4.0"
    core-js "^3.8.2"
    fast-deep-equal "^3.1.3"
    global "^4.4.0"
    history "5.0.0"
    lodash "^4.17.20"
    memoizerific "^1.11.3"
    qs "^6.10.0"
    react-router "^6.0.0"
    react-router-dom "^6.0.0"
    ts-dedent "^2.0.0"
```

So it looks like Storybook v6 requires `react-router` v6, and it has gone ahead and installed it. We have specifically not picked up `v6` of `react-router` yet because of incompatibilities. `v6` of `react-router` changes the type definitions, which is why we are now getting `tsc` failures. I'm pretty sure this means we cannot update to Storybook `v6` without also updating to `react-router` `v6`, which might not be possible at this moment.

Does that sound like an accurate assessment, @willdurand?

Do you still have your original branch hanging around? I wonder if you were able to get your branch passing as it existed.

willdurand:
> Does that sound like an accurate assessment, @willdurand?

Yeah, I checked out this patch locally to investigate and after some Apple M1-related traps, I found out that some typescript errors don't make sense. What you found here does explain why, though. Ugh!

bobsilverberg:
> > Does that sound like an accurate assessment, @willdurand?
> 
> Yeah, I checked out this patch locally to investigate and after some Apple M1-related traps, I found out that some typescript errors don't make sense. What you found here does explain why, though. Ugh!

Yes, I suppose we have to add this to our long list of interdependent dependencies which are blocked. At some point we will resolve some of these. 😢

willdurand:
We could force-install react-router v4 with this diff:

```diff
diff --git a/package.json b/package.json
index 4ed32cd..5517718 100644
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "react-ga": "3.3.0",
     "react-helmet": "6.1.0",
     "react-redux": "5.1.2",
+    "react-router": "4.3.1",
     "react-router-dom": "4.3.1",
     "react-scripts": "4.0.3",
     "react-textarea-autosize": "7.1.2",
```

bobsilverberg:
> We could force-install react-router v4 with this diff:
> 
> ```diff
> diff --git a/package.json b/package.json
> index 4ed32cd..5517718 100644
> --- a/package.json
> +++ b/package.json
> @@ -55,6 +55,7 @@
>      "react-ga": "3.3.0",
>      "react-helmet": "6.1.0",
>      "react-redux": "5.1.2",
> +    "react-router": "4.3.1",
>      "react-router-dom": "4.3.1",
>      "react-scripts": "4.0.3",
>      "react-textarea-autosize": "7.1.2",
> ```

I thought about trying that, but if Storybook `v6` really does required `react-router v6`, wouldn't that be a dangerous thing to do?

bobsilverberg:
> > We could force-install react-router v4 with this diff:
> > ```diff
> > diff --git a/package.json b/package.json
> > index 4ed32cd..5517718 100644
> > --- a/package.json
> > +++ b/package.json
> > @@ -55,6 +55,7 @@
> >      "react-ga": "3.3.0",
> >      "react-helmet": "6.1.0",
> >      "react-redux": "5.1.2",
> > +    "react-router": "4.3.1",
> >      "react-router-dom": "4.3.1",
> >      "react-scripts": "4.0.3",
> >      "react-textarea-autosize": "7.1.2",
> > ```
> 
> I thought about trying that, but if Storybook `v6` really does required `react-router v6`, wouldn't that be a dangerous thing to do?

Well, I tried it, and it works, and it doesn't *seem* to break Storybook, so I suppose we could give it a shot? I'm just worried that it might cause a problem down the road that might be very difficult to debug. I guess we can add a comment to `package.json` about why we did it.

bobsilverberg:
> 
> Well, I tried it, and it works, and it doesn't _seem_ to break Storybook, so I suppose we could give it a shot? I'm just worried that it might cause a problem down the road that might be very difficult to debug. I guess we can add a comment to `package.json` about why we did it.

Comments not allowed in `package.json` 😢

willdurand:
Let's land it and if Storybook goes back to bite us later, we'll discuss about removing it.

bobsilverberg:
> Let's land it and if Storybook goes back to bite us later, we'll discuss about removing it.

Sounds good. I've requested a review from @diox as we both worked on this patch.

diox (approved):
As mentioned on Matrix I have a weird scrollbar issue and https://github.com/mozilla/addons-code-manager/issues/2144 is annoying but that's not a dealbreaker, it seems to work.

Issue #2156, opened 2022-02-17
Support addons-server SessionIDAuthentication


eviljeff:
https://github.com/mozilla/addons-server/pull/18475 adds a new internal authentication method, that would be specified in the authorization header as Session xxxx (rather than Bearer xxxx). addons-fronted made the change in https://github.com/mozilla/addons-frontend/issues/11110.  The value is stored in the sessionID cookie - it's the django session ID. We need to change addons-code-manager too so it uses this cookie rather than frontend_auth_token and specifies the value in the authorization header.

note: The cookie/auth token is no longer a JWT, it's just a session id.

eviljeff:
QA: the changes here were the same as frontend - I wouldn't worry about the refresh and access tokens expiring, just focus on normal login and api requests.

ioanarusiczki:
@eviljeff I tried on -dev and -stage to login to a code manager page -> If I access "Content" from an addon review page I must login to code manager but after I enter credentials I still won't have access. 

I think this issue caused the regression: 

![code manager login fails](https://user-images.githubusercontent.com/33448286/155290112-ae47efcc-13c7-4c05-ad27-dceebaf55d72.gif)

eviljeff:
@ioanarusiczki it's hopefully working now on dev

ioanarusiczki:
@eviljeff Works on -dev , not yet on -stage.

eviljeff:
@ioanarusiczki thanks for the confirmation about -dev.  I've just cherry-picked the commit to stage now.

ioanarusiczki:
@eviljeff 
- Login works on AMO stage too
- I could browse on code manager pages using keyboard navigation or the code overview , leave comments, download a file and so on. 

I think it should be alright.

PR #2159, opened 2022-02-23
Fix session id auth: the data attribute in the HTML needs to match

diox:
Fixes https://github.com/mozilla/addons-code-manager/issues/2156

### DIFF
--- a/.env
+++ b/.env
@@ -5,7 +5,7 @@ REACT_APP_REVIEWERS_HOST=https://reviewers.addons.mozilla.org
 
 # This is the placeholder used in the `index.html` file to inject the
 # authentication token on runtime.
-REACT_APP_AUTH_TOKEN_PLACEHOLDER=__AUTH_TOKEN__
+REACT_APP_USER_AUTH_SESSION_ID_PLACEHOLDER=__AUTH_TOKEN__
 REACT_APP_AUTHENTICATION_COOKIE=sessionid
 
 # This is the language that will be sent with all API requests.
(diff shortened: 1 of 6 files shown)


src/reducers/api.tsx
```diff
@@ -9,7 +9,7 @@ type SetAuthSessionIdPayload = {
 };
 
 export const actions = {
-  setAuthSessionId: createAction('SET_AUTH_TOKEN', (resolve) => {
+  setAuthSessionId: createAction('SET_AUTH_SESSION_ID', (resolve) => {
```
eviljeff, line 12: I found this with a search but tbh I'm not 100% what is it for - we don't reference it elsewhere in the code.

.env
```diff
@@ -5,7 +5,7 @@ REACT_APP_REVIEWERS_HOST=https://reviewers.addons.mozilla.org
 
 # This is the placeholder used in the `index.html` file to inject the
 # authentication token on runtime.
-REACT_APP_AUTH_TOKEN_PLACEHOLDER=__AUTH_TOKEN__
+REACT_APP_USER_AUTH_SESSION_ID_PLACEHOLDER=__AUTH_TOKEN__
```
eviljeff, line 8: Should we rename `__AUTH_TOKEN__` too?  It's not referenced anywhere else so 🤷

src/reducers/api.tsx
eviljeff: I assume we're okay to change it then

bobsilverberg: It's just an internal constant to identify the action. It makes sense that it isn't used anywhere else.

bobsilverberg (approved):
Looks good, thanks @diox and @eviljeff for taking care of this!

Issue #2014, opened 2021-07-16
Update Circle CI config to use Node 14 and 16


willdurand:
We updated the Node version from 12 to 14 in #1984. We need to update the Circle config to change the Node versions used in CI.

diox:
We are no longer using Node 12, but we can't update to Node 16 yet - tests are not passing.

diox:
With Node 16, some tests are not passing because we've got unhandled promises. With Node 14, the following warning was triggered:
```
(node:223209) UnhandledPromiseRejectionWarning: TypeError: versions.filter is not a function
(Use `node --trace-warnings ...` to show where the warning was created)
(node:223209) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:223209) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code
```

With Node 16, it fails and exits right away.

Issue #1984, opened 2021-06-11
Node 14


willdurand:
With Node 16 already available, we can switch from Node 12 to Node 14 in the coming weeks. This issue is about doing it.

willdurand:
cc @bqbn

ioanarusiczki:
@willdurand I'm going to test this on AMO stage, at release testing.

ioanarusiczki:
@willdurand 
Verified on AMO stage - FF90(Win10).

PR #2163, opened 2022-03-15
Replace legacy circleci images (removing node 12 in the process)

### LINKED ISSUE #2162: Move away from machine: true and legacy convenience images in CircleCI config
We received the following email from CircleCI:
> Hello,
>
> We are deprecating Ubuntu 14.04-based machine images on CircleCI in preparation for an EOL on Tuesday, May 31, 2022 to ensure your builds remain secure. For a detailed overview of how this will affect your workflow, [read the blog article here](https://go.circleci.com/NDg1LVpNSC02MjYAAAGCndaubP1lrB_xKEs9K3iqzYYxUBqxzrHoMsjM7oN8hTLoyfIucsihEKChdSM-NVzcCvkPtwk=).
>
> We will also be conducting temporary brownouts on Tuesday, March 29, 2022, and again on Tuesday, April 26, 2022 during which these images will be unavailable.
> 
> We are contacting you because one or more of your projects has a job that either:
>
>    does not specify an image (uses machine: true in config)
>    explicitly uses an Ubuntu 14.04-based image
>
> 
> If you have specified an Ubuntu 14.04-based image or you are using machine: true in your config file, please see our [migration guide](https://go.circleci.com/NDg1LVpNSC02MjYAAAGCndaubAkpQkmFHfDYu-fp4EmnV8zWBJhxf8MPSWnVgraetk-HLjJopo1dxVGRR4O4citj3UQ=) to upgrade to a newer version of Ubuntu image in order to avoid any service disruption during the brownout & subsequent EOL.
>
> We will also be releasing a CircleCI Ubuntu 22.04 image on April 22nd offering the flexibility to upgrade to the latest LTS version of Ubuntu image before we remove older versions permanently. A beta version of the image will be available March 21st.

Like we did in https://github.com/mozilla/addons-server/issues/18843 we need to move away from `machine: true`.

On top of that, like in https://github.com/mozilla/addons-frontend/issues/11340 we're using some legacy convenience images (`circleci/*`) that [are deprecated](https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034) and that we need to change:
https://github.com/mozilla/addons-code-manager/blob/fcfc827a49673ff5341d41369b2ee048b4e20ddb/.circleci/config.yml#L15-L23

diox:
We're using node 14 in production, so don't need to test with 12 any longer. We're not ready for node 16 yet though, so comment that job out.

Fixes #2162

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,16 +14,18 @@ references:
     working_directory: ~/addons-code-manager
     docker:
       # This is the NodeJS version we run in production.
-      - image: circleci/node:12
+      - image: cimg/node:14.19
 
-  defaults-next: &defaults-next
-    working_directory: ~/addons-code-manager
-    docker:
-      # This is the next NodeJS version we will support.
-      - image: circleci/node:14
+  # https://github.com/mozilla/addons-code-manager/issues/2014
+  # defaults-next: &defaults-next
+  #   working_directory: ~/addons-code-manager
+  #   docker:
+  #     # This is the next NodeJS version we will support.
+  #     - image: cimg/node:16.14
 
   defaults-release: &defaults-release
-    machine: true
+    machine:
+      image: ubuntu-2004:202201-02
     working_directory: ~/addons-code-manager
 
   restore_build_cache: &restore_build_cache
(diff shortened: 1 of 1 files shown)


willdurand:
> We're not ready for node 16 yet though, so comment that job out.

What does that mean? And why is this commented out?

diox:
There is little point in keeping the node 12 tests, it's not a target anymore since https://github.com/mozilla/addons-code-manager/issues/1984, so I wanted to bump the `default` test job to node 14 to match production env.

Tests are not passing with node 16 yet though (https://github.com/mozilla/addons-code-manager/issues/2014) so I can't keep the test-next job around, hence why I commented it out.

willdurand:
> There is little point in keeping the node 12 tests, it's not a target anymore since #1984, so I wanted to bump the `default` test job to node 14 to match production env.

Yeah, that part I understood.



> Tests are not passing with node 16 yet though (#2014) so I can't keep the test-next job around, hence why I commented it out.

Right, I thought I was reviewing a patch for customs actually, which is why I didn't expect Node 16 to be commented out. This makes sense in the context of code-manager :p (this is a bit scary, though... we should really find out what's going on)

willdurand (approved):

willdurand:
(I am fine with commenting out the existing Node 16 config since we really want to add it back)

PR #2175, opened 2022-09-28
Update CI to run on Node 16 as well

diox:
Fixes https://github.com/mozilla/addons-code-manager/issues/2014

### DIFF
--- a/src/components/LinterProvider/index.spec.tsx
+++ b/src/components/LinterProvider/index.spec.tsx
@@ -197,6 +197,9 @@ describe(__filename, () => {
     const className = 'childClassName';
     const children = jest.fn().mockReturnValue(<div className={className} />);
     const store = configureStore();
+    // Pretend we already have a linter result - otherwise we'd try to fetch it
+    // and that would fail as the mocked empty API response would not be
+    // handled correctly in the reducer.
     _loadLinterResult({ store, messages: [] });
 
     const root = render({ children, store });
--- a/src/components/LinterProvider/index.spec.tsx
+++ b/src/components/LinterProvider/index.spec.tsx
@@ -195,9 +195,11 @@ describe(__filename, () => {
 
   it('renders child content', () => {
     const className = 'childClassName';
-    const root = render({
-      children: jest.fn().mockReturnValue(<div className={className} />),
-    });
+    const children = jest.fn().mockReturnValue(<div className={className} />);
+    const store = configureStore();
+    _loadLinterResult({ store, messages: [] });
+
+    const root = render({ children, store });
 
     expect(root).toHaveClassName(className);
   });
--- a/src/pages/Browse/index.spec.tsx
+++ b/src/pages/Browse/index.spec.tsx
@@ -69,6 +69,7 @@ describe(__filename, () => {
     _viewVersionFile,
     addonId = String(nextUniqueId()),
     history = createFakeHistory(),
+    shouldMockDispatch = false,
     store = configureStore(),
     versionId = String(nextUniqueId()),
   }: RenderParams = {}) => {
@@ -82,6 +83,12 @@ describe(__filename, () => {
       _log,
       _viewVersionFile,
     };
+
+    // Sometimes we need to mock `dispatch` to prevent it from dispatching
+    // actions which would result in an unhandled promise rejection.
+    if (shouldMockDispatch) {
+      spyOn(store, 'dispatch');
+    }
 
     return shallowUntilTarget(<Browse {...props} />, BrowseBase, {
       shallowOptions: {
@@ -312,9 +319,11 @@ describe(__filename, () => {
       }),
     );
 
-    spyOn(store, 'dispatch');
-
-    const root = render({ store, versionId: String(version.id) });
+    const root = render({
+      shouldMockDispatch: true,
+      store,
+      versionId: String(version.id),
+    });
 
     expect(root.find(CodeView)).toHaveLength(0);
     expect(root.find(CodeOverview)).toHaveLength(0);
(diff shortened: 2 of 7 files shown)


src/reducers/linter.tsx
```diff
@@ -220,6 +220,10 @@ export const fetchLinterMessagesIfNeeded = ({
       fileId: version.fileId,
     });
 
+    if (!response) {
```
bobsilverberg, line 223: Is this safe to add? It feels dangerous to be changing production code in order to fix a problem in a test.

bobsilverberg (commented):
I haven't looked at this code in years, so I'm not sure if my concern is valid, but I am uncomfortable adding production code into functions which will change the behaviour of those functions in order to address a test issue. Can we not fix the tests instead somehow?

Is the argument that, if `response` were to be falsey in any of those instances it would raise an exception, so we can be sure that we'd already know about it if `response` could actually be falsey?

src/reducers/linter.tsx
diox: I'm not entirely sure it only happens in tests, it's hard to know without digging deeper into the code. Dispatching an action here is clearly wrong - the underlying action code does not handle being passed those falsey `response` objects.

It was already bad before, it's just more apparent with Node 16: Node 14 used to just warn, Node 16 terminates the process because we end up with an exception thrown and not caught in an async function.

bobsilverberg: But is *this* the fix? Again, it feels dangerous to me to just `return` and act as if everything is ok.

willdurand:
I think it'd be best to fix the failing tests. For example, I noticed this one which was probably hitting the actual API layer (with fetch mock but still...). We can fix it by avoiding some actions to be dispatched:

```diff
diff --git a/src/components/LinterProvider/index.spec.tsx b/src/components/LinterProvider/index.spec.tsx
index a753fff..f1d8246 100644
--- a/src/components/LinterProvider/index.spec.tsx
+++ b/src/components/LinterProvider/index.spec.tsx
@@ -195,9 +195,11 @@ describe(__filename, () => {
 
   it('renders child content', () => {
     const className = 'childClassName';
-    const root = render({
-      children: jest.fn().mockReturnValue(<div className={className} />),
-    });
+    const children = jest.fn().mockReturnValue(<div className={className} />);
+    const store = configureStore();
+    _loadLinterResult({ store, messages: [] });
+
+    const root = render({ children, store });
 
     expect(root).toHaveClassName(className);
   });
```

diox:
Since you both agree, I'll look into doing this only in the tests

diox:
This test in `src/pages/Browse/index.spec.ts` is giving me trouble:
https://github.com/mozilla/addons-code-manager/blob/f917554210658d68812de5a78c3632caca588558/src/pages/Browse/index.spec.tsx#L574-L578

In order for this to work, we need to `Browse` with `version` being falsey, because of:
https://github.com/mozilla/addons-code-manager/blob/f917554210658d68812de5a78c3632caca588558/src/pages/Browse/index.tsx#L218-L224

But if we try calling `dispatchLoadVersionInfo` with `version` being `undefined` or `null`/`{}` before the `render()` in that test, other actions will be dispatched that end up failing, in `fileTree` or `createInternalVersionAddon`... And if we do nothing, we end up in the current situation, where sure, `version` is undefined, but we are also calling the (mock) API, which returns an empty response.

bobsilverberg:
If we don't actually need the data loaded for the test, it's better to just prevent `dispatch` from doing anything, rather than trying to set up the data so that dispatching an action doesn't result in an unhandled promise rejection. If we call `spyOn(store, 'dispatch');` before rendering the component, then `dispatch` does nothing, and it solves the problem, at least for the 3 tests in `src/pages/Browse/index.spec.ts`, and this will likely help us resolve the other problems as well. I have pushed a new commit with the fixes to `src/pages/Browse/index.spec.ts`.

src/pages/Browse/index.spec.tsx
```diff
@@ -83,6 +84,12 @@ describe(__filename, () => {
       _viewVersionFile,
     };
 
+    // Sometimes we need to mock `dispatch` to prevent it from dispatching
+    // actions which would result in an unhandled promise rejection.
```
willdurand, line 88: Although the unhandled rejection is obvious, the result is more that it's going to call a fetch mock that will return a falsey response, leading to the rejection if we try to read data from it

diox, line 88: How about this ?
```suggestion
    // Some tests require data to not have been loaded. Mocking dispatch
    // ensures actions that cause us to load data from the API aren't
    // triggered at all (that would raise an unhandled exception inside
    // async code and cause the node process to terminate, as tests mock API
    // responses but the reducers code usually doesn't handle API responses
    // being empty).
```

willdurand: LTGM

diox: @bobsilverberg are you ok with that comment ? If so I can copy/paste it where to the other files.

diox: Went ahead and did it in 3c962063a612040ee810090032374e4c5d1059ff

src/components/LinterProvider/index.spec.tsx
```diff
@@ -195,9 +195,11 @@ describe(__filename, () => {
 
   it('renders child content', () => {
     const className = 'childClassName';
-    const root = render({
-      children: jest.fn().mockReturnValue(<div className={className} />),
-    });
+    const children = jest.fn().mockReturnValue(<div className={className} />);
+    const store = configureStore();
+    _loadLinterResult({ store, messages: [] });
```
willdurand, line 200: I am wondering if we should mock dispatch here too.. although, that is working as intended. Only thing would be to add a comment here too, probably..

willdurand (approved):
r+wc

PR #2188, opened 2023-01-05
Disable storybook-deployer

### LINKED ISSUE #2169: Stop deploying storybook to gh-pages ; build it when needed instead
We're currently building and deploying storybook on each commit to `master`, through `addons-robot`, which has write access to the repo.

I'd like to remove that permission from addons-robot for safety. I propose we do something similar to [what was done in the linter](https://github.com/mozilla/addons-linter/issues/3853): build it ourselves when needed, and commit the result. CI will still build it to check that it hasn't changed, and fail the build if it did, forcing you to build and commit the result if it changed.

The drawback with this method is that sometimes the diff will be unnecessary large, especially as storybooks come with a lot of JavaScript we don't necessarily care about. I think it's an acceptable trade-off in the name of increased security, but please discuss if you disagree.

diox:
Fixes #2169

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -107,15 +107,6 @@ jobs:
       # See: https://support.circleci.com/hc/en-us/articles/360038192673-NodeJS-Builds-or-Test-Suites-Fail-With-ENOMEM-or-a-Timeout
       - run: yarn test --maxWorkers=2
 
-  deploy-storybook:
-    <<: *defaults
-    steps:
-      - checkout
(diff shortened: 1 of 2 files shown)


willdurand:
I think we should document the process to update Storybook on the `gh-pages` branch.

diox:
Added a basic explanation in c4e3e72290258eae475285dc1bfbf38184df9557

willdurand (approved):

